PR #1, opened 2017-08-28
unit tests for duplicate pods elimination

ravisantoshgudimetla:
/cc @aveshagarwal This covers the unit tests for duplicatePod removal strategy.

### DIFF
--- a/pkg/rescheduler/strategies/duplicates.go
+++ b/pkg/rescheduler/strategies/duplicates.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 
 	"k8s.io/kubernetes/pkg/api/v1"
+	//TODO: Change to client-go instead of generated clientset.
 	"k8s.io/kubernetes/pkg/client/clientset_generated/clientset"
 
 	"github.com/aveshagarwal/rescheduler/pkg/api"
@@ -31,14 +32,22 @@ import (
 //type creator string
 type DuplicatePodsMap map[string][]*v1.Pod
 
+// RemoveDuplicatePods removes the duplicate pods on node. This strategy evicts all duplicate pods on node.
+// A pod is said to be a duplicate of other if both of them are from same creator, kind and are within the same
+// namespace. As of now, this strategy won't evict daemonsets, mirror pods, critical pods and pods with local storages.
 func RemoveDuplicatePods(client clientset.Interface, strategy api.ReschedulerStrategy, policyGroupVersion string, nodes []*v1.Node) {
 	if !strategy.Enabled {
 		return
 	}
+	deleteDuplicatePods(client, policyGroupVersion, nodes)
+}
 
+// deleteDuplicatePods evicts the pod from node and returns the count of evicted pods.
+func deleteDuplicatePods(client clientset.Interface, policyGroupVersion string, nodes []*v1.Node) int {
+	podsEvicted := 0
 	for _, node := range nodes {
 		fmt.Printf("\nProcessing node: %#v\n", node.Name)
-		dpm := RemoveDuplicatePodsOnANode(client, node)
+		dpm := ListDuplicatePodsOnANode(client, node)
 		for creator, pods := range dpm {
 			if len(pods) > 1 {
 				fmt.Printf("%#v\n", creator)
@@ -47,17 +56,21 @@ func RemoveDuplicatePods(client clientset.Interface, strategy api.ReschedulerStr
 					//fmt.Printf("Removing duplicate pod %#v\n", k.Name)
 					success, err := evictions.EvictPod(client, pods[i], policyGroupVersion)
 					if !success {
+						//TODO: change fmt.Printf as glogs.
 						fmt.Printf("Error when evicting pod: %#v (%#v)\n", pods[i].Name, err)
 					} else {
+						podsEvicted++
 						fmt.Printf("Evicted pod: %#v (%#v)\n", pods[i].Name, err)
 					}
 				}
 			}
 		}
 	}
+	return podsEvicted
 }
 
-func RemoveDuplicatePodsOnANode(client clientset.Interface, node *v1.Node) DuplicatePodsMap {
+// ListDuplicatePodsOnANode lists duplicate pods on a given node.
+func ListDuplicatePodsOnANode(client clientset.Interface, node *v1.Node) DuplicatePodsMap {
 	pods, err := podutil.ListPodsOnANode(client, node)
 	if err != nil {
 		return nil
@@ -65,9 +78,9 @@ func RemoveDuplicatePodsOnANode(client clientset.Interface, node *v1.Node) Dupli
 	return FindDuplicatePods(pods)
 }
 
+// FindDuplicatePods takes a list of pods and returns a duplicatePodsMap.
 func FindDuplicatePods(pods []*v1.Pod) DuplicatePodsMap {
 	dpm := DuplicatePodsMap{}
-
 	for _, pod := range pods {
 		sr, err := podutil.CreatorRef(pod)
 		if err != nil || sr == nil {
--- a/pkg/rescheduler/strategies/duplicates_test.go
+++ b/pkg/rescheduler/strategies/duplicates_test.go
@@ -1,0 +1,177 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package strategies
+
+import (
+	"fmt"
+	"testing"
+
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	core "k8s.io/client-go/testing"
+	"k8s.io/kubernetes/pkg/api/v1"
+	"k8s.io/kubernetes/pkg/client/clientset_generated/clientset/fake"
+)
+
+// TODO:@ravisantoshgudimetla. As of now building some test pods here. This needs to
+// move to utils after refactor.
+// buildTestPod creates a test pod with given parameters.
+func buildTestPod(name string, cpu int64, memory int64, nodeName string) *v1.Pod {
+	pod := &v1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: "default",
+			Name:      name,
+			SelfLink:  fmt.Sprintf("/api/v1/namespaces/default/pods/%s", name),
+		},
+		Spec: v1.PodSpec{
+			Containers: []v1.Container{
+				{
+					Resources: v1.ResourceRequirements{
+						Requests: v1.ResourceList{},
+					},
+				},
+			},
+			NodeName: nodeName,
+		},
+	}
+	if cpu >= 0 {
+		pod.Spec.Containers[0].Resources.Requests[v1.ResourceCPU] = *resource.NewMilliQuantity(cpu, resource.DecimalSI)
+	}
+	if memory >= 0 {
+		pod.Spec.Containers[0].Resources.Requests[v1.ResourceMemory] = *resource.NewQuantity(memory, resource.DecimalSI)
+	}
+
+	return pod
+}
+
+// buildTestNode creates a node with specified capacity.
+func buildTestNode(name string, millicpu int64, mem int64) *v1.Node {
+	node := &v1.Node{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:     name,
+			SelfLink: fmt.Sprintf("/api/v1/nodes/%s", name),
+			Labels:   map[string]string{},
+		},
+		Status: v1.NodeStatus{
+			Capacity: v1.ResourceList{
+				v1.ResourcePods:   *resource.NewQuantity(100, resource.DecimalSI),
+				v1.ResourceCPU:    *resource.NewMilliQuantity(millicpu, resource.DecimalSI),
+				v1.ResourceMemory: *resource.NewQuantity(mem, resource.DecimalSI),
+			},
+			Allocatable: v1.ResourceList{
+				v1.ResourcePods:   *resource.NewQuantity(100, resource.DecimalSI),
+				v1.ResourceCPU:    *resource.NewMilliQuantity(millicpu, resource.DecimalSI),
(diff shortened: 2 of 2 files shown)


pkg/rescheduler/strategies/duplicates.go
```diff
@@ -21,49 +21,60 @@ import (
...
+// both of them have same name, kind and are within the same namespace. As of now, this strategy won't
+// evict daemonsets, mirror pods and pods with local storages.
+
 //type creator string
-type DuplicatePodsMap map[string][]*v1.Pod
+type duplicatePodsMap map[string][]*v1.Pod
 
-func RemoveDuplicatePods(client clientset.Interface, policyGroupVersion string, nodes []*v1.Node) error {
+// RemoveDuplicatePods removes the duplicate pods on node.
+func RemoveDuplicatePods(client clientset.Interface, policyGroupVersion string, nodes []*v1.Node) (int, error) {
```
aveshagarwal: Could you rebase this, as I changed it.

pkg/rescheduler/strategies/duplicates.go
```diff
@@ -21,49 +21,60 @@ import (
...
+	podsEvicted := 0
 	for _, node := range nodes {
 		fmt.Printf("\nProcessing node: %#v\n", node.Name)
-		dpm := RemoveDuplicatePodsOnANode(client, node)
+		dpm := listDuplicatePodsOnANode(client, node)
```
aveshagarwal: I have intentionally kept all exported for time being for faster development, as things will be changing fast.

pkg/rescheduler/strategies/duplicates.go
```diff
@@ -21,49 +21,60 @@ import (
...
 	podutil "github.com/aveshagarwal/rescheduler/pkg/rescheduler/pod"
 )
 
+// This strategy evicts all duplicate pods on node. A pod is said to be a duplicate of other if
+// both of them have same name, kind and are within the same namespace. As of now, this strategy won't
+// evict daemonsets, mirror pods and pods with local storages.
+
 //type creator string
-type DuplicatePodsMap map[string][]*v1.Pod
+type duplicatePodsMap map[string][]*v1.Pod
```
aveshagarwal: I have intentionally kept all exported for time being for faster development, as things will be changing fast.

pkg/rescheduler/strategies/duplicates.go
```diff
@@ -21,49 +21,60 @@ import (
...
 				fmt.Printf("%#v\n", creator)
 				// i = 0 does not evict the first pod
 				for i := 1; i < len(pods); i++ {
 					//fmt.Printf("Removing duplicate pod %#v\n", k.Name)
 					success, err := evictions.EvictPod(client, pods[i], policyGroupVersion)
 					if !success {
+						//TODO: change fmt.Printf as glogs.
 						fmt.Printf("Error when evicting pod: %#v (%#v)\n", pods[i].Name, err)
 					} else {
+						podsEvicted++
```
aveshagarwal: why do you need this, even for testing?

aveshagarwal:
thanks @ravisantoshgudimetla , just some comments, once you update, I will review and test again.

pkg/rescheduler/strategies/duplicates.go
ravisantoshgudimetla: > why do you need this, even for testing?

So as of now, we are silently failing for the pods that are not getting deleted but I thought we need to collect the errors and as well as pods that failed. This will be helpful for debugging later.

pkg/rescheduler/strategies/duplicates.go
ravisantoshgudimetla: Ok, makes sense. I will change them back.

ravisantoshgudimetla:
@aveshagarwal I reverted changes for naming except for signature of RemoveDuplicatePods. Please let me know, if you have some other idea to get the errors, removedPods.

pkg/rescheduler/strategies/duplicates.go
aveshagarwal: That is not the case, as its not about silently falling.. Pods not getting deleted is not an error always, It could be due to Pod Disruption Budget (PDB) too. Also rescheduler is best effort, that means that it just tries to evict a pod, but does not force.  Though, it would be good to know why a pod was not evicted, not sure how returning number of evicted pods is useful. Anyway let me think about it more and will let you know after you update. thanks.

ravisantoshgudimetla: I agree that a number will not be useful but a map with podName and error may be helpful is what I meant. I have rebased the branch so as to avoid conflicts.

pkg/rescheduler/strategies/duplicates.go
```diff
@@ -21,53 +21,64 @@ import (
...
+// This strategy evicts all duplicate pods on node. A pod is said to be a duplicate of other if
+// both of them have same name, kind and are within the same namespace. As of now, this strategy won't
+// evict daemonsets, mirror pods and pods with local storages.
+
 //type creator string
 type DuplicatePodsMap map[string][]*v1.Pod
 
-func RemoveDuplicatePods(client clientset.Interface, strategy api.ReschedulerStrategy, policyGroupVersion string, nodes []*v1.Node) {
+// RemoveDuplicatePods removes the duplicate pods on node.
+func RemoveDuplicatePods(client clientset.Interface, strategy api.ReschedulerStrategy, policyGroupVersion string, nodes []*v1.Node) (int, error) {
```
aveshagarwal: I want to keep all strategies call same (same input, no return) for now. So do no return here, see lownodeutilization for reference.

pkg/rescheduler/strategies/duplicates.go
aveshagarwal: I think It'd be better to have behind some glog levels once we move to glog from fmt. So only during debugging, it is enabled not by default.

pkg/rescheduler/strategies/duplicates.go
ravisantoshgudimetla: Ok, I was thinking of having a strategies function which calls the RemoveDuplicatePods so that we can have a single signature for all the strategies but I am fine with this approach as well. I need to think of another way to find the exact number of pods on the node.

aveshagarwal: I want to have strategy calls with same input and return types so that it'd be easier to move to a registration based approach, where when a rescheduler is created, strategies are registered first with it and then rescheduler executes them.

pkg/rescheduler/strategies/duplicates.go
ravisantoshgudimetla: Avesh, PTAL. I created another function which does the same thing but get called from RemoveDuplicatePods  and RemoveDuplicatePods won't return anything.

pkg/rescheduler/strategies/duplicates.go
```diff
@@ -21,24 +21,35 @@ import (
...
+// RemoveDuplicatePods removes the duplicate pods on node.
 func RemoveDuplicatePods(client clientset.Interface, strategy api.ReschedulerStrategy, policyGroupVersion string, nodes []*v1.Node) {
 	if !strategy.Enabled {
 		return
 	}
+	deleteDuplicatePods(client, policyGroupVersion, nodes)
+	return
+}
 
+func deleteDuplicatePods(client clientset.Interface, policyGroupVersion string, nodes []*v1.Node) (int, error) {
```
aveshagarwal: no error is being returned, so just return count of evicted pods.

pkg/rescheduler/strategies/duplicates.go
```diff
@@ -21,24 +21,35 @@ import (
...
 //type creator string
 type DuplicatePodsMap map[string][]*v1.Pod
 
+// RemoveDuplicatePods removes the duplicate pods on node.
 func RemoveDuplicatePods(client clientset.Interface, strategy api.ReschedulerStrategy, policyGroupVersion string, nodes []*v1.Node) {
 	if !strategy.Enabled {
 		return
 	}
+	deleteDuplicatePods(client, policyGroupVersion, nodes)
+	return
```
aveshagarwal: this return is not needed.

pkg/rescheduler/strategies/duplicates.go
```diff
@@ -21,24 +21,35 @@ import (
...
+	//TODO: Change to client-go instead of generated clientset.
 	"k8s.io/kubernetes/pkg/client/clientset_generated/clientset"
 
 	"github.com/aveshagarwal/rescheduler/pkg/api"
 	"github.com/aveshagarwal/rescheduler/pkg/rescheduler/evictions"
 	podutil "github.com/aveshagarwal/rescheduler/pkg/rescheduler/pod"
 )
 
+// This strategy evicts all duplicate pods on node. The a pod is said to be a duplicate of other if
+// both of them have same name, kind and are within the same namespace. As of now, this strategy won't
```
aveshagarwal: this is incorrect, pods dont have same name but belong to the same creator. For example, pods created by a RS (test) would have names like test-<random-id-1>,  test-<random-id-2> etc. I think it redundant to say same namespace.

pkg/rescheduler/strategies/duplicates.go
```diff
@@ -21,24 +21,35 @@ import (
...
+// evict daemonsets, mirror pods and pods with local storages.
```
aveshagarwal: also need to include critical pods.

pkg/rescheduler/strategies/duplicates.go
```diff
@@ -21,24 +21,35 @@ import (
...
+
```
aveshagarwal: remove this space.

pkg/rescheduler/strategies/duplicates_test.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+}
+
+// getDaemonSetAnnotaiton returns the annotation needed for daemonset pod.
+func getDaemonSetAnnotation() map[string]string {
+	return map[string]string{
+		"kubernetes.io/created-by": "{\"kind\":\"SerializedReference\",\"apiVersion\":\"v1\",\"reference\":{\"kind\":\"DaemonSet\"}}",
+	}
+}
+
```
aveshagarwal: you could add one for critical pod too.

pkg/rescheduler/strategies/duplicates_test.go
```diff
@@ -0,0 +1,166 @@
...
+// getNormalPodAnnotation returns the annotation needed for a normal pod. A normal pod is one without any references to
+func getNormalPodAnnotation() map[string]string {
+	return map[string]string{
+		"kubernetes.io/created-by": "{\"kind\":\"SerializedReference\",\"apiVersion\":\"v1\",\"reference\":{\"kind\":\"Pod\"}}",
+	}
+}
+
+// getReplicaSetAnnotation returns the annotation needed for replicaset pod.
+func getReplicaSetAnnotation() map[string]string {
```
aveshagarwal: Its ok for time being. Later you could add for RS, Deployment and Job too.

aveshagarwal:
@ravisantoshgudimetla some minor comments. overall LGTM. thanks.

pkg/rescheduler/strategies/duplicates.go
ravisantoshgudimetla: Thanks for the correction. I will make the change.

pkg/rescheduler/strategies/duplicates.go
ravisantoshgudimetla: Do, we include critical pods? I see that the check is not there in the FindDuplicatePods function.

aveshagarwal: Somehow I missed that, now I have added a check for critical pods too.

ravisantoshgudimetla: Its fine @aveshagarwal, I will include it.

ravisantoshgudimetla:
@aveshagarwal Thanks, I updated the PR with the changes you suggested. Included a critical pod check and test as well.

pkg/rescheduler/strategies/duplicates_test.go
ravisantoshgudimetla: Sure, I am focusing on code coverage as of now.

pkg/rescheduler/strategies/duplicates_test.go
ravisantoshgudimetla: Done.

pkg/rescheduler/strategies/duplicates.go
```diff
@@ -21,24 +21,35 @@ import (
...
+	//TODO: Change to client-go instead of generated clientset.
 	"k8s.io/kubernetes/pkg/client/clientset_generated/clientset"
 
-	"github.com/aveshagarwal/rescheduler/pkg/api"
 	"github.com/aveshagarwal/rescheduler/pkg/rescheduler/evictions"
 	podutil "github.com/aveshagarwal/rescheduler/pkg/rescheduler/pod"
+	"github.com/aveshagarwal/rescheduler/pkg/api"
 )
 
+// This strategy evicts all duplicate pods on node. The a pod is said to be a duplicate of other if
```
aveshagarwal: typos, and move it before func RemoveDuplicatePods.

pkg/rescheduler/strategies/duplicates_test.go
```diff
@@ -0,0 +1,177 @@
...
+	p5 := buildTestPod("p5", 100, 0, node.Name)
+	p6 := buildTestPod("p6", 100, 0, node.Name)
+	p7 := buildTestPod("p7", 100, 0, node.Name)
+
+	// All the following pods expect for one will be evicted.
+	p1.Annotations = getReplicaSetAnnotation()
+	p2.Annotations = getReplicaSetAnnotation()
+	p3.Annotations = getReplicaSetAnnotation()
+
+	// The following 3 pods won't get evicted.
```
aveshagarwal: Now 4 after inclusion of a critical pod too.

ravisantoshgudimetla:
Done @aveshagarwal. PTAL.

pkg/rescheduler/strategies/duplicates.go
```diff
@@ -47,33 +57,37 @@ func RemoveDuplicatePods(client clientset.Interface, strategy api.ReschedulerStr
...
 func FindDuplicatePods(pods []*v1.Pod) DuplicatePodsMap {
 	dpm := DuplicatePodsMap{}
-
 	for _, pod := range pods {
 		sr, err := podutil.CreatorRef(pod)
 		if err != nil || sr == nil {
 			continue
 		}
-		if podutil.IsMirrorPod(pod) || podutil.IsDaemonsetPod(sr) || podutil.IsPodWithLocalStorage(pod) {
+		if podutil.IsMirrorPod(pod) || podutil.IsDaemonsetPod(sr) || podutil.IsPodWithLocalStorage(pod) || podutil.IsCriticalPod(pod) {
```
aveshagarwal: Its failing merge because of this, can you rebase?

ravisantoshgudimetla: I see that there are no conflicts atleast in the github UI?

aveshagarwal:
Not sure if there is bug in github or what, I dont know. I will just merge it now and see what happens.

PR #4, opened 2017-09-02
Changes and file reorganization

ravisantoshgudimetla:
/cc @aveshagarwal. I wasn't sure about the node ready condition(https://github.com/aveshagarwal/rescheduler/blob/master/pkg/rescheduler/node/node.go#L80). For the rest, I refactored and reorganized functions. Please let me know your feedback.

### DIFF
--- a/pkg/rescheduler/node/node_test.go
+++ b/pkg/rescheduler/node/node_test.go
@@ -1,0 +1,82 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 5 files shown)


aveshagarwal:
@ravisantoshgudimetla the reason I had selected nodes with disk or memory pressure for consideration 
 because it is good to evict pods from them to alleviate their problems.

aveshagarwal:
@ravisantoshgudimetla could you discard changes in pkg/rescheduler/node/node.go in these commits?

aveshagarwal:
@ravisantoshgudimetla could you fix it soon? I did not realize that your previous PR was dependent upon this PR, and now its breaking unit tests.

ravisantoshgudimetla:
@aveshagarwal Done. PTAL.

PR #5, opened 2017-09-06
Evictions test case and small modification

ravisantoshgudimetla:
@aveshagarwal I believe this completes the unit tests. We have 3 more files without unit tests  which are client.go, policyconfig.go, rescheduler.go and I think none of them needs unit tests. Please let me know what you think. I think, we can add a travis.yaml file and code coverage badge from covers.io or someother code coverage tool.

### DIFF
--- a/pkg/rescheduler/evictions/evictions_test.go
+++ b/pkg/rescheduler/evictions/evictions_test.go
@@ -1,0 +1,40 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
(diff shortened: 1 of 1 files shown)


pkg/rescheduler/evictions/evictions.go
```diff
@@ -49,7 +49,7 @@ func EvictPod(client clientset.Interface, pod *v1.Pod, policyGroupVersion string
 	} else if apierrors.IsTooManyRequests(err) {
 		return false, fmt.Errorf("error when evicting pod (ignoring) %q: %v", pod.Name, err)
 	} else if apierrors.IsNotFound(err) {
-		return true, fmt.Errorf("pod not found when evicting %q: %v", pod.Name, err)
+		return false, fmt.Errorf("pod not found when evicting %q: %v", pod.Name, err)
```
aveshagarwal: please revert it, if the pod is not found, it is considered successful.

ravisantoshgudimetla: This is where I was confused, sure, I can revert but shouldn't it throw an error? Is this coming from k8s style?

resouer:
@ravisantoshgudimetla  Is this descheduler is the same thing we talked as `rescheduler` in sig-scheduling before?

aveshagarwal:
@resouer yes, rescheduler as discussed previously has been renamed to descheduler.  Also since this PR is already merged, i'd appreciate that for further discussion, we dont use it. thanks.

PR #10, opened 2017-09-13
Descheduler related modifications.

aveshagarwal:


aveshagarwal:
Did local testing and it works now.

PR #14, opened 2017-09-25
Fix travis yaml file.

aveshagarwal:
@ravisantoshgudimetla can you check if these changes are ok? Also the changes you had are really needed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,18 +8,6 @@ language: go
 go:
   - 1.8.3
 
-before_install:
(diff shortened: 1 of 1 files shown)


ravisantoshgudimetla:
@aveshagarwal I am fine with these changes. Do you want have unit tests & e2e's combined in the same "make test" command or want to split them? If you want to have them combined, this should be fine.

aveshagarwal:
@ravisantoshgudimetla I'd prefer separate: basicallt make test only for unit/integration and make e2e (or something similar) for e2e.

PR #19, opened 2017-10-03
Conversion to glog

ravisantoshgudimetla:

Conversion to glog.

cc @aveshagarwal.

### DIFF
--- a/cmd/descheduler/app/server.go
+++ b/cmd/descheduler/app/server.go
@@ -19,31 +19,40 @@ package app
 
 import (
 	"fmt"
+	"flag"
+	"io"
 
 	"github.com/kubernetes-incubator/descheduler/cmd/descheduler/app/options"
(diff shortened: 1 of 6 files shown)


aveshagarwal:
@ravisantoshgudimetla can you resolve conflicts and resubmit?

ravisantoshgudimetla:
Done. PTAL.

aveshagarwal:
@ravisantoshgudimetla thanks. why are all Infof in this PR? Can your further categorize them between errorf, warn, info (and these info as 0,1, 2 levels) as discussed before?

ravisantoshgudimetla:
So, most of the information we are having in fmt.Print seem to be just normal print statements. I thought, we agreed to have level 0 as it is good enough for printing information.

ravisantoshgudimetla:
Also, we need to pass --logtostderr=false when running descheduler or else logging would be redirected to stdout. Should I make this default?

PR #23, opened 2017-10-06
Modified permissions for gofmt

ravisantoshgudimetla:
@aveshagarwal Forgot to change permissions. Just checked it and changing it to 644.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,5 +2,6 @@ language: go
 go:
   - 1.8.3
 script:
+- hack/verify-gofmt.sh
 - make build
 - make test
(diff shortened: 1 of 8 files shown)


aveshagarwal:
why? 644 is incorrect. I dont see any issue with the current 755.

ravisantoshgudimetla:
I believe we need to have 744(since we are going to execute, apologies), should we allow others to read + execute? I believe its not needed.

ravisantoshgudimetla:
I have accidentally close the PR. I will create one more branch and raise a PR, in case you are fine with @aveshagarwal

aveshagarwal:
giving execute to others for this script is not an issue as it just verifies but does not change anything.

ravisantoshgudimetla:
Ok, I am fine too then.

Issue #24, opened 2017-10-09
Create a version for descheduler


ravisantoshgudimetla:
As of now, we don't have version flag for descheduler. We need to have one. I will submit a PR for this.

cc @aveshagarwal

ravisantoshgudimetla:
#25 merged.

PR #25, opened 2017-10-09
Introducing versioning in descheduler

ravisantoshgudimetla:

/cc @aveshagarwal. For versioning.

### DIFF
--- a/cmd/descheduler/app/version.go
+++ b/cmd/descheduler/app/version.go
@@ -1,0 +1,86 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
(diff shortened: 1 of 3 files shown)


ravisantoshgudimetla:
Addresses #24.

cmd/descheduler/app/version.go
```diff
@@ -0,0 +1,37 @@
...
+
+package app
+
+import (
+	"fmt"
+	"github.com/spf13/cobra"
+)
+
+// This becomes easy for exporting later. In fact, this could be moved to top level directory.
+const DeschedulerVersion = "0.1.0"
```
aveshagarwal: Instead of hard coding, could you check how to print something like, for example, last tag+commits since last tag or something similar. You could also see how it is done in kube/openshift, and have something similar.

ravisantoshgudimetla: Sure, I initially wanted to do that, looking at the kube code but thought it was unnecessary at this stage but I can.

aveshagarwal: actually hard coding 0.1.0 might give wrong information even at this stage, and also it would require updating whenever we have a new release with a new tag.

ravisantoshgudimetla:
/cc @aveshagarwal Addressed the comments. Now, the output would be similar to kubectl version. Sample output:

$_output/bin/descheduler version
`Descheduler version {Major:0 Minor:2+ GitCommit:9f448ce12756bb0b9d71d82c50b04e66a91237df GitVersion:v0.2.0-2-g9f448ce BuildDate:2017-10-24T11:07:42-0400 GoVersion:go1.8.3 Compiler:gc Platform:linux/amd64}`

PR #26, opened 2017-10-10
Update readme with description to run descheduler as a job inside a pod.

aveshagarwal:
@ravisantoshgudimetla in case if you have any feedback.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,109 @@ For more information about available options run:
 $ ./_output/bin/descheduler --help
 ```
 
+## Running Descheduler as a Job Inside of a Pod
+
+Descheduler can be run as a job inside of a pod. It has the advantage of
+being able to be run multiple times without needing user intervention.
+Descheduler pod is run as a critical pod to avoid being evicted by itself,
+or by kubelet due to an eviction event. Since critical pods are created in
+`kube-system` namespace, descheduler job and its pod will also be created
+in `kube-system` namespace.
+
+###  Create a container image
+
+First we create a simple Docker image utilizing the Dockerfile found in the root directory:
+
+```
+$ make image
+```
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -40,6 +40,109 @@ For more information about available options run:
 $ ./_output/bin/descheduler --help
 ```
 
+## Running Descheduler as a Job Inside of a Pod
+
+Descheduler can be run as a job inside of a pod. It has the advantage of
+being able to be run multiple times without needing user intervention.
+Descheduler pod is run as a critical pod to avoid being evicted by itself,
+or by kubelet due to an eviction event. Since critical pods are created in
+`kube-system` namespace, Descheduler job and its pod will also be created
```
ravisantoshgudimetla: s/Descheduler/descheduler

README.md
```diff
@@ -40,6 +40,109 @@ For more information about available options run:
 $ ./_output/bin/descheduler --help
 ```
 
+## Running Descheduler as a Job Inside of a Pod
```
ravisantoshgudimetla: Can we change the name to incluster configuration for descheduler?

README.md
```diff
@@ -40,6 +40,109 @@ For more information about available options run:
...
+
+First we create a simple Docker image utilizing the Dockerfile found in the root directory:
+
+```
+$ make image
+```
+
+### Create a cluster role
+
+To give necessary permissions to the descheduler pod, create a cluster role:
```
ravisantoshgudimetla: s/descheduler pod/for descheduler to work in a pod/ ?

README.md
aveshagarwal: i think many users might not be familar with the term incluster configuration.

PR #31, opened 2017-10-30
E2E using gce first pass

ravisantoshgudimetla:
@aveshagarwal I am able to create a 3 node cluster in gce with pod being evicted for lownodeutilization case.  Please review when you find time. I believe, travis ci might complain because of encrypted file(client.json.enc) which will probably not be decrypted on this repo. 

Some changes:

- Now instead of running unit-tests, travis will run unit-tests + e2e's when using `make test`.
- Before running `make test`, we need to ensure that gcloud_create_cluster.sh has been run. I have included that in the run-tests.sh(file renamed from run-unittests.sh)

Sample Output after running e2e:

`evicting pods from node %#v with usage: %#v descheduler-b148f4c6-bd94-11e7-8e84-c85b76333877 map[cpu:2264 memory:263.78542244169597 pods:81.81818181818181]`

### DIFF
--- a/hack/e2e-gce/gcloud_create_cluster.sh
+++ b/hack/e2e-gce/gcloud_create_cluster.sh
@@ -1,0 +1,95 @@
+#!/bin/bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+echo "Make sure that uuid package is installed"
+
+master_uuid=$(uuid)
+node1_uuid=$(uuid)
+node2_uuid=$(uuid)
+kube_apiserver_port=6443
+kube_version=1.9.4
+
+DESCHEDULER_ROOT=$(dirname "${BASH_SOURCE}")/../../
+E2E_GCE_HOME=$DESCHEDULER_ROOT/hack/e2e-gce
+
+
+create_cluster() {
+	echo "#################### Creating instances ##########################"
+        gcloud compute instances create descheduler-$master_uuid --image="ubuntu-1604-xenial-v20180306" --image-project="ubuntu-os-cloud" --zone=us-east1-b
+	# Keeping the --zone here so as to make sure that e2e's can run locally.
+	echo "gcloud compute instances delete descheduler-$master_uuid --zone=us-east1-b --quiet" > $E2E_GCE_HOME/delete_cluster.sh
+       
+	gcloud compute instances create descheduler-$node1_uuid --image="ubuntu-1604-xenial-v20180306" --image-project="ubuntu-os-cloud" --zone=us-east1-b
+	echo "gcloud compute instances delete descheduler-$node1_uuid --zone=us-east1-b --quiet" >> $E2E_GCE_HOME/delete_cluster.sh
+
+	gcloud compute instances create descheduler-$node2_uuid --image="ubuntu-1604-xenial-v20180306" --image-project="ubuntu-os-cloud" --zone=us-east1-b
+	echo "gcloud compute instances delete descheduler-$node2_uuid --zone=us-east1-b --quiet" >> $E2E_GCE_HOME/delete_cluster.sh
+
+	# Delete the firewall port created for master.
+	echo "gcloud compute firewall-rules delete kubeapiserver-$master_uuid --quiet" >> $E2E_GCE_HOME/delete_cluster.sh
+	chmod 755 $E2E_GCE_HOME/delete_cluster.sh
+}
+
+
+generate_kubeadm_instance_files() {
+	# TODO: Check if they have come up. awk $6 contains the state(RUNNING or not).
+	master_public_ip=$(gcloud compute instances list | grep $master_uuid|awk '{print $5}')
+	node1_public_ip=$(gcloud compute instances list | grep $node1_uuid|awk '{print $5}')
+	node2_public_ip=$(gcloud compute instances list | grep $node2_uuid|awk '{print $5}')
+	echo "kubeadm init --kubernetes-version=${kube_version} --apiserver-advertise-address=${master_public_ip}" --skip-preflight-checks --pod-network-cidr=10.96.0.0/12 > $E2E_GCE_HOME/kubeadm_install.sh
+}
+
+
+transfer_install_files() {
+	gcloud compute scp  $E2E_GCE_HOME/kubeadm_preinstall.sh descheduler-$master_uuid:/tmp --zone=us-east1-b
+	gcloud compute scp $E2E_GCE_HOME/kubeadm_install.sh descheduler-$master_uuid:/tmp --zone=us-east1-b
+	gcloud compute scp  $E2E_GCE_HOME/kubeadm_preinstall.sh descheduler-$node1_uuid:/tmp --zone=us-east1-b
+	gcloud compute scp  $E2E_GCE_HOME/kubeadm_preinstall.sh descheduler-$node2_uuid:/tmp --zone=us-east1-b
+}
+
+
+install_kube() {
+	# Docker installation.
+	gcloud compute ssh descheduler-$master_uuid --command "sudo apt-get update; sudo apt-get install -y docker.io" --zone=us-east1-b
+	gcloud compute ssh descheduler-$node1_uuid --command "sudo apt-get update; sudo apt-get install -y docker.io" --zone=us-east1-b
+	gcloud compute ssh descheduler-$node2_uuid --command "sudo apt-get update; sudo apt-get install -y docker.io" --zone=us-east1-b
+	# kubeadm installation.
+	# 1. Transfer files to master, nodes.
+	transfer_install_files
+	# 2. Install kubeadm.
+	#TODO: Add rm /tmp/kubeadm_install.sh
+	# Open port for kube API server	
+	gcloud compute firewall-rules create kubeapiserver-$master_uuid --allow tcp:6443 --source-tags=descheduler-$master_uuid  --source-ranges=0.0.0.0/0 --description="Opening api server port" 
+
+	gcloud compute ssh descheduler-$master_uuid --command "sudo chmod 755 /tmp/kubeadm_preinstall.sh; sudo /tmp/kubeadm_preinstall.sh" --zone=us-east1-b
+	kubeadm_join_command=$(gcloud compute ssh descheduler-$master_uuid --command "sudo chmod 755 /tmp/kubeadm_install.sh; sudo /tmp/kubeadm_install.sh" --zone=us-east1-b|grep 'kubeadm join')
+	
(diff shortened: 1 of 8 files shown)


aveshagarwal:
@ravisantoshgudimetla thanks ravi, will look into it soon.

aveshagarwal:
ok to test

ravisantoshgudimetla:
ok to test

aveshagarwal:

ravisantoshgudimetla:
ok to test

ravisantoshgudimetla:
retest this please

ravisantoshgudimetla:
@aveshagarwal These are the latest changes:

- test suite has been split into 2. Travis CI runs unit tests alone and Jenkins runs e2e's.
- Makefile has been changed to reflect the new changes.
- instead of `make test`, we should run `make test-unit` to run unit tests, `make test-e2e`. The e2e framework uses GCE to create a cluster.
- github vendoring has been changed to reflect it.

test/e2e/e2e_test.go
```diff
@@ -0,0 +1,150 @@
...
+// waitForPodToScheduleWithTimeout waits for a pod to get scheduled and returns
+// an error if it does not scheduled within the given timeout.
+func waitForPodToScheduleWithTimeout(cs clientset.Interface, pod *v1.Pod, timeout time.Duration) error {
+	return wait.Poll(time.Second, timeout, podScheduled(cs, pod.Namespace, pod.Name))
+}
+
+func TestE2E(t *testing.T) {
+	// If we have reached here, it means cluster would have been already setup and the kubeconfig file should
+	// be in /tmp directory.
+	clientSet, _ := client.CreateClient("/tmp/admin.conf")
```
ingvagabund:
```go
clientSet, err := client.CreateClient("/tmp/admin.conf")
if err != nil {
	panic(err)
}
```

please so the test does not panic with nil dereference but with `Failed to parse kubeconfig file: open /tmp/admin.conf: no such file or directory` error message

test/e2e/e2e_test.go
```diff
@@ -0,0 +1,150 @@
...
+	nodeList, _ := clientSet.Core().Nodes().List(metav1.ListOptions{})
```
ingvagabund: The same here

test/e2e/e2e_test.go
```diff
@@ -0,0 +1,150 @@
...
+	podList := generatePods(clientSet, 4)
+	// This is to ensure that all pods got scheduled.
+	for _, pod := range podList {
+		waitForPodToScheduleWithTimeout(clientSet, pod, time.Second)
+	}
+	// Marking the node which was unschedulable earlier to schedulable again, so that pods
+	// could be moved to this node.
+	for _, node := range nodeList.Items {
+		if len(node.Spec.Taints) == 0 && node.Spec.Unschedulable == true {
+			node.Spec.Unschedulable = false
```
ingvagabund: Should not the code call Patch at least so the change takes effect in the cluster?

test/e2e/e2e_test.go
```diff
@@ -0,0 +1,150 @@
...
+func TestE2E(t *testing.T) {
+	// If we have reached here, it means cluster would have been already setup and the kubeconfig file should
+	// be in /tmp directory.
+	clientSet, _ := client.CreateClient("/tmp/admin.conf")
+	nodeList, _ := clientSet.Core().Nodes().List(metav1.ListOptions{})
+	// We may have a 3 node cluster by now. Let us mark one node as unschedulable so that, we can actually
+	// make sure that one node is heavily utilized.
+	for _, node := range nodeList.Items {
+		if len(node.Spec.Taints) == 0 && node.Spec.Unschedulable == false {
+			node.Spec.Unschedulable = true
```
ingvagabund: Should not the code call Patch at least so the change takes effect in the cluster?

test/e2e/e2e_test.go
ravisantoshgudimetla: AFAIK no, because we are partially updating the node object in the client side and we are operating on the node list again on the client side.

ravisantoshgudimetla:
So, the e2e is working on 1.9.3 cluster @ingvagabund @aveshagarwal. Let me know, if you have any other concerns.

test/e2e/e2e_test.go
```diff
@@ -0,0 +1,136 @@
...
+		t.Errorf("Error while listing nodes with %v", err)
+	}
+	podsToBeCreated := 4
+	podList := generatePods(clientSet, podsToBeCreated)
+	// This is to ensure that all pods got scheduled.
+	for _, pod := range podList {
+		waitForPodToScheduleWithTimeout(clientSet, pod, time.Second)
+	}
+
+	startEndToEndForLowNodeUtilization(clientSet)
```
ingvagabund: So the pods are now scheduled without any limitation. How can it test the descheduler functionality now? My original understanding was to make one node unschedulable (so it is pod-free), schedule 4 pods to the remaining two nodes and then set the descheduling policies the way at least one of the scheduled pods must be de-scheduled. Is that no longer the intended behavior?

test/e2e/e2e_test.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+func TestE2E(t *testing.T) {
+	// If we have reached here, it means cluster would have been already setup and the kubeconfig file should
+	// be in /tmp directory.
+	clientSet, err := client.CreateClient("/tmp/admin.conf")
+	if err != nil {
+		t.Errorf("Error during client creation with %v", err)
+	}
+	if err != nil {
```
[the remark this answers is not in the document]
ravisantoshgudimetla: Ahh, thanks for catching it. I think don't need that any more.

test/e2e/e2e_test.go
ravisantoshgudimetla: So, I thought about that yesterday. The problem here is since we are not simulating cluster, rather creating pods on existing cluster, depending on the size of node, the thresholds may vary making these tests unreliable if we rely on number of pods being evicted. So, I thought of removing the checks and in the logs of tests, since I added a -v flag, we could see if atleast one pod got evicted.

ingvagabund:
WIP JJ: https://ci.openshift.redhat.com/jenkins/job/ci-kubernetes-descheduler-e2e-gce/
WIP JJ PR: https://github.com/openshift/aos-cd-jobs/pull/1232
Prow configs updated: https://github.com/openshift/release/pull/683

aveshagarwal:
@ravisantoshgudimetla this PR updates vendor dir, but where are changes to glide.lock file?

ravisantoshgudimetla:
@aveshagarwal I removed the dependency on those vendored files. I copied the function which I need to the test so it should run independently.

hack/e2e-gce/gcloud_create_cluster.sh
```diff
@@ -0,0 +1,95 @@
...
+	node2_public_ip=$(gcloud compute instances list | grep $node2_uuid|awk '{print $5}')
+	echo "kubeadm init --kubernetes-version=${kube_version} --apiserver-advertise-address=${master_public_ip}" --skip-preflight-checks> $E2E_GCE_HOME/kubeadm_install.sh
+}
+
+
+transfer_install_files() {
+	gcloud compute copy-files  $E2E_GCE_HOME/kubeadm_preinstall.sh descheduler-$master_uuid:/tmp --zone=us-east1-b
+	gcloud compute copy-files $E2E_GCE_HOME/kubeadm_install.sh descheduler-$master_uuid:/tmp --zone=us-east1-b
+	gcloud compute copy-files  $E2E_GCE_HOME/kubeadm_preinstall.sh descheduler-$node1_uuid:/tmp --zone=us-east1-b
+	gcloud compute copy-files  $E2E_GCE_HOME/kubeadm_preinstall.sh descheduler-$node2_uuid:/tmp --zone=us-east1-b
```
ingvagabund: WARNING: `gcloud compute copy-files` is deprecated.  Please use `gcloud compute scp` instead.  Note that `gcloud compute scp` does not have recursive copy on by default.  To turn on recursion, use the `--recurse` flag.

ravisantoshgudimetla: Thanks for catching it. These were not deprecated back when I wrote them back in October. I can add a TODO or update them.

ravisantoshgudimetla: Done!

ingvagabund:
CI job merged.

aveshagarwal:
@ingvagabund @ravisantoshgudimetla  is it ready for merge?

test/e2e/e2e_test.go
```diff
@@ -0,0 +1,154 @@
...
+	// be in /tmp directory.
+	clientSet, err := client.CreateClient("/tmp/admin.conf")
+	if err != nil {
+		t.Errorf("Error during client creation with %v", err)
+	}
+	nodeList, err := clientSet.Core().Nodes().List(metav1.ListOptions{})
+	if err != nil {
+		t.Errorf("Error listing node with %v", err)
+	}
+	leastLoadedNode := nodeList.Items[2]
```
ingvagabund: Nit: We should check if the cluster has at least 3 nodes. Otherwise this will panic.

Btw. how do you the third node in the list is actually the least loaded node?

hack/e2e-gce/gcloud_sdk_configure.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
ravisantoshgudimetla: This is needed if someone wants to run e2e using GCE from local system. This is similar to kube-up.sh in upstream.

test/e2e/e2e_test.go
ravisantoshgudimetla: > Nit: We should check if the cluster has at least 3 nodes. Otherwise this will panic.

We would have failed in gcloud_cluster_create.sh if we don't have 3 nodes. I thought failing early is better. 


> Btw. how do you the third node in the list is actually the least loaded node?

I have added a comment here.

https://github.com/kubernetes-incubator/descheduler/pull/31/files/e0d3b2663509dde3266daed9ecd0867f6fc4fdb0#diff-6413e1dbafd6dd63b5d907ea3ae6f6f9R151

ingvagabund: Ok, let's put the comment here. Smth like:
```
// Assumption: kubeadm brings all the master components onto master node. So, the last node would have least utilization.
```

ravisantoshgudimetla: Sure.

PR #33, opened 2017-11-09
 Implement node selectors to retrieve node list based on provided query.

aveshagarwal:


aveshagarwal:
@ravisantoshgudimetla let me know if you have any feedback.

cmd/descheduler/app/options/options.go
```diff
@@ -53,4 +53,5 @@ func (rs *DeschedulerServer) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&rs.KubeconfigFile, "kubeconfig-file", rs.KubeconfigFile, "File with  kube configuration.")
 	fs.StringVar(&rs.PolicyConfigFile, "policy-config-file", rs.PolicyConfigFile, "File with descheduler policy configuration.")
 	fs.BoolVar(&rs.DryRun, "dry-run", rs.DryRun, "execute descheduler in dry run mode.")
+	fs.StringVar(&rs.NodeSelector, "node-selector", rs.NodeSelector, "Selector (label query) to filter on, supports '=', '==', and '!='.(e.g. -l key1=value1,key2=value2)")
```
ravisantoshgudimetla: Can you add a comment, which tells how we intend to use this flag like "Descheduler would be targeted on these node which are chosen from this selector, if not specified it will be run on all the nodes which could be potentially dangerous"

aveshagarwal: sure.

pkg/descheduler/node/node.go
```diff
@@ -31,16 +31,23 @@ import (
 
 // ReadyNodes returns ready nodes irrespective of whether they are
 // schedulable or not.
-func ReadyNodes(client clientset.Interface, stopChannel <-chan struct{}) ([]*v1.Node, error) {
+func ReadyNodes(client clientset.Interface, nodeSelector string, stopChannel <-chan struct{}) ([]*v1.Node, error) {
 	nl := GetNodeLister(client, stopChannel)
-	nodes, err := nl.List(labels.Everything())
+
```
ravisantoshgudimetla: Can you add a unit test to this function?

ravisantoshgudimetla:
@aveshagarwal Overall LGTM, just some minor comments, wrt documentation and unit test, once you have them, we can merge this.

pkg/descheduler/node/node.go
aveshagarwal: yes will add unit tests soon.

aveshagarwal:
@ravisantoshgudimetla updated PR based on your feedback.

aveshagarwal:
merging as tests passed.

pkg/descheduler/node/node_test.go
```diff
@@ -80,3 +80,20 @@ func TestReadyNodes(t *testing.T) {
...
+	node2 := test.BuildTestNode("node2", 1000, 2000, 9)
+	node2.Labels = map[string]string{"type": "infra"}
+
+	fakeClient := fake.NewSimpleClientset(node1, node2)
+
+	nodeSelector := "type=compute"
+	stopChannel := make(chan struct{})
+	nodes, _ := ReadyNodes(fakeClient, nodeSelector, stopChannel)
+
+	if nodes[0].Name != "node1" {
```
ravisantoshgudimetla: Probably ` && len(nodes) ==1 ` would make it clear.

PR #34, opened 2017-11-10
Remove unused test code.

aveshagarwal:
@ravisantoshgudimetla can you check why you had this code as it was unused as I tested?

### DIFF
--- a/pkg/descheduler/node/node_test.go
+++ b/pkg/descheduler/node/node_test.go
@@ -17,18 +17,14 @@ limitations under the License.
 package node
 
 import (
-	"fmt"
 	"testing"
 
 	"github.com/kubernetes-incubator/descheduler/test"
-	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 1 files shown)


pkg/descheduler/node/node_test.go
```diff
@@ -41,25 +37,6 @@ func TestReadyNodes(t *testing.T) {
 	node6 := test.BuildTestNode("node6", 1000, 2000, 9)
 	node6.Status.Conditions = []v1.NodeCondition{{Type: v1.NodeReady, Status: v1.ConditionFalse}}
 
-	fakeClient.Fake.AddReactor("get", "nodes", func(action core.Action) (bool, runtime.Object, error) {
```
ravisantoshgudimetla: So, this is needed once we move to table driven tests, we will have a single IsReady(node) in a loop after returning list and then let it return either true are false. I believe adding a TODO would have been good previously. If you are not ok with it, we can delete it as well and add it later.

PR #35, opened 2017-11-10
Fix to not process empty node list and also fix error reporting in other places.

aveshagarwal:
@ravisantoshgudimetla


pkg/descheduler/node/node.go
```diff
@@ -32,26 +32,32 @@ import (
...
+	if nl != nil {
+		// err is defined above
+		if nodes, err = nl.List(ns); err != nil {
+			return []*v1.Node{}, err
+		}
 	}
 
 	if len(nodes) == 0 {
-		var err error
+		glog.V(2).Infof("node lister returned empty list, now fetch directly")
```
ravisantoshgudimetla: Any specific reason for choosing V(2)? I thought, we have fixed on 1 & 0 as levels.

ravisantoshgudimetla:
This looks good @aveshagarwal

PR #36, opened 2017-11-21
kubeconfig file correction

ravisantoshgudimetla:
kubeconfig file flag change.


ravisantoshgudimetla:
I believe just for sake of consistency with other components in kubernetes ecosystem, this PR should go in ahead of #37, @aveshagarwal @containscafeine WDYT?

aveshagarwal:
Not sure why I had `kubeconfig-file`, but @ravisantoshgudimetla  if you have verified that `kubeconfig` is everywhere instead of  `kubeconfig-file`, lets go ahead with this, and close the other ones.

Issue #42, opened 2017-11-25
Nodes with scheduling disabled should not be taken into consideration for LowNodeUtilization


containscafeine:
I have the following nodes -

```console
$ kubectl get nodes
NAME                           STATUS                     ROLES     AGE       VERSION
kubernetes-master              Ready,SchedulingDisabled   <none>    56m       v1.8.4-dirty
kubernetes-minion-group-5rrh   Ready                      <none>    56m       v1.8.4-dirty
kubernetes-minion-group-fb8c   Ready                      <none>    56m       v1.8.4-dirty
kubernetes-minion-group-t1r3   Ready,SchedulingDisabled   <none>    56m       v1.8.4-dirty
```

The worker node `kubernetes-minion-group-t1r3` was cordoned and marked as unschedulable, however it fulfilled the criteria for being an underutilized node according to the following policy file -

```yaml
apiVersion: "descheduler/v1alpha1"
kind: "DeschedulerPolicy"
strategies:
  "LowNodeUtilization":
     enabled: true
     params:
       nodeResourceUtilizationThresholds:
         thresholds:  # any node below the following percentages is considered underutilized
           "cpu" : 40
           "memory": 40
           "pods": 40
         targetThresholds: # any node above the following percentages is considered overutilized
           "cpu" : 30
           "memory": 2
           "pods": 1
```

When I ran the descheduler, `kubernetes-minion-group-t1r3` (the cordoned node) was taken into account and marked as underutilized and multiple pods were evicted from other nodes in the hope that the scheduler will schedule on `kubernetes-minion-group-t1r3`, but that never happened since the node was cordoned.

Does it make sense to not take a cordoned node into consideration while looking for underutilized nodes?

I ran the descheduler like the following -

```console
$ _output/bin/descheduler --kubeconfig-file /var/run/kubernetes/admin.kubeconfig --policy-config-file examples/custom.yaml -v 5 
I1125 18:58:46.014381    2813 reflector.go:198] Starting reflector *v1.Node (1h0m0s) from github.com/kubernetes-incubator/descheduler/pkg/descheduler/node/node.go:83
I1125 18:58:46.016167    2813 node.go:50] node lister returned empty list, now fetch directly
I1125 18:58:46.017010    2813 reflector.go:236] Listing and watching *v1.Node from github.com/kubernetes-incubator/descheduler/pkg/descheduler/node/node.go:83
[...]

PR #43, opened 2017-11-25
Ignore unschedulable nodes for eviction decisions

containscafeine:
This commit makes descheduler ignore the unschedulable nodes from
being considered for taking eviction decisions

### DIFF
--- a/pkg/descheduler/node/node.go
+++ b/pkg/descheduler/node/node.go
@@ -105,9 +105,9 @@ func IsReady(node *v1.Node) bool {
 		}*/
 	}
 	// Ignore nodes that are marked unschedulable
-	/*if node.Spec.Unschedulable {
+	if node.Spec.Unschedulable {
 		glog.V(4).Infof("Ignoring node %v since it is unschedulable", node.Name)
 		return false
-	}*/
+	}
 	return true
 }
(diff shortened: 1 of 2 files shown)


aveshagarwal:

ravisantoshgudimetla:
@containscafeine  You need to fix the test case and also please look at the new flag node-selector https://github.com/kubernetes-incubator/descheduler/pull/33, through which you can ignore certain nodes based on labels. 

Also, there is a plan in upstream community to replace UnSchedulable field with unschedulable  taint and in descheduler we plan to have taint based descheduling.

containscafeine:
@ravisantoshgudimetla ack, fixed the failing test.

Does it make sense that till the time spec.unschedulable is removed, we use it in descheduler?

ravisantoshgudimetla:
@containscafeine 

> Does it make sense that till the time spec.unschedulable is removed, we use it in descheduler?

A `readynode` in descheduler's context means we have a node on which descheduler could run. Even if the node is unschedulable, it could have some pods running(unless it has been drained) which can cause it to be over `highUtilization threshold`. In this  case, descheduler should run against it.

 IOW, descheduler is just a pod evictor. For eviction, we don't need to check if the node is unschedulable.

If you don't want node to be considered by descheduler, you should label the node and ask it to be excluded by descheduler.

aveshagarwal:
>A readynode in descheduler's context means we have a node on which descheduler could run. Even if the node is unschedulable, it could have some pods running(unless it has been drained) which can cause it to be over highUtilization threshold. In this case, descheduler should run against it.

Yes that was the intention to consider unschedulable but ready nodes. However, for LowNodeUtilization strategy, if unschedulable nodes are under utilized, we must ignore them as pods wont be scheduled on them.  IOW unschedulable nodes must only be considered for eviction if they are above target utilization but we must not consider unschedulable nodes if they are under low utilization threshold. If it is not the case, we should fix the LowNodeUtilization strategy.

ravisantoshgudimetla:
Ok after reading #42, I understood @containscafeine wants to do here but I believe the change in code should be in utilization strategy file. I will send a PR soon.

containscafeine:
Yay, thanks @ravisantoshgudimetla. Closing in favor of #45

PR #44, opened 2017-11-25
Add log messages for target, underutilized nodes

containscafeine:
This commit adds log messages to be printed whenever a node is
identified as a target node from which the pods can be evicted or
as an underutilized node to which the evicted pods can be
scheduled, if at all.

This should help with debugging as well as with letting the user
know the identified nodes.


aveshagarwal:

ravisantoshgudimetla:
Thanks @containscafeine.

PR #49, opened 2017-11-28
Small refactor to accomodate descheduling only evictable pods

### LINKED ISSUE #48: pod antiaffinity strategy evicts all pods.
Pod anti affinity strategy doesn't have a check on type of pod to be evicted. It can evict critical, mirror pods.  As this is some functionality that needs to be respected by all strategies in descheduler, I am planning to move this to pods.go to avoid code duplication so that people implementing strategies won't have to think about them.

ravisantoshgudimetla:
cc @aveshagarwal This is to ensure that all strategies will take into account only evictable pods. 

Fixes #48

### DIFF
--- a/pkg/descheduler/strategies/duplicates.go
+++ b/pkg/descheduler/strategies/duplicates.go
@@ -70,7 +70,7 @@ func deleteDuplicatePods(client clientset.Interface, policyGroupVersion string, 
 
 // ListDuplicatePodsOnANode lists duplicate pods on a given node.
 func ListDuplicatePodsOnANode(client clientset.Interface, node *v1.Node) DuplicatePodsMap {
-	pods, err := podutil.ListPodsOnANode(client, node)
+	pods, err := podutil.ListEvictablePodsOnNode(client, node)
 	if err != nil {
 		return nil
 	}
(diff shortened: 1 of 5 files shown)


pkg/descheduler/strategies/duplicates.go
```diff
@@ -85,9 +85,6 @@ func FindDuplicatePods(pods []*v1.Pod) DuplicatePodsMap {
 		if err != nil || sr == nil {
 			continue
 		}
```
aveshagarwal: remove this check also.

ravisantoshgudimetla: I need sr in the next line, so I had to include that.

aveshagarwal: but the check is not needed i think, because all pods with sr==nil or err != nil are already checked in ListEvictablePodsOnNode so no pod at this stage should have sr==nil or err != nil.

ravisantoshgudimetla: The reason is - I have to ignore err which is second argument and take sr only in order to remove this check. Will make the change now.

ravisantoshgudimetla:
Done @aveshagarwal. PTAL.

Issue #53, opened 2017-11-30
Enable profiling in descheduler


ravisantoshgudimetla:
/cc @aveshagarwal - As per our offline discussion, I think first step would be to enable profiling. I am planning to add flag(s) which enable profiling. I will try to avoid starting a httpserver based profiling in the initial stages.

aveshagarwal:
> I will try to avoid starting a httpserver based profiling in the initial stages.

sgtm.

PR #55, opened 2017-11-30
Introducing profiling for descheduler

ravisantoshgudimetla:
/cc @aveshagarwal . This is to introduce profiling. Please have a look when you find time. 

Fixes: #53

### DIFF
--- a/cmd/descheduler/app/server.go
+++ b/cmd/descheduler/app/server.go
@@ -29,6 +29,10 @@ import (
 
 	aflag "k8s.io/apiserver/pkg/util/flag"
 	"k8s.io/apiserver/pkg/util/logs"
+
+	"os"
+	"runtime/pprof"
+	"time"
 )
 
 // NewDeschedulerCommand creates a *cobra.Command object with default parameters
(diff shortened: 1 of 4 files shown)


aveshagarwal:
@ravisantoshgudimetla thanks.

cmd/descheduler/app/server.go
```diff
@@ -57,5 +61,20 @@ func NewDeschedulerCommand(out io.Writer) *cobra.Command {
...
+		// As of now, adding only CPU & memory profiles, we can include
+		// more profiles like blocking, trace, mutex. When we move to http
+		// based profiling all those profiles are automatically added.
+		cpuprofile, err := os.Create("cpu" + timeNow + ".pprof")
+		// The pprof file is of format cpu20171130152506.pprof
+		heapprofile, err := os.Create("heap" + timeNow + ".pprof")
+		if err != nil {
+			glog.Errorf("%v", err)
+		}
+		pprof.StartCPUProfile(cpuprofile)
```
aveshagarwal: can we create these files when running in a pod?

ravisantoshgudimetla: I didn't test in a pod, were you getting an error when running descheduler in a pod?

aveshagarwal: yes permission denied error.

PR #60, opened 2017-12-04
Introducing caching for serialized reference

ravisantoshgudimetla:
/cc @aveshagarwal - While working on duplicate pod strategy, I realized we need benchmarking of existing strategies. After looking at the call graphs it seems we are spending lot of time on serialization of json in CreatorRef function. So, I cached that in a map. This reduced the benchmarking time from 86sec to 56sec. 

Attaching the call graphs for reference.

Before: 
[cpu5.pdf](https://github.com/kubernetes-incubator/descheduler/files/1527815/cpu5.pdf)

After:
[cpu6.pdf](https://github.com/kubernetes-incubator/descheduler/files/1527817/cpu6.pdf)

### DIFF
--- a/pkg/descheduler/pod/pods.go
+++ b/pkg/descheduler/pod/pods.go
@@ -27,6 +27,13 @@ import (
 	"k8s.io/kubernetes/pkg/kubelet/types"
 )
 
+// Map to hold pod Name to serialized reference.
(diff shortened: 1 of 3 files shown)


ravisantoshgudimetla:
This can be closed as we are using ownerRef now.

PR #72, opened 2018-01-04
Owner ref switch

### LINKED ISSUE #52: RemoveDuplicates strategy is not working with kubernetes 1.9
Currently RemoveDuplicates evict pod according to `metadata.annotations.kubernetes.io/created-by`, but created-by is deprecated (https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.9.md)

https://github.com/kubernetes-incubator/descheduler/blob/master/pkg/descheduler/pod/pods.go#L113

ravisantoshgudimetla:
/cc @aveshagarwal  - Since creatorRef is removed in 1.9, switching to ownerRef now. This works with both 1.8 and 1.9, I have tested it against a running 1.9 cluster. Fixes #52

### DIFF
--- a/pkg/descheduler/pod/pods.go
+++ b/pkg/descheduler/pod/pods.go
@@ -19,7 +19,6 @@ package pod
 import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
-	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/kubernetes/pkg/api"
 	"k8s.io/kubernetes/pkg/api/v1"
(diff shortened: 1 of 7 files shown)


aveshagarwal:
@ravisantoshgudimetla thanks, can you please test against 1.7 too if possible?

ravisantoshgudimetla:
Sure, @aveshagarwal , I will set one up tomorrow.

ravisantoshgudimetla:
@aveshagarwal I just confirmed on 1.7 cluster. This change is working.

ravisantoshgudimetla:
@aveshagarwal Let me know, if you want this change to be backported to 0.1-0.3 releases. I think this should go in.

PR #56, opened 2017-12-01
add requiredDuringSchedulingIgnoredDuringExecution for RemovePodsViolatingNodeAffinity strategy

containscafeine:
Yikes, @aveshagarwal @ravisantoshgudimetla, honestly, I've really not thought this through. Currently, it's more of a hack to just get it barely working. The code is bad, and there are no tests, but if you pass a YAML like the following, then it works.

How to test -
1. label a node x=y
2. schedule pods on it using node affinity as requiredDuringSchedulingIgnoredDuringExecution
3. remove label from the node and set it to another node 
4. run the descheduler, and if the new node is schedulable, it evicts the pods

```yaml
apiVersion: "descheduler/v1alpha1"
kind: "DeschedulerPolicy"
strategies:
  "RemovePodsViolatingNodeAffinity":
    enabled: true
    params:
      nodeAffinityType:
      - "requiredDuringSchedulingIgnoredDuringExecution"
```

Would love the comments on design for this new strategy.

To Do:
- [ ] Cleanup and abstract parts of code
- [ ] Add unit tests, maybe e2e can go in a separate PR
- [ ] Add documentation

P. S. I'm away for most of next week, so will get back to this once I'm back :tada:

PR #74, opened 2018-01-11
Update vendor directory using glide

containscafeine:
This commit is a direct result of running `glide update -v` on the
current master.


aveshagarwal:

containscafeine:
So, I got the point that updating the vendor for https://github.com/kubernetes-incubator/descheduler/pull/56 will be a mess, however the functions that I need from predicates package are available in the current vendor, but upon using them, the code not compile. However, it works fine when I do a `glide update -v` on the current master, I think some missing dependencies are satisfied by doing this.

There are 2 ways I can proceed with #56, one is copying the functions from predicates, another is by just updating the vendor/ on current master.

This is different from what I did in #56 previously, before I had modified glide.yml, while in this commit I've simply run `glide update -v`.

If updating the current vendor is fine, then this commit should unblock #56, if not, then I need to copy the relevant functions from predicates, but then we have to maintain them in the project subsequently.

PTAL, WDYT?

containscafeine:
Actually no, this is not working without updating the entire thing. Resorting to copying the code.

Issue #76, opened 2018-01-23
Consider priorities while evicting pods


ravisantoshgudimetla:
Now that 1.9 rebase happened, we can use priorities(https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/) which are in alpha while evicting pods.

ravisantoshgudimetla:
This has been addressed by #105 

/close

PR #78, opened 2018-02-09
Descheduler should not run when cluster size is 1

ravisantoshgudimetla:
This ensures that no strategy is run in first place when the cluster size is 1.

/cc @aveshagarwal @containscafeine


ravisantoshgudimetla:
@aveshagarwal - Merging this as this closes #76

PR #79, opened 2018-02-15
Scratch image + multistage builds.

jelmersnoeck:
By disabling CGO, we can use the scratch image instead of the fedora
image, allowing a lighter weight image.

By setting up a multistage docker build, we can create the container image
in a single command. This eliminates external setup and allows us to build
this automatically on registries.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,10 +11,9 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
-FROM fedora
+FROM scratch
 
 MAINTAINER Avesh Agarwal <avagarwa@redhat.com>
 
 COPY _output/bin/descheduler /bin/descheduler
-CMD ["/bin/descheduler --help"]
+CMD ["/bin/descheduler", "--help"]
--- a/README.md
+++ b/README.md
@@ -57,8 +57,18 @@ in `kube-system` namespace.
 First we create a simple Docker image utilizing the Dockerfile found in the root directory:
 
 ```
+$ make dev-image
+```
+
+This creates an image based off the binary we've built before. To build both the
+binary and image in one step you can run the following command:
+
+```
 $ make image
 ```
+
+This eliminates the need to have Go installed locally and builds the binary
+within it's own container.
 
 ### Create a cluster role
 
@@ -150,7 +160,7 @@ $ kubectl create -f descheduler-job.yaml
 ```
 
 ## Policy and Strategies
- 
+
 Descheduler's policy is configurable and includes strategies to be enabled or disabled.
 Three strategies, `RemoveDuplicates`, `LowNodeUtilization`, `RemovePodsViolatingInterPodAntiAffinity` are currently implemented.
 As part of the policy, the parameters associated with the strategies can be configured too.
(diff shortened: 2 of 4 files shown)


aveshagarwal:

jelmersnoeck:
CLA signed.

Dockerfile
```diff
@@ -1,3 +1,9 @@
+FROM golang:1.9.2
+
+WORKDIR /go/src/github.com/kubernetes-incubator/descheduler
+COPY . .
+RUN make
+
```
aveshagarwal: can you move this below the license?

jelmersnoeck:
@aveshagarwal done! I've also updated the README to reflect the new make target, `make docker`.

README.md
```diff
@@ -57,7 +57,7 @@ in `kube-system` namespace.
 First we create a simple Docker image utilizing the Dockerfile found in the root directory:
 
 ```
-$ make image
+$ make docker
```
aveshagarwal: Could you update README to include both `make image` and `make docker` targets and explain the difference? Also I was wondering if instead of using `docker` if we could use some other name similar to image or container-image or along these lines?

aveshagarwal:
@jelmersnoeck thanks looks really good. Just a comment for README.

Dockerfile
```diff
@@ -11,10 +11,16 @@
...
 # See the License for the specific language governing permissions and
 # limitations under the License.
+FROM golang:1.9.2
 
-FROM fedora
+WORKDIR /go/src/github.com/kubernetes-incubator/descheduler
+COPY . .
+RUN make
+
+FROM scratch
```
cesartl: do you really need this `FROM scratch` here?

cesartl:
I have the same goal.

Perhaps we could optimise the docker image and remove the source code after compiling. It could also be done in a single command so the data doesn't go into the docker layer.

For example before seeing this PR I had done something like that

```Dockerfile
RUN    git clone https://github.com/kubernetes-incubator/descheduler.git kubernetes-incubator/descheduler \
    && make -C kubernetes-incubator/descheduler \
    && cp kubernetes-incubator/descheduler/_output/bin/descheduler /bin/descheduler \
    && rm -rf kubernetes-incubator/descheduler
```

here I'm cloning rather than using `COPY` which is not ideal. Even with the COPY we should delete the redundant source code.

jelmersnoeck:
@cesartl to answer both your questions: this is what the [multi-stage build](https://docs.docker.com/develop/develop-images/multistage-build/) setup is for.

In the first container we'll do the build by copying in the source code into a `golang` container. This is then able to build the binary, which we store in `/go/src/github.com/kubernetes-incubator/descheduler/_output/bin/descheduler`.

The second `FROM scratch` then starts a second image - a `scratch` image in this case - with nothing in it. We just copy over the binary we've created before. This second build is what the final image will look like, a scratch image with only the binary in it, nothing else.

README.md
jelmersnoeck: I initially thought of renaming the `image` target to `dev-image` but didn't want to make a breaking change. If that's something you're happy with, we could have `dev-image` which builds the way it is currently before this patch, and then `image` which does the multi-stage build.

👍 on updating the README to document both targets.

cesartl:
@jelmersnoeck that's amazing, I didn't know this new feature. Much cleaner and easy to use.

README.md
aveshagarwal: sgtm

jelmersnoeck:
@aveshagarwal renamed the targets to `dev-image` and `image` and updated the README accordingly.

aveshagarwal:
@jelmersnoeck thanks. just merged it.

PR #80, opened 2018-02-16
Universal Docker build

cesartl:
I'm opening this PR to start a discussion.
This project currently supports building a Docker image, but it only works on linux distribution compatible with Docker. This is because the Go  binaries are compiled on the host machine and then moved to docker.

For example if you run `make` and `make image` on a Mac, then try to run the Docker image you get an error like this:

```
/bin/sh: /bin/descheduler: cannot execute binary file: Exec format error
```

I think a better approach would be to compile the Go code directly in the docker container.

I've added the Dockerfile which I use to build the Descheduler. This is just an example. If we were to integrate with this project we should probably use a `COPY` command rather than cloning from git.

Let me know your thoughts and I'd happy to tweak my Dockerfile to make it fit this project

### DIFF
--- a/dockerbuild/Dockerfile
+++ b/dockerbuild/Dockerfile
@@ -1,0 +1,12 @@
+FROM golang:1.9
+
+WORKDIR /golang
+ENV GOPATH /golang
+WORKDIR /golang/src/github.com
+
+RUN    git clone https://github.com/kubernetes-incubator/descheduler.git kubernetes-incubator/descheduler \
+    && make -C kubernetes-incubator/descheduler \
+    && cp kubernetes-incubator/descheduler/_output/bin/descheduler /bin/descheduler \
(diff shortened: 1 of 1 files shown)


aveshagarwal:

cesartl:
Just to be clear, the proposed Dockerfile is not supposed to be the end version. As described above it probably needs changing but I thought it would be a starting point for the discussion

aveshagarwal:
@cesartl thanks, yes it'd be good to have some portable build process. Could you see another PR https://github.com/kubernetes-incubator/descheduler/pull/79 that might be doing what you are looking for or if not you could create something additional on top of that PR?

cesartl:
Ha yes it seems to be doing the same thing. The goal is the same at least.

Let's close that PR and focus on #79.

I'll see if it does what I need.

PR #83, opened 2018-03-12
Evict pods which has no owner references.

ravisantoshgudimetla:
While working on e2e's, I realized that pods that have no owner references are not evictable. So, I removed the check for that and added a regression test.

### DIFF
--- a/pkg/descheduler/pod/pods.go
+++ b/pkg/descheduler/pod/pods.go
@@ -54,7 +54,7 @@ func IsLatencySensitivePod(pod *v1.Pod) bool {
 // IsEvictable checks if a pod is evictable or not.
 func IsEvictable(pod *v1.Pod) bool {
 	ownerRefList := OwnerRef(pod)
-	if IsMirrorPod(pod) || IsPodWithLocalStorage(pod) || len(ownerRefList) == 0 || IsDaemonsetPod(ownerRefList) || IsCriticalPod(pod) {
+	if IsMirrorPod(pod) || IsPodWithLocalStorage(pod) || IsDaemonsetPod(ownerRefList) || IsCriticalPod(pod) {
 		return false
 	}
 	return true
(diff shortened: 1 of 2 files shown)


ingvagabund:
> pods that have no owner references are not evictable.

it kinda contradicts the summary saying `Evict pods which has no owner references`

So what is the PR about again?

ravisantoshgudimetla:
> it kinda contradicts the summary saying Evict pods which has no owner references

So, as of now, we are not evicting pods that have no owner references but we should be able to evict them. This PR ensures that we can evict them.

ingvagabund:
> So, as of now, we are not evicting pods that have no owner references but we should be able to evict them. This PR ensures that we can evict them.

If we evict them they will not be re-created. Is that the expected behavior?

ravisantoshgudimetla:
Yes.

aveshagarwal:
@ravisantoshgudimetla but we had decided to not evict those pods as they wont be recreated, so why to evict them now?

ravisantoshgudimetla:
I believe that was not the reason. When we made the switch to owner references as v1.CreatedBy Annotation of pod got removed, we added this ownerRef check. Only pods that are daemonsets, having local storage, critical pods etc will not be removed by descheduler, rest of pods created by any controller should be evicted.

aveshagarwal:
@ravisantoshgudimetla but if a pod does not have any owner ref, how do we know if it has any controller or not?

ravisantoshgudimetla:
@aveshagarwal Yes you are right but if it is one of the pods that we don't want to evict like daemonset etc, it will have ownerRef. If it is not having one, it could be generic pod as well and we are not evicting them currently. This is the case I ran into when I created pause pod without owner reference. They were not getting evicted by descheduler.

aveshagarwal:
@ravisantoshgudimetla what is "generic pod"? We had decided to not evict any pod that wont be recreated, and unless we see any need for this, we are going to do this as it could lead to loss of service.

ravisantoshgudimetla:
what is "generic pod"? - Any pod without ownerRef or controllerRef. 

> We had decided to not evict any pod that wont be recreated.

So, going forward say if we implement priorities, even then we won't be evicting those kind of pods to make room for other pods?



I think we mentioned that descheduler won't evict pods

ravisantoshgudimetla:
Update: Had an offline discussion, it seems we don't evict these kind of pods in descheduler. We will think about that when priorities are implemented.

PR #84, opened 2018-03-20
Max pods per node

ingvagabund:
Building on top of https://github.com/kubernetes-incubator/descheduler/pull/69

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,3 +46,9 @@ test-unit:
 
 test-e2e:
 	./test/run-e2e-tests.sh
+
+gen:
+	./hack/update-codecgen.sh
+	./hack/update-generated-conversions.sh
+	./hack/update-generated-deep-copies.sh
+	./hack/update-generated-defaulters.sh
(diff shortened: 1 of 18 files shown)


aveshagarwal:

ingvagabund:
@ravisantoshgudimetla @aveshagarwal PTAL
Let's do the Victor the `Evictor` kinda refactoring in another PR.

Makefile
```diff
@@ -46,3 +46,9 @@ test-unit:
 
 test-e2e:
 	./test/run-e2e-tests.sh
```
ravisantoshgudimetla: Thanks for adding this

cmd/descheduler/app/options/options.go
```diff
@@ -55,4 +55,6 @@ func (rs *DeschedulerServer) AddFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&rs.DryRun, "dry-run", rs.DryRun, "execute descheduler in dry run mode.")
 	// node-selector query causes descheduler to run only on nodes that matches the node labels in the query
 	fs.StringVar(&rs.NodeSelector, "node-selector", rs.NodeSelector, "Selector (label query) to filter on, supports '=', '==', and '!='.(e.g. -l key1=value1,key2=value2)")
+	// max-no-pods-to-evict limits the maximum number of pods to be evicted per node by descheduler.
+	fs.IntVar(&rs.MaxNoOfPodsToEvictPerNode, "max-no-pods-to-evict", rs.MaxNoOfPodsToEvictPerNode, "Limits the maximum number of pods to be evicted per node by descheduler")
```
aveshagarwal: what about MaxPodsToEvictPerNode?

aveshagarwal:
I just glanced and in general looks good to me.

cmd/descheduler/app/options/options.go
ingvagabund: Given it is a variable name and not the flag/option name, we can change it later to make the name more clear. Though, `MaxNoOfPodsToEvictPerNode` ~ `MaxPodsToEvictPerNode`.

aveshagarwal: That's fine. But now i think to make the flag : max-pods-to-evict-per-node to make it clear in the flag itself that its per node

ingvagabund: Done

PR #88, opened 2018-04-23
Fix dryRun issue and avoid unnecessary log msg when node affinity strategy is disabled.

aveshagarwal:
@ravisantoshgudimetla @ingvagabund

### DIFF
--- a/pkg/descheduler/strategies/node_affinity.go
+++ b/pkg/descheduler/strategies/node_affinity.go
@@ -27,8 +27,7 @@ import (
 )
 
 func RemovePodsViolatingNodeAffinity(ds *options.DeschedulerServer, strategy api.DeschedulerStrategy, evictionPolicyGroupVersion string, nodes []*v1.Node, nodePodCount nodePodEvictedCount) {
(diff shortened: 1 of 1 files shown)


pkg/descheduler/strategies/node_affinity.go
```diff
@@ -70,5 +69,6 @@ func removePodsViolatingNodeAffinityCount(ds *options.DeschedulerServer, strateg
 			return evictedPodCount
 		}
 	}
+	glog.V(1).Infof("Evicted %v pods", evictedPodCount)
```
ingvagabund: How does it help to move the log message here? It is logged in both cases under the same conditions.

aveshagarwal: previously this msg was being logged even when this strategy was disabled so want to avoid it.

ingvagabund: I see now:
```
 	if !strategy.Enabled {
 		return evictedPodCount
 	}
```

aveshagarwal:
@ingvagabund thanks.

PR #89, opened 2018-04-25
add remove too many restarts policy

liubin:
This PR will eviction pods having too many restarts, due to the host-originated problem ( network/storage/firewall ). Usually, these pods should be re-scheduled to other pods.


aveshagarwal:

aveshagarwal:
@liubin thanks for this PR. I will review it soon.

pkg/descheduler/strategies/toomanyrestarts.go
```diff
@@ -0,0 +1,98 @@
...
+			}
+
+			restarts, initRestarts := calcContainerRestarts(pod)
+			if strategy.Params.PodsHavingTooManyRestarts.IncludingInitContainers {
+				if restarts+initRestarts <= strategy.Params.PodsHavingTooManyRestarts.PodeRestartThresholds {
+					continue
+				}
+			} else if restarts <= strategy.Params.PodsHavingTooManyRestarts.PodeRestartThresholds {
+				continue
+			}
```
ingvagabund: Init container is a special case of a container, so the `calcContainerRestarts` could be defined as:

```go
func calcContainerRestarts(pod *v1.Pod, countInitContainers bool) int32 {
	var restarts int32 = 0

	for _, cs := range pod.Status.ContainerStatuses {
		restarts += cs.RestartCount
	}

	if countInitContainers {
		for _, cs := range pod.Status.InitContainerStatuses {
			restarts += cs.RestartCount
		}
	}

	return restarts
}
```

which simplifies the code to:

```go
			restarts := calcContainerRestarts(pod, params.IncludingInitContainers)
			params := strategy.Params.PodsHavingTooManyRestarts
			if restarts <= params.PodeRestartThresholds {
				continue
			}
```

pkg/descheduler/strategies/toomanyrestarts.go
```diff
@@ -0,0 +1,98 @@
...
+
+			glog.V(1).Infof("RemovePodsHavingTooManyRestarts will evicted pod: %#v, container restarts: %d, initContainer restarts: %d", pod.Name, restarts, initRestarts)
```
ingvagabund: @aveshagarwal do we need to distinguish between ordinary and init containers? We could just report the total number of containers that got restarted. WDYT?

pkg/descheduler/strategies/toomanyrestarts_test.go
```diff
@@ -0,0 +1,209 @@
...
+func initPods(node *v1.Node) []v1.Pod {
+
+	pods := make([]v1.Pod, 0)
+
+	var i int32 = 0
+	for i <= 10 {
+		pod := test.BuildTestPod(fmt.Sprintf("pod-%d", i), 100, 0, node.Name)
+		pod.ObjectMeta.OwnerReferences = test.GetNormalPodOwnerRefList()
+
+		// pod i will has 25 * i restarts.
```
ingvagabund: s/has/have

pkg/descheduler/strategies/toomanyrestarts.go
aveshagarwal: Yeah thats fine, it's helpful to distinguish in the log message.

aveshagarwal:
@liubin In general, it looks good to me but still need to review the code more thoroughly.

One question I have that what is this strategy relying on that makes sure that evicted pods would get scheduled to some other nodes by the default scheduler but not on the same node? I am `NOT` looking for a guarantee but some higher probability, as there is no predicate/priority function that takes restart into account. So for the same, I wonder, if you really need to look into the cause of restart and then decide which pod to evict than evicting any pod with several restarts?

liubin:
@aveshagarwal Good question. I have the sam question, and some others: restarts is a cumulative value, 1000 restarts in one year and 1000 restarts in 1 hour have different meanings.

I will do more research and try to find the most practical answer.

liubin:
@ingvagabund I tried:

```
# ./hack/verify-gofmt.sh 
# echo $?
0
# go version
go version go1.8.7 linux/amd64

# ./hack/verify-gofmt.sh 
# echo $?
0
# go version
go version go1.8.3 linux/amd64

$ hack/verify-gofmt.sh 
$ go version
go version go1.10.1 darwin/amd64
```

Is the CI has some different settings, I can find the failed reason.

aveshagarwal:
@liubin meanwhile could you split it in 2 commits during your next update: 1 commit for auto-generated files and 1 commit for main code changes for easier review.

ingvagabund:
@aveshagarwal @ravisantoshgudimetla we should bump to go version to 1.9.* at least.

ingvagabund:
https://github.com/kubernetes-incubator/descheduler/pull/94

ingvagabund:
@liubin 

```sh
$ go version
go version go1.10.2 linux/amd64
$ ./hack/verify-gofmt.sh
$ echo $?
0
```

ravisantoshgudimetla:
@liubin  would you interested in picking this up again as I see there are some use-cases around it?

damemi:
@liubin sorry to resurrect an old thread, but are you still interested in rebasing this PR? If not, we can still pick your commits into a new PR

damemi:
Since there's still interest in this, I've picked @liubin's commits into a new, rebased PR here: https://github.com/kubernetes-sigs/descheduler/pull/254

Thank you for this work @liubin! If it's okay, I'm going to close this PR now and we can move discussion to the new one.

/close

PR #103, opened 2018-07-09
WIP rescheduler mode

ravisantoshgudimetla:
This ensures descheduler is run rescheduler mode so that it can make some room for critical pods created by DS controller. Will add unit tests in the same PR.

/cc @aveshagarwal

### DIFF
--- a/pkg/descheduler/strategies/rescheduler_mode.go
+++ b/pkg/descheduler/strategies/rescheduler_mode.go
@@ -1,0 +1,199 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package strategies
+
+import (
+	"github.com/golang/glog"
+
+	"github.com/kubernetes-incubator/descheduler/cmd/descheduler/app/options"
+	"github.com/kubernetes-incubator/descheduler/pkg/api"
+	"github.com/kubernetes-incubator/descheduler/pkg/descheduler/evictions"
+	podutil "github.com/kubernetes-incubator/descheduler/pkg/descheduler/pod"
+	"k8s.io/api/core/v1"
+	clientset "k8s.io/client-go/kubernetes"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	priorityutil "k8s.io/kubernetes/plugin/pkg/scheduler/algorithm/priorities/util"
+)
+
+const (
+	criticalAddonsOnlyTaintKey = "CriticalAddonsOnly"
+)
+
+func RunInReschedulerMode(ds *options.DeschedulerServer, strategy api.DeschedulerStrategy, policyGroupVersion string, nodes []*v1.Node, nodepodCount nodePodEvictedCount) {
+	if !strategy.Enabled {
+		return
+	}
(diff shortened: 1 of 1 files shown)


pkg/descheduler/strategies/rescheduler_mode.go
```diff
@@ -0,0 +1,199 @@
...
+			continue
+		}
+		for _, criticalDSPod := range criticalDSPods {
+			if err := makeRoomForCriticalDaemonSetPod(client, criticalDSPod, node, policyGroupVersion, dryRun); err != nil {
+				glog.Fatal("Error while making room for critical DS pods")
+				continue
+			}
+		}
+		// Release critical taint from node.
+		releaseTaintsFromNode(client, node)
```
ravisantoshgudimetla: Note to myslf: Need to add number of pods evicted to ensure criticalDS pod runs smoothly.

pkg/descheduler/strategies/rescheduler_mode.go
```diff
@@ -0,0 +1,199 @@
...
+	if !strategy.Enabled {
+		return
+	}
+	evictPodsOnlyInReschedulerMode(ds.Client, policyGroupVersion, nodes, ds.DryRun, nodepodCount, ds.MaxNoOfPodsToEvictPerNode)
+}
+
+func evictPodsOnlyInReschedulerMode(client clientset.Interface, policyGroupVersion string, nodes []*v1.Node, dryRun bool, nodepodCount nodePodEvictedCount, maxPodsToEvict int) int {
+	for _, node := range nodes {
+		criticalDSPods := getCriticalDaemonSetPodsOnNode(client, node)
+		if len(criticalDSPods) < 0 || !IsNodeUnderResourcePressure(client, node) {
```
rphillips: Looks like this should be `<= 0`.

pkg/descheduler/strategies/rescheduler_mode.go
```diff
@@ -0,0 +1,199 @@
...
+		for _, criticalDSPod := range criticalDSPods {
+			if err := makeRoomForCriticalDaemonSetPod(client, criticalDSPod, node, policyGroupVersion, dryRun); err != nil {
+				glog.Fatal("Error while making room for critical DS pods")
+				continue
+			}
+		}
+		// Release critical taint from node.
+		releaseTaintsFromNode(client, node)
+	}
+	return 0
```
rphillips: evictPodsOnlyInReschedulerMode always returns 0... Might be a good candidate for Aggregate errors instead.

pkg/descheduler/strategies/rescheduler_mode.go
ravisantoshgudimetla: You are right, I wanted to remove < and make it == but removed the `=`. Thanks for catching it.

pkg/descheduler/strategies/rescheduler_mode.go
```diff
@@ -0,0 +1,199 @@
...
+	// qos here to make this algo better.
+	for _, evictablePod := range evictablePods {
+		if totalMemoryReclaimed >= neededMemory && totalCPUReclaimed >= neededCPU {
+			// We shouldn't enter this in the first iteration.
+			glog.Infof("We don't need to evict any pod futher as critical Pod has enough resources to run on this node")
+			return nil
+		}
+		// As of know, priority is not taken into consideration while evicting.
+		success, err := evictions.EvictPod(client, evictablePod, policyGroupVersion, dryRun)
+		if !success {
```
rphillips: err should be checked for non nil

pkg/descheduler/strategies/rescheduler_mode.go
ravisantoshgudimetla: Please see my `note to myself`. I should update it with no of pods evicted.

pkg/descheduler/strategies/rescheduler_mode.go
ravisantoshgudimetla: yeah, I just copied that code from existing place. Will modify it. Thanks again.

aveshagarwal:
@ravisantoshgudimetla given that the target is promote "schedule DaemonSetPods by default scheduler" feature to beta in 1.12https://github.com/kubernetes/kubernetes/issues/66526, i wonder what would be the usecase for this? I remember we discussed not going this path last time, unless something changed.

@sjenning

ravisantoshgudimetla:
/close

Closing this as we want to wait till kube 1.12 so that we can remove rescheduler.

PR #105, opened 2018-08-02
Low node utilization to respect priority while evicting pods

ravisantoshgudimetla:
This PR introduces the notion of priority while evicting pods. Also, updated existing test cases to reflect the same. Some of the e2e's may fail which I will work on.

### DIFF
--- a/pkg/descheduler/strategies/lownodeutilization.go
+++ b/pkg/descheduler/strategies/lownodeutilization.go
@@ -41,6 +41,7 @@ type NodeUsageMap struct {
 	bPods            []*v1.Pod
 	gPods            []*v1.Pod
 }
+
 type NodePodsMap map[*v1.Node][]*v1.Pod
 
 func LowNodeUtilization(ds *options.DeschedulerServer, strategy api.DeschedulerStrategy, evictionPolicyGroupVersion string, nodes []*v1.Node, nodepodCount nodePodEvictedCount) {
@@ -59,7 +60,7 @@ func LowNodeUtilization(ds *options.DeschedulerServer, strategy api.DeschedulerS
 		return
 	}
 
-	npm := CreateNodePodsMap(ds.Client, nodes)
+	npm := createNodePodsMap(ds.Client, nodes)
 	lowNodes, targetNodes := classifyNodes(npm, thresholds, targetThresholds)
 
 	glog.V(1).Infof("Criteria for a node under utilization: CPU: %v, Mem: %v, Pods: %v",
@@ -151,6 +152,9 @@ func classifyNodes(npm NodePodsMap, thresholds api.ResourceThresholds, targetThr
 	return lowNodes, targetNodes
 }
 
+// evictPodsFromTargetNodes evicts pods based on priority, if all the pods on the node have priority, if not
+// evicts them based on QoS as fallback option.
+// TODO: @ravig Break this function into smaller functions.
 func evictPodsFromTargetNodes(client clientset.Interface, evictionPolicyGroupVersion string, targetNodes, lowNodes []NodeUsageMap, targetThresholds api.ResourceThresholds, dryRun bool, maxPodsToEvict int, nodepodCount nodePodEvictedCount) int {
 	podsEvicted := 0
 
@@ -191,12 +195,27 @@ func evictPodsFromTargetNodes(client clientset.Interface, evictionPolicyGroupVer
 		glog.V(3).Infof("evicting pods from node %#v with usage: %#v", node.node.Name, node.usage)
 		currentPodsEvicted := nodepodCount[node.node]
 
-		// evict best effort pods
-		evictPods(node.bePods, client, evictionPolicyGroupVersion, targetThresholds, nodeCapacity, node.usage, &totalPods, &totalCpu, &totalMem, &currentPodsEvicted, dryRun, maxPodsToEvict)
-		// evict burstable pods
-		evictPods(node.bPods, client, evictionPolicyGroupVersion, targetThresholds, nodeCapacity, node.usage, &totalPods, &totalCpu, &totalMem, &currentPodsEvicted, dryRun, maxPodsToEvict)
-		// evict guaranteed pods
-		evictPods(node.gPods, client, evictionPolicyGroupVersion, targetThresholds, nodeCapacity, node.usage, &totalPods, &totalCpu, &totalMem, &currentPodsEvicted, dryRun, maxPodsToEvict)
+		// Check if one pod has priority, if yes, assume that all pods have priority and evict pods based on priority.
+		if node.allPods[0].Spec.Priority != nil {
+			glog.V(1).Infof("All pods have priority associated with them. Evicting pods based on priority")
+			evictablePods := make([]*v1.Pod, 0)
+			evictablePods = append(append(node.bPods, node.bePods...), node.gPods...)
+
+			// sort the evictable Pods based on priority. This also sorts them based on QoS. If there are multiple pods with same priority, they are sorted based on QoS tiers.
+			sortPodsBasedOnPriority(evictablePods)
+			evictPods(evictablePods, client, evictionPolicyGroupVersion, targetThresholds, nodeCapacity, node.usage, &totalPods, &totalCpu, &totalMem, &currentPodsEvicted, dryRun, maxPodsToEvict)
+		} else {
+			// TODO: Remove this when we support only priority.
+			//  Falling back to evicting pods based on priority.
+			glog.V(1).Infof("Evicting pods based on QoS")
+			glog.V(1).Infof("There are %v non-evictable pods on the node", len(node.nonRemovablePods))
+			// evict best effort pods
+			evictPods(node.bePods, client, evictionPolicyGroupVersion, targetThresholds, nodeCapacity, node.usage, &totalPods, &totalCpu, &totalMem, &currentPodsEvicted, dryRun, maxPodsToEvict)
+			// evict burstable pods
+			evictPods(node.bPods, client, evictionPolicyGroupVersion, targetThresholds, nodeCapacity, node.usage, &totalPods, &totalCpu, &totalMem, &currentPodsEvicted, dryRun, maxPodsToEvict)
+			// evict guaranteed pods
+			evictPods(node.gPods, client, evictionPolicyGroupVersion, targetThresholds, nodeCapacity, node.usage, &totalPods, &totalCpu, &totalMem, &currentPodsEvicted, dryRun, maxPodsToEvict)
+		}
 		nodepodCount[node.node] = currentPodsEvicted
 		podsEvicted = podsEvicted + nodepodCount[node.node]
 		glog.V(1).Infof("%v pods evicted from node %#v with usage %v", nodepodCount[node.node], node.node.Name, node.usage)
@@ -269,7 +288,30 @@ func SortNodesByUsage(nodes []NodeUsageMap) {
 	})
 }
 
-func CreateNodePodsMap(client clientset.Interface, nodes []*v1.Node) NodePodsMap {
+// sortPodsBasedOnPriority sorts pods based on priority and if their priorities are equal, they are sorted based on QoS tiers.
+func sortPodsBasedOnPriority(evictablePods []*v1.Pod) {
+	sort.Slice(evictablePods, func(i, j int) bool {
+		if evictablePods[i].Spec.Priority == nil && evictablePods[j].Spec.Priority != nil {
+			return true
+		}
+		if evictablePods[j].Spec.Priority == nil && evictablePods[i].Spec.Priority != nil {
+			return false
+		}
+		if (evictablePods[j].Spec.Priority == nil && evictablePods[i].Spec.Priority == nil) || (*evictablePods[i].Spec.Priority == *evictablePods[j].Spec.Priority) {
+			if podutil.IsBestEffortPod(evictablePods[i]) {
+				return true
+			}
+			if podutil.IsBurstablePod(evictablePods[i]) && podutil.IsGuaranteedPod(evictablePods[j]) {
+				return true
+			}
+			return false
+		}
+		return *evictablePods[i].Spec.Priority < *evictablePods[j].Spec.Priority
+	})
+}
+
+// createNodePodsMap returns nodepodsmap with evictable pods on node.
+func createNodePodsMap(client clientset.Interface, nodes []*v1.Node) NodePodsMap {
 	npm := NodePodsMap{}
 	for _, node := range nodes {
 		pods, err := podutil.ListPodsOnANode(client, node)
@@ -308,6 +350,7 @@ func IsNodeWithLowUtilization(nodeThresholds api.ResourceThresholds, thresholds 
 	return true
 }
 
+// Nodeutilization returns the current usage of node.
 func NodeUtilization(node *v1.Node, pods []*v1.Pod) (api.ResourceThresholds, []*v1.Pod, []*v1.Pod, []*v1.Pod, []*v1.Pod, []*v1.Pod) {
 	bePods := []*v1.Pod{}
 	nonRemovablePods := []*v1.Pod{}
--- a/pkg/descheduler/strategies/lownodeutilization_test.go
+++ b/pkg/descheduler/strategies/lownodeutilization_test.go
@@ -28,10 +28,11 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes/fake"
 	core "k8s.io/client-go/testing"
+	"reflect"
 )
 
 // TODO: Make this table driven.
-func TestLowNodeUtilization(t *testing.T) {
+func TestLowNodeUtilizationWithoutPriority(t *testing.T) {
 	var thresholds = make(api.ResourceThresholds)
 	var targetThresholds = make(api.ResourceThresholds)
 	thresholds[v1.ResourceCPU] = 30
@@ -110,7 +111,7 @@ func TestLowNodeUtilization(t *testing.T) {
 		return true, nil, fmt.Errorf("Wrong node: %v", getAction.GetName())
 	})
 	expectedPodsEvicted := 3
-	npm := CreateNodePodsMap(fakeClient, []*v1.Node{n1, n2, n3})
+	npm := createNodePodsMap(fakeClient, []*v1.Node{n1, n2, n3})
 	lowNodes, targetNodes := classifyNodes(npm, thresholds, targetThresholds)
 	if len(lowNodes) != 1 {
 		t.Errorf("After ignoring unschedulable nodes, expected only one node to be under utilized.")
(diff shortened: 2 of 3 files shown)


aveshagarwal:
@ravisantoshgudimetla I do not see that the pods with same priority are being evicted based on their QoS order (be, busrtable, and guaranteed)? did you remove the existing part?

aveshagarwal:
@ravisantoshgudimetla nevermind, I see them now.

ravisantoshgudimetla:
@aveshagarwal  - Any other major concerns? The unit tests seems to be passing for me locally, will look more into e2e's and units, unless you have any other comments.

aveshagarwal:
@ravisantoshgudimetla is it backward compatible?

ravisantoshgudimetla:
@aveshagarwal - Depends, I think as long as we mention that the next release of descheduler to make is compatible with kube 1.11+ where priority is enabled, we should be good. If someone has enabled priority in older version of kube clusters and used this release of descheduler, they will see surprises.

pkg/descheduler/strategies/lownodeutilization.go
```diff
@@ -308,26 +356,34 @@ func IsNodeWithLowUtilization(nodeThresholds api.ResourceThresholds, thresholds
...
+		}
+	}
+	return bePods, bPods, gPods
+}
+
+// Nodeutilization returns the current usage of node.
+func NodeUtilization(node *v1.Node, pods []*v1.Pod) api.ResourceThresholds {
 	totalReqs := map[v1.ResourceName]resource.Quantity{}
 	for _, pod := range pods {
+		glog.Infof("Ravig %v", pod.Name)
```
ravisantoshgudimetla: Not to myself: Need to remove this.

aveshagarwal:
> If someone has enabled priority in older version of kube clusters and used this release of descheduler, they will see surprises.

I did not mean that infact I dont think so anyone is going to enable priority in older releases. I meant the code after this PR keeps working with older releases?

ravisantoshgudimetla:
> I meant the code after this PR keeps working with older releases?

Got it. Thanks. I think this line could cause a problem https://github.com/kubernetes-incubator/descheduler/pull/105/files#diff-36906f0d7bfe55a6cab3b3b86aa63bacR194 but I will fix it. I need add a check for nil for pod.Spec.Priority. I will fix it.

ravisantoshgudimetla:
@aveshagarwal - I have ensured this strategy is backward compatible with some tunit ests to catch regressions. PTAL.

ravisantoshgudimetla:
@aveshagarwal - I have updated unit tests to catch regressions and e2e's are passing as well. Can you PTAL?

pkg/descheduler/strategies/lownodeutilization.go
```diff
@@ -146,11 +146,13 @@ func classifyNodes(npm NodePodsMap, thresholds api.ResourceThresholds, targetThr
 		} else {
 			glog.V(2).Infof("Node %#v is appropriately utilized with usage: %#v", node.Name, usage)
 		}
-		glog.V(2).Infof("allPods:%v, nonRemovablePods:%v, bePods:%v, bPods:%v, gPods:%v", len(allPods), len(nonRemovablePods), len(bePods), len(bPods), len(gPods))
 	}
 	return lowNodes, targetNodes
 }
 
+// evictPodsFromTargetNodes evicts pods based on priority, if all the pods on the node have priority, if not
+// evicts them based on QoS as fallback option.
```
aveshagarwal: The above is not correct, it should be "pods are evicted in the order of their priorities from low to high. Pods with same priority are evicted based on their QoS." QoS is not a fallback option, both priority and QoS work together.

pkg/descheduler/strategies/lownodeutilization.go
```diff
@@ -308,28 +382,37 @@ func IsNodeWithLowUtilization(nodeThresholds api.ResourceThresholds, thresholds
 	return true
 }
 
-func NodeUtilization(node *v1.Node, pods []*v1.Pod) (api.ResourceThresholds, []*v1.Pod, []*v1.Pod, []*v1.Pod, []*v1.Pod, []*v1.Pod) {
```
aveshagarwal: I am still trying to understand why you have split this function "NodeUtilization" into 2 functions (classifyPods and NodeUtilization). The reason I had combined them for "performance reasons". I wanted to traverse all nodes and their pods "just once". So please make sure you are not traversing the lists of pods more than once. Try to create anything just based on one traversal of all pods on all nodes.

aveshagarwal:
@ravisantoshgudimetla I need more time to review this, mainly I am trying to understand the changes in the PR does not impact performance. Traversing pods on node multiple time (or even more than once) must be avoided, even if that means consuming more memory (though I assume that it would not lead to an unacceptable memory consumption).

pkg/descheduler/strategies/lownodeutilization.go
ravisantoshgudimetla: So, this is when any of the pods on the node have no priority associated with them, if we have atleast one pod with no priority then we will evict all the pods based on QoS.

pkg/descheduler/strategies/lownodeutilization.go
ravisantoshgudimetla: I thought about that when I was making changes. If you think performance is a bottleneck, I can revert the change but I was trying to make code a bit more readable.

ravisantoshgudimetla:
@aveshagarwal  - Sure, I can wait, I am just curious, if you have some performance numbers in your mind, I can try tweeking the code so as to fit in that range. If not, I can try writing benchmarks. Let me know either way.

aveshagarwal:
> Sure, I can wait, I am just curious, if you have some performance numbers in your mind, I can try tweeking the code so as to fit in that range.

I do not have any performance numbers in my mind. But doing something obvious that would impact performance on large clusters should be avoided in the first place.

> If not, I can try writing benchmarks. Let me know either way.

Its upto you.

pkg/descheduler/strategies/lownodeutilization.go
aveshagarwal: > If you think performance is a bottleneck, I can revert the change but I was trying to make code a bit more readable.

This code is not really large that we can not maintain readability. You could always split that function into multiple functions but can call new functions inside the bigger function to improve readability, You do not have to separate them.

pkg/descheduler/strategies/lownodeutilization.go
aveshagarwal: > So, this is when any of the pods on the node have no priority associated with them, if we have atleast one pod with no priority then we will evict all the pods based on QoS.

That approach seems incorrect. We should not do that. Anyway I think, either all pods have priority or no pods have priority (mainly in older release). If priority admission plugin is enabled, there will not be any mixed scenario. 

Even if there is a chance of mixed case, we should consider all pods with out priority as with same priority and evict them based on their QoS.

ravisantoshgudimetla: > out priority as with same priority

You mean update the priority to 0 or something else? If yes, I think we shouldn't touch priority field.

pkg/descheduler/strategies/lownodeutilization.go
ravisantoshgudimetla: Its more on the lines of one function doing one thing but I see your point. I can make the changes after writing the benchmark.

pkg/descheduler/strategies/lownodeutilization.go
aveshagarwal: > You mean update the priority to 0 or something else? If yes, I think we shouldn't touch priority field.

I did not mention about updating priority field at all. I just mentioned "treating/grouping them as with the same priority".

ravisantoshgudimetla: Can you please elaborate. We are sorting based on priority field.

ravisantoshgudimetla:
@aveshagarwal - I reverted the change I made for `NodeUtilization` function. PTAL.

pkg/descheduler/strategies/lownodeutilization.go
```diff
@@ -191,19 +198,84 @@ func evictPodsFromTargetNodes(client clientset.Interface, evictionPolicyGroupVer
...
+					podGroupByPriority[priority] = append(podGroupByPriority[priority], pod)
+				}
+			}
+		}
+	}
+}
+
+// checkIfAllPodsHavePriority checks if all the pods in the given pod list have a priority associated with them.
+func checkIfAllPodsHavePriority(podList []*v1.Pod) bool {
+	for _, pod := range podList {
```
aveshagarwal: why are you checking it for all pods? if one pod has priority, that means all have priority, if one does not have, that means, all dont have priority, so just checking for one pod should be enough?

ravisantoshgudimetla: Well, I am not sure, if priority admission plugin is enabled after some pods are created, there could be some pods that don't have priority.(Because admission controller is run on create/update). While I did not test this scenario, I think this could be a problem, that's why I ran for entire set of pods on the node. This also ensures that the subsequent sorting of pods based on priority won't fail.

pkg/descheduler/strategies/lownodeutilization.go
```diff
@@ -191,19 +198,84 @@ func evictPodsFromTargetNodes(client clientset.Interface, evictionPolicyGroupVer
...
-		evictPods(node.bePods, client, evictionPolicyGroupVersion, targetThresholds, nodeCapacity, node.usage, &totalPods, &totalCpu, &totalMem, &currentPodsEvicted, dryRun, maxPodsToEvict)
-		// evict burstable pods
-		evictPods(node.bPods, client, evictionPolicyGroupVersion, targetThresholds, nodeCapacity, node.usage, &totalPods, &totalCpu, &totalMem, &currentPodsEvicted, dryRun, maxPodsToEvict)
-		// evict guaranteed pods
-		evictPods(node.gPods, client, evictionPolicyGroupVersion, targetThresholds, nodeCapacity, node.usage, &totalPods, &totalCpu, &totalMem, &currentPodsEvicted, dryRun, maxPodsToEvict)
+		// Check if all pods have priority, if yes, go ahead and sort them based on priority.
+		if checkIfAllPodsHavePriority(node.allpods) {
+			glog.V(1).Infof("All pods have priority associated with them. Evicting pods based on priority")
+			// sort the evictable Pods based on priority.
+			sortPodsBasedOnPriority(node.allpods)
```
aveshagarwal: I am thinking if you could find out how to sort based on 2 parameters: priority and qos field, that way, you will have just one sorted pod list, and you would not have to create any priorityToPodMap or classify pods again. For example, lets say there are 5 unsorted pods as follows (convention is like pod-name(priority, qos) just for example) : p1(5,G), p2(3,G), p3(5,be), p4(2,G)  p5(3,burstable), and after sorting based on priority and qos field, the sorted list would look like: p4(2,G), p5(3,burstable), p2(3,G), p3(5,be), p1(5,G), Once we have this sorted list, we could just start evicting from the first one till the point in the sorted list when our target is met. I think it would simplify the logic a lot.

aveshagarwal: @ravisantoshgudimetla I have another idea too which could be a way better from performance point of view. Fetch the list of the priority classes from apiserver. This way we already know the list of priorities. Create same number of buckets(slices) as number of priorities. Traverse the list of pods one and put each pod in the respective bucket (counting sort). The same algorithm (counting sort which is o(n)) I am using in NodeUtilization for guaranteed, best effort and burstable pods.

aveshagarwal:
@ravisantoshgudimetla are you working on this, or if you want, i can work on it?

pkg/descheduler/strategies/lownodeutilization.go
ravisantoshgudimetla: Even though, the 2nd option LGTM in terms of performance, I want to something on the lines of first option.

```go
type podPriorityMap struct {
priority int32
podList []*v1.Pod
}
```

and then we will sort podPriorityMap struct based on priorities. After this, we can evict pods from podPriorityMap.podList based on QoS tiers. This will ensure that there is one sorting and then one linear search while classification.

aveshagarwal: @ravisantoshgudimetla IMO, we dont need podPriorityMap. Just sorting pod list based on priority and QoS is enough.

ravisantoshgudimetla: I think, we can do it, if we allow multiple traversals of the list and you wanted this to be in single traversal unless I am missing something.

ravisantoshgudimetla: So, I was trying to do the samething but I see that less works on both the keys that we want to sort. I will try to simplify it and see. https://gist.github.com/ravisantoshgudimetla/b3ed1da28c125abf9d43b3ca1b5b8ed3

ravisantoshgudimetla:
@aveshagarwal - I have updated PR as per your suggestions. PTAL, if this is the same thing you have in your mind.

pkg/descheduler/strategies/lownodeutilization.go
```diff
@@ -269,7 +298,23 @@ func SortNodesByUsage(nodes []NodeUsageMap) {
...
-func CreateNodePodsMap(client clientset.Interface, nodes []*v1.Node) NodePodsMap {
+// sortPodsBasedOnPriority sorts pods based on priority and if their priorities are equal, they are sorted based on QoS tiers.
+func sortPodsBasedOnPriority(evictablePods []*v1.Pod) {
+	sort.Slice(evictablePods, func(i, j int) bool {
+		if *evictablePods[i].Spec.Priority == *evictablePods[j].Spec.Priority {
+			if podutil.IsBestEffortPod(evictablePods[i]) && (podutil.IsGuaranteedPod(evictablePods[j]) || podutil.IsBurstablePod(evictablePods[j])) {
+				return true
+			} else if podutil.IsBurstablePod(evictablePods[i]) && podutil.IsGuaranteedPod(evictablePods[j]) {
+				return true
+			}
```
aveshagarwal: both if and else are returning true, why to use if-else then?

aveshagarwal:
> @aveshagarwal - I have updated PR as per your suggestions. PTAL, if this is the same thing you have in your mind.

yes.

pkg/descheduler/strategies/lownodeutilization.go
```diff
@@ -35,12 +35,13 @@ import (
 type NodeUsageMap struct {
 	node             *v1.Node
 	usage            api.ResourceThresholds
-	allPods          []*v1.Pod
+	allpods          []*v1.Pod
```
aveshagarwal: is there any reason to change this?

pkg/descheduler/strategies/lownodeutilization.go
```diff
@@ -191,19 +195,44 @@ func evictPodsFromTargetNodes(client clientset.Interface, evictionPolicyGroupVer
...
 		nodepodCount[node.node] = currentPodsEvicted
 		podsEvicted = podsEvicted + nodepodCount[node.node]
 		glog.V(1).Infof("%v pods evicted from node %#v with usage %v", nodepodCount[node.node], node.node.Name, node.usage)
 	}
 	return podsEvicted
 }
 
+// checkIfAllPodsHavePriority checks if all the pods in the given pod list have a priority associated with them.
+func checkIfAllPodsHavePriority(podList []*v1.Pod) bool {
+	for _, pod := range podList {
```
aveshagarwal: i don't think that we need to check it for all pods as priority admission plugin is enabled by default upstream so either all will have priority or none will have priority?

pkg/descheduler/strategies/lownodeutilization.go
```diff
@@ -191,19 +195,44 @@ func evictPodsFromTargetNodes(client clientset.Interface, evictionPolicyGroupVer
...
-		evictPods(node.bPods, client, evictionPolicyGroupVersion, targetThresholds, nodeCapacity, node.usage, &totalPods, &totalCpu, &totalMem, &currentPodsEvicted, dryRun, maxPodsToEvict)
-		// evict guaranteed pods
-		evictPods(node.gPods, client, evictionPolicyGroupVersion, targetThresholds, nodeCapacity, node.usage, &totalPods, &totalCpu, &totalMem, &currentPodsEvicted, dryRun, maxPodsToEvict)
+		// Check if all pods have priority, if yes, go ahead and sort them based on priority.
+		if checkIfAllPodsHavePriority(node.allpods) {
+			glog.V(1).Infof("All pods have priority associated with them. Evicting pods based on priority")
+			evictablePods := make([]*v1.Pod, 0)
+			evictablePods = append(append(node.bPods, node.bePods...), node.gPods...)
+
+			// sort the evictable Pods based on priority. This also sorts them based on QoS. If there multiple pods with same priority, they are sorted based on QoS tiers.
```
aveshagarwal: typo in this comment: "If there multiple pods with same priority"

pkg/descheduler/strategies/lownodeutilization.go
ravisantoshgudimetla: Yup, another if would be good enough.

pkg/descheduler/strategies/lownodeutilization.go
ravisantoshgudimetla: Nothing, typo when I added it back.

pkg/descheduler/strategies/lownodeutilization.go
ravisantoshgudimetla: Ok, I will remove this, check if one pod priority, the downside is program will crash without it.

pkg/descheduler/strategies/lownodeutilization.go
ravisantoshgudimetla: Sure.

pkg/descheduler/strategies/lownodeutilization.go
```diff
@@ -269,7 +288,30 @@ func SortNodesByUsage(nodes []NodeUsageMap) {
 	})
 }
 
-func CreateNodePodsMap(client clientset.Interface, nodes []*v1.Node) NodePodsMap {
+// sortPodsBasedOnPriority sorts pods based on priority and if their priorities are equal, they are sorted based on QoS tiers.
+func sortPodsBasedOnPriority(evictablePods []*v1.Pod) {
+	sort.Slice(evictablePods, func(i, j int) bool {
+		if evictablePods[i].Spec.Priority == nil {
+			*evictablePods[i].Spec.Priority = 0 // This should have been handled by priority admission controller but to be on safe-side, we are setting its priority to 0.
```
aveshagarwal: I dont think assignment here is necessary, you could just return true here.

pkg/descheduler/strategies/lownodeutilization.go
```diff
@@ -269,7 +288,30 @@ func SortNodesByUsage(nodes []NodeUsageMap) {
...
+		}
+		if evictablePods[j].Spec.Priority == nil {
+			*evictablePods[j].Spec.Priority = 0 // This should have been handled by priority admission controller but to be on safe-side, we are setting its priority to 0.
```
aveshagarwal: similarly, you could just return false here.

pkg/descheduler/strategies/lownodeutilization.go
```diff
@@ -269,7 +288,30 @@ func SortNodesByUsage(nodes []NodeUsageMap) {
...
+		}
+		if *evictablePods[i].Spec.Priority == *evictablePods[j].Spec.Priority {
+			if podutil.IsBestEffortPod(evictablePods[i]) && (podutil.IsGuaranteedPod(evictablePods[j]) || podutil.IsBurstablePod(evictablePods[j])) {
+				return true
```
aveshagarwal: just "if podutil.IsBestEffortPod(evictablePods[i])" is enough, and no need to check for "podutil.IsGuaranteedPod(evictablePods[j]) || podutil.IsBurstablePod(evictablePods[j])"

pkg/descheduler/strategies/lownodeutilization.go
```diff
@@ -269,7 +288,30 @@ func SortNodesByUsage(nodes []NodeUsageMap) {
...
+			}
+			if podutil.IsBurstablePod(evictablePods[i]) && podutil.IsGuaranteedPod(evictablePods[j]) {
+				return true
```
aveshagarwal: you have not checked for the scenario when `j` is best effort, and `i` is burstable.

aveshagarwal: Nevermind, it is covered by false below.

pkg/descheduler/strategies/lownodeutilization.go
ravisantoshgudimetla: Ya, I was just being over-cautious

pkg/descheduler/strategies/lownodeutilization.go
ravisantoshgudimetla: I think the problem will be if both pods don't have priority and one is best effort and other is guaranteed, we will them going other way around, if we return immediately.

pkg/descheduler/strategies/lownodeutilization.go
ravisantoshgudimetla: Yes.

pkg/descheduler/strategies/lownodeutilization.go
```diff
@@ -269,7 +288,30 @@ func SortNodesByUsage(nodes []NodeUsageMap) {
...
+// sortPodsBasedOnPriority sorts pods based on priority and if their priorities are equal, they are sorted based on QoS tiers.
+func sortPodsBasedOnPriority(evictablePods []*v1.Pod) {
+	sort.Slice(evictablePods, func(i, j int) bool {
+		if evictablePods[i].Spec.Priority == nil && evictablePods[j].Spec.Priority != nil {
+			return true
+		}
+		if evictablePods[j].Spec.Priority == nil && evictablePods[i].Spec.Priority != nil {
+			return false
+		}
+		if (evictablePods[j].Spec.Priority == nil && evictablePods[i].Spec.Priority == nil) || (*evictablePods[i].Spec.Priority == *evictablePods[j].Spec.Priority) {
```
ravisantoshgudimetla: @aveshagarwal - The above covers all the 3 scenarios when one of the pods doesn't have a priority and when both doesn't have priority. Please let me know, if you have anything else in your mind.

pkg/descheduler/strategies/lownodeutilization.go
aveshagarwal: yeah why to have unnecessary check, so removing that would be good.

pkg/descheduler/strategies/lownodeutilization.go
aveshagarwal: this looks better for sure. The only thing I am not able to understand when can we hit (evictablePods[j].Spec.Priority == nil), can you explain the use cases?

ravisantoshgudimetla: So, say we have 2 pods for comparing and first pod has priority where as 2nd one is not priority, we need this case.

pkg/descheduler/strategies/lownodeutilization.go
```diff
@@ -269,7 +288,30 @@ func SortNodesByUsage(nodes []NodeUsageMap) {
...
+			if podutil.IsBestEffortPod(evictablePods[i]) && (podutil.IsGuaranteedPod(evictablePods[j]) || podutil.IsBurstablePod(evictablePods[j])) {
```
aveshagarwal: can you just update this and then we can merge it?

ravisantoshgudimetla: Done.

ravisantoshgudimetla:
e2es passing. Merging the PR.

PR #108, opened 2018-08-21
Fix deprecated warning in CI

ravisantoshgudimetla:
I am seeing the following errors in CI continuously.

Flag --skip-preflight-checks has been deprecated, it is now equivalent to --ignore-preflight-errors=all

This should fix the behaviour.


aveshagarwal:
I am fine as far as you think its ok.

ravisantoshgudimetla:
Alright. Thanks @aveshagarwal . e2es passing and now we are not seeing this error. So, merging it.

PR #112, opened 2018-08-24
add kubernetes yaml files

nicholasklem:
Here is a working setup with yaml files ready to be applied with 
`kubectl apply -f .`

### DIFF
--- a/kubernetes/rbac.yaml
+++ b/kubernetes/rbac.yaml
@@ -1,0 +1,37 @@
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1beta1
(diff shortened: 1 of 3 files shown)


aveshagarwal:
@nicholasklem thanks for this PR. I will take a look soon.

kubernetes/rbac.yaml
```diff
@@ -0,0 +1,37 @@
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1beta1
```
aveshagarwal: rbac.authorization.k8s.io/v1

PR #129, opened 2019-01-08
add preferredDuringSchedulingIgnoredDuringExecution option

krudskoy:


aveshagarwal:

krudskoy:
Hello @ravisantoshgudimetla, I can't access the link [104.196.33.47:8080/job/descheduler-CI/165/](url) and see what happend with default test. Does the service work properly?

ravisantoshgudimetla:
Seems, the link is still pointing to old IP address, I will look into this today/tomorrow.

krudskoy:
Hello @ravisantoshgudimetla, are there some changes?

ravisantoshgudimetla:
> Hello @ravisantoshgudimetla, are there any changes?

I have updated the jenkins end point to point to latest ip, I will some restrictions once the test cases pass.

krudskoy:
Thanks

pkg/descheduler/strategies/node_affinity.go
```diff
@@ -26,19 +26,29 @@ import (
 	"k8s.io/api/core/v1"
 )
 
+func Contains(a []string, x string) bool {
```
ravisantoshgudimetla: Any specific reason for making this function public?

pkg/descheduler/strategies/node_affinity.go
```diff
@@ -53,21 +63,52 @@ func removePodsViolatingNodeAffinityCount(ds *options.DeschedulerServer, strateg
 					if maxPodsToEvict > 0 && nodepodCount[node]+1 > maxPodsToEvict {
 						break
 					}
-					if pod.Spec.Affinity != nil && pod.Spec.Affinity.NodeAffinity != nil && pod.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution != nil {
+					if pod.Spec.Affinity != nil && pod.Spec.Affinity.NodeAffinity != nil {
```
ravisantoshgudimetla: You can combine both the blocks under switch cases to function and re-use them.

pkg/descheduler/strategies/node_affinity.go
krudskoy: Actually no, changed it.

pkg/descheduler/strategies/node_affinity.go
krudskoy: Got it.

yoda:
> Thank you @krudskoy for the PR. I am curious if you have any use-case for evicting preferred pods too.

@ravisantoshgudimetla Use case would be where you have 2 autoscaler groups of nodes where one is a spot instance type and the other is standard. The spot nodes get terminated on price going above threshold resulting in the preferredDuringScheduling being invalidated resulting in scheduling onto standard nodes. Over time the price on the spots goes back down and the descheduler can cause the rescheduling of them back onto spot instances.

seanmalloy:
@yoda thanks for the feedback. It looks like feature request #211 is currently opened for tracking this. You might want to comment here with any thoughts or feedback. Thanks!

PR #134, opened 2019-01-31
Evict duplicate pods only if not all nodes are used

nitzs:
Currently, if the total number of pods running on a cluster is greater than the number of nodes and each node has atleast one pod allocated, the descheduler will evict the pods from the nodes where duplicates are found. This behavior is unnecessary as the evicted pods will be allocated to some nodes again. Further, if the descheduler runs periodically as a CronJob, the cluster never stabilizes and remains in a constant state of flux.

This PR adds a condition to check if a _creator_ has its pods on all possible nodes (the pods satisfy the affinity and taints on these nodes). If it has, we call the creator _saturated_ and no pods of this creator would be evicted from any nodes. This helps stabilize the cluster.

### DIFF
--- a/pkg/descheduler/node/node.go
+++ b/pkg/descheduler/node/node.go
@@ -165,3 +165,21 @@ func PodFitsCurrentNode(pod *v1.Pod, node *v1.Node) bool {
 	glog.V(3).Infof("Pod %v fits on node %v", pod.Name, node.Name)
 	return true
 }
+
+// PodToleratesNodeTaints check if the given pod tolerates the taints on the given node
+func PodToleratesNodeTaints(pod *v1.Pod, node *v1.Node) bool {
+	ok, err := utils.PodToleratesNodeTaints(pod, node)
+
+	if err != nil {
+		glog.Error(err)
+		return false
+	}
+
+	if !ok {
+		glog.V(1).Infof("Pod %v does not tolerate taints on node %v", pod.Name, node.Name)
+		return false
+	}
+
+	glog.V(3).Infof("Pod %v tolerates taints on node %v", pod.Name, node.Name)
+	return true
+}
--- a/pkg/descheduler/strategies/duplicates.go
+++ b/pkg/descheduler/strategies/duplicates.go
@@ -27,6 +27,7 @@ import (
 	"github.com/kubernetes-incubator/descheduler/cmd/descheduler/app/options"
 	"github.com/kubernetes-incubator/descheduler/pkg/api"
 	"github.com/kubernetes-incubator/descheduler/pkg/descheduler/evictions"
+	nodeutil "github.com/kubernetes-incubator/descheduler/pkg/descheduler/node"
 	podutil "github.com/kubernetes-incubator/descheduler/pkg/descheduler/pod"
 )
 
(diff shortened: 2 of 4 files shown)


aveshagarwal:

pkg/descheduler/node/node.go
```diff
@@ -165,3 +165,21 @@ func PodFitsCurrentNode(pod *v1.Pod, node *v1.Node) bool {
...
+func PodToleratesNodeTaints(pod *v1.Pod, node *v1.Node) bool {
+	ok, err := utils.PodToleratesNodeTaints(pod, node)
+
+	if err != nil {
+		glog.Error(err)
+		return false
+	}
+
+	if !ok {
+		glog.V(1).Infof("Pod %v does not tolerate taints on node %v", pod.Name, node.Name)
```
aveshagarwal: Please move this log level to 2 or 3.

pkg/descheduler/node/node.go
```diff
@@ -165,3 +165,21 @@ func PodFitsCurrentNode(pod *v1.Pod, node *v1.Node) bool {
 	glog.V(3).Infof("Pod %v fits on node %v", pod.Name, node.Name)
 	return true
 }
+
+// PodToleratesNodeTaints check if the given pod tolerates the taints on the given node
```
aveshagarwal: nit: should be checks I think.

aveshagarwal:
>  Currently, if the total number of pods running on a cluster is greater than the number of nodes and each node has atleast one pod allocated, the descheduler will evict the pods from the nodes where duplicates are found. This behavior is unnecessary as the evicted pods will be allocated to some nodes again. Further, if the descheduler runs periodically as a CronJob, the cluster never stabilizes and remains in a constant state of flux.

> This PR adds a condition to check if a _creator_ has its pods on all possible nodes (the pods satisfy the affinity and taints on these nodes). If it has, we call the creator _saturated_ and no pods of this creator would be evicted from any nodes. This helps stabilize the cluster.

I agree with your explanation. thanks for this PR. I will review it soon.

anton-johansson:
Any news on this pull request? I haven't hit this issue **yet**, but I have a use case where I probably will. :)

pkg/descheduler/strategies/duplicates.go
```diff
@@ -72,6 +74,39 @@ func deleteDuplicatePods(client clientset.Interface, policyGroupVersion string,
...
+			creatorAssignedNodes[creator] = append(creatorAssignedNodes[creator], node)
+		}
+	}
+
+	creatorPossibleNodes := make(map[string][]*v1.Node)
+	for creator, nodeList := range creatorAssignedNodes {
+		creatorNode := nodeList[0]
+		creatorPod := dpmByNode[creatorNode][creator][0]
+		for _, node := range nodes {
+			if nodeutil.PodFitsCurrentNode(creatorPod, node) && nodeutil.PodToleratesNodeTaints(creatorPod, node) {
```
ravisantoshgudimetla: I think these 2 are not the only criteria for a pod to land on land, we should try to check all available predicates. IIUC, you want to evict pods only if the node has no duplicate on every node, so why can't we check if more than one copy exists on every node and if yes, don't do anything or else go with evicting the pod? To be clear, why just this check?

aveshagarwal:
@nitzs could you resolve conflicts?

towolf:
@nitzs could you reopn this issue and get it merged. We need this too.

We have two special nodes where 4 copies of a pod are running on each. After a node maintenance all pods are always stuck on one node and we want to rebalance 8 pods as 4 + 4.

PR #136, opened 2019-02-19
Care about namespace when evaluating pods in the duplicate strategy

mrwulf:
If two deployments in different namespaces have identical template hashes, the replicaset name is the same -- with the same name the **Remove Duplicates** strategy will erroneously evict pods.

### DIFF
--- a/pkg/descheduler/strategies/duplicates.go
+++ b/pkg/descheduler/strategies/duplicates.go
@@ -50,19 +50,20 @@ func deleteDuplicatePods(client clientset.Interface, policyGroupVersion string, 
 		glog.V(1).Infof("Processing node: %#v", node.Name)
 		dpm := ListDuplicatePodsOnANode(client, node)
 		for creator, pods := range dpm {
-			if len(pods) > 1 {
-				glog.V(1).Infof("%#v", creator)
-				// i = 0 does not evict the first pod
-				for i := 1; i < len(pods); i++ {
-					if maxPodsToEvict > 0 && nodepodCount[node]+1 > maxPodsToEvict {
-						break
-					}
-					success, err := evictions.EvictPod(client, pods[i], policyGroupVersion, dryRun)
-					if !success {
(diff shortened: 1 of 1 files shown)


aveshagarwal:

mrwulf:
I signed it

pkg/descheduler/strategies/duplicates.go
```diff
@@ -50,19 +50,21 @@ func deleteDuplicatePods(client clientset.Interface, policyGroupVersion string,
...
-						break
-					}
-					success, err := evictions.EvictPod(client, pods[i], policyGroupVersion, dryRun)
-					if !success {
-						glog.Infof("Error when evicting pod: %#v (%#v)", pods[i].Name, err)
-					} else {
-						nodepodCount[node]++
-						glog.V(1).Infof("Evicted pod: %#v (%#v)", pods[i].Name, err)
+			if len(pods) > 0 {
+				for i := 0; i < len(pods); i++ {
```
aveshagarwal: why did you switch the order? what is the issue with the existing code?

mrwulf: I switched it so I could log all of the pods that were being evaluated

mrwulf:
Is there anything else that needs to be done for this?

pkg/descheduler/strategies/duplicates.go
```diff
@@ -50,19 +50,21 @@ func deleteDuplicatePods(client clientset.Interface, policyGroupVersion string,
...
+					glog.V(5).Infof("Evaluating pod: %#v : %#v", creator, pods[i].Name)
```
aveshagarwal: reduce glog level to 2.

aveshagarwal:
squash your commits into 1.

aveshagarwal:
> If two deployments in different namespaces have identical template hashes, the replicaset name is the same -- with the same name the Remove Duplicates strategy will erroneously evict pods.

could you give/show me an example where you noticed this?

PR #140, opened 2019-03-30
Allow for custom critical annotation

theMagicalKarp:
## Problem
We have pods that we consider to be critical which don't run in the kube-system namespace. This is a problem for the current implementation of the descheduler.  The descheduler only classifies a pod as critical if it has the "scheduler.alpha.kubernetes.io/critical-pod" annotation and runs in the kube-system namespace. Ideally we'd like another way to classify pods as critical so they won't be evicted by the descheduler.

## Solution
Have the descheduler look for a custom annotation to determine if a pod is critical or not.
For example we could say for any pod with the "descheduler.kubernetes.io/critical-pod" annotation should be considered critical. (this would remove the constraint of having to run in the kube-system namespace)

## Notes
It's also worth noting that "scheduler.alpha.kubernetes.io/critical-pod" is a deprecated annotation in 1.13 and will be removed 1.14.  https://kubernetes.io/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods/ Soon pods will only be deemed "critical" if they have priority of SystemCriticalPriority or higher.

To also help with flexibility I added a `--annotations-prefix` flag which allows for the caller to determine the annotation prefix. (this resulted a lot of changes for such a small feature) Alternatively I could just statically look for "descheduler.kubernetes.io/critical-pod" in "pkg/descheduler/pod/pods.go" instead of having it be dynamic. (this would result in a lot less code changes) However I thought it'd be nice to provide the option to whoever is invoking the descheduler.


aveshagarwal:

theMagicalKarp:
I signed it

aveshagarwal:
@theMagicalKarp thanks for this PR. i will review it soon.

README.md
```diff
@@ -48,9 +48,7 @@ $ ./_output/bin/descheduler --help
 Descheduler can be run as a job inside of a pod. It has the advantage of
 being able to be run multiple times without needing user intervention.
 Descheduler pod is run as a critical pod to avoid being evicted by itself,
-or by kubelet due to an eviction event. Since critical pods are created in
-`kube-system` namespace, descheduler job and its pod will also be created
-in `kube-system` namespace.
+or by kubelet due to an eviction event.
```
ravisantoshgudimetla: As of now, critical pods are limited to `kube-system` namespace. I think having another annotation specific to descheduler to define critical pods may confuse end-users. I am working on ensuring that critical pods could be created in any namespace that admin wants to be in using quota(https://github.com/kubernetes/kubernetes/pull/76310). I will try to get to it sometime next week. After that, I think this problem should be solved too.

aveshagarwal:
@theMagicalKarp could you resolve the conflicts?

theMagicalKarp:
@ravisantoshgudimetla @aveshagarwal I'll go ahead and close this PR in favor of pushing for something like, https://github.com/kubernetes/kubernetes/pull/76310

Issue #146, opened 2019-05-07
Policy for Balancing Pods across topology domains


krmayankk:
The currently well known topology domains labels on the nodes are 
- failure-domain.beta.kubernetes.io/region
- failure-domain.beta.kubernetes.io/zone

Descheduler currently supports the following options: `RemoveDuplicates` , `LowNodeUtilization`, `RemovePodsViolatingInterPodAntiAffinity` and `RemovePodsViolatingNodeAffinity`.  None of these options support balancing pods across topology domains. The closest one to mimic  this behavior would be LowNodeUtilization which basically based on two thresholds, evicts pods which don't fall between a range but is sub optimal. 

Currently there is a KEP for even pod spreading, but that will take some time to come in. In the meantime , we need a policy in descheduler which does  the following:-
- forms a topology domain which is combination of region and zone
- figures out nodes in the same bucket by using the bucket key in step 1
- selects which pods this policy applies based on a labelSelector
- finds buckets, where the difference in the number of pods(selected by labelSelector) in that bucket and any other bucket is maximum, and starts evicting pods in that bucket in the hopes that it will schedule on a node in a different bucket

krmayankk:
FYI @ravisantoshgudimetla @bsalamat @Huang-Wei

bsalamat:
@krmayankk As you know, we are adding even pod spreading in 1.15. Given that effort, I don't think it makes sense to add the same (or a very similar) feature in parallel to another component. If timing is a concern, you may want to fork a component and customize it for you use-cases in the meantime that we release even pod spreading.
That being said, it makes sense to add support for even pod spreading to Descheduler. The scheduler enforces even pod spreading at the time of scheduling pods. Descheduler can enforce the feature at run-time. If we go with this approach, we should ensure that Descheduler uses the same API and does not add its own API for the feature.

krmayankk:
> I don't think it makes sense to add the same (or a very similar) feature in parallel to another component.

Not sure i understand. descheduler will just evict if the pod imbalance is detected. Are you against adding this in descheduler ?

> Descheduler can enforce the feature at run-time. If we go with this approach, we should ensure that Descheduler uses the same API and does not add its own API for the feature.

Descheduler will just do this at runtime based on a new policy. Not sure i understand about the same API, since descheduler cannot work on an API that is being still worked on . May you are saying, that later descheduler can be modified to understand the pod topology fields being introduced in the even pod spreading . Is my understanding correct ?

[...]

PR #154, opened 2019-05-23
Runtime Even Pod Spreading 

krmayankk:
Fixes https://github.com/kubernetes-incubator/descheduler/issues/146

API based on [KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-scheduling/20190221-even-pods-spreading.md)

The Descheduler policy is basically TopologyConstraints per namespace. Its described as follows:
```
kind: "DeschedulerPolicy"
strategies:
  "TopologySpreadConstraint":
     enabled: true
     params:
        namespacedtopologyspreadconstraints:
         - namespace: sam-system
           topologyspreadconstraints:
            - maxSkew: 1
              topologyKey: failure-domain.beta.kubernetes.io/zone
              labelSelector:
                      matchLabels:
                              apptype: server
```


krmayankk:
FYI @ravisantoshgudimetla @bsalamat @Huang-Wei @aveshagarwal this PR is to initiate API discussion

pkg/descheduler/strategies/topologyspreadyconstraint.go
```diff
@@ -0,0 +1,150 @@
+/*
+Copyright 2017 The Kubernetes Authors.
```
krmayankk: TODO: Add UT once we reach consensus on the api

pkg/api/types.go
```diff
@@ -58,3 +61,32 @@ type NodeResourceUtilizationThresholds struct {
...
+	// MaxSkew describes the degree to which pods may be unevenly distributed.
+	// It's the maximum permitted difference between the number of matching pods in
+	// any two topology domains of a given topology type.
+	// For example, in a 3-zone cluster, currently pods with the same labelSelector
+	// are "spread" such that zone1 and zone2 each have one pod, but not zone3.
+	// - if MaxSkew is 1, incoming pod can only be scheduled to zone3 to become 1/1/1;
+	// scheduling it onto zone1(zone2) would make the ActualSkew(2) violate MaxSkew(1)
+	// - if MaxSkew is 2, incoming pod can be scheduled to any zone.
+	// It's a required value. Default value is 1 and 0 is not allowed.
+	MaxSkew int32
```
ravisantoshgudimetla: @Huang-Wei - I believe this is inline with what you're proposing...

pkg/descheduler/strategies/topologyspreadyconstraint.go
ravisantoshgudimetla: Change it to 2019..

pkg/descheduler/strategies/topologyspreadyconstraint.go
```diff
@@ -0,0 +1,150 @@
...
+	"github.com/kubernetes-incubator/descheduler/pkg/descheduler/evictions"
+	podutil "github.com/kubernetes-incubator/descheduler/pkg/descheduler/pod"
+)
+
+func TopologySpreadConstraint(ds *options.DeschedulerServer, strategy api.DeschedulerStrategy, policyGroupVersion string, nodes []*v1.Node, nodePodCount nodePodEvictedCount) {
+	if !strategy.Enabled {
+		return
+	}
+
+	fmt.Printf("Found following parameters for TopologySpreadConstraint %v\n", strategy)
```
ravisantoshgudimetla: Please change to klog or glog..

pkg/descheduler/strategies/topologyspreadyconstraint.go
```diff
@@ -0,0 +1,150 @@
...
+// add the pod with key as the combination of topokey and its value
+// find the min value of pods with any key
+// iterate through all keys and diff currentPods -minPods <=maxSkew
+// if diff > maxSkew, select a pod in the current bucket for eviction
+//
+func evictPodsViolatingSpreadConstraints(client clientset.Interface, policyGroupVersion string, nodes []*v1.Node, dryRun bool, nodePodCount nodePodEvictedCount,
+	constraint api.NamespacedTopologySpreadConstraint) {
+
+	if len(constraint.TopologySpreadConstraints) != 1 {
+		glog.V(1).Infof("We currently only support 1 topology spread constraint per namespace")
```
ravisantoshgudimetla: I think, we need to explicitly document this along with reason.

pkg/descheduler/strategies/topologyspreadyconstraint.go
```diff
@@ -0,0 +1,150 @@
...
+		if err != nil {
+		}
+		for _, pod := range pods {
+			if pod.Namespace != constraint.Namespace {
+				continue
+			}
+			// does this pod labels match the constraint label selector
+			// TODO: This is intentional that it only looks at the first constraint
+			selector, err := metav1.LabelSelectorAsSelector(constraint.TopologySpreadConstraints[0].LabelSelector)
+			if err != nil {
```
ravisantoshgudimetla: log and continue..

pkg/descheduler/strategies/topologyspreadyconstraint.go
```diff
@@ -0,0 +1,150 @@
...
+				continue
+			}
+			if !selector.Matches(labels.Set(pod.Labels)) {
+				continue
```
ravisantoshgudimetla: same as above, if you think, you're going to log heavily, please increase the log-level or append all the failures and log them at the end

pkg/descheduler/strategies/topologyspreadyconstraint.go
```diff
@@ -0,0 +1,150 @@
...
+			minPodsForGivenTopo = len(v)
+		}
+	}
+
+	for _, v := range topologyPairToPods {
+		podsInTopo := len(v)
+		if int32(podsInTopo-minPodsForGivenTopo) >= constraint.TopologySpreadConstraints[0].MaxSkew {
+			//we need to evict maxSkew-(podsInTopo-minPodsForGivenTopo))
+			countToEvict := constraint.TopologySpreadConstraints[0].MaxSkew - int32(podsInTopo-minPodsForGivenTopo)
+			podsListToEvict := GetPodsToEvict(countToEvict, v)
```
ravisantoshgudimetla: Why is this public method? Do you want to expose it for testing?

pkg/descheduler/strategies/topologyspreadyconstraint.go
```diff
@@ -0,0 +1,150 @@
...
+
+func addTopologyPair(topoMap map[topologyPair]podSet, pair topologyPair, pod *v1.Pod) {
+	if topoMap[pair] == nil {
+		topoMap[pair] = make(map[*v1.Pod]struct{})
+	}
+	topoMap[pair][pod] = struct{}{}
+}
+
+// GetPodFullName returns a name that uniquely identifies a pod.
+func GetPodFullName(pod *v1.Pod) string {
```
ravisantoshgudimetla: Where are we using this function?

pkg/descheduler/strategies/topologyspreadyconstraint.go
```diff
@@ -0,0 +1,150 @@
...
+			// does this pod labels match the constraint label selector
+			// TODO: This is intentional that it only looks at the first constraint
+			selector, err := metav1.LabelSelectorAsSelector(constraint.TopologySpreadConstraints[0].LabelSelector)
+			if err != nil {
+				continue
+			}
+			if !selector.Matches(labels.Set(pod.Labels)) {
+				continue
+			}
+			// TODO: Need to determine if the topokey already present in the node or not
```
ravisantoshgudimetla: I think, we need to address this as part of the PR.

pkg/api/types.go
Huang-Wei: Yes.

@krmayankk if the upstream API is available, I think you can simply vendor that?

pkg/descheduler/descheduler.go
```diff
@@ -60,11 +60,13 @@ func Run(rs *options.DeschedulerServer) error {
 		return nil
 	}
 
+	glog.V(1).Infof("Reached here \n")
```
Huang-Wei: TODO: remove this upon merging.

pkg/descheduler/strategies/topologyspreadyconstraint.go
```diff
@@ -0,0 +1,150 @@
...
+	podutil "github.com/kubernetes-incubator/descheduler/pkg/descheduler/pod"
+)
+
+func TopologySpreadConstraint(ds *options.DeschedulerServer, strategy api.DeschedulerStrategy, policyGroupVersion string, nodes []*v1.Node, nodePodCount nodePodEvictedCount) {
+	if !strategy.Enabled {
+		return
+	}
+
+	fmt.Printf("Found following parameters for TopologySpreadConstraint %v\n", strategy)
+	for _, topoConstraints := range strategy.Params.NamespacedTopologySpreadConstraints {
```
Huang-Wei: s/topoConstraints/topoConstraint/

pkg/descheduler/strategies/topologyspreadyconstraint.go
```diff
@@ -0,0 +1,150 @@
...
+type topologyPair struct {
+	key   string
+	value string
+}
+
+type podSet map[*v1.Pod]struct{}
+
+type topologyPairSet map[topologyPair]struct{}
+
+// finnd all nodes
```
Huang-Wei: Please reword the comments.

pkg/descheduler/strategies/topologyspreadyconstraint.go
Huang-Wei: The `topoConstraints` is sort of "aggregated" topologyConstraints group by namespace?

Huang-Wei:
Please remove pkg/descheduler/strategies/.pod_antiaffinity.go.swp.

pkg/descheduler/strategies/topologyspreadyconstraint.go
Huang-Wei: Agree. This is a significant limitation.

Huang-Wei: I'm lean to delivering a full implementation which respects all constraints.

pkg/descheduler/strategies/topologyspreadyconstraint.go
```diff
@@ -0,0 +1,150 @@
...
+	constraint api.NamespacedTopologySpreadConstraint) {
+
+	if len(constraint.TopologySpreadConstraints) != 1 {
+		glog.V(1).Infof("We currently only support 1 topology spread constraint per namespace")
+		return
+	}
+	topologyPairToPods := make(map[topologyPair]podSet)
+	for _, node := range nodes {
+		glog.V(1).Infof("Processing node: %#v\n", node.Name)
+		pods, err := podutil.ListEvictablePodsOnNode(client, node, false)
```
Huang-Wei: I believe clientset is able to filter pods by namespace. We can pass in `namespace` hence don't need to check if `pod.Namespace != constraint.Namespace` later.

pkg/descheduler/strategies/topologyspreadyconstraint.go
```diff
@@ -0,0 +1,150 @@
...
+		if err != nil {
```
Huang-Wei: Incomplete? Should log and continue.

pkg/descheduler/strategies/topologyspreadyconstraint.go
Huang-Wei: Listing pods for each node will cause N API requests (N=len(nodes)). It's better to change the logic to:

```
for each topologySpreadConstraint
    pre-fetch all qualified pods globally (or use similar cache in descheduler if appropriate)
    process the pods list:
	(1) filter out the pods whose node doesn't have needed topologyKey
	(2) so that we can get a map[nodeName]podSet, and also we know the minimum match number
	for nodeName, podSet := range map[nodeName]podSet
		needEvictNum := len(podSet) - minMatch - maxSkew
		if needEvictNum > 0
			evict needEvictNum pods from this Node
			Note: this math is sort of brute force, we can come up with better math later.
			For example, for a 5/1/3 cluster, and maxSkew is 1; the brute force math above will
			evict 3/0/1 pods from each topology, but if we consider the math "dynamically", we
			should only evict 2/0/0 pods, so that eventually it can become 3/3/3.
```

pkg/descheduler/strategies/topologyspreadyconstraint.go
```diff
@@ -0,0 +1,150 @@
...
+	minPodsForGivenTopo := math.MaxInt32
+	for _, v := range topologyPairToPods {
+		if len(v) < minPodsForGivenTopo {
+			minPodsForGivenTopo = len(v)
+		}
+	}
+
+	for _, v := range topologyPairToPods {
+		podsInTopo := len(v)
+		if int32(podsInTopo-minPodsForGivenTopo) >= constraint.TopologySpreadConstraints[0].MaxSkew {
```
Huang-Wei: It should be `>`. Otherwise, for a 2/1/1 cluster, you're gonna evict one pod from topology1 - which is unnecessary.

pkg/descheduler/strategies/topologyspreadyconstraint.go
Huang-Wei: When multiple constraints are considered, it'd good to postpone the pickup process, so that it would be possible to pick the same Pod satisfying multiple constraints. For example, for a 2-constraint case, the calculation result decides to evict 3/2 pods. If we postpone the pickup logic, we may come up with a solution evicting 3 pods, instead of 5(3+2) pods.

pkg/descheduler/strategies/topologyspreadyconstraint.go
```diff
@@ -0,0 +1,150 @@
...
+			//we need to evict maxSkew-(podsInTopo-minPodsForGivenTopo))
+			countToEvict := constraint.TopologySpreadConstraints[0].MaxSkew - int32(podsInTopo-minPodsForGivenTopo)
```
Huang-Wei: It should be "len(podsInTopo) - minPodsForGivenTopo - countToEvict".

pkg/api/types.go
krmayankk: Currently the descheduler is looking at TopologySpreadConstraint's as defined in DeschedulerPolicy. Later these constraints will come directly from the Pod itself. So there is nothing to vendor, just the source of the TopologySpreadConstraint will change @Huang-Wei

pkg/descheduler/strategies/topologyspreadyconstraint.go
krmayankk: I have described the DeschedulerPolicy in the description. Here is what it looks like: Basically this loop is just reading the per namespace constraints specified in the policy.  May be `TopologySpreadConstraintsPerNamespace` ?

```
kind: "DeschedulerPolicy"
strategies:
  "TopologySpreadConstraint":
     enabled: true
     params:
        namespacedtopologyspreadconstraints:
         - namespace: sam-system
           topologyspreadconstraints:
            - maxSkew: 1
              topologyKey: failure-domain.beta.kubernetes.io/zone
              labelSelector:
                      matchLabels:
                              apptype: server
```

pkg/descheduler/strategies/topologyspreadyconstraint.go
krmayankk: @Huang-Wei so per namespace, do an AND of all constraints ?

Huang-Wei: @krmayankk Yes.

Huang-Wei:
@krmayankk I'm not sure I like the API design in policy "namespacedtopologyspreadconstraints".

I checked existing descheduler policy [examples](https://github.com/kubernetes-incubator/descheduler/tree/master/examples), and all of them offers pretty simple parameters to evict pods violating that kind of policy.

IMO we should offer a neat policy for EvenPodsSpread, i.e. just enabled or disable, or probably additionally provides an option like AntiAffinity in case upstream feature supports more `v1.UnsatisfiableConstraintAction` options:

https://github.com/kubernetes-incubator/descheduler/blob/9e28f0b362ea5afa6ef4ec15f95cd5fc7eaf108a/examples/node-affinity.yml#L4-L8

With current API, users must explicitly specify every topologySpreadConstraints which isn't practical.

So I want to stop reviewing here to get a concensus on the API first.

cc @ravisantoshgudimetla on the API design.

pkg/api/types.go
```diff
@@ -48,6 +48,9 @@ type DeschedulerStrategy struct {
 type StrategyParameters struct {
 	NodeResourceUtilizationThresholds NodeResourceUtilizationThresholds
 	NodeAffinityType                  []string
+	// TopologySpreadConstraints describes how a group of pods should be spread across topology
+	// domains. Descheduler will use these constraints to decide which pods to evict.
+	NamespacedTopologySpreadConstraints []NamespacedTopologySpreadConstraint
```
bsalamat: Isn't this name unnecessarily long? It feels a bit like adding part of its documentation to the name.

seanmalloy:
@krmayankk are you planning to continue working on this pull request?

It would be great if you could rebase and resolve the merge conflicts. I believe the even pod spreading feature in the scheduler is being promoted to beta in the k8s v1.18, so this will be a very useful feature once k8s v1.18 is released.

Huang-Wei:
@seanmalloy Correct, the PodTopologySpread (featuregate EvenPodsSpread) is gonna be beta in 1.18. So it makes great sense to vendor the 1.18 k/k codebase upon the implementation of this PR.

seanmalloy:
@krmayankk the master branch has been updated with the k/k v1.18 vendor dependencies. Are you planning on continuing to work on this pull request?

I'm willing to continue working on this feature and use your original commits as a starting point if you do not have time to complete this work.

Thanks!

seanmalloy:
/close

I started working on the updated code based on this PR. The new branch is is here: https://github.com/KohlsTechnology/descheduler/tree/evenpod. I'm hoping to submit a new PR in the next few weeks. See also, my comment regarding possible API changes https://github.com/kubernetes-sigs/descheduler/issues/146#issuecomment-637287180.

PR #156, opened 2019-05-24
Added ownerRef.UID for evaluating duplicate Pods

### LINKED ISSUE #147: Identical deployments in different namespaces
I have two identical statefulsets (would be the same for deployment too, I suppose), in two different namespaces. They have the same name of the set, same name of the service, and they have the same labels.

They both work in terms of "pure" k8s. They control their own three pods, respectively.

However, one of the sets keep getting evicted by descheduler. I assume the descheduler does not consider namespace of the pods? Is this intended?

tammert:
PR fixes #147.

In short, added ownerRef.UID in addition to ownerRef.Kind and ownerRef.Name when evaluating duplicates. On clusters with multiple Namespaces that have an identical Deployment or StatefulSet (think `dev` and `test` namespaces), the added UID ensures that only Pods that are managed by the same ReplicaSet are considered for eviction.

Tested on a 3-node DigitalOcean cluster:
```
$ kubectl get rs --all-namespaces | grep nginx
default       nginx-7db9fccd9b             3         3         3       8m57s
testing       nginx-7db9fccd9b             3         3         3       74s
$ ./_output/bin/descheduler --kubeconfig k8s-1-14-2-do-0-fra1-1558715482863-kubeconfig.yaml --policy-config-file policyfile.yaml --v=3
I0524 18:48:54.138997    7838 reflector.go:202] Starting reflector *v1.Node (1h0m0s) from github.com/kubernetes-incubator/descheduler/pkg/descheduler/node/node.go:84
I0524 18:48:54.139373    7838 reflector.go:240] Listing and watching *v1.Node from github.com/kubernetes-incubator/descheduler/pkg/descheduler/node/node.go:84
I0524 18:48:54.239137    7838 duplicates.go:50] Processing node: "pool-6g7junthk-ya8q"
I0524 18:48:54.491948    7838 duplicates.go:50] Processing node: "pool-6g7junthk-ya3y"
I0524 18:48:54.560602    7838 duplicates.go:50] Processing node: "pool-6g7junthk-ya3f"

### DIFF
--- a/pkg/descheduler/strategies/duplicates_test.go
+++ b/pkg/descheduler/strategies/duplicates_test.go
@@ -31,26 +31,43 @@ import (
 func TestFindDuplicatePods(t *testing.T) {
 	node := test.BuildTestNode("n1", 2000, 3000, 10)
 	p1 := test.BuildTestPod("p1", 100, 0, node.Name)
+	p1.Namespace = "dev"
 	p2 := test.BuildTestPod("p2", 100, 0, node.Name)
+	p2.Namespace = "dev"
 	p3 := test.BuildTestPod("p3", 100, 0, node.Name)
+	p3.Namespace = "dev"
 	p4 := test.BuildTestPod("p4", 100, 0, node.Name)
 	p5 := test.BuildTestPod("p5", 100, 0, node.Name)
 	p6 := test.BuildTestPod("p6", 100, 0, node.Name)
 	p7 := test.BuildTestPod("p7", 100, 0, node.Name)
+	p7.Namespace = "kube-system"
 	p8 := test.BuildTestPod("p8", 100, 0, node.Name)
+	p8.Namespace = "test"
 	p9 := test.BuildTestPod("p9", 100, 0, node.Name)
+	p9.Namespace = "test"
+	p10 := test.BuildTestPod("p10", 100, 0, node.Name)
+	p10.Namespace = "test"
 
-	// All the following pods expect for one will be evicted.
-	p1.ObjectMeta.OwnerReferences = test.GetReplicaSetOwnerRefList()
-	p2.ObjectMeta.OwnerReferences = test.GetReplicaSetOwnerRefList()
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by aveshagarwal** [comment]
@tammert thanks for this PR. Could you add some tests too?

**Comment by tammert** [comment]
@aveshagarwal thanks for the suggestion, I've added the tests for this scenario (and performed a little cleanup on the `duplicates_test.go` file). It wasn't actually working correctly due to an incorrect `maxPodsToEvict` setting, so I've rectified that as well.

When adding the test, I came to the conclusion that it's more natural to include the Namespace of the Pod instead of the UID of the ReplicaSet; you should see this in the second commit.

Furthermore, I've opened issue #157 from what I saw while working on this. I didn't want to include it in this PR to prevent big scope creep. Let me know if you agree with that?

File: pkg/descheduler/strategies/duplicates_test.go [hunk 1]
```diff
@@ -62,24 +79,30 @@ func TestFindDuplicatePods(t *testing.T) {
...
+		return true, &v1.PodList{Items: []v1.Pod{*p1, *p2, *p3, *p4, *p5, *p6, *p7, *p8, *p9, *p10}}, nil
 	})
 	fakeClient.Fake.AddReactor("get", "nodes", func(action core.Action) (bool, runtime.Object, error) {
 		return true, node, nil
 	})
+
+	expectedEvictedPodCount := 4
 	npe := nodePodEvictedCount{}
 	npe[node] = 0
-	podsEvicted := deleteDuplicatePods(fakeClient, "v1", []*v1.Node{node}, false, npe, 2, false)
```
tammert: Since `maxPodsEvicted` was 2, the test would pretty much always succeed due to `expectedEvictedPodCount` being the same value. I've upped `maxPodsEvicted` to 10 now, the amount of Pods currently created.

tammert:
@aveshagarwal is there anything I can do to move this PR forwards? Otherwise I'm awaiting for review and approval :).

aveshagarwal:
/lgtm
/approve

@tammert thanks for this PR.

PR #158, opened 2019-06-04
Adding support for go modules

dndungu:
This PR adds support for Golang Modules.

I created the PR by running.
`go mod init github.com/kubernetes-incubator/descheduler`
Go mod init automatically detected that there is a `glide.lock` file in the repo. It them imported the dependencies defined in the lock file into `go.mod`

I then run `go test ./...` from the root directory of the repo to generate Go Module lock file `go.sum`

[Go Modules](https://github.com/golang/go/wiki/Modules) is now the officially supported dependency management tool for Golang.


seanmalloy:
The `Dockerfile` references Go Lang version 1.9.2. I think the `Dockerfile` will need to be updated to at least use Go Lang version 1.11.

seanmalloy:
I think files `glide.lock` and `glide.yaml` should also be removed. The `go build` command will also need to be updated in the `Makefile` with environment variable`GO111MODULE=on`.

seanmalloy:
Can the `vendor` directory be removed too?

dndungu:
@seanmalloy, 

Thanks for the review.

I have removed `glide.lock`, `glide.yaml`, and `vendor` and updated `Dockerfile` `.travis.yml` to use Go 1.12

seanmalloy:
@dndungu I think I gave you incorrect advice to remove the `vendor` directory. It looks like the Kubernetes project is converting to Go modules, but keeping the `vendor` directory. See the below links for details.

https://github.com/kubernetes/enhancements/blob/master/keps/sig-architecture/2019-03-19-go-modules.md
https://github.com/kubernetes/kubernetes/pull/74877

Dockerfile
```diff
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-FROM golang:1.9.2
+FROM golang:1.12
```
seanmalloy: Maybe change this to `golang:1.12.6`?

@ravisantoshgudimetla and @aveshagarwal what do you think?

seanmalloy:
@dndungu I was able to successfully build the binary and run the unit tests. I'm having some difficulty running the the e2e tests on my machine.

Were you able to run the e2e tests(`make test-e2e`)?

I'm going to try and run the built binary from this branch against some clusters to see if it is working as expected. Hoping to do this by the end of this week.

Dockerfile
seanmalloy: Looks like Kubernetes 1.15 is built using `go1.12.5`.

yeya24:
I am curious about the status of this PR. Can we move forward this?

damemi:
this PR would be superceded by https://github.com/kubernetes-sigs/descheduler/pull/216, which also switches us to go modules and bumps dependencies
/close

PR #159, opened 2019-06-04
Adds K8s manifests

dndungu:
This PR adds K8s manifests to get Descheduler running as a CronJob.


dndungu:
/close 
🤦‍♂ just realized there is a directory `kubernetes` with some manifests.

PR #162, opened 2019-06-11
Improved logic for duplicates testing

### LINKED ISSUE #157: Test logic for duplicates can be improved
While adding a test for PR #156 is noticed the test logic in `duplicates_test.go` is slightly flawed. Currently, a lot of different tests are performed and only the cumulative result is considered. By doing this, if two tests fail (one Pod that should not be evicted *is*, and a Pod that should be evicted *is not*) they could cancel each other out so the test would pass when it should not.

I would suggest rewriting the tests so that each individual scenario is run and considered separately. If you agree, I'd be happy to do this so you can assign this to me.

tammert:
Fixes #157.

Note for reviewer: renamed `tests` to `testCases` because of `test` import collision.

### DIFF
--- a/pkg/descheduler/strategies/duplicates_test.go
+++ b/pkg/descheduler/strategies/duplicates_test.go
@@ -86,64 +86,58 @@ func TestFindDuplicatePods(t *testing.T) {
 	// A Critical Pod.
 	p7.Annotations = test.GetCriticalPodAnnotation()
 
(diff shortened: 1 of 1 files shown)


aveshagarwal:
thanks, will review it soon.

aveshagarwal:
thanks @tammert 

/lgtm
/approve

Issue #144, opened 2019-04-16
Official Docker image


anton-johansson:
Where is the official Docker image located? The [job example](../blob/master/kubernetes/job.yaml) says `descheduler:latest`, however, when I try to pull that image I just get:
```
➜ docker pull descheduler:latest
Error response from daemon: pull access denied for descheduler, repository does not exist or may require 'docker login'
```

warmchang:
DIY?

![image](https://user-images.githubusercontent.com/13084624/57191730-49a88a00-6f5b-11e9-9577-70b0a8b083de.png)

anton-johansson:
Yeah, I was hoping that we could avoid building and hosting the image ourselves. :(

aveshagarwal:
@anton-johansson yes we have not pushed official descheduler images. In case you would want to help, you could look into the location/process kubernetes is using to push official images and we could use the same.

aveshagarwal:
@rhockenbury

rhockenbury:
It would be great to explore pushing an image to gcr.io, but in the meantime, would it be reasonable  to throw the latest image up on dockerhub? - https://hub.docker.com/r/aveshagarwal/descheduler

aveshagarwal:
@rhockenbury  i have pushed 0.9.0 to docker hub: docker.io/aveshagarwal/descheduler:0.9.0

seanmalloy:
I'm interested in working on this issue. Looks like the official container image would be something like `k8s.gcr.io/descheduler:0.9.0`. I'll try to research how to get this setup.

aveshagarwal:
@seanmalloy thanks for working on this issue.

seanmalloy:
I submitted #160 to get the `Makefile` updated with an option to push the container images.

seanmalloy:
The `Makefile` changes have been merged to master.

I attended the SIG Scheduling meeting today, and it was recommend to reach out to @ravisantoshgudimetla to facilitate getting the official image pushed to k8s.gcr.io.

@ravisantoshgudimetla can you provide any insight on how to push the container image to k8s.gcr.io Or how would we setup an automated process to push the container images k8s.gcr.io?

seanmalloy:
I submitted PR https://github.com/kubernetes/k8s.io/pull/296 to get the GCE staging repo created.

rhockenbury:
Is this now waiting on just a new release, and maybe a readme update?

seanmalloy:
[...]

PR #164, opened 2019-06-21
Update example Job with k8s.gcr.io image location

seanmalloy:


seanmalloy:
Should not merge until descheduler container image is pushed to k8s.gce.io. See #144 for progress updates on container image upload.

PR #168, opened 2019-07-11
Remove hardcoded node selection in E2e test 

Bowenislandsong:
/kind flake

The e2e test was hardcoded to selecting the 3rd node as default leastLoadedNode. This hard coding does not guarantee to select the least loaded node if the testing cluster is set up differently.

Fixes # This fix help selects the actual least loaded node based on the number of pods assigned to each node after RC is set up. It also skips the master node.

/cc @ravisantoshgudimetla

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ sudo: false
 language: go
 
 go:
-  - 1.9.1
+  - "1.10"
 services:
   - docker
 before_script:
@@ -19,6 +19,7 @@ before_script:
   - cp "$(kind get kubeconfig-path --name="kind")" /tmp/admin.conf
 script:
 - hack/verify-gofmt.sh
+- make lint
 - make build
 - make test-unit
 - make test-e2e
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,22 @@ sudo: false
 language: go
 
 go:
-- "1.10"
-
+  - 1.9.1
+services:
+  - docker
+before_script:
+  - curl -Lo kubectl \https://storage.googleapis.com/kubernetes-release/release/v1.13.0/bin/linux/amd64/kubectl && chmod +x kubectl && sudo mv kubectl /usr/local/bin/
+  - wget https://github.com/kubernetes-sigs/kind/releases/download/0.2.1/kind-linux-amd64
+  - chmod +x kind-linux-amd64
+  - mv kind-linux-amd64 kind
+  - export PATH=$PATH:$PWD
+  - kind create cluster --config=$TRAVIS_BUILD_DIR/hack/kind_config.yaml
+  - export KUBECONFIG="$(kind get kubeconfig-path)"
+  - docker pull kubernetes/pause
+  - kind load docker-image kubernetes/pause
+  - cp "$(kind get kubeconfig-path --name="kind")" /tmp/admin.conf
 script:
-  - make lint
-  - make build
-  - make test-unit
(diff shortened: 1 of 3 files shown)


test/e2e/e2e_test.go
```diff
@@ -119,7 +121,10 @@ func startEndToEndForLowNodeUtilization(clientset clientset.Interface) {
 
 func TestE2E(t *testing.T) {
 	// If we have reached here, it means cluster would have been already setup and the kubeconfig file should
-	// be in /tmp directory.
+	// be in /tmp directory as admin.conf.
+
```
ravisantoshgudimetla: Remove white spaces

test/e2e/e2e_test.go
```diff
@@ -130,20 +135,31 @@ func TestE2E(t *testing.T) {
...
-	leastLoadedNode := nodeList.Items[2]
 	rc := RcByNameContainer("test-rc", int32(15), map[string]string{"test": "app"}, nil)
 	_, err = clientSet.CoreV1().ReplicationControllers("default").Create(rc)
 	if err != nil {
 		t.Errorf("Error creating deployment %v", err)
 	}
-	podsOnleastUtilizedNode, err := podutil.ListPodsOnANode(clientSet, &leastLoadedNode)
-	if err != nil {
-		t.Errorf("Error listing pods on a node %v", err)
+	podsBefore := math.MaxInt16
```
ravisantoshgudimetla: nit: Perhaps having a map with node[podCount] would be easier.

test/e2e/e2e_test.go
```diff
@@ -130,20 +135,31 @@ func TestE2E(t *testing.T) {
...
+		if tmpLoads := len(podsOnANode); tmpLoads < podsBefore {
+			leastLoadedNode = nodeList.Items[i]
+			podsBefore = tmpLoads
+		}
 	}
-	podsBefore := len(podsOnleastUtilizedNode)
 	t.Log("Eviction of pods starting")
 	startEndToEndForLowNodeUtilization(clientSet)
-	podsOnleastUtilizedNode, err = podutil.ListPodsOnANode(clientSet, &leastLoadedNode)
+	podsOnleastUtilizedNode, err := podutil.ListPodsOnANode(clientSet, &leastLoadedNode)
```
ravisantoshgudimetla: Change this statement to `https://github.com/kubernetes-incubator/descheduler/pull/168/files#diff-6413e1dbafd6dd63b5d907ea3ae6f6f9R168`

.travis.yml
```diff
@@ -3,9 +3,23 @@ sudo: false
...
+  - export KUBECONFIG="$(kind get kubeconfig-path)"
+  - docker pull kubernetes/pause
+  - kind load docker-image kubernetes/pause
+  - cp "$(kind get kubeconfig-path --name="kind")" /tmp/admin.conf
 script:
-  - make lint
-  - make build
-  - make test-unit
+- hack/verify-gofmt.sh
+- make lint
```
ravisantoshgudimetla: Why this commit coming in? Can you check this @Bowenislandsong

Bowenislandsong: this is to accommodate https://github.com/kubernetes-incubator/descheduler/commit/dee89a6cc1ec00ed699a09ed632ca3ed68232a92 It asked for 'gofmt' which I didn't do. I will use 'gofmt' for all the code in the update and then it should not be a problem anymore.

PR #169, opened 2019-07-11
Remove hardcoded node selection in E2e test 

Bowenislandsong:
/kind flake

The e2e test was hardcoded to selecting the 3rd node as default leastLoadedNode. This hard coding does not guarantee to select the least loaded node if the testing cluster is set up differently.

Fixes # This fix help selects the actual least loaded node based on the number of pods assigned to each node after RC is set up. It also skips the master node.

/cc @ravisantoshgudimetla

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ sudo: false
 language: go
 
 go:
-  - 1.9.1
+  - 1.11.1
 services:
   - docker
 before_script:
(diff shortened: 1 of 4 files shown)


Bowenislandsong:
> /lgtm
> 
> Thanks for the @Bowenislandsong . Can you check if the pods are actually getting evicted if tested locally? Meaning can you manually increase the log level and verify that pods are getting evicted?

I did check by stepping through the code in debug mode, I will run the test again in a bit by showing the logging. Thank you.

Bowenislandsong:
We expect the number of pods on the least utilized node to be 7 before eviction test and 8 after.

PR #171, opened 2019-07-22
Adding Pod Antiaffinity e2e test

Bowenislandsong:
Adding e2e test for descheduler strategy based on Pod Anti-Affinity rule. The RC is created with pod anti-affinity. After pods started to run on nodes, we label a pod that matches the anti-affinity label and then invoke evicting strategy. 

Pods will be evicted by checking if any pods within the node has a label that matches exactly the label from pod anti-affinity. By labeling one pod inside the node, all other pods will be evicted leaving the pod with the label change inside the node.

The NodePodEvictedCount is changed to a public type so that we can both pass the strategy functions

/cc @ravisantoshgudimetla

### DIFF
--- a/pkg/descheduler/pod/pods.go
+++ b/pkg/descheduler/pod/pods.go
@@ -78,7 +78,7 @@ func ListEvictablePodsOnNode(client clientset.Interface, node *v1.Node, evictLoc
 }
 
 func ListPodsOnANode(client clientset.Interface, node *v1.Node) ([]*v1.Pod, error) {
-	fieldSelector, err := fields.ParseSelector("spec.nodeName=" + node.Name + ",status.phase!=" + string(api.PodSucceeded) + ",status.phase!=" + string(api.PodFailed))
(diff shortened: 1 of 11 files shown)


pkg/descheduler/pod/pods.go
```diff
@@ -78,7 +78,7 @@ func ListEvictablePodsOnNode(client clientset.Interface, node *v1.Node, evictLoc
 }
 
 func ListPodsOnANode(client clientset.Interface, node *v1.Node) ([]*v1.Pod, error) {
-	fieldSelector, err := fields.ParseSelector("spec.nodeName=" + node.Name + ",status.phase!=" + string(api.PodSucceeded) + ",status.phase!=" + string(api.PodFailed))
+	fieldSelector, err := fields.ParseSelector("spec.nodeName=" + node.Name + ",status.phase=" + string(api.PodRunning))
```
ravisantoshgudimetla: Can you explain why this is needed?

/cc @aveshagarwal

PR #175, opened 2019-08-12
Strategy to consider taints and tolerations in Descheduler

swatisehgal:
PR to address: https://github.com/kubernetes-incubator/descheduler/issues/131

### DIFF
--- a/pkg/descheduler/strategies/node_taint.go
+++ b/pkg/descheduler/strategies/node_taint.go
@@ -1,0 +1,138 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package strategies
+
+import (
+	"sigs.k8s.io/descheduler/cmd/descheduler/app/options"
+	"sigs.k8s.io/descheduler/pkg/api"
+	"sigs.k8s.io/descheduler/pkg/descheduler/evictions"
+	podutil "sigs.k8s.io/descheduler/pkg/descheduler/pod"
+
+	"k8s.io/api/core/v1"
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/klog"
+)
+
+const (
+	TolerationOpExists v1.TolerationOperator = "Exists"
+	TolerationOpEqual  v1.TolerationOperator = "Equal"
+)
+
+// RemovePodsViolatingNodeTaints with elimination strategy
+func RemovePodsViolatingNodeTaints(ds *options.DeschedulerServer, strategy api.DeschedulerStrategy, policyGroupVersion string, nodes []*v1.Node, nodePodCount nodePodEvictedCount) {
+	if !strategy.Enabled {
+		return
+	}
+	deletePodsViolatingNodeTaints(ds.Client, policyGroupVersion, nodes, ds.DryRun, nodePodCount, ds.MaxNoOfPodsToEvictPerNode, ds.EvictLocalStoragePods)
+}
+
+// deletePodsViolatingNodeTaints evicts pods on the node which violate NoSchedule Taints on nodes
+func deletePodsViolatingNodeTaints(client clientset.Interface, policyGroupVersion string, nodes []*v1.Node, dryRun bool, nodePodCount nodePodEvictedCount, maxPodsToEvict int, evictLocalStoragePods bool) int {
+	podsEvicted := 0
+	for _, node := range nodes {
+		klog.V(1).Infof("Processing node: %#v\n", node.Name)
+		pods, err := podutil.ListEvictablePodsOnNode(client, node, evictLocalStoragePods)
+		if err != nil {
+			//no pods evicted as error encountered retrieving evictable Pods
+			return 0
+		}
+		totalPods := len(pods)
+		for i := 0; i < totalPods; i++ {
+			if maxPodsToEvict > 0 && nodePodCount[node]+1 > maxPodsToEvict {
+				break
+			}
+			if !checkPodsSatisfyTolerations(pods[i], node) {
+				success, err := evictions.EvictPod(client, pods[i], policyGroupVersion, dryRun)
+				if !success {
+					klog.Errorf("Error when evicting pod: %#v (%#v)\n", pods[i].Name, err)
+				} else {
+					nodePodCount[node]++
+					klog.V(1).Infof("Evicted pod: %#v (%#v)", pods[i].Name, err)
+				}
+			}
+		}
+		podsEvicted += nodePodCount[node]
+	}
+	return podsEvicted
+}
+
+// checkPodsSatisfyTolerations checks if the node's taints (NoSchedule) are still satisfied by pods' tolerations.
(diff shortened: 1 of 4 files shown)


pkg/descheduler/strategies/node_taint.go
```diff
@@ -0,0 +1,148 @@
...
+	if !allTolerated {
+		glog.V(2).Infof("Not all taints are tolerated after update for Pod %v on node %v", pod.Name, node.Name)
+		return false
+	}
+	return true
+}
+
+// getNoScheduleTaints return a slice of NoSchedule taints from the a slice of taints that it receives.
+func getNoScheduleTaints(taints []v1.Taint) []v1.Taint {
+
```
damemi: extra line

pkg/descheduler/strategies/node_taint.go
```diff
@@ -0,0 +1,148 @@
...
+	result := []v1.Taint{}
+	for i := range taints {
+		if taints[i].Effect == v1.TaintEffectNoSchedule {
+			result = append(result, taints[i])
+		}
+	}
+	return result
+
```
damemi: extra line

pkg/descheduler/strategies/node_taint.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+//toleratesTaint returns true if a toleration tolerates a taint, or false otherwise
+func toleratesTaint(toleration *v1.Toleration, taint *v1.Taint) bool {
+
```
damemi: extra line

pkg/descheduler/strategies/node_taint.go
```diff
@@ -0,0 +1,148 @@
...
+// checkPodsSatisfyTolerations checks if the node's taints (NoSchedule) are still satisfied by pods' tolerations.
+func checkPodsSatisfyTolerations(pod *v1.Pod, node *v1.Node) bool {
+	tolerations := pod.Spec.Tolerations
+	taints := node.Spec.Taints
+	if len(taints) == 0 {
+		return true
+	}
+	noScheduleTaints := getNoScheduleTaints(taints)
+	allTolerated, _ := getMatchingTolerations(noScheduleTaints, tolerations)
```
damemi: is `getMatchingTolerations` used anywhere else to actually get the matching tolerations? Seems like it doesn't need to return a result and could just be renamed `allTaintsTolerated` and return a bool

then [line 83 to 88](https://github.com/kubernetes-sigs/descheduler/pull/175/files#diff-09a4f67796acb58a4e2a45466dcc11e7R83-R88) could be simplified to just `return allTaintsTolerated(...)`

pkg/descheduler/strategies/node_taint.go
```diff
@@ -0,0 +1,148 @@
...
+//toleratesTaint returns true if a toleration tolerates a taint, or false otherwise
+func toleratesTaint(toleration *v1.Toleration, taint *v1.Taint) bool {
+
+	if len(toleration.Effect) > 0 && toleration.Effect != taint.Effect {
+		return false
+	}
+
+	if len(toleration.Key) > 0 && toleration.Key != taint.Key {
+		return false
+	}
```
damemi: could these be simplified to a single compounded `or` check?

pkg/descheduler/strategies/node_taint.go
```diff
@@ -0,0 +1,148 @@
...
+		return false, []v1.Toleration{}
+	}
+	result := []v1.Toleration{}
+	for i := range taints {
+		tolerated := false
+		for j := range tolerations {
+			if toleratesTaint(&tolerations[j], &taints[i]) {
+				result = append(result, tolerations[j])
+				tolerated = true
+				break
```
damemi: If we're not using the result anywhere, we could remove the whole `result` logic and change this section to simply `if !toleratesTaint(..) { return false }`

damemi:
Hi @swatisehgal, sorry this took so long for a review. I found a couple nits (commented) but everything else looks solid. If you can update those this lgtm

damemi:
cc @ravisantoshgudimetla

pkg/descheduler/strategies/node_taint.go
```diff
@@ -0,0 +1,157 @@
...
+			}
+		}
+		if !tolerated {
+			return false
+		}
+	}
+	return true
+}
```
damemi: This doesn't look right. You're `break`ing then returning `true` as soon as you find one taint that can be tolerated, that doesn't test if *all* taints are tolerated. This section can be simplified:
```
for i := range taints {
  for j := range tolerations {
    if !toleratesTaint(&tolerations[j], &taints[i]) {
      return false
    }
  }
}
return true
```
No reason to check `if !tolerated` outside the loop then return `false`, because `if !tolerated` would evaluate to `false` in this case anyway, right? And `toleratesTaint` returns the value you're seeking, so you don't need to store that in a variable anyway.

pkg/descheduler/strategies/node_taint.go
```diff
@@ -0,0 +1,157 @@
...
+
+// checkPodsSatisfyTolerations checks if the node's taints (NoSchedule) are still satisfied by pods' tolerations.
+func checkPodsSatisfyTolerations(pod *v1.Pod, node *v1.Node) bool {
+	tolerations := pod.Spec.Tolerations
+	taints := node.Spec.Taints
+	if len(taints) == 0 {
+		return true
+	}
+	noScheduleTaints := getNoScheduleTaints(taints)
+	allTolerated := allTaintsTolerated(noScheduleTaints, tolerations)
```
damemi: Don't need to declare a variable here, you could just check `if !allTaintsTolerated(noScheduleTaints, tolerations) {...`

pkg/descheduler/strategies/node_taint.go
```diff
@@ -0,0 +1,157 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package strategies
+
+import (
+	"github.com/kubernetes-incubator/descheduler/cmd/descheduler/app/options"
+	"github.com/kubernetes-incubator/descheduler/pkg/api"
+	"github.com/kubernetes-incubator/descheduler/pkg/descheduler/evictions"
```
damemi: these should be `kubernetes-sigs` now

pkg/descheduler/strategies/node_taint.go
```diff
@@ -0,0 +1,158 @@
...
+import (
+	"sigs.k8s.io/descheduler/cmd/descheduler/app/options"
+	"sigs.k8s.io/descheduler/pkg/api"
+	"sigs.k8s.io/descheduler/pkg/descheduler/evictions"
+	podutil "sigs.k8s.io/descheduler/pkg/descheduler/pod"
+
+	"k8s.io/klog"
+
+	"k8s.io/api/core/v1"
+	clientset "k8s.io/client-go/kubernetes"
```
damemi: klog should be grouped with the other k8s deps, no line between them

pkg/descheduler/strategies/node_taint.go
damemi: Actually, I was completely mistaken and I apologize for that. You are correct in that we can't simple break on the first failure, because we need to check against all tolerations. This section lgtm

pkg/descheduler/strategies/node_taint.go
swatisehgal: oh, apologies I misunderstood your previous comment. Thanks for your suggestion and code snippet to rectify this. Will update now :)

swatisehgal:
@damemi Thanks for your review. I noticed that the travis-ci is failing. Is there anything I need to do to ensure all is okay from my end?

damemi:
@swatisehgal it's an odd failure, I think it may be a flake but I'm not sure how to trigger the to run another build. I think closing and reopening the PR might trigger another run
/close

damemi:
Looks like reopening got past that issue, must have been a network hiccup on travis. If it goes green now, @aveshagarwal could you approve?

swatisehgal:
Thanks @damemi  and @aveshagarwal !!

Issue #185, opened 2019-10-01
Descheduler evict annotation proposal


ksimon1:
We are looking for a solution that would let descheduler evict pods with a special annotation e.g. descheduler.alpha.kubernetes.io/evict. In our use case the pod represents a virtual machine (pod with local storage). The pod is managed by Kubevirt (https://kubevirt.io/, part of [CNCF](https://www.cncf.io/)). With this approach, the user can control which VM pods can be evicted and which not. We hope to combine this with kubernetes eviction webhooks into a VM load balancing solution.

We were checking the evict-local-storage-pods option, which would also be able to evict the Kubevirt VMs, but this option has a couple of drawbacks for us: It might not be enabled in cluster-wide descheduler, it would affect all local storage pods in the cluster and we want to allow each user to select which of his VM pods can be evicted and which can’t.

Implementation:
It may be enough to modify this line: https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/descheduler/pod/pods.go#L57
And add the check for pod annotation there.

/cc @MarSik, @fromanirh, @aveshagarwal, @k82cn, @ravisantoshgudimetla

ksimon1:
I am closing this proposal, because functionality is already merged

PR #186, opened 2019-10-03
Evict annotation

ksimon1:
This PR adds new functionality, which enables special pod annotation. 
User can control which pods can be evicted by setting annotation descheduler.alpha.kubernetes.io/evict. With this annotation user can override checks, which prevents descheduler to evict pods (user should know what he is doing).

This PR is POC for this proposal: https://github.com/kubernetes-sigs/descheduler/issues/185

/cc @MarSik, @fromanirh, @aveshagarwal, @k82cn, @ravisantoshgudimetla

### DIFF
--- a/README.md
+++ b/README.md
@@ -259,6 +259,9 @@ never evicted because these pods won't be recreated.
 * Pods associated with DaemonSets are never evicted.
 * Pods with local storage are never evicted.
 * Best efforts pods are evicted before Burstable and Guaranteed pods.
+* All types of pods with annotation descheduler.alpha.kubernetes.io/evict are evicted. This 
+annotation is used to override checks which prevent eviction and user can select which pod is evicted. 
+User should know how and if the pod will be recreated.
 
 ### Pod disruption Budget (PDB)
 Pods subject to Pod Disruption Budget (PDB) are not evicted if descheduling violates its pod
--- a/pkg/descheduler/pod/pods.go
+++ b/pkg/descheduler/pod/pods.go
@@ -17,13 +17,17 @@ limitations under the License.
 package pod
 
 import (
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	clientset "k8s.io/client-go/kubernetes"
 	api "k8s.io/kubernetes/pkg/apis/core"
 	"k8s.io/kubernetes/pkg/apis/core/v1/helper/qos"
 	"k8s.io/kubernetes/pkg/kubelet/types"
+)
+
+const (
+	evictPodAnnotationKey = "descheduler.alpha.kubernetes.io/evict"
 )
 
 // checkLatencySensitiveResourcesForAContainer checks if there are any latency sensitive resources like GPUs.
@@ -54,7 +58,7 @@ func IsLatencySensitivePod(pod *v1.Pod) bool {
 // IsEvictable checks if a pod is evictable or not.
 func IsEvictable(pod *v1.Pod, evictLocalStoragePods bool) bool {
 	ownerRefList := OwnerRef(pod)
-	if IsMirrorPod(pod) || (!evictLocalStoragePods && IsPodWithLocalStorage(pod)) || len(ownerRefList) == 0 || IsDaemonsetPod(ownerRefList) || IsCriticalPod(pod) {
+	if !HaveEvictAnnotation(pod) && (IsMirrorPod(pod) || (!evictLocalStoragePods && IsPodWithLocalStorage(pod)) || len(ownerRefList) == 0 || IsDaemonsetPod(ownerRefList) || IsCriticalPod(pod)) {
 		return false
 	}
 	return true
@@ -124,6 +128,12 @@ func IsDaemonsetPod(ownerRefList []metav1.OwnerReference) bool {
 // IsMirrorPod checks whether the pod is a mirror pod.
 func IsMirrorPod(pod *v1.Pod) bool {
 	_, found := pod.ObjectMeta.Annotations[types.ConfigMirrorAnnotationKey]
+	return found
+}
+
+// HaveEvictAnnotation checks if the pod have evict annotation
+func HaveEvictAnnotation(pod *v1.Pod) bool {
+	_, found := pod.ObjectMeta.Annotations[evictPodAnnotationKey]
 	return found
 }
 
--- a/pkg/descheduler/pod/pods_test.go
+++ b/pkg/descheduler/pod/pods_test.go
@@ -19,11 +19,160 @@ package pod
 import (
 	"testing"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"sigs.k8s.io/descheduler/test"
 )
 
+func TestIsEvictable(t *testing.T) {
+	n1 := test.BuildTestNode("node1", 1000, 2000, 13)
+	type testCase struct {
+		pod                   *v1.Pod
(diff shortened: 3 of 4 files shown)


k82cn:
how do we resolve the conflict?

ksimon1:
@k82cn which conflict you mean?

k82cn:
> which conflict you mean?

e.g. daemonset with this annotation. As you known, therer'e some cases that descehduler will not take eviction; we need to have a suggested behaviour when configuration conflict.

ksimon1:
I am aware of this use case. This functionality is for users who know what they are doing. It is not permanently enabled. It needs some effort to enable it. And if someone uses it and doesn't know what it is doing, then he should accept consequences. I will try to add more information to the documentation to explain how this functionality works and how to use it.

ravisantoshgudimetla:
The general direction looks good to me. This would bring us on parity with `drain` where we can forcefully let DS pods, pods with storage be deleted. Should we have an annotation or flag is what I am thinking. I personally feel that a `--force` flag would be better suited, instead of relying on every pod to have that annotation.

WDYT? @k82cn @aveshagarwal 

/cc @damemi

ksimon1:
I want to mention that we need something like pod annotation, or pod label, because we don't deploy descheduler as a part of Kubevirt. We are relying on descheduler which is already deployed, so we cannot control which functionality is allowed and which not.

MarSik:
@ravisantoshgudimetla The whole point of the proposed annotation is to let the pod owner say "Feel free to evict this pod if you think it is needed, my app knows how to handle that". We want to give the user (as opposed to the admin) a way to opt-in to descheduler initiated evictions.

Handling DaemonSet evictions on cluster level (admin decision) is an interesting idea, but it was not our goal when proposing this PR.

k82cn:
btw, do you consider `spec.Priority` for that?


> I will try to add more information to the documentation to explain how this functionality works and how to use it.

That's ok to me

> Should we have an annotation or flag is what I am thinking. I personally feel that a --force flag would be better suited, instead of relying on every pod to have that annotation.

Do you mean descheduler will evict pods by ignoring the criteria? If so, I think that'll be ping-pong: DS create pods, descheduler evict it ......

MarSik:
@k82cn spec.Priority itself does not help when the Pod is not evictable according to the general rules (local storage, no known replica controller...). We first need to convince descheduler it wants to touch the Pod. The rest of the logic (including priority) then stays the same.

k82cn:
> spec.Priority itself does not help when the Pod is not evictable according to the general rules (local storage, no known replica controller...).

I means you can define a lower priority for descheduler instead of introducing an annotation.

MarSik:
@k82cn How does that help with our use case? We want descheduler to evict Pods that are normally not evictable.

k82cn:
descheduler evict pods with priority below a configurable value forcely.

ksimon1:
@k82cn setting priority to pods affects scheduling of these pods. We don't want this behavior. We need mechanism, which doesn't require set special flag to deployment of descheduler (e.g. evict-local-storage-pods) and user is able to simply tell to descheduler, that pods with special annotation can be evicted.

MarSik:
@k82cn But we do not want to evict the pods immediately or as the first candidate. We simply want to tell descheduler that it should treat the marked Pods as any other evictable pods independently of what it thinks about them (critical, local storage, ..). We simply want to allow descheduler to ignore the non-evictability rules for marked pods.

Changing the priority would probably not work (the Pod has local storage and no known way of restarting so it will be ignored anyway) and it would negatively affect other parts of our integration with kubernetes (scheduling, resources..).

k82cn:
> setting priority to pods affects scheduling of these pods. 

That makes sense :)

ravisantoshgudimetla:
I am fine with the changes being made. The last question I've is using annotations vs labels? I thought as a community, we've decided to move away from annotations to well known labels?

@damemi and @ingvagabund - Can you review this PR?

damemi:
This PR lgtm if @ravisantoshgudimetla's question regarding labels/annotations can be answered. The syntax should be essentially the same for either so I'm also curious what the semantics are between choosing one or the other

MarSik:
@ravisantoshgudimetla @damemi We just followed the kubernetes docs wrt annotations: https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/

According to the page labels are used for selecting objects and we did not think that would be necessary as descheduler iterates over all the Pods anyway. Annotations are random data that are not used for querying.

This would be a trivial change either way. Labels are heavier though (k8s has some kind of index to make searching faster iirc) and I do not think we would get useful enough reward if we use them.

damemi:
@MarSik thanks for the explanation, I thought labels were more intended for selection. If we do decide that would be useful in the future it would be a simple change, but for now I think annotations accomplish what we want without the extra weight

/lgtm

ksimon1:
@aveshagarwal Everything is explained in this proposal(https://github.com/kubernetes-sigs/descheduler/issues/185) and in comments here

aveshagarwal:
I read this comment: https://github.com/kubernetes-sigs/descheduler/pull/186#issuecomment-539964165 and i think we are trying to evict "non-evictable pods". IMHO, It really seems contrary to what is the main purpose of not evicting those pods in the first place. 

From https://github.com/kubernetes-sigs/descheduler/issues/185, it seems that the main target is to  evict only specific pods with local storage. If that is case, why not just make this change specific to IsPodWithLocalStorage instead of generalizing it to all non-evictable pods?

MarSik:
@aveshagarwal 

> I read this comment: [#186 (comment)](https://github.com/kubernetes-sigs/descheduler/pull/186#issuecomment-539964165) and i think we are trying to evict "non-evictable pods". IMHO, It really seems contrary to what is the main purpose of not evicting those pods in the first place.

The reason the descheduler logic is not good enough is that it does not and cannot know about products built on top of kubernetes. We have custom controllers that make sure the Pod eviction is OK. So even though descheduler thinks the pod is not evictable, the user knows better and this annotation adds a way to express that.

aveshagarwal:
> @aveshagarwal
> 
> > I read this comment: [#186 (comment)](https://github.com/kubernetes-sigs/descheduler/pull/186#issuecomment-539964165) and i think we are trying to evict "non-evictable pods". IMHO, It really seems contrary to what is the main purpose of not evicting those pods in the first place.
> 
> The reason the descheduler logic is not good enough is that it does not and cannot know about products built on top of kubernetes. We have custom controllers that make sure the Pod eviction is OK. So even though descheduler thinks the pod is not evictable, the user knows better and this annotation adds a way to express that.

I agree with it, and its not new as its very common for lower layers to not know the intention of upper layers.  Also I am not against this PR. However there are other aspects (security and performance) we need to think about it, for example:

If we dont have a proper protection who can assign this new annotation, it could be misused. Imagine a multi tenant cluster where a user with lower priority pods assigns this annotation to its pods, that means those pods would not be evicted by descheduler. Also that means descheduler might evict higher priority pods before lower priority pods. Thats why I am suggesting to limit the use of this new annotation.

MarSik:
> user with lower priority pods assigns this annotation to its pods, that means those pods would not be evicted by descheduler

Quite the contrary. The annotation tells descheduler to consider evicting the Pod even though it would normally not do that. For example because of an active local storage volume.

> Also that means descheduler might evict higher priority pods before lower priority pods. Thats why I am suggesting to limit the use of this new annotation.

Why? The priority logic is not changed by this change.

aveshagarwal:
> > user with lower priority pods assigns this annotation to its pods, that means those pods would not be evicted by descheduler
> 
> Quite the contrary. The annotation tells descheduler to consider evicting the Pod even though it would normally not do that. For example because of an active local storage volume.
> 
> > Also that means descheduler might evict higher priority pods before lower priority pods. Thats why I am suggesting to limit the use of this new annotation.
> 
> Why? The priority logic is not changed by this change.

ok that sounds good and makes sense.

/approve

Issue #187, opened 2019-10-09
Update to latest k8s deps and switch to go mod


damemi:
The dependencies in the current latest version are out of date and should be updated to bring in new k8s functionality. With the update to 1.16 will bring the option of switching to go mod, which we should do to be in line with other kubernetes components.

This change should probably be a new version branch

- [x] bump to k8s 1.16 (https://github.com/kubernetes-sigs/descheduler/pull/188)
- [ ] switch to go.mod (https://github.com/kubernetes-sigs/descheduler/issues/207)

damemi:
closing for #207

PR #188, opened 2019-10-10
Bump to k8s 1.16

damemi:
Issue: https://github.com/kubernetes-sigs/descheduler/issues/187

Deps were last bumped (from what I can tell) sometime around k8s 1.9, so there are a couple small changes that need to be done, especially around our API registration.

After everything is settled and updated, I'll open another PR to follow up with a switch to go mod.


damemi:
(this is still in progress while I figure out the changes that need to be done for API registration, specifically:
* [componentconfig install](https://github.com/kubernetes-sigs/descheduler/blob/f1127541aa7702c67c7ea53cd35c920706486f94/pkg/apis/componentconfig/install/install.go)
* [api install](https://github.com/kubernetes-sigs/descheduler/blob/f1127541aa7702c67c7ea53cd35c920706486f94/pkg/apis/componentconfig/install/install.go)
* [descheduler scheme](https://github.com/kubernetes-sigs/descheduler/blob/f1127541aa/pkg/descheduler/scheme/scheme.go)

all of which are done with outdated patterns using the `k8s.io/apimachinery/pkg/apimachinery/announced` and `k8s.io/apimachinery/pkg/apimachinery/registered` packages which are now deprecated)

glide.yaml
```diff
@@ -1,25 +1,14 @@
 package: sigs.k8s.io/descheduler
 import:
 - package: k8s.io/client-go
-  version: 78700dec6369ba22221b72770783300f143df150
+  version: kubernetes-1.16.0
```
krmayankk: @Huang-Wei Does this version also include the even spread api or is that in some alpha branch ?

Huang-Wei: Yes.

damemi:
Squashed this into more reasonable commits, it was green so hopefully it still is and ready for review

aveshagarwal:
@damemi thanks for this PR and updating descheduler to 1.16. Its been pending for long long time so thanks a lot to you on working on this. It lgtm. Lets see if others have any feedback, and based on that we can merge this PR. Also once this PR is merged, i will cut a new descheduler release too.

ravisantoshgudimetla:
/lgtm 

again as I joined the org now.

damemi:
bump, any more review for this?

cc @krmayankk @ravisantoshgudimetla

Dockerfile
```diff
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-FROM golang:1.9.2
+FROM golang:1.12.0
```
seanmalloy: I believe this tag should be `1.12` or specify a newer minor version tag like `1.12.12`. Personally I think it would be best to use `1.12.12` here.

seanmalloy: Here is the tag info for 1.12.0 from DockerHub ... https://hub.docker.com/_/golang?tab=tags&page=1&name=1.12.0

Makefile
```diff
@@ -37,7 +37,7 @@ IMAGE_GCLOUD:=$(REGISTRY)/descheduler:$(VERSION)
 all: build
 
 build:
-	CGO_ENABLED=0 go build ${LDFLAGS} -o _output/bin/descheduler sigs.k8s.io/descheduler/cmd/descheduler
+	CGO_ENABLED=0 GODEBUG=tls13=1 go build ${LDFLAGS} -o _output/bin/descheduler sigs.k8s.io/descheduler/cmd/descheduler
```
krmayankk: Why is tls13 needed ?

Dockerfile
damemi: Sounds good to me, updated

Makefile
damemi: Sorry, I've had to add this on similar updates and thought it was needed here too, but it is not. Removed

damemi:
Feedback added, thanks. this will need another lgtm tag if that's all

Dockerfile
seanmalloy: LGTM

aveshagarwal:
@damemi  is it good to go?

damemi:
@aveshagarwal all good from me if you can approve it!

krmayankk:
Awesome super cool thanks for your work

Issue #145, opened 2019-04-17
Switch from glog to klog


ravisantoshgudimetla:
Upstream has moved from glog to klog. We need to follow them.

damemi:
I think this may be causing a panic related to the addition of the `component-base` dependency after the bump to 1.16:

```
$ ./_output/bin/descheduler --help
./_output/bin/descheduler flag redefined: log_dir
panic: ./_output/bin/descheduler flag redefined: log_dir

goroutine 1 [running]:
flag.(*FlagSet).Var(0xc0000c0060, 0x2244cc0, 0x2f19f90, 0x206f4d3, 0x7, 0x2099e27, 0x2f)
    /usr/local/go/src/flag/flag.go:850 +0x4af
flag.(*FlagSet).StringVar(...)
    /usr/local/go/src/flag/flag.go:753
sigs.k8s.io/descheduler/vendor/k8s.io/klog.InitFlags(0xc0000c0060)
    /Users/mdame/go/src/sigs.k8s.io/descheduler/vendor/k8s.io/klog/klog.go:428 +0x9d
sigs.k8s.io/descheduler/vendor/k8s.io/component-base/logs.init.0()
    /Users/mdame/go/src/sigs.k8s.io/descheduler/vendor/k8s.io/component-base/logs/logs.go:35 +0x2d
```
/remove-lifecycle rotten
/priority important-soon
/kind bug

PR #192, opened 2019-10-28
Switch from glog to klog

damemi:
Fixes https://github.com/kubernetes-sigs/descheduler/issues/145, including a new panic from the bump to 1.16


k82cn:
/approve

Thanks for the work :)

PR #201, opened 2019-11-28
Fix intentation on yaml

dileep-p:
Intentation is wrong on job yaml


dileep-p:
Done CLA

damemi:
/lgtm
thanks for catching this, removing those spaces looks right to me

Issue #195, opened 2019-11-11
Filter by label


killianmuldoon:
I'm looking to filter pods by label before applying a descheduling strategy to them. This allows us to slice our descheduling operations helping with testing and with only descheduling workloads that are actively identified as deschedulable based on developer/operator knowledge. This should be opt in - i.e. with no label selector in the policy descheduler will work as now.

I'm happy to write a PR for this - but I wanted to ask for advice on where this logic should be implemented. 

One option is in the ListEvictablePodsOnNode method, or one of its sub methods, - but it would need a change to the function signature which would break current uses of this package. 

Another option is to implement the label selector with a call in every strategy. There's duplication of work here that may be tough to maintain in future.

A third option is to wrap the Kubernetes Client in an LabelledClient Interface that embeds the kubernetes.Interface but also holds a labelSelector. This could then be called in the ListEvictablePodsOnNode, but this requires a type assertion in that method. 

I'd love to get some feedback on this - or any implementation ideas I might be missing.

killianmuldoon:
PR to resolve this at #202

seanmalloy:
The proposal has been reviewed by SIG scheduling and this feature can now be implemented. Descheduling by pod label will be implemented using a strategy parameter, not a CLI option. The `labelSelector` field in the YAML should be a [standard k8s labelSelector](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#labelselector-v1-meta).

Here is an example of what the YAML might look like:
```
---
apiVersion: "descheduler/v1alpha1"
kind: "DeschedulerPolicy"
strategies:
  "PodLifeTime":
     enabled: true
     params:
        maxPodLifeTimeSeconds: 86400
	labelSelector:
          matchLabels:
            component: redis
          matchExpressions:
            - {key: tier, operator: In, values: [cache]}
            - {key: environment, operator: NotIn, values: [dev]}

```

The proposal document can be found here: https://bit.ly/k8s-descheduler-labels-namespaces

@hanumanthan you had mentioned that you were interested in implementing this feature. Feel free to start working on it. Let us know if you have any questions. Let us know if you are able to work on this or not. Thanks!

hanumanthan:
@seanmalloy I couldnt find the standard implementation for labels parsing. Can you please provide any doc or code for parsing labels.
Found this https://godoc.org/k8s.io/apimachinery/pkg/labels#Selector but want to know the standard way to parse this from the input config

```
matchLabels:
[...]

Issue #163, opened 2019-06-11
POD Label Selector evaluation to be taken into account to decide POD eviction


VF-mbrauer:
It would be very helpful if we could decide upon a label which PODs are allowed to be evicted to an another node, rather than any POD on a node where the Threshold are met.
So this would give more flexibility. Imagine a Cluster consists of nonPROD and PROD PODs.
We could control that the nonPROD ones will be moved first or only.
Critical POD definition is not an option to mark PODS as important in order no to be evicted.
This will bring possibly other side effects with the normal scheduling of the nodes, etc.

aveshagarwal:
Regarding differentiation among pods which to evict, more than label selectors, i'd say that should be done based on their priority. In your scenario, nonPROD pods should be getting lower priority than PROD ones. And descheduler should evict lower priority pods first. We already have eviction based on priority here: https://github.com/kubernetes-incubator/descheduler/blob/master/pkg/descheduler/strategies/lownodeutilization.go#L198 .

So you could try assigning low priority to your nonProd pods than Prod pods, see if descheduler does what you are expecting.

Regarding eviction based on label selectors, (though one simple way could be that having a a script that chooses pods based on their label selectors and then deletes them), I'd like to understand, what are benefits or use cases you have in mind that can not be achieved via priorities, and we can discuss them here so that we understand adding eviction based on label selectors makes sense.

VF-mbrauer:
Thanks for your answer on this matter. 

Yes, we took also in consideration to use the `priority` option, to flag the PODs in order to evict according to this. But we think that could have a drawback (risk) also as setting this flag in the Deployments will also affect the normal `scheduler` and probably other sections, which are consuming this option/flag. Therefore in case of Node Shutdown, or any other activities this `prior-flag` will always play a role inside the whole cluster scheduling processes. 

An another use-case of having a Label-Selector could be, that we just mark the ones which we would like to evict as part of the `descheduler` process. And this independent of the Priority and any other relationship dependency. So it could be a special `test` or `staging` environment and there we just want to have some of them handled.
I think Priority is an important flag, but not for this use-case of re-shuffling PODs due to resource re-balancing. 

So do you see that as an option to integrate the Label-Selector as a further decision criteria?
Thanks a lot.

aveshagarwal:
> Thanks for your answer on this matter.
> 
[...]

PR #202, opened 2019-12-10
Add pod label selector to cli and strategies

killianmuldoon:
this PR was created to address #195 and #163 

It adds "pod-selector" - a Kubernetes Label Selector - as a command line argument and propagates it through the existing strategies. This allows slicing the cluster into pods the descheduler is interested in and those it isn't interested in e.g. "app=test". 

This PR mirrors the implementation of the node selector arg.

### DIFF
--- a/pkg/descheduler/strategies/lownodeutilization.go
+++ b/pkg/descheduler/strategies/lownodeutilization.go
@@ -19,8 +19,9 @@ package strategies
 import (
 	"sort"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/labels"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/klog"
 	helper "k8s.io/kubernetes/pkg/api/v1/resource"
@@ -60,7 +61,7 @@ func LowNodeUtilization(ds *options.DeschedulerServer, strategy api.DeschedulerS
 		return
 	}
 
-	npm := createNodePodsMap(ds.Client, nodes)
+	npm := createNodePodsMap(ds.Client, ds.PodSelector, nodes)
 	lowNodes, targetNodes := classifyNodes(npm, thresholds, targetThresholds, ds.EvictLocalStoragePods)
 
 	klog.V(1).Infof("Criteria for a node under utilization: CPU: %v, Mem: %v, Pods: %v",
@@ -311,10 +312,15 @@ func sortPodsBasedOnPriority(evictablePods []*v1.Pod) {
 }
 
 // createNodePodsMap returns nodepodsmap with evictable pods on node.
-func createNodePodsMap(client clientset.Interface, nodes []*v1.Node) NodePodsMap {
+func createNodePodsMap(client clientset.Interface, podSelector string, nodes []*v1.Node) NodePodsMap {
 	npm := NodePodsMap{}
+	selector, err := labels.Parse(podSelector)
+	if err != nil {
+		return npm
+	}
+
 	for _, node := range nodes {
-		pods, err := podutil.ListPodsOnANode(client, node)
+		pods, err := podutil.ListPodsOnANode(client, selector, node)
 		if err != nil {
 			klog.Warningf("node %s will not be processed, error in accessing its pods (%#v)", node.Name, err)
 		} else {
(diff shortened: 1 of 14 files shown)


killianmuldoon:
Signed CLA

pkg/descheduler/strategies/lownodeutilization.go
```diff
@@ -311,10 +312,15 @@ func sortPodsBasedOnPriority(evictablePods []*v1.Pod) {
...
+func createNodePodsMap(client clientset.Interface, podSelector string, nodes []*v1.Node) NodePodsMap {
 	npm := NodePodsMap{}
+	selector, err := labels.Parse(podSelector)
+	if err != nil {
+		return npm
+	}
+
 	for _, node := range nodes {
-		pods, err := podutil.ListPodsOnANode(client, node)
+		pods, err := podutil.ListPodsOnANode(client, selector, node)
```
aveshagarwal: pod selectors might not work right for lownodeutilization startgey this way as in this strategy, a node's load is computed by considering all pods. So computing loads on each node should be done by considering all pods. Perhaps during only eviction, pod selectors might be used.

killianmuldoon: Would it be better if I moved the implementation of this into podutil.isEvictable rather than podutil.ListPodsOnNode? I considered this, but it means a label check on each pod which slows down the check as pod numbers go up. Would it work to make lownodeutilization a special case? I guess it already works quite a bit differently to the other strategies.

aveshagarwal:
IMHO, I am not sure adding pod selectors is a good idea in general, or at least might need a bit more thinking than just introducing pod selectors like this. kube-scheduler makes its decisions by considering all pods in a cluster, and descheduler strategies must follow same. As I am not sure how evicting only selected pods can help with balancing a cluster which is the main goal of descheduler. In any case, even if we would like to have pod selectors, that must be done after careful review, abd by considering real use cases. 

So please feel free to write down concrete use cases and how that would help with balancing whole cluster when pod selectors are used.

swatisehgal:
@aveshagarwal Its understandable that the goals of descehduler is to balance the cluster but there might be scenarios/use case where descheduler is used for other purposes. E.g. the use case we are targeting involves decheduling a set of pods in case of an event. This would be useful in case of [Telemetry Aware Scheduling](https://github.com/intel/telemetry-aware-scheduling) where the TAS controller ensures that the telemetry policy corresponding to a pod is not violated, violation involves the controller to label pods which can be then descheduled by the descehuler. Other than that this could be very useful for slicing the cluster for testing purposes and also for scenarios where you would like to prevent workloads from being descheduled. But regardless of that, this pod label selector is provided as an optional argument meaning the current behavior of the descheduler is not altered. Also, if you have suggestions to improve the design/ implementation of this feature we would be more than happy to make changes.

killianmuldoon:
+1 to @swatisehgal here - Telemetry Aware Scheduling is the main use I was considering with this PR. I'd like to add more generally that slicing the cluster for descheduling in this way would move the cluster to a more balanced state while allowing a step based approach.

ravisantoshgudimetla:
I think the best way to do this would be to write a google doc and share it with sig-scheduling mailing list and ask for the feedback, if there are lots of people who have use-cases like this, we can consider this

WDYT @aveshagarwal @seanmalloy @damemi

seanmalloy:
@ravisantoshgudimetla I agree that submitting a proposal to the sig-scheduling mailing list would be a good next step. @killianmuldoon is this something is this something you would be able to do?

killianmuldoon:
@ravisantoshgudimetla @seanmalloy I'll get working on this and submit it soon. Thanks for the direction!

aviorma:
Any update on this ? i need this in order to run descheduler on namespace scope

seanmalloy:
> @ravisantoshgudimetla @seanmalloy I'll get working on this and submit it soon. Thanks for the direction!

@killianmuldoon are you still planning on creating a design proposal and sharing it with sig-scheduling? Is there anything we can do to help?


CC: @damemi @ravisantoshgudimetla @aveshagarwal

killianmuldoon:
@seanmalloy  Sorry - this fell by the wayside a little bit! I'll put this on the agenda for an upcoming sig-scheduling meeting. Thanks for the reminder!

seanmalloy:
@killianmuldoon I created a Google Doc proposal and shared it with the sig-scheduling mailing list. I've also added this to the sig-scheduling meeting agenda for July 2nd.

seanmalloy:
@swatisehgal and @killianmuldoon are there specific descheduler strategies you would like to use with this feature?

Any additional details you can provide about real world use cases would be useful as I build out the proposal document for review by SIG scheduling. Thanks!

killianmuldoon:
@seanmalloy Thanks for following up on this - it completely fell off my plate. I updated the doc with some details of our use case, and I'll join the meeting later on in case there's any questions I can help answer.

seanmalloy:
/close

@killianmuldoon after review by SIG scheduling we would like to implement this as a descheduler strategy parameters instead of a CLI option. So I'm going to close this PR. Additional details can be found in this comment: https://github.com/kubernetes-sigs/descheduler/issues/195#issuecomment-656995247.

PR #213, opened 2020-01-17
Update examples to support v1.16.0+

mdavidsen:
Support for using the `scheduler.alpha.kubernetes.io/critical-pod` annotation was
deprectated in 1.13 and finally removed in 1.16

Also alignes the format of the yaml-blob for descheduler-job.yaml in readme.md to the one in examples/descheduler-job.yaml 

I'm not sure if there was made consiuous decision made around the `image:` property here; is there a reason they differed? I kept that difference for the time being atleast.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,19 +133,18 @@ spec:
   template:
     metadata:
       name: descheduler-pod
-      annotations:
-        scheduler.alpha.kubernetes.io/critical-pod: ""
     spec:
-        containers:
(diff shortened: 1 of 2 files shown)


seanmalloy:
@aveshagarwal  I was able to successfully apply the updated example from the README.md on a k8s 1.16 cluster. Please review when you have some time. Thanks!

aveshagarwal:
/approve

thanks @mdavidsen for the PR.

PR #217, opened 2020-01-25
Automate Container Image Builds Using GCB

seanmalloy:
Adds a GCB(Google Cloud Build) configuraiton file that automates the
building and pushing of descheduler container images to the official
Google Container Registry for k8s.

Reference documentation:
https://github.com/kubernetes/test-infra/blob/master/config/jobs/image-pushing/README.md

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -1,0 +1,24 @@
+# See https://cloud.google.com/cloud-build/docs/build-config
+
+# this must be specified in seconds. If omitted, defaults to 600s (10 mins)
+timeout: 1200s
+# this prevents errors if you don't use both _GIT_TAG and _PULL_BASE_REF,
+# or any new substitutions added in the future.
+options:
+  substitution_option: ALLOW_LOOSE
+steps:
+  - name: 'gcr.io/k8s-testimages/gcb-docker-gcloud:v20190906-745fed4'
+    entrypoint: make
+    env:
+    - DOCKER_CLI_EXPERIMENTAL=enabled
+    - VERSION=$_GIT_TAG
+    - BASE_REF=$_PULL_BASE_REF
+    args:
+    - push
+substitutions:
+  # _GIT_TAG will be filled with a git-based tag for the image, of the form vYYYYMMDD-hash, and
+  # can be used as a substitution
+  _GIT_TAG: '12345'
+  # _PULL_BASE_REF will contain the ref that was pushed to to trigger this build -
+  # a branch like 'master' or 'release-0.2', or a tag like 'v0.2'.
+  _PULL_BASE_REF: 'master'
--- a/docs/releasing.md
+++ b/docs/releasing.md
@@ -1,0 +1,21 @@
+# Release process
+
+## Semi-automatic
+
+1. Make sure your repo is clean by git's standards
+2. Tag the repository and push the tag `VERSION=v0.10.0 git tag -m $VERSION $VERSION; git push origin $VERSION`
+3. Publish a draft release using the tag you just created
+4. Perform the [image promotion process](https://github.com/kubernetes/k8s.io/tree/master/k8s.gcr.io#image-promoter)
+5. Publish release
+6. Email `kubernetes-sig-scheduling@googlegroups.com` to announce the release
+
+## Manual
+
+1. Make sure your repo is clean by git's standards
+2. Tag the repository and push the tag `VERSION=v0.10.0 git tag -m $VERSION $VERSION; git push origin $VERSION`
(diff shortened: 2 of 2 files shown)


seanmalloy:
This has **NOT** been tested, so please don't merge it yet.  I think right now it will only build the container image, but won't push it. Some tweaks might need to be made to the `Makefile`, but not 100% sure yet.

seanmalloy:
/hold cancel

I made one small change to the GCB config to also push the container image to the staging registry. I believe it will work correctly now. However it is not clear to be how to test the GCB job prior to merging this to the master branch.

@damemi @aveshagarwal @ravisantoshgudimetla this is ready for review now.

cloudbuild.yaml
```diff
@@ -0,0 +1,24 @@
...
+    env:
+    - DOCKER_CLI_EXPERIMENTAL=enabled
+    - VERSION=$_GIT_TAG
+    - BASE_REF=$_PULL_BASE_REF
+    args:
+    - push
+substitutions:
+  # _GIT_TAG will be filled with a git-based tag for the image, of the form vYYYYMMDD-hash, and
+  # can be used as a substitution
+  _GIT_TAG: '12345'
```
ravisantoshgudimetla: What does this mean? Will this be substituted by the git-tag `v20190906-745fed4` ?

seanmalloy: Short answer ...

Yeah it looks like it will be vYYYYMMDD-GIT_TAG. For example v20200131-v0.10.0. I think this is not what we want. When a descheduler version is released the tag should be something like v0.10.0. I'll have to figure out how to change this.

Long answer ...

It looks like this a [GCB user defined substitution](https://cloud.google.com/cloud-build/docs/configuring-builds/substitute-variable-values#using_user-defined_substitutions). The container image listed on line 10 must be going something special with the variable.

I believe I tracked it down to this directory https://github.com/kubernetes/test-infra/tree/master/images/builder. More specifically here ... https://github.com/kubernetes/test-infra/blob/09a81c4adf5b662b2643dccb9d285ad3102af4af/images/builder/main.go#L48-L56.

It seems like most other k8s projects using image build automation are using _GIT_TAG. See [here](https://github.com/kubernetes/kops/blob/master/ci/postsubmit/push-to-staging/cloudbuild.yaml#L10), [here](https://github.com/kubernetes-sigs/external-dns/blob/master/cloudbuild.yaml#L10) and [here](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/cloudbuild.yaml#L10).

seanmalloy: One more thought ...

We could merge this PR as is and sort out git/docker tag configuration is a follow up PR. This would at least enable the automated builds for the master branch and let us confirm that the build automation is working.

Also be aware that the container build automation only pushes container images the GCR staging registry. A person will still need to submit a pull request for each container image that needs to be promoted to the production GCR registry.

seanmalloy:
/hold

I need to add some basic docs explaining how this works for both the automated and manual process.

cloudbuild.yaml
Katharine: Another thought (which is where the design comes from): it doesn't really matter what you tag things in the staging registry - when the image promoter promotes them, they can be given a different tag to what they had in staging.

seanmalloy: @Katharine thanks for adding your comment. This makes sense to me now. We don't need to worry about the tags in the staging registry.

@ravisantoshgudimetla let me know if you have any additional questions on this topic.

seanmalloy:
/hold cancel

I added commit ac089fe5ce4154eb0c6c0d95aaf610d38d0e6a33 to this pull request that documents the new release process. This PR is now ready for review. @ravisantoshgudimetla @aveshagarwal and @damemi please review this commit to make sure the release documentation is clear. Thanks!

cloudbuild.yaml
damemi: So, with this as it is we'll get staging repo tags in the format `v20190906-v0.9.0-76-g30d05382b` (based on my own run of the function you linked to [here](https://github.com/kubernetes/test-infra/blob/09a81c4adf5b662b2643dccb9d285ad3102af4af/images/builder/main.go#L48-L56)).

That seems good enough for me in the staging repo, it's descriptive of the version. And we choose our published tag when promoted, so it will be cleaner (like `v0.9.0`). Sgtm

damemi:
@seanmalloy thanks for this work, this all looks good to me. The instructions you linked are pretty clear and are definitely helpful to add.
/lgtm

ravisantoshgudimetla:
@seanmalloy - Thanks for working on this.

/lgtm

Issue #220, opened 2020-01-29
--descheduling-interval flag not functional


damemi:
The [`--descheduling-interval` flag](https://github.com/kubernetes-sigs/descheduler/blob/dd54f1a6560f2945502af3b5347beaeceb1449a2/cmd/descheduler/app/options/options.go#L51) doesn't seem to be wired up to anything.

As a user, I would expect this flag to set the descheduler to run continuously, with the interval between runs determined by the flag's value.

Doing so would allow the descheduler to be run as a standalone pod instead of a CronJob, which are not currently GA. Is there any previous discussion on this flag and its implementation?

aveshagarwal:
cronjob are widely used so not sure it helps much.

aveshagarwal:
and thats the reason to not implement it in the first place though initially we had thought about it.

PR #221, opened 2020-01-29
Wire --descheduling-interval to run descheduler on a loop

damemi:
Fixes https://github.com/kubernetes-sigs/descheduler/issues/220

This wires the `--descheduling-interval` flag to actually run the descheduler on a loop in the interval specified. If the value is unset (which defaults to `0`) the descheduler only runs once, maintaining consistency with the current expectation. 

One benefit to this is that it removes the need to use a CronJob for a continuously running controller.

### DIFF
--- a/cmd/descheduler/app/options/options.go
+++ b/cmd/descheduler/app/options/options.go
@@ -48,7 +48,7 @@ func NewDeschedulerServer() *DeschedulerServer {
 
 // AddFlags adds flags for a specific SchedulerServer to the specified FlagSet
 func (rs *DeschedulerServer) AddFlags(fs *pflag.FlagSet) {
-	fs.DurationVar(&rs.DeschedulingInterval, "descheduling-interval", rs.DeschedulingInterval, "time interval between two consecutive descheduler executions")
+	fs.DurationVar(&rs.DeschedulingInterval, "descheduling-interval", rs.DeschedulingInterval, "Time interval between two consecutive descheduler executions. Setting this value instructs the descheduler to run in a continuous loop at the interval specified.")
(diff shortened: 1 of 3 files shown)


aveshagarwal:
> Fixes #220
> 
> This wires the `--descheduling-interval` flag to actually run the descheduler on a loop in the interval specified. If the value is unset (which defaults to `0`) the descheduler only runs once, maintaining consistency with the current expectation.
> 
> One benefit to this is that it removes the need to use a CronJob for a continuously running controller.

CronJob are widely used so not sure it helps much. So have an option where if someone uses descheduler as cronjob, they can disable it.

damemi:
@aveshagarwal the way I've written this, the default value for `descheduling-interval` (0s) still runs the descheduler only once, then exits. This makes the new control-loop behavior opt-in, and shouldn't see any change for current users unless. However we could make this clearer or more difficult to "accidentally" switch on if you'd like.

I think the ability to execute multiple descheduler runs in the same pod does have some other advantages too, such as easier debugging and the ability to filter through logs of multiple runs, along with simplifying deployment.

ravisantoshgudimetla:
@damemi - I am fine with the changes, thanks for working on this. @aveshagarwal is right that we thought about it but didn't implement since no one was asking for it.

Please update the docs string at https://github.com/kubernetes-sigs/descheduler/blob/dd54f1a6560f2945502af3b5347beaeceb1449a2/cmd/descheduler/app/options/options.go#L51, to state our intent clearly.

damemi:
@ravisantoshgudimetla thanks, I updated the doc string and also added an E2E which confirms that by default, the descheduler should still only run once and exit (this took some minor refactoring of our main run function)

PR #226, opened 2020-02-04
Fix typo in function name

ankon:
There is also an inconsistency in the test utilities where it is capitalized as "Un_S_chedulable", I left that alone.


aveshagarwal:
@ankon thanks for this PR.

/lgtm
/approve

PR #240, opened 2020-02-13
Turn StrategyParameters.NodeResourceUtilizationThresholds field into a pointer

ingvagabund:
The field is intended to be omitempty when not set. Without a pointer the strategy
serialized into json string looks like:

```json
strategies:
  LowNodeUtilization:
    enabled: true
    params:
        numberOfNodes: 1
        targetThresholds:
          cpu: 50
          memory: 50
          pods: 20
        thresholds:
          cpu: 50
          memory: 50
          pods: 20
  RemoveDuplicates:
    enabled: true
    params:
      nodeResourceUtilizationThresholds: {}
    enabled: true
    params:
      nodeResourceUtilizationThresholds: {}
    enabled: true
    params:
      nodeAffinityType:
      nodeResourceUtilizationThresholds: {}
    enabled: true
    params:
      nodeResourceUtilizationThresholds: {}
```

It's preferred to have the following json string instead:
```
strategies:
  LowNodeUtilization:
    enabled: true
    params:
        numberOfNodes: 1
        targetThresholds:
          cpu: 50
          memory: 50
          pods: 20
        thresholds:
          cpu: 50
          memory: 50
          pods: 20
  RemoveDuplicates:
    enabled: true
    enabled: true
    enabled: true
    params:
      nodeAffinityType:
    enabled: true
```

This effectively changes the exported data type so any user importing `StrategyParameters` data type and setting/reading `NodeResourceUtilizationThresholds` field will need to adjust affected code.


go.mod
```diff
@@ -5,6 +5,7 @@ go 1.13
 require (
 	github.com/spf13/cobra v0.0.5
 	github.com/spf13/pflag v1.0.5
+	gopkg.in/yaml.v2 v2.2.8 // indirect
```
damemi: any idea where this came from? says it's indirect but just curious which of these changes might have brought it in

seanmalloy:
@damemi it is too late for v0.10.0. The tag has already been created https://github.com/kubernetes-sigs/descheduler/releases/tag/v0.10.0.

go.mod
seanmalloy: I do not know the answer to this question. I did compile this code using Go 1.13.6 and can confirm that the `go.mod` and `go.sum` files are clean. I also verified `go mod tidy` and it is good.

seanmalloy:
@aveshagarwal and @ravisantoshgudimetla please take a look at this when you have some time.

damemi:
To resurrect this, do you think we should do the same with the `PodsHavingTooManyRestarts` strategy params? I think we'll hit the same problem there too now

ingvagabund:
> To resurrect this, do you think we should do the same with the PodsHavingTooManyRestarts strategy params? I think we'll hit the same problem there too now

Yeah, given the Strategy data types is shared among all strategies, more pointers, the better!!!

Issue #245, opened 2020-02-28
RemovePodsViolatingNodeTaints policy not working with --descheduling-interval  option 


nshekhar221:
While trying to run Descheduler with `RemovePodsViolatingNodeTaints` policy and `--descheduling-interval` option set to 5m , we are observing that descheduler is caching the nodes status/taints etc. at its first run and that cache did not get updated in the subsequent runs.

Due to this, any changes made to nodes taints(after the descheduler's first run) did not get picked up by Descheduler and hence pods are not getting Evicted from that node.

nshekhar221:
Some logs for reference -

> I0228 06:10:30.925414       1 reflector.go:432] pkg/mod/k8s.io/client-go@v0.17.0/tools/cache/reflector.go:108: Watch close - *v1.Node total 51 items received
I0228 06:10:39.869716       1 reflector.go:278] pkg/mod/k8s.io/client-go@v0.17.0/tools/cache/reflector.go:108: forcing resync
I0228 06:10:53.968840       1 node_taint.go:48] Processing node: "vmss-agent-worker-nshekhartest-tjmrw00000g"
I0228 06:10:53.986100       1 node_taint.go:48] Processing node: "vmss-agent-worker-nshekhartest-tjmrw00000i"
I0228 06:10:54.070078       1 node_taint.go:48] Processing node: "vmss-agent-worker-nshekhartest-tjmrw00000e"
I0228 06:10:54.081994       1 node_taint.go:48] Processing node: "vmss-agent-worker-nshekhartest-tjmrw00000h"
I0228 06:10:54.095838       1 node_taint.go:48] Processing node: "vmss-agent-worker-nshekhartest-tjmrw00000f"
I0228 06:10:54.166821       1 node_taint.go:48] Processing node: "vmss-agent-worker-nshekhartest-tjmrw00000j"
I0228 06:11:54.185717       1 node_taint.go:48] Processing node: "vmss-agent-worker-nshekhartest-tjmrw00000g"
I0228 06:11:54.204834       1 node_taint.go:48] Processing node: "vmss-agent-worker-nshekhartest-tjmrw00000i"
I0228 06:11:54.222188       1 node_taint.go:48] Processing node: "vmss-agent-worker-nshekhartest-tjmrw00000e"
I0228 06:11:54.266629       1 node_taint.go:48] Processing node: "vmss-agent-worker-nshekhartest-tjmrw00000h"
[...]

PR #249, opened 2020-03-03
List nodes through informer in every iteration

ingvagabund:
Also, refactor the code a bit so it can be tested without checking for eviction support.

Fixes: https://github.com/kubernetes-sigs/descheduler/issues/245

### DIFF
--- a/pkg/descheduler/descheduler.go
+++ b/pkg/descheduler/descheduler.go
@@ -22,6 +22,7 @@ import (
 	"k8s.io/klog"
 
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/informers"
 	"sigs.k8s.io/descheduler/cmd/descheduler/app/options"
 	"sigs.k8s.io/descheduler/pkg/api"
 	"sigs.k8s.io/descheduler/pkg/descheduler/client"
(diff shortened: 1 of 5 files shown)


pkg/descheduler/descheduler.go
```diff
@@ -56,18 +57,29 @@ func RunDeschedulerStrategies(rs *options.DeschedulerServer, deschedulerPolicy *
...
-	nodePodCount := utils.InitializeNodePodCount(nodes)
 	wait.Until(func() {
+		nodes, err := nodeutil.ReadyNodes(rs.Client, nodeInformer, rs.NodeSelector, stopChannel)
+		if err != nil {
+			klog.V(1).Infof("Unable to get ready nodes: %v", err)
+			close(stopChannel)
+			return
+		}
+
+		if len(nodes) <= 1 {
```
dharmab: Is number this a good candidate for a configurable option in a future PR? e.g. a user might want to set this to match the number of AZs in their cluster.

ingvagabund: Two nodes is a minimal condition for running the descheduler. In case you need to also verify each AZ has at list one node available, what are the chances you will not have it before you run the descheduler. Also, it's maybe more convenient for individual strategies. No strategy should evict a pod if there's no other node where the new pod can lend. Including constraining strategies to be zone aware.

ingvagabund:
@damemi  @aveshagarwal PTAL

aveshagarwal:
/lgtm
/approve

thanks @ingvagabund

Issue #62, opened 2017-12-06
deschedule pods that fail to start or restart too often


kabakaev:
It is not uncommon that pods get scheduled on nodes that are not able to start it.
For example, a node may have network issues and unable to mount a networked persistent volume, or cannot pull a docker image, or has some docker configuration issue which is seen only on container startup.

Another common issue is when a container gets restarted by liveliness check because of some local node issue (e.g. wrong routing table, slow storage, network latency or packet-drop). In that case, a pod is unhealthy most of the time and hangs in a restart state forever without a chance of being migrated to another node.

As of now, there is no possibility to re-schedule pods with faulty containers. It may be helpful to introduce two new Strategies:
* container-restart-rate: re-schedule a pod if it is unhealthy since `$notReadyPeriod` seconds and one of its containers was restarted `$maxRestartCount` times. 
* pod-startup-failure: a pod was scheduled on a node, but was unable to start all of its containers since `$maxStartupTime` seconds.

The similar issue is filed against kubernetes: https://github.com/kubernetes/kubernetes/issues/13385

ravisantoshgudimetla:
Seems like a reasonable ask. @kabakaev I am planning to defer this to 0.6 release or later. Hope you are ok with that.

bgrant0607:
Ref also https://github.com/kubernetes/kubernetes/issues/14796

mbdas:
A whole bunch of issues were referred to this and then this gets auto closed. Should the users just write a controller and delete pod after too many restarts etc

ravisantoshgudimetla:
https://github.com/kubernetes-sigs/descheduler/pull/89 tried addressing this. Let's make sure that we're getting this in.

riking:
Please remember to implement safeguards against a "pod of death" where attempting to start it crashes kubelet. From what I understand, this would look identical to a pod that simply can't be started on that node.

Wrt restart too often (CrashLoopBackoff), the value of this is questionable because the majority of cases where that happens is an application bug.

mbdas:
True it is app error most times but there are valid scenarios where restarting a sidecar (assuming that crashed) may not fix the pod as a whole and may need a fresh launch and in real life often host misconfigurations/bad upgrade may cause the pod failures and good to move away from those bad hosts in the fleet.

pohly:
[...]

PR #254, opened 2020-03-30
Add RemoveTooManyRestarts policy

damemi:
This rebases the changes from https://github.com/kubernetes-sigs/descheduler/pull/89 to the current master branch

See issue https://github.com/kubernetes-sigs/descheduler/issues/62


damemi:
/cc @ingvagabund 
ptal

pkg/descheduler/strategies/toomanyrestarts.go
```diff
@@ -0,0 +1,99 @@
...
+func RemovePodsHavingTooManyRestarts(ds *options.DeschedulerServer, strategy api.DeschedulerStrategy, policyGroupVersion string, nodes []*v1.Node, nodePodCount utils.NodePodEvictedCount) {
+	evictionCount := removePodsHavingTooManyRestarts(ds, strategy, policyGroupVersion, nodes, nodePodCount, ds.EvictLocalStoragePods)
+	glog.V(1).Infof("RemovePodsHavingTooManyRestarts evicted %v pods", evictionCount)
+}
+
+func removePodsHavingTooManyRestarts(ds *options.DeschedulerServer, strategy api.DeschedulerStrategy, policyGroupVersion string, nodes []*v1.Node, nodePodCount utils.NodePodEvictedCount, evictLocalStoragePods bool) int {
+	podsEvicted := 0
+
+	if !strategy.Enabled || strategy.Params.PodsHavingTooManyRestarts.PodeRestartThresholds < 1 {
+		glog.Info("RemovePodsHavingTooManyRestarts policy is disabled.")
```
seanmalloy: I believe all references to `glog` should be converted to `klog`.

README.md
```diff
@@ -170,6 +170,22 @@ strategies:
...
+
+```
+apiVersion: "descheduler/v1alpha1"
+kind: "DeschedulerPolicy"
+strategies:
+  "RemovePodsHavingTooManyRestarts":
+     enabled: true
+     params:
+       podsHavingTooManyRestarts:
+         podeRestartThresholds: 100
```
seanmalloy: `podeRestartThresholds` looks like there is an extra `e` after `pod`.

examples/policy.yaml
```diff
@@ -17,3 +17,9 @@ strategies:
            "cpu" : 50
            "memory": 50
            "pods": 50
+  "RemovePodsHavingTooManyRestarts":
+     enabled: true
+     params:
+       podsHavingTooManyRestarts:
+         podeRestartThresholds: 100
```
seanmalloy: Same here, extra `e` after `pod`.

pkg/api/types.go
```diff
@@ -58,3 +59,8 @@ type NodeResourceUtilizationThresholds struct {
 	TargetThresholds ResourceThresholds
 	NumberOfNodes    int
 }
+
+type PodsHavingTooManyRestarts struct {
+	PodeRestartThresholds   int32
```
seanmalloy: Same here, extra `e` after `pod`.

pkg/api/v1alpha1/types.go
```diff
@@ -58,3 +59,8 @@ type NodeResourceUtilizationThresholds struct {
 	TargetThresholds ResourceThresholds `json:"targetThresholds,omitempty"`
 	NumberOfNodes    int                `json:"numberOfNodes,omitempty"`
 }
+
+type PodsHavingTooManyRestarts struct {
+	PodeRestartThresholds   int32 `json:"podeRestartThresholds,omitempty"`
```
seanmalloy: Same here, extra `e` after `pod`.

pkg/descheduler/strategies/toomanyrestarts.go
```diff
@@ -0,0 +1,99 @@
...
+		return podsEvicted
+	}
+
+	for _, node := range nodes {
+		glog.V(1).Infof("RemovePodsHavingTooManyRestarts Processing node: %#v", node.Name)
```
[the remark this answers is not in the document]
ingvagabund: Also node.Name is a string, so s/%#v/%v or %q

pkg/descheduler/strategies/toomanyrestarts.go
```diff
@@ -0,0 +1,99 @@
...
+		}
+		podsEvicted += nodePodCount[node]
+	}
+	return podsEvicted
+}
+
+// calcContainerRestarts get container restarts and init container restarts.
+func calcContainerRestarts(pod *v1.Pod) (int32, int32) {
+	var (
+		restarts     int32 = 0
```
ingvagabund: No need for `0` as it's the default value:
```
var restarts, initRestarts int32
```

pkg/descheduler/strategies/toomanyrestarts_test.go
```diff
@@ -0,0 +1,207 @@
...
+			maxPodsToEvict:          0,
+		},
+		{
+			description:             "One pod have total restarts bigger than threshold, 6 pod evictions",
+			strategy:                createStrategy(true, true, 1),
+			expectedEvictedPodCount: 6,
+			maxPodsToEvict:          0,
+		},
+		{
+			description:             "Nine pods have total restarts equals threshold(includingInitContainers=true), 5 pods evictions",
```
ingvagabund: The strategy has the following condition for deciding if the threshold is met:
```
if strategy.Params.PodsHavingTooManyRestarts.IncludingInitContainers {
	if restarts+initRestarts <= strategy.Params.PodsHavingTooManyRestarts.PodeRestartThresholds {
		continue
	}
} else if restarts <= strategy.Params.PodsHavingTooManyRestarts.PodeRestartThresholds {
	continue
}
```
Which means a pod is evicted only when the number of restarts is bigger than the threshold. Is this the intention? I would assume a pod to be evicted as soon as the threshold is met.

pkg/descheduler/strategies/toomanyrestarts.go
```diff
@@ -0,0 +1,99 @@
...
+					continue
+				}
+			} else if restarts <= strategy.Params.PodsHavingTooManyRestarts.PodeRestartThresholds {
+				continue
+			}
+
+			glog.V(1).Infof("RemovePodsHavingTooManyRestarts will evicted pod: %#v, container restarts: %d, initContainer restarts: %d", pod.Name, restarts, initRestarts)
+			success, err := evictions.EvictPod(ds.Client, pod, policyGroupVersion, ds.DryRun)
+			if !success {
+				glog.Infof("RemovePodsHavingTooManyRestarts Error when evicting pod: %#v (%#v)", pod.Name, err)
```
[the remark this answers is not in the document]
damemi: I don't disagree, but all of our other policies format the errors with `(%#v)` so for consistency in error reporting we should probably keep it. (Personally I prefer `%+v` for errors)

pkg/descheduler/strategies/toomanyrestarts.go
```diff
@@ -0,0 +1,99 @@
...
+			} else {
+				nodePodCount[node]++
+				glog.V(1).Infof("RemovePodsHavingTooManyRestarts Evicted pod: %#v (%#v)", pod.Name, err)
```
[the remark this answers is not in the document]
damemi: same as https://github.com/kubernetes-sigs/descheduler/pull/254#discussion_r401040516

pkg/descheduler/strategies/toomanyrestarts.go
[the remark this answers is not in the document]
damemi: Is %s fine?

This is also something that's copied in all the other strategies (using %#v for nodename here)

damemi:
@seanmalloy @ingvagabund thanks, I've addressed your feedback

README.md
```diff
@@ -170,6 +170,22 @@ strategies:
...
+
+```
+apiVersion: "descheduler/v1alpha1"
+kind: "DeschedulerPolicy"
+strategies:
+  "RemovePodsHavingTooManyRestarts":
+     enabled: true
+     params:
+       podsHavingTooManyRestarts:
+         podRestartThresholds: 100
```
seanmalloy, line 185: In my opinion it makes more sense for this to be singular instead of plural. Can you change this to `podRestartThreshold`?

pkg/descheduler/strategies/toomanyrestarts.go
```diff
@@ -0,0 +1,92 @@
...
+	evictionCount := removePodsHavingTooManyRestarts(ds, strategy, policyGroupVersion, nodes, nodePodCount, ds.EvictLocalStoragePods)
+	klog.V(1).Infof("evicted %v pods", evictionCount)
+}
+
+func removePodsHavingTooManyRestarts(ds *options.DeschedulerServer, strategy api.DeschedulerStrategy, policyGroupVersion string, nodes []*v1.Node, nodePodCount utils.NodePodEvictedCount, evictLocalStoragePods bool) int {
+	podsEvicted := 0
+	for _, node := range nodes {
+		klog.V(1).Infof("Processing node: %s", node.Name)
+		pods, err := podutil.ListEvictablePodsOnNode(ds.Client, node, evictLocalStoragePods)
+
```
seanmalloy, line 46: nit: remove this blank line

pkg/descheduler/strategies/toomanyrestarts.go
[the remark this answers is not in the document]
ingvagabund: Good point. No more against keeping %#v. It's better to change all the occurrences at once.

pkg/descheduler/strategies/toomanyrestarts.go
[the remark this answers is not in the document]
ingvagabund: No more against keeping %#v. It's better to change all the occurrences at once.

README.md
ingvagabund: +1 for `podRestartThreshold`

damemi:
@seanmalloy thanks, updated. I'll squash down to 1 commit later today if there's no more feedback

damemi:
squashed to 1 commit

aveshagarwal:
just curious why it includes so many vendor changes. Did it update some existing dependencies. 
Also please split the commit into 2 commit, one with main changes and other related to vendor changes as it helps with the review.

damemi:
@aveshagarwal I think the vendor changes come from our generator scripts, see:
```
$ make gen
./hack/update-generated-deep-copies.sh
./hack/update-generated-defaulters.sh
#undo go mod changes caused by above.
go mod tidy

$ git status
On branch toomanyrestarts
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   go.mod
	modified:   go.sum

no changes added to commit (use "git add" and/or "git commit -a")

$ git diff
diff --git a/go.mod b/go.mod
index 2dba8b5ba..c93820bd8 100644
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module sigs.k8s.io/descheduler
 go 1.13
 
 require (
+       github.com/gogo/protobuf v1.3.1 // indirect
        github.com/spf13/cobra v0.0.5
        github.com/spf13/pflag v1.0.5
        k8s.io/api v0.17.0
@@ -11,4 +12,6 @@ require (
        k8s.io/client-go v0.17.0
        k8s.io/component-base v0.17.0
        k8s.io/klog v1.0.0
+       k8s.io/kube-openapi v0.0.0-20200121204235-bf4fb3bd569c // indirect
+       sigs.k8s.io/yaml v1.2.0 // indirect
 )
```

Specifically, the `go mod tidy` at the end of our generator. I committed these changes with `go mod vendor` to make sure our `vendor/` stays up to date with local changes

seanmalloy:
@aveshagarwal and @damemi running `make gen` on the master branch without doing any other code changes is modifying `go.mod` and `go.sum`. It looks like the `go build` commands in the code generation hack scripts([here](https://github.com/kubernetes-sigs/descheduler/blob/3cabb690141cda1cee3114d8141958ab48e2fc7d/hack/update-generated-defaulters.sh#L4), [here](https://github.com/kubernetes-sigs/descheduler/blob/3cabb690141cda1cee3114d8141958ab48e2fc7d/hack/update-generated-deep-copies.sh#L4) and [here](https://github.com/kubernetes-sigs/descheduler/blob/3cabb690141cda1cee3114d8141958ab48e2fc7d/hack/update-generated-conversions.sh#L4)) are adding the dependencies for conversion-gen, deepcopy-gen, and defaulter-gen to `go.mod` and `go.sum`. The `go mod tidy` run at the end of `make gen` does not undo all of the changes.

It looks like cluster-api-provider-vsphere has a nice solution to this problem. Have a look at `Makefile`, `go.mod` and `tools.go` in their [hack/tools directory](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/tree/5c13ae4b4604f1985cb486795ecf9eafd142ab2d/hack/tools). I think we should consider implementing something similar for descheduler.

damemi:
@seanmalloy I agree, and in OpenShift we have a similar approach to that by including a [dependencymagnet package](https://github.com/openshift/cluster-kube-scheduler-operator/tree/master/pkg/dependencymagnet) that allows us to pin deps like this in our main [`go.mod`](https://github.com/openshift/cluster-kube-scheduler-operator/blob/master/go.mod#L8). Not quite as organized as the one you posted but it is a bit simpler imo. Still the same idea though

damemi:
@aveshagarwal also, for the client-go changes, I believe https://github.com/kubernetes-sigs/descheduler/pull/249 forgot to commit the informer deps (sorry @ingvagabund)

damemi:
Updated this to remove the client-go vendor changes, now it only includes vendor changes from the code generators modifying go.mod/go.sum

seanmalloy:
/lgtm

@ravisantoshgudimetla and @aveshagarwal this is ready for you to review/merge now.

seanmalloy:
@aveshagarwal and @ravisantoshgudimetla please take a look at this PR when you have some time.

seanmalloy:
/lgtm cancel

@damemi this PR needs a rebase now that #266 has been merged.

damemi:
@seanmalloy thanks, rebased, and got the travis passing. Good for another look now

seanmalloy:
@aveshagarwal and @ravisantoshgudimetla please take a look when you have some time. This PR is ready for you to review again.

aveshagarwal:
sorry if it has been discussed before. Is the goal with this strategy that the evicted pods that had too many restarts would be placed on some other nodes? I am concerned if it does not happen that descheduler might go into loop?

ingvagabund:
> Is the goal with this strategy that the evicted pods that had too many restarts would be placed on some other nodes?

With some likelihood, yes.

> I am concerned if it does not happen that descheduler might go into loop?

As long as the descheduler is run once in a while, it will not.

damemi:
@aveshagarwal the strategy itself doesn't explicitly place failed pods onto other nodes, that is still up to the owning controller to recreate the pod and the scheduler to place it. But the goal is that when the scheduler picks up the "stuck" pod again, it will make a better decision this time.

If the pod does land on the same node again, it could cause a loop. But like @ingvagabund said this depends on the frequency the descheduler is being run. Ultimately though the pod is already in a CrashLoop anyway, so this will at least make an attempt to un-stick it.

aveshagarwal:
> > Is the goal with this strategy that the evicted pods that had too many restarts would be placed on some other nodes?
> 
> With some likelihood, yes.

Actually that likeliood is one of the main goals behind each descheduler strategy and what makes, among other things, descheduler useful. The idea is to have likelihood as high as possible so that descheduler really helps with its balancing act. If likelihood is low then the strategy is not going to be much helpful. And that is my main concern with this strategy. 

Could we do something to increase the likelihood of this strategy? Right now it seems to me that likelihood is low. Please let me know if you think otherwise. 

> 
> > I am concerned if it does not happen that descheduler might go into loop?
> 
> As long as the strategy is run once in a while, it will not.

aveshagarwal:
> @aveshagarwal the strategy itself doesn't explicitly place failed pods onto other nodes, that is still up to the owning controller to recreate the pod and the scheduler to place it. 

I am aware of that. I am just wondering as I mentioned in my previous comment to Jan that is there anyway to increase the likelihood of success for this strategy. Because if likelihood is low then this strategy might not be much helpful.

But the goal is that when the scheduler picks up the "stuck" pod again, it will make a better decision this time.
> 
> If the pod does land on the same node again, it could cause a loop. But like @ingvagabund said this depends on the frequency the descheduler is being run. Ultimately though the pod is already in a CrashLoop anyway, so this will at least make an attempt to un-stick it.

ingvagabund:
> Actually that likeliood is one of the main goals behind each descheduler strategy and what makes, among other things, descheduler useful. The idea is to have likelihood as high as possible so that descheduler really helps with its balancing act. If likelihood is low then the strategy is not going to be much helpful. And that is my main concern with this strategy.

Some strategies test if there's at least one additional node that can schedule evicted pod (e.g. taints are tolerated, node set pointed by a node selector is not empty, etc.). However, strategies do not keep a state. We might make a cache of pods for a given strategy, check if the pod was evicted from the same node again and "try" to make a better decision. Though, as long as scheduler is not aware of pod's re-incarnation history, it still will be random.

@aveshagarwal you could use the same argument for any of already existing strategies. As long as re-incarnated pod do not have additional information for the scheduler, I don't think we can do anything about the likelihood.

aveshagarwal:
> @aveshagarwal you could use the same argument for any of already existing strategies. As long as re-incarnated pods do not have additional information for the scheduler, I don't think we can do anything about the likelihood.

I agree. But I'd say that with other strategies, likelihood is still very high the way they function. 

If you look at the reasons for  toomanyrestart errors, in the issue https://github.com/kubernetes-sigs/descheduler/issues/62 this PR is trying to address, you could notice that there are cases like application bug etc where this blanket strategy might not be much helpful. 

I can approve it if you guys would like to go ahead with this PR as it is. I really believe that it would be more helpful to target those cases where success of this strategy could be higher. Please let me know.

ingvagabund:
> If you look at the reasons for toomanyrestart errors, in the issue #62 this PR is trying to address, you could notice that there are cases like application bug etc where this blanket strategy might not be much helpful.

I agree. Container restarts count does not say anything about the cause. Also, have consumers decide when the strategy is helpful is not the right justification. We still need to provide a stable solution that will have more positive than negative effects in a cluster. Minimizing disturbance, etc. What can happen in the worst case? Assuming an application has a bug causing the pod to panic (e.g. right away), the pod will enter the crash loop. IINM, kubelet has a mechanism to backoff crashing containers. So, in the worst case the scheduler will schedule the pod again and again. However, the scheduler will not enter an active loop since it backoffs as well. The only drawback I can see now is the pod will never gets to run properly. Though, nor the kubelet, nor the scheduler will enter a downward spiral even when the descheduler period is set to 1s (in which case every strategy will be causing havoc in the cluster anyway).

aveshagarwal:
> > I really believe that it would be more helpful to target those cases where success of this strategy could be higher. Please let me know.
> 
> I don't think any pod instance has information that can help increase the likelihood. Only kubelet knows the reason why given container failed. Also, we don't consume any third party diagnostics (e.g. node problem detector). All the strategies are quite naive and simple in their decision making.

thanks @ingvagabund for your explanation.
/approve

PR #256, opened 2020-04-02
Fix early return from PodFitsAnyNode check

damemi:
The `PodFitsAnyNode` function returned `false` early if it hit any node with matched the selectors but was also unschedulable (before finding a suitable node).

This PR also includes vendor updates that weren't committed earlier, in order to compile.

### DIFF
--- a/pkg/descheduler/node/node.go
+++ b/pkg/descheduler/node/node.go
@@ -110,12 +110,9 @@ func PodFitsAnyNode(pod *v1.Pod, nodes []*v1.Node) bool {
 		if err != nil || !ok {
 			continue
 		}
-		if ok {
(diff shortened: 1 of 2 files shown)


pkg/descheduler/node/node.go
```diff
@@ -115,7 +115,6 @@ func PodFitsAnyNode(pod *v1.Pod, nodes []*v1.Node) bool {
 				klog.V(2).Infof("Pod %v can possibly be scheduled on %v", pod.Name, node.Name)
 				return true
 			}
-			return false
```
ingvagabund, line 118: You can drop entire `if ok` block, ok has to be true at this point.

damemi: true, updated

PR #258, opened 2020-04-02
Add verify-vendor scripts to CI

### LINKED ISSUE #257: Add CI test to verify vendor changes are committed
A test that runs `go mod tidy && go mod vendor` then compares the git diff would help prevent cases where we end up with mystery vendor changes in later PRs

damemi:
Fixes #257 

This adds a simple `verify-vendor.sh` script that ensures `go mod tidy` and `go mod vendor` have been properly run. It mostly copies from the upstream verify script: https://github.com/kubernetes/kubernetes/blob/60df45fa55746fa2e9c9c116d09c5280cba9059f/hack/verify-vendor.sh


Makefile
```diff
@@ -12,6 +12,8 @@
 # # See the License for the specific language governing permissions and
 # # limitations under the License.
 
+include ./vendor/github.com/openshift/build-machinery-go/make/targets/openshift/deps.mk
```
seanmalloy, line 15: I think it would be useful to add a comment here to note that this adds the `verify-deps` target.

damemi:
I've updated this to use simple shell scripts, similar to the pattern in upstream (actually mostly copied from k/k) in order to not depend on the openshift build target.

ingvagabund:
Before the PR gets merged, we need to make sure travis is running again.

damemi:
@ingvagabund good point, travis doesn't seem to be running on any PRs... @aveshagarwal are you aware of anything that could be causing this?

damemi:
Actually, it seems like travis *is* running but it's not being reported in the PR, see: https://travis-ci.org/github/kubernetes-sigs/descheduler/pull_requests (in which this PR actually is currently failing its own check due to the missing client-go deps that inspired it). So maybe it's a github issue?

damemi:
Following up with Ravi offline, it seems like the travis issue may be related to this intermittent problem: https://travis-ci.community/t/github-status-not-posted-on-commits-on-repositories-using-legacy-service-integration/7798

Closing and re-opening the PR looks like it fixed it, so I've also added a commit that adds the currently missing deps for client go to this PR in order to get it to pass

ravisantoshgudimetla:
To be clear, we're bringing in openshift library into descheduler, hope the other contributors are fine with it. I see @seanmalloy is fine. If anyone else has any objections, please let us know. If not, I'll approve it tomorrow.

seanmalloy:
/lgtm

The OpenShift dependencies have been removed. Simple shell scripts are being used to verify vendor dependencies now.

hack/verify-vendor.sh
```diff
@@ -0,0 +1,88 @@
...
+  if [ "${KEEP_TMP}" == "true" ]; then
+    echo "Leaving ${_tmpdir} for you to examine or copy. Please delete it manually when finished. (rm -rf ${_tmpdir})"
+  else
+    echo "Removing ${_tmpdir}"
+    rm -rf "${_tmpdir}"
+  fi
+}
+trap "cleanup" EXIT
+
+_kubetmp="${_tmpdir}"
```
ravisantoshgudimetla, line 46: nit: Why is this `kubetmp`?

hack/verify-vendor.sh
```diff
@@ -0,0 +1,88 @@
...
+mkdir -p "${_kubetmp}"
+
+git archive --format=tar --prefix=kubernetes/ "$(git write-tree)" | (cd "${_kubetmp}" && tar xf -)
+_kubetmp="${_kubetmp}/kubernetes"
```
ravisantoshgudimetla, line 50: nit: Same as above? Shouldn't it be `descheduler`?

PR #262, opened 2020-04-14
Order pods by priority only over over utilized nodes

ingvagabund:
Only over utilized nodes need clasification of pods into categories. Thus, skipping categorizing of pods which saves computation time in cases where the number of over utilized nodes makes less than 50% of all nodes or their fraction.

Also, refactoring the test code a bit in order to make lownodeutilization testing table driven.

### DIFF
--- a/pkg/descheduler/strategies/lownodeutilization.go
+++ b/pkg/descheduler/strategies/lownodeutilization.go
@@ -33,13 +33,9 @@ import (
 )
 
 type NodeUsageMap struct {
-	node             *v1.Node
-	usage            api.ResourceThresholds
-	allPods          []*v1.Pod
-	nonRemovablePods []*v1.Pod
-	bePods           []*v1.Pod
-	bPods            []*v1.Pod
-	gPods            []*v1.Pod
+	node    *v1.Node
+	usage   api.ResourceThresholds
+	allPods []*v1.Pod
 }
 
 type NodePodsMap map[*v1.Node][]*v1.Pod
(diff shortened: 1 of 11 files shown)


pkg/descheduler/strategies/lownodeutilization.go
```diff
@@ -197,11 +214,14 @@ func evictPodsFromTargetNodes(client clientset.Interface, evictionPolicyGroupVer
 		klog.V(3).Infof("evicting pods from node %#v with usage: %#v", node.node.Name, node.usage)
 		currentPodsEvicted := nodepodCount[node.node]
 
+		nonRemovablePods, bePods, bPods, gPods := clasifyPods(node.allPods, evictLocalStoragePods)
```
damemi, line 217: typo in function name, should be `classifyPods`

Also, could you change these to better variable names and comment on `classifyPods` how they are classified? `bePods` can be `bestEffortPods`, same with burstable. I'm not sure what `gPods` stands for though

ingvagabund: DOne

ingvagabund:
@aveshagarwal @ravisantoshgudimetla PTAL

pkg/descheduler/strategies/lownodeutilization.go
```diff
@@ -33,13 +33,9 @@ import (
 )
 
 type NodeUsageMap struct {
-	node             *v1.Node
-	usage            api.ResourceThresholds
-	allPods          []*v1.Pod
-	nonRemovablePods []*v1.Pod
-	bePods           []*v1.Pod
-	bPods            []*v1.Pod
-	gPods            []*v1.Pod
```
aveshagarwal, line 42: why are we removing the above info from node usage map?

ingvagabund: Since all three QoS pods are computed per each node by `classifyPods` when they are needed. Not globally for all nodes. Please see invocation of `classifyPods` in the code.

ingvagabund: To more clarify on the matter. Right now, all be/b/g pods are computed globally for all nodes. However, only over utilized nodes access all the removed fields. So, instead of classifying the pods for all nodes, they are classified only when over utilized nodes are processed. Thus, there's no longer a need to access all the removed fields. Instead, they are computed on-the-fly when they are needed.

aveshagarwal: I agree with this PR. I was just wondering if that statistics (be, burstable, guaranteed) remains part of node usage info (could be stored 0 for underutilized nodes), then it might be useful sometimes to expose that statistics to end users. Dont remember if we are already doing it now.

aveshagarwal: We can have that as part of follow up PR too. I was just bringing it up for discussion.

ingvagabund: Good point. Yeah. With the code in this PR, we can just call classifyPods even for remaining nodes in case it became configurable for admins/end users.

PR #161, opened 2019-06-11
Making test table driven

reetasingh:
making test table drive and adding more test cases

PR #266, opened 2020-04-20
Move maximum-pods-per-nodes-evicted logic under a single invocation

ingvagabund:
Each strategy implements a test for checking if a maximum number of pods per node was already evicted. The test duplicates a code that can be put under a single invocation. Thus, reducing the number of arguments passed to each strategy given EvictPod call can encapsulate both the check and the invocation of the pod eviction itself.


ingvagabund:
@damemi @seanmalloy PTAL

ingvagabund:
Integrating part of https://github.com/kubernetes-sigs/descheduler/pull/161/files as well.

ingvagabund:
Also, if this PR gets merged first, https://github.com/kubernetes-sigs/descheduler/pull/254 will need to be rebased.

pkg/descheduler/evictions/evictions.go
```diff
@@ -32,6 +32,72 @@ import (
...
+		nodepodCount:       nodePodCount,
+	}
+}
+
+// Evicted gives a number of pods evicted for node
+func (pe *PodEvictor) NodeEvicted(node *v1.Node) int {
+	return pe.nodepodCount[node]
+}
+
+// Evicted gives a number of pods evicted for node
```
seanmalloy, line 73: Can you up this comment to describe what the `TotalEvicted` method does?

pkg/descheduler/evictions/evictions.go
```diff
@@ -32,6 +32,72 @@ import (
...
+func (pe *PodEvictor) TotalEvicted() int {
+	var total int
+	for _, count := range pe.nodepodCount {
+		total += count
+	}
+	return total
+}
+
+// EvictPod returns non-nil error only when evicting a pod on a node is no
```
seanmalloy, line 82:
```suggestion
// EvictPod returns non-nil error only when evicting a pod on a node is not
```

pkg/descheduler/evictions/evictions.go
```diff
@@ -32,6 +32,72 @@ import (
...
+	}
+
+	success, err = EvictPod(pe.client, pod, pe.policyGroupVersion, pe.dryRun)
+	if success {
+		pe.nodepodCount[node]++
+		klog.V(1).Infof("Evicted pod: %#v (%#v)", pod.Name, err)
+		return success, nil
+	}
+	// err is used only for logging purposes
+	klog.Infof("Error when evicting pod: %#v (%#v)", pod.Name, err)
```
seanmalloy, line 97:
```suggestion
	klog.Errorf("Error when evicting pod: %#v (%#v)", pod.Name, err)
```

pkg/descheduler/strategies/node_taint.go
```diff
@@ -29,44 +29,33 @@ import (
...
 		for i := 0; i < totalPods; i++ {
-			if maxPodsToEvict > 0 && nodePodCount[node]+1 > maxPodsToEvict {
-				break
-			}
 			if !utils.TolerationsTolerateTaintsWithFilter(
 				pods[i].Spec.Tolerations,
 				node.Spec.Taints,
 				func(taint *v1.Taint) bool { return taint.Effect == v1.TaintEffectNoSchedule },
 			) {
-				klog.V(2).Infof("Not all taints with NoSchedule effect are tolerated after update for pod %v on node %v", pods[i].Name, node.Name)
```
seanmalloy, line 59: Are you sure you intend to remove this log message?

ingvagabund: Good catch!!!

seanmalloy:
/lgtm
/assign @damemi 

I think this looks good. I think it would be a good idea if @damemi reviews these changes too.

PR #267, opened 2020-04-20
Drop descheduler server from strategies

ingvagabund:
Removing `sigs.k8s.io/descheduler/cmd/descheduler/app/options` dependency from all strategies so they can be run independently of `options.DeschedulerServer`.

Building on top of https://github.com/kubernetes-sigs/descheduler/pull/266

### DIFF
--- a/pkg/descheduler/descheduler.go
+++ b/pkg/descheduler/descheduler.go
@@ -19,6 +19,8 @@ package descheduler
 import (
 	"fmt"
 
+	"k8s.io/api/core/v1"
+	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/klog"
 
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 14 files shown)


pkg/descheduler/descheduler.go
```diff
@@ -77,13 +77,29 @@ func RunDeschedulerStrategies(rs *options.DeschedulerServer, deschedulerPolicy *
...
+		}
+		if s := deschedulerPolicy.Strategies["RemovePodsViolatingInterPodAntiAffinity"]; s.Enabled {
+			strategies.RemovePodsViolatingInterPodAntiAffinity(rs.Client, nodes, rs.EvictLocalStoragePods, podEvictor)
+		}
+		if s := deschedulerPolicy.Strategies["RemovePodsViolatingNodeAffinity"]; s.Enabled {
+			strategies.RemovePodsViolatingNodeAffinity(rs.Client, s, nodes, rs.EvictLocalStoragePods, podEvictor)
+		}
+		if s := deschedulerPolicy.Strategies["RemovePodsViolatingNodeTaints"]; s.Enabled {
+			strategies.RemovePodsViolatingNodeTaints(rs.Client, nodes, rs.EvictLocalStoragePods, podEvictor)
+		}
```
damemi, line 102: maybe we could initialize a map of `name:strategyFunction` and put these all in a loop? That would look more like what the scheduler does and could open us up for better user configuration in the future (like being able to specify a custom ordering of the strategies to run). Something like
```
type strategyFunction func(clientset.Interface, []v1*Node, bool, *evictions.PodEvictor)

strategyFuncs := map[string]strategyFunction{
  "RemoveDuplicates": strategies.RemoveDuplicatePods,
  "LowNodeUtilization": strategies.LowNodeUtilization,
  "RemovePodsViolatingInterPodAntiAffinity": strategies.RemovePodsViolatingInterPodAntiAffinity,
  "RemovePodsViolatingNodeAffinity": strategies.RemovePodsViolatingNodeAffinity,
  "RemovePodsViolatingNodeTaints": strategies.RemovePodsViolatingNodeTaints,
}

for name, f := range strategyFuncs {
  if s := deschedulerPolicy.Strategies[name]; s.Enabled {
    f(rs.Client, nodes, rs.EvictLocalStoragePods, podEvictor)
  }
}
```

The map could be declared at the package level so we have a good `DefaultStrategy` map

ingvagabund: Yeah, I wanted to talk about this in more detail. Right now every strategy has free will about arguments that go in. Should we create an interface for a strategy? Or, use a data type for it as you suggest? Interface might be too restrictive atm. Private `strategyFunction` is a low hanging fruit that can be extended as other strategies goes in. 

tldr; Though, the list of arguments might grow too much. Later, we might implement some initialization/factory part that will get the config and produces a list of enabled strategies so we don't have to check for `s.Enabled` at all. Each strategy might register itself (as is done in the scheduler) as plugings. The factory will just accept rs.EvictLocalStoragePods and rs.Client for now. Each strategy implementing `Run([]*v1.Node, *evictions.PodEvictor)`

damemi: I agree, there is a lot of discussion that could be had around implementing more of a "framework" for the descheduler strategies. I was just thinking my change could be a quick fix to clean up the chain of if-statements, but with the side benefit of suggesting a starting framework somewhere down the line. I don't think we need to get too much into it for this PR though

pkg/descheduler/descheduler.go
```diff
@@ -56,13 +58,24 @@ func Run(rs *options.DeschedulerServer) error {
...
+
 func RunDeschedulerStrategies(rs *options.DeschedulerServer, deschedulerPolicy *api.DeschedulerPolicy, evictionPolicyGroupVersion string, stopChannel chan struct{}) error {
 	sharedInformerFactory := informers.NewSharedInformerFactory(rs.Client, 0)
 	nodeInformer := sharedInformerFactory.Core().V1().Nodes()
 
 	sharedInformerFactory.Start(stopChannel)
 	sharedInformerFactory.WaitForCacheSync(stopChannel)
 
+	strategyFuncs := map[string]strategyFunction{
+		"RemoveDuplicates":                        strategies.RemoveDuplicatePods,
```
ravisantoshgudimetla, line 71: Do you also want to make Strategies constants?

ingvagabund: We will probably switch to auto-registering strategies eventually. So the constant would be deleted anyway. I would keep it as it is for now.

PR #274, opened 2020-05-01
Add New PodLifeTime Strategy

seanmalloy:
The new PodLifeTime descheduler strategy can be used to evict pods that were created more than the configured number of seconds ago.

In the below example pods created more than 24 hours ago will be evicted.
````
kind: "DeschedulerPolicy"
strategies:
  "PodLifeTime":
    enabled: true
    params:
      maxPodLifeTimeSeconds: 86400
````

Implements feature #205.

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -5,8 +5,10 @@ Starting with descheduler release v0.10.0 container images are available in thes
 * `eu.gcr.io/k8s-artifacts-prod/descheduler/descheduler`
 * `us.gcr.io/k8s-artifacts-prod/descheduler/descheduler`
 
+## Policy Configuration Examples
 The [examples](https://github.com/kubernetes-sigs/descheduler/tree/master/examples) directory has descheduler policy configuration examples.
 
+## CLI Options
 The descheduler has many CLI options that can be used to override its default behavior.
 ```
 descheduler --help
@@ -44,4 +46,45 @@ Flags:
       --vmodule moduleSpec               comma-separated list of pattern=N settings for file-filtered logging
 
 Use "descheduler [command] --help" for more information about a command.
+```
+
+## Production Use Cases
+This section contains descriptions of real world production use cases.
+
+### Balance Cluster By Pod Age
+When initially migrating applications from a static virtual machine infrastructure to a cloud native k8s
+infrastructure there can be a tendency to treat application pods like static virtual machines. One approach
+to help prevent developers and operators from treating pods like virtual machines is to ensure that pods
+only run for a fixed amount
+of time.
+
+The `PodLifeTime` strategy can be used to ensure that old pods are evicted. It is recommended to create a
+[pod disruption budget](https://kubernetes.io/docs/tasks/run-application/configure-pdb/) for each
+application to ensure application availability.
+```
+descheduler -v=3 --evict-local-storage-pods --policy-config-file=pod-life-time.yml
+```
+
+This policy configuration file ensures that pods created more than 7 days ago are evicted.
+```
+---
+apiVersion: "descheduler/v1alpha1"
+kind: "DeschedulerPolicy"
+strategies:
+  "LowNodeUtilization":
+    enabled: false
+  "RemoveDuplicates":
+    enabled: false
+  "RemovePodsViolatingInterPodAntiAffinity":
+    enabled: false
+  "RemovePodsViolatingNodeAffinity":
+    enabled: false
+  "RemovePodsViolatingNodeTaints":
+    enabled: false
+  "RemovePodsHavingTooManyRestarts":
+    enabled: false
+  "PodLifeTime":
+    enabled: true
+    params:
+      maxPodLifeTimeSeconds: 604800 # pods run for a maximum of 7 days
 ```
--- a/pkg/api/v1alpha1/types.go
+++ b/pkg/api/v1alpha1/types.go
@@ -17,7 +17,7 @@ limitations under the License.
 package v1alpha1
 
 import (
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
@@ -49,6 +49,7 @@ type StrategyParameters struct {
 	NodeResourceUtilizationThresholds *NodeResourceUtilizationThresholds `json:"nodeResourceUtilizationThresholds,omitempty"`
 	NodeAffinityType                  []string                           `json:"nodeAffinityType,omitempty"`
 	PodsHavingTooManyRestarts         *PodsHavingTooManyRestarts         `json:"podsHavingTooManyRestarts,omitempty"`
+	MaxPodLifeTimeSeconds             *uint                              `json:"maxPodLifeTimeSeconds,omitempty"`
 }
 
 type Percentage float64
--- a/pkg/descheduler/descheduler.go
+++ b/pkg/descheduler/descheduler.go
@@ -74,6 +74,7 @@ func RunDeschedulerStrategies(rs *options.DeschedulerServer, deschedulerPolicy *
 		"RemovePodsViolatingNodeAffinity":         strategies.RemovePodsViolatingNodeAffinity,
 		"RemovePodsViolatingNodeTaints":           strategies.RemovePodsViolatingNodeTaints,
 		"RemovePodsHavingTooManyRestarts":         strategies.RemovePodsHavingTooManyRestarts,
+		"PodLifeTime":                             strategies.PodLifeTime,
 	}
 
 	wait.Until(func() {
--- a/pkg/descheduler/strategies/pod_lifetime.go
+++ b/pkg/descheduler/strategies/pod_lifetime.go
@@ -1,0 +1,69 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package strategies
+
+import (
+	v1 "k8s.io/api/core/v1"
+	v1meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/klog"
+
+	"sigs.k8s.io/descheduler/pkg/api"
+	"sigs.k8s.io/descheduler/pkg/descheduler/evictions"
+	podutil "sigs.k8s.io/descheduler/pkg/descheduler/pod"
+)
+
+// PodLifeTime evicts pods on nodes that were created more than strategy.Params.MaxPodLifeTimeSeconds seconds ago.
+func PodLifeTime(client clientset.Interface, strategy api.DeschedulerStrategy, nodes []*v1.Node, evictLocalStoragePods bool, podEvictor *evictions.PodEvictor) {
+	if strategy.Params.MaxPodLifeTimeSeconds == nil {
+		klog.V(1).Infof("MaxPodLifeTimeSeconds not set")
+		return
+	}
+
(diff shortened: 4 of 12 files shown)

### CODE REVIEW

**Comment by seanmalloy** [comment]
/hold

I'm going to test this code out on some real world clusters before asking for a review. All the unit tests that I created are passing, so it appears the code is "correct", but I'd like to try it out in a real cluster to see how it behaves.

**Comment by seanmalloy** [comment]
Closing to re-run Travis CI tests.

**Comment by seanmalloy** [comment]
@damemi thanks for the feedback. I'm fine with using an `int` for the `maxPodLifeTimeSeconds` field or switching it to a more generic `maxPodLifeTime` field and using a `time.Duration`. Either way works for me. If the consensus is that a `time.Duration` would be better then I'd be more than happy to change it.

File: pkg/descheduler/strategies/pod_lifetime.go [hunk 1]
```diff
@@ -0,0 +1,63 @@
...
+
+func listOldPodsOnNode(client clientset.Interface, node *v1.Node, maxAge uint, evictLocalStoragePods bool) []*v1.Pod {
+	pods, err := podutil.ListEvictablePodsOnNode(client, node, evictLocalStoragePods)
+	if err != nil {
+		return nil
+	}
+
+	var oldPods []*v1.Pod
+	for _, pod := range pods {
+		podAgeSeconds := uint(v1meta.Now().Sub(pod.GetCreationTimestamp().Local()).Seconds())
```
seanmalloy, line 56: I'm open to suggestions on a more elegant way get the pod age.

seanmalloy:
/hold cancel
/assign @damemi @ingvagabund 

@damemi and @ingvagabund PTAL when you have some time. Any feedback you have would be greatly appreciated. I was able to successfully test the new strategy on a real cluster, it seems to be working correctly.

pkg/descheduler/strategies/pod_lifetime.go
```diff
@@ -0,0 +1,63 @@
...
+	podutil "sigs.k8s.io/descheduler/pkg/descheduler/pod"
+)
+
+// PodLifeTime evicts pods on nodes that were created more than strategy.Params.MaxPodLifeTimeSeconds seconds ago.
+func PodLifeTime(client clientset.Interface, strategy api.DeschedulerStrategy, nodes []*v1.Node, evictLocalStoragePods bool, podEvictor *evictions.PodEvictor) {
+	for _, node := range nodes {
+		klog.V(1).Infof("Processing node: %#v", node.Name)
+		pods := listOldPodsOnNode(client, node, strategy.Params.MaxPodLifeTimeSeconds, evictLocalStoragePods)
+		for _, pod := range pods {
+			success, err := podEvictor.EvictPod(pod, node)
```
ingvagabund, line 36: If possible, we should check if any evicted pod is still schedulable to at least one node. In this case it can be the same.

pkg/descheduler/strategies/pod_lifetime.go
```diff
@@ -0,0 +1,63 @@
...
+			if success {
+				klog.V(1).Infof("Evicted pod: %#v\n because it was created more than %v seconds ago", pod.Name, strategy.Params.MaxPodLifeTimeSeconds)
+			}
+
+			if err != nil {
+				klog.Errorf("Error evicting pod: (%#v)", err)
```
ingvagabund, line 42: Right now, non-nil error means the maximal number of pods evicted per node was exceeded. In future we might test for a specific error. So I suggest to break here. Yet, it's optional since, in the worst case you will see a bunch of `Error eviction pods` messages.

pkg/descheduler/strategies/pod_lifetime.go
ingvagabund: Can't think of anything better. As long as all the clocks are synced, it's safe. Apiserver is the component responsible for the creation timestamp. Unless apiserver provides means to get the current ts :P, it's only best-effort.

pkg/descheduler/strategies/pod_lifetime.go
seanmalloy: Fixed in 7a88755cd9be07a3e287cc752811afe1afefb82e

pkg/api/v1alpha1/types.go
```diff
@@ -49,6 +49,7 @@ type StrategyParameters struct {
 	NodeResourceUtilizationThresholds *NodeResourceUtilizationThresholds `json:"nodeResourceUtilizationThresholds,omitempty"`
 	NodeAffinityType                  []string                           `json:"nodeAffinityType,omitempty"`
 	PodsHavingTooManyRestarts         PodsHavingTooManyRestarts          `json:"podsHavingTooManyRestarts,omitempty"`
+	MaxPodLifeTimeSeconds             uint                               `json:"maxPodLifeTimeSeconds"`
```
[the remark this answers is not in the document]
seanmalloy: Fixed in 0918eee5f7c20e4e119c48ae01b0346b479c0924

pkg/descheduler/strategies/pod_lifetime.go
seanmalloy: @ingvagabund I want to make sure I understand your suggestion correctly. You think the code should check if the pod can be rescheduled on at least one other node before evicting it?

Something like ...
```go
// podCanBeRescheduled is a bool that determines if the pod that will be evicted can be scheduled on at least one other node
if podCanBeRescheduled {
    success, err := podEvictor.EvictPod(pod, node)
} else {
    klog.V(1).Infof("Skipping pod: %#v\n because it cannot be rescheduled", pod.Name)
}
```

ingvagabund: > Something like ...
yes,

e.g. if there's still a node whose taints are respected. Some function/method that will be called across all strategies. Though, not a blocker for the PR.

pkg/descheduler/strategies/pod_lifetime_test.go
```diff
@@ -83,6 +83,9 @@ func TestPodLifeTime(t *testing.T) {
 	p5.ObjectMeta.OwnerReferences = ownerRef4
 	p6.ObjectMeta.OwnerReferences = ownerRef4
 
+	var maxLifeTime *uint
+	maxLifeTime = new(uint)
+	*maxLifeTime = 600
```
ingvagabund, line 88:
```
var maxLifeTime uint = 600
...
MaxPodLifeTimeSeconds: &maxLifeTime
```

will do as well.

seanmalloy: Fixed in 51eed624bf7370917428581acb3f4d62d3b114ad

seanmalloy:
@ingvagabund and @damemi please take another look when you have some time. I added two commits since you reviewed last.
* 51eed624bf7370917428581acb3f4d62d3b114ad refactor to simplify pointer usage in unit tests
* aee13457a3e0e1c39bfa71a30af693967c980292 prevent nil pointer dereference when MaxPodLifeTimeSeconds is not set in the policy config file

I'd also like to squash commits prior to merging, so the commit history is not so messy.

ingvagabund:
@seanmalloy lgtm, +1 for squashing

seanmalloy:
/assign @ravisantoshgudimetla @aveshagarwal 

Commits have been squashed. @ravisantoshgudimetla and @aveshagarwal please review when you have some time. Thanks!

seanmalloy:
Closing to re-run Travis CI tests.

seanmalloy:
@ravisantoshgudimetla and @aveshagarwal the Travis CI test are all passing. Ready for review let me know if you have any questions.

aveshagarwal:
@seanmalloy could you explain what is the use case for this or in what practical scenarios, this strategy is going to be useful?

seanmalloy:
@aveshagarwal here is my real world use case ....

Running a k8s cluster in an on-prem data center(not a public cloud) I would like to be able to ensure that application pods are resilient to restarts. Therefore I would like to evict all pods that are considered old(i.e. older than 7 days). This helps prevent application teams from treating pods like pets. Treating pods like pets instead of cattle can happen when an application initially migrates from virtual machines into k8s. It also helps make underlying node patching less of a special event because all support teams are conditioned to pods being evicted constantly. The new descheduler strategy implemented in this PR could be used to ensure that old pods get evicted.

This new descheduler strategy could also be used in a k8s cluster running in a public cloud, but in my opinion it is easier to run [k8s cluster autoscaler](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler) and just delete old nodes(i.e. older than 7 days). Deleting old nodes could be accomplished using [node-problem-detetor](https://github.com/kubernetes/node-problem-detector#remedy-systems) and the descheduler `RemovePodsViolatingNodeTaints` strategy.

I do have some proprietary code that currently implements deleting k8s nodes older than X days in a public cloud environment. I'm hoping to leverage the descheduler in public cloud and on-prem to ensure that old pods are evicted regularly.

Also, I did present this new strategy in a SIG scheduling meeting earlier this year. A formal proposal Google Doc was not created for SIG scheduling to review.

pkg/descheduler/descheduler.go
```diff
@@ -19,7 +19,7 @@ package descheduler
 import (
 	"fmt"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
aveshagarwal, line 22: why this change?

aveshagarwal:
> @aveshagarwal here is my real world use case ....
> 
> Running a k8s cluster in an on-prem data center(not a public cloud) I would like to be able to ensure that application pods are resilient to restarts. Therefore I would like to evict all pods that are considered old(i.e. older than 7 days). This helps prevent application teams from treating pods like pets. Treating pods like pets instead of cattle can happen when an application initially migrates from virtual machines into k8s. It also helps make underlying node patching less of a special event because all support teams are conditioned to pods being evicted constantly. The new descheduler strategy implemented in this PR could be used to ensure that old pods get evicted.

hmm.. interesting use case. Seems like a training strategy for applications (and also application developers/admins) for setting their expectations right in k8s environments.

Could you add something about this use case to README too where you explained about this strategy? 

> 
> This new descheduler strategy could also be used in a k8s cluster running in a public cloud, but in my opinion it is easier to run [k8s cluster autoscaler](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler) and just delete old nodes(i.e. older than 7 days). Deleting old nodes could be accomplished using [node-problem-detetor](https://github.com/kubernetes/node-problem-detector#remedy-systems) and the descheduler `RemovePodsViolatingNodeTaints` strategy.
> 

But it might not be feasible to always to delete node as they might contain pods with various lifetimes some old some new. Also, it might not be feasible in every cluster. But I get your point i think.

> I do have some proprietary code that currently implements deleting k8s nodes older than X days in a public cloud environment. I'm hoping to leverage the descheduler in public cloud and on-prem to ensure that old pods are evicted regularly.
> 
> Also, I did present this new strategy in a SIG scheduling meeting earlier this year. A formal proposal Google Doc was not created for SIG scheduling to review.

Thats good to know.

pkg/descheduler/descheduler.go
seanmalloy: I believe VS Code made this change automatically. I can change it back.

seanmalloy: Fixed.

seanmalloy:
> Could you add something about this use case to README too where you explained about this strategy?

@aveshagarwal see commit 05496b2f7b75aa68a5c7de3461745cf56636c46e. I decided to add this to the user guide instead of the README. I'm afraid the README will get to long.

docs/user-guide.md
```diff
@@ -44,4 +46,44 @@ Flags:
...
\ No newline at end of file
+```
+
+## Production Use Cases
+This section contains descriptions of real world production use cases.
+
+### Balance Cluster By Pod Age
+When initially migrating applications from a static virtual machine infrastructure to a cloud native k8s
+infrastructure there can be a tendency to treat application pods like static virtual machines. One approach
+to helping developers and operators break this bad habit is to ensure that pods only run for a fixed amount
```
aveshagarwal, line 57: nit: @seanmalloy could you rephrase this line a bit specifically avoiding "bad habit"?

seanmalloy: Fixed. The phrase "bad habit" has been removed.

seanmalloy:
@aveshagarwal I fixed your last review comment. Let me know if any other changes are required.

seanmalloy:
Closing to re-run Travis CI tests.

aveshagarwal:
thanks @seanmalloy for this PR.
/lgtm

Issue #230, opened 2020-02-05
DuplicatePods strategy might be too broad


damemi:
Looking at the code for the `RemoveDuplicatePods` strategy, it seems to consider pods as duplicates of each other if they have the same owner and are in the same namespace.

However, wouldn't it be possible that 2 entirely different pods could have the same owner? Say for example I have a controller that creates pods that serve different functions, but are both owned by the same controller. Would RemoveDuplicates consider these to be duplicate, and if so how could I know which one is going to be deleted?

ingvagabund:
> However, wouldn't it be possible that 2 entirely different pods could have the same owner? Say for example I have a controller that creates pods that serve different functions, but are both owned by the same controller. Would RemoveDuplicates consider these to be duplicate, and if so how could I know which one is going to be deleted?

Pods owned by the same controller (running on the same node) originating from the same RC/RS/... are seen as cattle. Is your question about "which one is going to be deleted" still valid under that assumption?

aveshagarwal:
when pods belong to same controller and are more than one on same node, then except one, rest are deleted and selection is random.

vincent-pli:
I think the `Duplicate` should defined by customer, could we introduce list which contains all `owner`s  which could generate `Duplicate`  pod?
For example, a build in list should contain: `Deployment`, `ReplicaSet`, it should editable by end user.

seanmalloy:
> I think the `Duplicate` should defined by customer, could we introduce list which contains all `owner`s which could generate `Duplicate` pod?
> For example, a build in list should contain: `Deployment`, `ReplicaSet`, it should editable by end user.

@vincent-pli this sounds like an interesting feature request. Can you open a new separate issue for this feature request? This will allow us to track it more easily. Thanks!

vincent-pli:
@seanmalloy 
Sure, done, thanks

damemi:
As another example use case for this: in OpenShift, [DeploymentConfigs](https://docs.openshift.com/container-platform/4.1/applications/deployments/what-deployments-are.html#deployments-and-deploymentconfigs_what-deployments-are) create a separate `deploy` pod to handle the rollout of new deployments. This deployer pod has the same owner as the workload pods, but will be considered a duplicate under the current strategy. If it's deleted however, it is not recreated and breaks deployments.

So to add to the suggestion in https://github.com/kubernetes-sigs/descheduler/issues/243, it would be helpful if a user could set other criteria/fields to compare against when considering duplicates.

[...]

Issue #243, opened 2020-02-24
Introduce list for "Duplicate" decision


vincent-pli:
As mentioned in #230 , the `DuplicatePods` strategy is too broad, 2 pods in same `namespace` with same `owner` will be treated as `Douplicate`, that's rough. 
I know project [tekton-pipeline](https://github.com/tektoncd/pipeline), a `Taskrun` resource will generate unlimited `pod` and they all have deep relationship with each other.

So I suggest introduce a list for `DuplicatePods` strategy, let user decide which resource as `Owner` then it's children should be treated as `Douplicate`, we could has several default candidates (Deployment, ReplicaSet), and user could reset it.

damemi:
This parameter may be added in https://github.com/kubernetes-sigs/descheduler/pull/275 (specifically https://github.com/kubernetes-sigs/descheduler/pull/275/commits/c11cb4f52bea5c71e3fa519b6a58aabdcdce13a4), please feel free to add any review or feedback you may have to that PR

damemi:
closed via #275

PR #275, opened 2020-05-01
Consider pod image in duplicates strategy

damemi:
Issue: https://github.com/kubernetes-sigs/descheduler/issues/230

This adds the images in a pod to the key for determining duplicates along with the ownerref. The intended effect of this is that now instead of just checking if 2 pods have the same owner, they will also need to both have a container with the same image.


damemi:
This strategy should probably be refactored a bit, but for a quick and dirty fix I think this will tighten up the duplicates edge case

damemi:
@ingvagabund thanks, you are right that it would lead to multiple attempts to delete the same pod, so I added your changes, please take a look. I think I was actually able to simplify from the code you suggested, so if you can verify that I understood your point that would help

damemi:
Actually, I think my previous changes would have caused pods with more than 1 container to automatically match on themselves, so I've made the following logic changes:

* within the containers loop, build a `podKeys` list of fully-qualified owner/image keys that will be matched in `dpm`
* at the end of the pod loop, add the pod to `dpm` for each key, as before (but now only in the pod loop, not the containers loop)
* if we find a match, add this pod to the list of duplicates (this will ensure that the pod only matches on other pods). Then continue adding the rest of the keys (in case future pods may match on them) but we don't need to append this pod anymore

damemi:
I also added tests to confirm this works as expected

@aveshagarwal could you take a look as well?

damemi:
cc @ingvagabund could you take another look at the solution we came to here? I slightly modified your suggestions (explained above) and added a test which looks like it confirms this is now working as expected

pkg/descheduler/strategies/duplicates.go
```diff
@@ -42,40 +43,51 @@ func RemoveDuplicatePods(
...
+	duplicatePods := make([]*v1.Pod, 0, len(pods))
+	dpm := map[string][]*v1.Pod{}
 	for _, pod := range pods {
 		ownerRefList := podutil.OwnerRef(pod)
+		podKeys := make([]string, 0, len(ownerRefList)*len(pod.Spec.Containers))
 		for _, ownerRef := range ownerRefList {
-			// Namespace/Kind/Name should be unique for the cluster.
-			s := strings.Join([]string{pod.ObjectMeta.Namespace, ownerRef.Kind, ownerRef.Name}, "/")
-			dpm[s] = append(dpm[s], pod)
+			for _, container := range pod.Spec.Containers {
```
ingvagabund, line 70: There's a strong assumption the owner is either RC, RS, SS or alike. I.e. one template for all the instances. Also, there's no policy disallowing to create an owner that's not template based. I.e. all pods owned by the same owner  kind/name but each pod created from a different template. As you now mention in the strategy description: `A pod is said to be a duplicate of other if both of them are from same creator, kind and are within the same namespace, and have at least one container with the same image.` It's perfectly valid for template based owners. However, for non-template based owners, one might face the following situation:
pod1:
- container: image1
- container: image2
pod2:
- container: image1
- container: image3
Which, based on the current definition of a duplicate, `pod1` is a duplicate of `pod2`. What about to extend the definition to say:
```
A pod is a duplicate of another pod if both pods have:
- the same owner kind/name
- the same namespace
- the same list of containers (the same list of names, each two equal names have the same image).
```
?

damemi: @ingvagabund This was actually a problem I considered, whether we want to consider duplicates when there is an exact matching list of images or if just one match is enough. I originally went with the latter, but can see that there is a case for matching exact lists.

So I have pushed https://github.com/kubernetes-sigs/descheduler/pull/275/commits/365e355e28aac45587b4b14056665e54223a4c9b, which updates the strategy to the strictest definition, where pods are only duplicates if they have the exact same list of owners and images.

But because now we're checking uniqueness against a list of strings (instead of just one string), the map we had before wouldn't work (unless we came up with a way to represent the entire list as a unique string, which could get long and messy).

So now, I'm building the list of all the ownerref/image keys for a pod, then sorting that list. I'm keying the duplicates map on the first entry of that list (because if another pod doesn't have the same first entry, it's not a duplicate). The value for that key in the map is a list of all the lists which contain the same first key. If we find a matching list, the pod is a duplicate.

I was going to just make a simple loop using the existing map, where we check if all the key entries have >1 existing match like before. But this could give a false positive for pods with shorter lists.

Please take a look at the new commit (I've kept it separate for now to review easier)

pkg/descheduler/strategies/duplicates.go
```diff
@@ -75,18 +89,21 @@ func listDuplicatePodsOnANode(ctx context.Context, client clientset.Interface, n
 				podKeys = append(podKeys, s)
 			}
 		}
+		sort.Strings(podKeys)
```
ingvagabund, line 92: Assuming the following pods owned by the same (namespace, kind, owner) triple:
```
- pod1:
  - image1
  - image2
- pod2:
  - image2
  - image1
```
Do we want to consider both pods to be duplicates of each other?

Also, the following case independent of ordering where one image contains multiple binaries:
```
- pod1:
  - image1 (containername=withbinary1)
  - image1 (containername=withbinary2)
- pod2:
  - image1 (containername=withbinary3)
  - image1 (containername=withbinary4)
```
Each binary responsible for a different aspect of a customer solution. We could keep going and discuss the case where two pods have the exact same list of containers (wrt. image and container name), yet both pods running different binaries. Which will make the logic unnecessarily complicated. 

This can eventually lead to customers reporting strange behavior for their custom controllers. As long as the owner kind is a known controller (RC, RS, SS, Deployment, ..), we are safe. However, we need to stress in the strategy description it's meant to be run only over template based owners. Or, allow to explicitly skip specific list of kinds. @damemi what do you think about adding a new `ignoreOwnerKinds` param to the strategy?

pkg/descheduler/strategies/duplicates.go
```diff
@@ -62,7 +63,20 @@ func listDuplicatePodsOnANode(ctx context.Context, client clientset.Interface, n
...
+	//
+	// duplicateKeysMap maps the first Namespace/Kind/Name/Image in a pod's list to a 2D-slice of all the other lists where that is the first key
+	// (Since we sort each pod's list, we only need to key the map on the first entry in each list. Any pod that doesn't have
+	// the same first entry is clearly not a duplicate. This makes lookup quick and minimizes storage needed).
+	// If any of the existing lists for that first key matches the current pod's list, the current pod is a duplicate.
+	// If not, then we add this pod's list to the list of lists for that key.
+	duplicateKeysMap := map[string][][]string{}
 	for _, pod := range pods {
 		ownerRefList := podutil.OwnerRef(pod)
 		podKeys := make([]string, 0, len(ownerRefList)*len(pod.Spec.Containers))
```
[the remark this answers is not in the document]
damemi: I want to make it clear that this is a collection of container keys essentially representing a pod, so I went with `podContainerKeys`

pkg/descheduler/strategies/duplicates.go
damemi: We talked about this a little offline so for reference here, I think that those 2 cases are reasonably out of the scope for what we should take into consideration for a duplicate pod.

The ordering of containers in a pod doesn't have any effect on their actual output, so I think it's safe to ignore that. And we can't really know what binaries are in the image, but we could look at the commands that are being run in the container. I think that's excessive though, but if there ends up being demand for that level of specificity I'm open to revisiting it.

Ultimately with these changes we have already made the strategy much stricter than it was before, which is a big improvement. I also pushed commit https://github.com/kubernetes-sigs/descheduler/pull/275/commits/c11cb4f52bea5c71e3fa519b6a58aabdcdce13a4, which adds a parameter to exclude certain kinds of ownerrefs as you suggested (and also addresses the request in https://github.com/kubernetes-sigs/descheduler/issues/243). 

One caveat to note about this parameter is that it becomes all-or-nothing for considering a pod a duplicate. That is, if the pod has multiple ownerrefs and one of them is an excluded kind, we have to consider the whole pod ineligible for eviction. Because for example, if we just skipped adding the keys for that particular ownerref, it would actually make the pod *more* likely to match another pod (as it would have less keys identifying its uniqueness)

Please let me know if the new parameter matches what you had in mind, and if so I can squash the commits down.

ingvagabund: Looks good!!! Please squash, I will lgtm then.

damemi:
Squashed, and updated with a better commit message to describe the changes.

PR #279, opened 2020-05-12
Add Additional Details To Pod Eviction Log Messages

seanmalloy:
* 3a204dec97f3c4f020ef11d3e97e1aa6686e944d - all strategies log error from EvictPod method
* 55de4417949b59c204b388b1175f2173b7de280a - add pod namespace to pod eviction log messages

### DIFF
--- a/pkg/descheduler/evictions/evictions.go
+++ b/pkg/descheduler/evictions/evictions.go
@@ -83,25 +83,26 @@ func (pe *PodEvictor) TotalEvicted() int {
 // EvictPod returns non-nil error only when evicting a pod on a node is not
 // possible (due to maxPodsToEvict constraint). Success is true when the pod
 // is evicted on the server side.
-func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod, node *v1.Node) (success bool, err error) {
+func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod, node *v1.Node) (bool, error) {
 	if pe.maxPodsToEvict > 0 && pe.nodepodCount[node]+1 > pe.maxPodsToEvict {
 		return false, fmt.Errorf("Maximum number %v of evicted pods per %q node reached", pe.maxPodsToEvict, node.Name)
 	}
 
-	success, err = EvictPod(ctx, pe.client, pod, pe.policyGroupVersion, pe.dryRun)
-	if success {
-		pe.nodepodCount[node]++
-		klog.V(1).Infof("Evicted pod: %#v in namespace %#v", pod.Name, pod.Namespace)
-		return success, nil
+	err := evictPod(ctx, pe.client, pod, pe.policyGroupVersion, pe.dryRun)
+	if err != nil {
+		// err is used only for logging purposes
+		klog.Errorf("Error evicting pod: %#v in namespace %#v (%#v)", pod.Name, pod.Namespace, err)
+		return false, nil
 	}
-	// err is used only for logging purposes
-	klog.Errorf("Error evicting pod: %#v in namespace %#v (%#v)", pod.Name, pod.Namespace, err)
-	return false, nil
+
+	pe.nodepodCount[node]++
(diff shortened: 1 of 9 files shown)


seanmalloy:
Looks like the Travis CI tests did not get triggered :-(

seanmalloy:
Close and re-open to try and make Travis CI work.

pkg/descheduler/evictions/evictions.go
```diff
@@ -90,11 +90,11 @@ func (pe *PodEvictor) EvictPod(pod *v1.Pod, node *v1.Node) (success bool, err er
 	success, err = EvictPod(pe.client, pod, pe.policyGroupVersion, pe.dryRun)
 	if success {
 		pe.nodepodCount[node]++
-		klog.V(1).Infof("Evicted pod: %#v (%#v)", pod.Name, err)
+		klog.V(1).Infof("Evicted pod: %#v in namespace %#v", pod.Name, pod.Namespace)
 		return success, nil
 	}
 	// err is used only for logging purposes
-	klog.Errorf("Error when evicting pod: %#v (%#v)", pod.Name, err)
+	klog.Errorf("Error evicting pod: %#v in namespace %#v (%#v)", pod.Name, pod.Namespace, err)
```
damemi, line 97: If you're adding logs everywhere `EvictPod` is called to log the error, do you also need to log it here? I'd say we could get rid of this and keep the logs elsewhere

ingvagabund: `err` logged here is different from the error logged outside. If `EvictPod` returns non-nil error, this line is not executed.

pkg/descheduler/evictions/evictions.go
```diff
@@ -90,11 +90,11 @@ func (pe *PodEvictor) EvictPod(pod *v1.Pod, node *v1.Node) (success bool, err er
 	success, err = EvictPod(pe.client, pod, pe.policyGroupVersion, pe.dryRun)
 	if success {
```
ingvagabund, line 91: Based on the current implementation of `EvictPod`, when `success` is `true`, `err` is always `nil`, when `success` is `false`, `err` is never `nil`. @seanmalloy since you started, can you remove the `bool` return argument from EvictPod and simply return just the error?

pkg/descheduler/evictions/evictions.go
damemi: I think with your point in https://github.com/kubernetes-sigs/descheduler/pull/279/files#r424363568 though it might be redundant. So if we change `EvictPod` to just return an error, then `podEvictor.EvictPod` here can also be changed to just return an error right? No need for the bool anymore

ingvagabund: If the error is not nil, `EvictPod` signals the caller to stop evicting any pod. In our case it's due to reaching the maximum number of pods descheduler is allowed to evict per a node. Any other attempt to evict a pod from the same node will result in non-nil error. Thus, there's no reason to iterate through other pods. Whereas, in case of nil error, both `(true, nil)` and `(false, nil)` are valid cases. Negative success allows the strategy to react on the fact a pod was not evicted. In case it's not a blocker for a particular strategy, it may decide to continue evicting. Other strategies may decide to break and continue processing another node.

damemi: Ah I see, in that case this lgtm. Thanks for clarifying @ingvagabund

pkg/descheduler/evictions/evictions.go
seanmalloy: Fixed in commit 7039b6c8aafbad333fa21b3604ba1dae0a7f9b15

pkg/descheduler/evictions/evictions.go
seanmalloy: Sounds like no changes are required. Marking as resolved.

seanmalloy:
@damemi and @ingvagabund PTAL when you have some time. I believe I addressed all of your feedback.

ingvagabund:
@seanmalloy thanks for addressing all the comments
/lgtm

PR #285, opened 2020-05-15
Change break condition and thresholds validation for lowUtilization

lixiang233:
We should stop evicting pods when any of totalCPU/totalMem/totalPods ran out, otherwise we may make underutilized nodes overutilized which is not efficient. To do this, there should be a value of totalCPU/Mem if their targetThreshold is not configured, here I set it to all of the remaining resources 
 on these nodes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,8 @@ usage is below threshold for all (cpu, memory, and number of pods), the node is 
 Currently, pods request resource requirements are considered for computing node resource utilization.
 
 There is another configurable threshold, `targetThresholds`, that is used to compute those potential nodes
-from where pods could be evicted. Any node, between the thresholds, `thresholds` and `targetThresholds` is
+from where pods could be evicted. If a node's usage is above targetThreshold for any (cpu, memory, or number of pods),
+the node is considered over utilized. Any node between the thresholds, `thresholds` and `targetThresholds` is
 considered appropriately utilized and is not considered for eviction. The threshold, `targetThresholds`,
 can be configured for cpu, memory, and number of pods too in terms of percentage.
 
@@ -113,6 +114,15 @@ strategies:
            "memory": 50
            "pods": 50
 ```
+
+Policy should pass the following validation checks:
+* Only three types of resources are supported: `cpu`, `memory` and `pods`.
+* `thresholds` or `targetThresholds` can not be nil and they must configure exactly the same types of resources.
+* The valid range of the resource's percentage value is \[0, 100\]
+* Percentage value of `thresholds` can not be greater than `targetThresholds` for the same resource.
+
+If any of the resource types is not specified, all its thresholds default to 100% to avoid nodes going
+from underutilized to overutilized.
 
 There is another parameter associated with the `LowNodeUtilization` strategy, called `numberOfNodes`.
 This parameter can be configured to activate the strategy only when the number of under utilized nodes
--- a/pkg/descheduler/strategies/lownodeutilization.go
+++ b/pkg/descheduler/strategies/lownodeutilization.go
@@ -18,6 +18,7 @@ package strategies
 
 import (
 	"context"
+	"fmt"
 	"sort"
 
 	v1 "k8s.io/api/core/v1"
@@ -40,6 +41,11 @@ type NodeUsageMap struct {
 
 type NodePodsMap map[*v1.Node][]*v1.Pod
 
+const (
+	MinResourcePercentage = 0
+	MaxResourcePercentage = 100
+)
+
 func LowNodeUtilization(ctx context.Context, client clientset.Interface, strategy api.DeschedulerStrategy, nodes []*v1.Node, evictLocalStoragePods bool, podEvictor *evictions.PodEvictor) {
 	if !strategy.Enabled {
 		return
@@ -52,12 +58,23 @@ func LowNodeUtilization(ctx context.Context, client clientset.Interface, strateg
 	}
 
 	thresholds := strategy.Params.NodeResourceUtilizationThresholds.Thresholds
-	if !validateThresholds(thresholds) {
+	targetThresholds := strategy.Params.NodeResourceUtilizationThresholds.TargetThresholds
+	if err := validateStrategyConfig(thresholds, targetThresholds); err != nil {
+		klog.Errorf("LowNodeUtilization config is not valid: %v", err)
 		return
 	}
-	targetThresholds := strategy.Params.NodeResourceUtilizationThresholds.TargetThresholds
-	if !validateTargetThresholds(targetThresholds) {
-		return
+	// check if Pods/CPU/Mem are set, if not, set them to 100
+	if _, ok := thresholds[v1.ResourcePods]; !ok {
+		thresholds[v1.ResourcePods] = MaxResourcePercentage
+		targetThresholds[v1.ResourcePods] = MaxResourcePercentage
+	}
+	if _, ok := thresholds[v1.ResourceCPU]; !ok {
+		thresholds[v1.ResourceCPU] = MaxResourcePercentage
+		targetThresholds[v1.ResourceCPU] = MaxResourcePercentage
+	}
+	if _, ok := thresholds[v1.ResourceMemory]; !ok {
+		thresholds[v1.ResourceMemory] = MaxResourcePercentage
+		targetThresholds[v1.ResourceMemory] = MaxResourcePercentage
 	}
 
 	npm := createNodePodsMap(ctx, client, nodes)
@@ -102,37 +119,46 @@ func LowNodeUtilization(ctx context.Context, client clientset.Interface, strateg
 	klog.V(1).Infof("Total number of pods evicted: %v", podEvictor.TotalEvicted())
 }
 
-func validateThresholds(thresholds api.ResourceThresholds) bool {
-	if thresholds == nil || len(thresholds) == 0 {
-		klog.V(1).Infof("no resource threshold is configured")
-		return false
+// validateStrategyConfig checks if the strategy's config is valid
+func validateStrategyConfig(thresholds, targetThresholds api.ResourceThresholds) error {
+	// validate thresholds and targetThresholds config
+	if err := validateThresholds(thresholds); err != nil {
+		return fmt.Errorf("thresholds config is not valid: %v", err)
 	}
-	for name := range thresholds {
-		switch name {
-		case v1.ResourceCPU:
-			continue
-		case v1.ResourceMemory:
-			continue
-		case v1.ResourcePods:
-			continue
-		default:
-			klog.Errorf("only cpu, memory, or pods thresholds can be specified")
-			return false
+	if err := validateThresholds(targetThresholds); err != nil {
+		return fmt.Errorf("targetThresholds config is not valid: %v", err)
+	}
+
+	// validate if thresholds and targetThresholds have same resources configured
+	if len(thresholds) != len(targetThresholds) {
+		return fmt.Errorf("thresholds and targetThresholds configured different resources")
+	}
+	for resourceName, value := range thresholds {
+		if targetValue, ok := targetThresholds[resourceName]; !ok {
+			return fmt.Errorf("thresholds and targetThresholds configured different resources")
+		} else if value > targetValue {
+			return fmt.Errorf("thresholds' %v percentage is greater than targetThresholds'", resourceName)
 		}
 	}
-	return true
+	return nil
 }
 
-//This function could be merged into above once we are clear.
-func validateTargetThresholds(targetThresholds api.ResourceThresholds) bool {
-	if targetThresholds == nil {
-		klog.V(1).Infof("no target resource threshold is configured")
-		return false
-	} else if _, ok := targetThresholds[v1.ResourcePods]; !ok {
-		klog.V(1).Infof("no target resource threshold for pods is configured")
-		return false
+// validateThresholds checks if thresholds have valid resource name and resource percentage configured
+func validateThresholds(thresholds api.ResourceThresholds) error {
+	if thresholds == nil || len(thresholds) == 0 {
+		return fmt.Errorf("no resource threshold is configured")
 	}
-	return true
+	for name, percent := range thresholds {
+		switch name {
+		case v1.ResourceCPU, v1.ResourceMemory, v1.ResourcePods:
+			if percent < MinResourcePercentage || percent > MaxResourcePercentage {
+				return fmt.Errorf("%v threshold not in [%v, %v] range", name, MinResourcePercentage, MaxResourcePercentage)
+			}
+		default:
+			return fmt.Errorf("only cpu, memory, or pods thresholds can be specified")
+		}
+	}
+	return nil
 }
 
 // classifyNodes classifies the nodes into low-utilization or high-utilization nodes. If a node lies between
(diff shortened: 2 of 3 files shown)


pkg/descheduler/strategies/lownodeutilization_test.go
```diff
@@ -141,6 +144,64 @@ func TestLowNodeUtilization(t *testing.T) {
 			},
 			expectedPodsEvicted: 3,
 		},
+		{
+			name: "without priorities stop if any required capacity (cpu, memory, pods) is moved",
```
ingvagabund, line 148: without priorities stop when cpu capacity is depleted

pkg/descheduler/strategies/lownodeutilization_test.go
```diff
@@ -141,6 +144,64 @@ func TestLowNodeUtilization(t *testing.T) {
...
+					},
+				},
+				n2NodeName: {
+					Items: []v1.Pod{
+						*test.BuildTestPod("p9", 400, 2100, n1NodeName, setRSOwnerRef),
+					},
+				},
+				n3NodeName: {},
+			},
+			expectedPodsEvicted: 3,
```
ingvagabund, line 203: Please add a comment: `4 pods available for eviction based on v1.ResourcePods, only 3 pods can be evicted before cpu is depleted`

ingvagabund:
@lixiang233 can you also squash both commits into one?

ingvagabund:
@lixiang233 you will also need to call `LowNodeUtilization` directly in the `TestLowNodeUtilization` which will require to refactor `TestLowNodeUtilization` a bit.

lixiang233:
done. @ingvagabund

README.md
```diff
@@ -91,7 +91,8 @@ Currently, pods request resource requirements are considered for computing node
 There is another configurable threshold, `targetThresholds`, that is used to compute those potential nodes
 from where pods could be evicted. Any node, between the thresholds, `thresholds` and `targetThresholds` is
 considered appropriately utilized and is not considered for eviction. The threshold, `targetThresholds`,
-can be configured for cpu, memory, and number of pods too in terms of percentage.
+can be configured for cpu, memory, and number of pods too in terms of percentage. Notice that `pods` must be
+configured and if `cpu` or `memory` not configured, they'll be set to default value `100`.
```
ingvagabund, line 95: s/not configured/are not configured/

pkg/descheduler/strategies/lownodeutilization.go
```diff
@@ -59,6 +59,13 @@ func LowNodeUtilization(ctx context.Context, client clientset.Interface, strateg
 	if !validateTargetThresholds(targetThresholds) {
 		return
 	}
+	// check if CPU/Mem not set in targetThresholds and set it to 100
```
ingvagabund, line 62: check if CPU/Mem not set in targetThresholds and set it to 100
check if CPU/Mem are set in targetThresholds, if not, set them to 100

pkg/descheduler/strategies/lownodeutilization.go
```diff
@@ -187,16 +194,13 @@ func evictPodsFromTargetNodes(
 		totalPods += ((float64(podsPercentage) * float64(nodeCapacity.Pods().Value())) / 100)
 
 		// totalCPU capacity to be moved
-		if _, ok := targetThresholds[v1.ResourceCPU]; ok {
-			cpuPercentage := targetThresholds[v1.ResourceCPU] - node.usage[v1.ResourceCPU]
-			totalCPU += ((float64(cpuPercentage) * float64(nodeCapacity.Cpu().MilliValue())) / 100)
-		}
+		// CPU and Mem in targetThresholds have already validated in LowNodeUtilization
```
ingvagabund, line 197: No need for the comment as it's assumed both targetThresholds are already set.

lixiang233:
> Just nits, otherwise lgtm

lixiang233:
Changed, thanks for your help @ingvagabund

README.md
```diff
@@ -91,7 +91,8 @@ Currently, pods request resource requirements are considered for computing node
 There is another configurable threshold, `targetThresholds`, that is used to compute those potential nodes
 from where pods could be evicted. Any node, between the thresholds, `thresholds` and `targetThresholds` is
 considered appropriately utilized and is not considered for eviction. The threshold, `targetThresholds`,
-can be configured for cpu, memory, and number of pods too in terms of percentage.
+can be configured for cpu, memory, and number of pods too in terms of percentage. Notice that `pods` must be
+configured and if `cpu` or `memory` are not configured, they'll be set to default value `100`.
```
damemi, line 95: why is `pods` required now, and we should add some validation to ensure that it's set

lixiang233: I only added default value to `cpu` and `mem` and didn't change the validate func .Current `validateTargetThresholds` func only checks wheather `pods` is configured, if not, `LowNodeUtilization` will stop, so only `pods` is required. Should we treat`pods` as same as `cpu` and `mem`, and `validateTargetThresholds` func returns true if any of these resources configured?

lixiang233: Maybe we can use `validateThresholds` to check `targetThresholds`

lixiang233:
What about the latest commit? I unified the verification method for `targetThresholds` and `thresholds`, and set the allowed range of resource percentage to [0, 100]. @damemi @ingvagabund

pkg/descheduler/strategies/lownodeutilization.go
```diff
@@ -117,11 +125,20 @@ func validateThresholds(thresholds api.ResourceThresholds) bool {
 	for name := range thresholds {
 		switch name {
 		case v1.ResourceCPU:
-			continue
+			if !validateResourcePercentage(thresholds[v1.ResourceCPU]) {
```
ingvagabund, line 128: Can you refactor the function to return `fmt.Errorf("cpu threshold not valid")` instead of bool? Logging Info/Error and return bool is not practical. Instead, called of `validateThresholds` can decide what to do about non-nil error.

pkg/descheduler/strategies/lownodeutilization.go
```diff
@@ -56,15 +61,18 @@ func LowNodeUtilization(ctx context.Context, client clientset.Interface, strateg
 		return
 	}
 	targetThresholds := strategy.Params.NodeResourceUtilizationThresholds.TargetThresholds
-	if !validateTargetThresholds(targetThresholds) {
+	if !validateThresholds(targetThresholds) {
```
ingvagabund, line 64: Better but not sufficient. `thresholds` has to be less than `targetThresholds` per a resource type. @lixiang233  Can you add check for that as well?

Also, `thresholds` and `targetThresholds` have gave me a lot of headache before learning what they actually mean. I would prefer to rename both of them into `lowThreshold` and `highThreshold` at least on the code level. Also given the strategy is configured as:
```
  "LowNodeUtilization":
     enabled: true
     params:
         thresholds:
           "cpu" : 20
           "memory": 20
           "pods": 20
         targetThresholds:
           "cpu" : 50
           "memory": 50
           "pods": 50
```
we can drop `thresholds` and `targetThresholds` and have:
```
  "LowNodeUtilization":
     enabled: true
     params:
         low:
           "cpu" : 20
           "memory": 20
           "pods": 20
         high:
           "cpu" : 50
           "memory": 50
           "pods": 50
```

@damemi @seanmalloy anything against?

pkg/descheduler/strategies/lownodeutilization.go
```diff
@@ -130,16 +147,12 @@ func validateThresholds(thresholds api.ResourceThresholds) bool {
...
-//This function could be merged into above once we are clear.
-func validateTargetThresholds(targetThresholds api.ResourceThresholds) bool {
-	if targetThresholds == nil {
-		klog.V(1).Infof("no target resource threshold is configured")
-		return false
-	} else if _, ok := targetThresholds[v1.ResourcePods]; !ok {
-		klog.V(1).Infof("no target resource threshold for pods is configured")
-		return false
+// validateResourcePercentage checks if resource percentage is in the range [0, 100]
+func validateResourcePercentage(percent api.Percentage) bool {
```
ingvagabund, line 151: I don't think `validateResourcePercentage` is the right function to unit test given its length and complexity. Can you move the distinct cases into TestValidateThresholds and drop `TestValidateResourcePercentage` completely?

pkg/descheduler/strategies/lownodeutilization.go
lixiang233: Should we allow `thresholds` and `targetThresholds` configuring different resources? For instance, `thresholds` configured `cpu` and `mem` while `targetThresholds` only configured `pods`.

pkg/descheduler/strategies/lownodeutilization.go
lixiang233: done

pkg/descheduler/strategies/lownodeutilization.go
lixiang233: done

pkg/descheduler/strategies/lownodeutilization.go
ingvagabund: Good point. Assuming low threshold has a cpu set but high does not, you mark a node as underutilized wrt. cpu. You then find over-utilized node(s) wrt. memory. In the process of moving workload from over-utilized (since you want to decrease memory consumption), you move some pods to the underutilized nodes with a hope there's at least some memory left on those nodes. So, in practice this scenario is possible though you are removing any guarantee the list of underutilized nodes is selected effectively. I.e. the nodes have sufficient memory left. Since there's no low threshold defaulting for memory, you might consume all the memory left and end up with nodes which are memory over utilized. Also, selecting under-utilized nodes based on cpu becomes irrelevant since you then may pick any subset of nodes since you still have no guarantee there's sufficient memory left.

So, the answer is no.

pkg/descheduler/strategies/lownodeutilization_test.go
```diff
@@ -453,7 +543,7 @@ func TestValidateThresholds(t *testing.T) {
 	}
 
 	for _, test := range tests {
-		isValid := validateThresholds(test.input)
+		isValid := (validateThresholds(test.input) == nil)
 
 		if isValid != test.succeed {
```
ingvagabund, line 548: Please replace `succeed bool` with `error err` and set it to nil or a specific error instead given `validateThresholds` now returns `err`.

lixiang233: Got it. I'll commit later.

README.md
damemi: It looks like you pushed a change to set a default for `pods`, is that correct? If so please remember to update this readme note

lixiang233: Yes, I'll update it in later commit

lixiang233:
One more question, after I added new validation check, `TestWithTaints` failed because strategy's `TargetThresholds` is changed in `LowNodeUtilization` while `thresholds` not. Should I set `thresholds` to 100 while setting `TargetThresholds` to make sure strategy is always valid or just change the testcase? I prefer the first one. @ingvagabund

lixiang233:
Updated README and validation checks for the two thresholds @damemi @ingvagabund

pkg/descheduler/strategies/lownodeutilization.go
```diff
@@ -102,37 +119,62 @@ func LowNodeUtilization(ctx context.Context, client clientset.Interface, strateg
...
+
+// validateThresholds checks if thresholds have valid resource name and resource percentage configured
+func validateThresholds(thresholds api.ResourceThresholds) error {
 	if thresholds == nil || len(thresholds) == 0 {
-		klog.V(1).Infof("no resource threshold is configured")
-		return false
+		return fmt.Errorf("no resource threshold is configured")
 	}
 	for name := range thresholds {
 		switch name {
```
ingvagabund, line 152: Given all valid resources execute the same code, you can consolidate them:
```
switch name {
case v1.ResourceCPU:
case v1.ResourceMemory:
case v1.ResourcePods:
	if !validateResourcePercentage(thresholds[name]) {
		return fmt.Errorf("%v threshold not valid", name)
	}
default:
	return fmt.Errorf("only cpu, memory, or pods thresholds can be specified")
}
```
Also, given `validateResourcePercentage` gets to used only once, you can inline its implementation:

```
switch name {
case v1.ResourceCPU:
case v1.ResourceMemory:
case v1.ResourcePods:
	percent := thresholds[name]
	if percent < MinResourcePercentage || percent > MaxResourcePercentage {
		return fmt.Errorf("%v threshold not in [%v, %v] range", name, MinResourcePercentage, MaxResourcePercentage)
	}
default:
	return fmt.Errorf("only cpu, memory, or pods thresholds can be specified")
}
```

ingvagabund:
@lixiang233 that's the last nit I found. Sorry for the review taking so long. Thanks for asking the right questions!!!

pkg/descheduler/strategies/lownodeutilization.go
lixiang233: Wow,that's much better!

lixiang233:
Simplified `validateThresholds` and squashed all commits into one.

README.md
```diff
@@ -89,7 +89,8 @@ usage is below threshold for all (cpu, memory, and number of pods), the node is
 Currently, pods request resource requirements are considered for computing node resource utilization.
 
 There is another configurable threshold, `targetThresholds`, that is used to compute those potential nodes
-from where pods could be evicted. Any node, between the thresholds, `thresholds` and `targetThresholds` is
+from where pods could be evicted. If a node's usage is above targetThreshold for any (cpu, memory, or number of pods),
+the node is considered over utilized. Any node, between the thresholds, `thresholds` and `targetThresholds` is
```
ingvagabund, line 93: Any node between the thresholds `thresholds` and `targetThresholds` is ... (no need for the comas)

README.md
```diff
@@ -114,6 +115,15 @@ strategies:
            "pods": 50
 ```
 
+Policy should pass the following validation checks:
+* Only three types of resource are supported: `cpu`, `memory` and `pods`.
```
ingvagabund, line 119: s/resource/resources

README.md
```diff
@@ -114,6 +115,15 @@ strategies:
...
+* `thresholds` or `targetThresholds` can not be nil and they must configure exactly the same types of resources.
+* The valid range of the resource's percentage value is \[0, 100\]
+* Percentage value of `thresholds` can not be greater than `targetThresholds` for the same resource.
+
+You can configure any number of these three resources, if one resource is not configured, it will be set to default
```
ingvagabund, line 124: `If any of the resource types is not specified, all its thresholds default to 100% to avoid nodes going from underutilized to overutilized`.

lixiang233: All these modified

pkg/descheduler/strategies/lownodeutilization.go
```diff
@@ -52,13 +58,24 @@ func LowNodeUtilization(ctx context.Context, client clientset.Interface, strateg
...
+		targetThresholds[v1.ResourcePods] = MaxResourcePercentage
+	}
+	if _, ok := thresholds[v1.ResourceCPU]; !ok {
+		thresholds[v1.ResourceCPU] = MaxResourcePercentage
+		targetThresholds[v1.ResourceCPU] = MaxResourcePercentage
+	}
+	if _, ok := thresholds[v1.ResourceMemory]; !ok {
+		thresholds[v1.ResourceMemory] = MaxResourcePercentage
+		targetThresholds[v1.ResourceMemory] = MaxResourcePercentage
+	}
```
damemi, line 78: nit: I feel like this section could probably be moved to `validateStrategyConfig` but I won't block on it

Issue #271, opened 2020-04-28
Desheduler needs to check the number of replicas before evicting the pod.


pbannykh:
Vendor applications often run in a single replica. Descheduler should not evict pods that work in a single replica.

seanmalloy:
/kind feature

@pbannykh thanks for opening this issue!

Based on my understanding of the descheduler I believe you are correct the descheduler does not take the number of replicas into consideration. For your reference the [IsEvictable](https://github.com/kubernetes-sigs/descheduler/blob/78eef6c343cc77a5c11c9ada3bf73be89861624a/pkg/descheduler/pod/pods.go#L31-L38) function shows the various conditions that determines if a pod can be evicted or not.

Here are three possible solutions ...
1. Create a [PDB](https://kubernetes.io/docs/tasks/run-application/configure-pdb/) to prevent the pod from being evicted
2. Mark the pod as [critical pod](https://kubernetes.io/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods/)
3. Add a new feature to descheduler to take replicas into consideration

Would options 1 or 2 work for you? Or do you feel that you really need a configuration option for descheduler to take replicas into consideration?

pbannykh:
@seanmalloy thanks for the proposed solutions!

We cannot allow users to mark their pods as critical, because users will mark almost all of their pods as critical.

Creating a PDB for a single replica makes node drain difficult.

Add a new feature to descheduler to take replicas into consideration is not a bad decision for me and I am ready to discuss this. Enabling this option may leave some nodes unbalanced. This will create unnecessary work for descheduler the next time it starts.

seanmalloy:
@pbannykh would you want the number of replicas to be configurable? Or is a simple hard coded exception for ignoring single replica pods good enough?

One more idea ...

Would adding a label selector configuration option to descheduler meet your requirement? See pull request #202. I believe if the label section feature was implemented it would allow you to tell descheduler to ignore pods with certain labels.

Right now I'm undecided if this feature should be added or not. I'd like some feedback from @damemi @aveshagarwal @ravisantoshgudimetla or @ingvagabund to get some additional perspectives.

damemi:
There are valid cases for descheduling a single replica, for example if affinity conditions change or even rebalancing resource usage, so I don't think this would work as a default behavior. It could be configurable though if you want to enable it.

seanmalloy:
[...]

PR #293, opened 2020-05-26
Add PDB check in `IsEvictable`

oif:
Adding PDB check for pods to avoid trigger 'NotAllowed' eviction

Ths PR created reference with #271

### DIFF
--- a/pkg/descheduler/pod/pods.go
+++ b/pkg/descheduler/pod/pods.go
@@ -18,11 +18,14 @@ package pod
 
 import (
 	"context"
+
+	"sigs.k8s.io/descheduler/pkg/utils"
+
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/labels"
 	clientset "k8s.io/client-go/kubernetes"
-	"sigs.k8s.io/descheduler/pkg/utils"
 )
 
 const (
@@ -30,12 +33,12 @@ const (
 )
 
 // IsEvictable checks if a pod is evictable or not.
-func IsEvictable(pod *v1.Pod, evictLocalStoragePods bool) bool {
+func IsEvictable(ctx context.Context, client clientset.Interface, pod *v1.Pod, evictLocalStoragePods bool) (bool, error) {
 	ownerRefList := OwnerRef(pod)
 	if !HaveEvictAnnotation(pod) && (IsMirrorPod(pod) || (!evictLocalStoragePods && IsPodWithLocalStorage(pod)) || len(ownerRefList) == 0 || IsDaemonsetPod(ownerRefList) || IsCriticalPod(pod)) {
-		return false
+		return false, nil
 	}
-	return true
+	return IsPDBAllowsEviction(ctx, client, pod)
 }
 
 // ListEvictablePodsOnNode returns the list of evictable pods on node.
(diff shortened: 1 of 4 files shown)


damemi:
/assign @ingvagabund 
Jan could you please take a look at this?

pkg/descheduler/pod/pods.go
```diff
@@ -124,3 +131,21 @@ func IsPodWithLocalStorage(pod *v1.Pod) bool {
 func OwnerRef(pod *v1.Pod) []metav1.OwnerReference {
 	return pod.ObjectMeta.GetOwnerReferences()
 }
+
+// IsPDBAllowsEviction returns true when pod eviction is allowed by PDB
+func IsPDBAllowsEviction(ctx context.Context, client clientset.Interface, pod *v1.Pod) (bool, error) {
+	PDBList, err := client.PolicyV1beta1().PodDisruptionBudgets(pod.Namespace).List(ctx, metav1.ListOptions{})
```
lixiang233, line 137: Better to use `PodDisruptionBudgetInformer` instead of clientSet to get better performance. `GetPodPodDisruptionBudgets` method is perfect for this situation

pkg/descheduler/pod/pods_test.go
```diff
@@ -17,19 +17,26 @@ limitations under the License.
...
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/client-go/kubernetes/fake"
 )
 
 func TestIsEvictable(t *testing.T) {
 	n1 := test.BuildTestNode("node1", 1000, 2000, 13, nil)
 	type testCase struct {
 		pod                   *v1.Pod
 		runBefore             func(*v1.Pod)
+		PDB                   *policyv1beta1.PodDisruptionBudget
```
lixiang233, line 39: Maybe better to set a list of PDBs here rather than a single one？. And can u add some testcases for it?

pkg/descheduler/strategies/lownodeutilization.go
```diff
@@ -209,7 +211,10 @@ func evictPodsFromTargetNodes(
 		}
 		klog.V(3).Infof("evicting pods from node %#v with usage: %#v", node.node.Name, node.usage)
 
-		nonRemovablePods, bestEffortPods, burstablePods, guaranteedPods := classifyPods(node.allPods, evictLocalStoragePods)
+		nonRemovablePods, bestEffortPods, burstablePods, guaranteedPods, err := classifyPods(ctx, client, node.allPods, evictLocalStoragePods)
+		if err != nil {
+			klog.Warningf("Unable to classify pods: %s", err)
+		}
```
lixiang233, line 217: Log error and return? For all these pod lists are empty if error occurs, it makes no sense to do eviction.

aveshagarwal:
could you/anyone please explain why is this PR needed? May be I am missing something.

Descheduler uses eviction subresource for deleting/evicting pods. Eviction sub resource honors PDB. They why is this change required?

oif:
@aveshagarwal Although eviction API comply with PDB, but strategies still get not evictable pods from `podutil.ListEvictablePodsOnNode`, so to avoid giving a PDB.Allowed < 1 (zero) pod to strategies to calculate add this filter inside.

pkg/descheduler/pod/pods.go
oif: Yep, I'm preffer using informer instead, but considering losts get/list actions use clientset directly so I think maybe descheduler DON'T intend to use too much informers.

Never mind, modify this later

aveshagarwal:
But descheduler mostly ignore errors when deleting pods and just moves on to other pods. So getting not allowed warning clearly shows that the pod is not evictable. I am still not sure how adding this change is going to be helpful?

ingvagabund:
@oif I agree with @aveshagarwal , if not PDB there may be other reasons not allowing a pod to be evicted in the future. That's what the eviction subresource is for. I.e. to take responsibility for deciding when a pod can be evicted. So it's ok for a strategy to list more pods than it can actually evict as long as the list itself is not used to calculate a condition which might negatively affect a cluster if computed incorrectly.

oif:
@aveshagarwal @ingvagabund 

Yes, that view described above is reasonable and acceptable. So maybe close this  PR

PR #296, opened 2020-05-27
Make DeschedulerStrategy.Params a pointer

ingvagabund:
To avoid empty params fields when serializing:
```
  RemoveDuplicates:
    enabled: true
    params: {}
  RemovePodsHavingTooManyRestarts:
    enabled: true
    params:
      podsHavingTooManyRestarts: {}
  RemovePodsViolatingInterPodAntiAffinity:
    enabled: true
    params: {}
  RemovePodsViolatingNodeAffinity:
    enabled: true
    params:
      nodeAffinityType:
      - requiredDuringSchedulingIgnoredDuringExecution
  RemovePodsViolatingNodeTaints:
    enabled: true
    params: {}
```

Issue #248, opened 2020-03-03
Official Helm Chart


stevehipwell:
Now that there is an official Docker image (#144) it would be really useful if there was an official Helm chart.

seanmalloy:
Here are the required details that are needed to make this issue qualify as a good first issue.

Please use helm v3, not v2.

The helm chart needs to create a `CronJob`, `ConfigMap`, `ClusterRole`, `ServiceAccount`, and `ClusterRoleBinding` in the `kube-system` namespace. The `.spec.schedule` field needs to be configurable. The `ConfigMap` needs to support specifying a custom descheduler policy. See the files in the [kubernetes directory](https://github.com/kubernetes-sigs/descheduler/blob/master/kubernetes/rbac.yaml) for a good starting point.

The descheduler container image can be pulled from the below registry locations.

`asia.gcr.io/k8s-artifacts-prod/descheduler/descheduler:v0.10.0`
`eu.gcr.io/k8s-artifacts-prod/descheduler/descheduler:v0.10.0`
`us.gcr.io/k8s-artifacts-prod/descheduler/descheduler:v0.10.0`

stevehipwell:
@seanmalloy I'm interested in picking this up and am slightly curious around the Helm v3 requirement? I understand that in a perfect world we'd all have moved to Helm v3 and all of the chart authors would have upgraded their charts to v3. However based on the current available charts it seems that most people are still using v2 charts even if they are applying them via v3. My suggestion to add this chart was to get closer alignment and it seems like only supporting Helm v3 might not be the best way to do this?

Would it be worth creating two charts (v2 & v3) or due to the limited scope of this chart would it be better sticking with Helm v2 for now?

Also do you have a repository for this new chart as the [stable repo is no longer taking new chart submissions](https://github.com/helm/charts#deprecation-timeline)?

Odernor:
Hi,
i have written an small descheduler helm chart for me. Im newbie in this..
Perhaps it can help an little: https://github.com/Odernor/descheduler_helm
its helm v2 but is installable with v3 too

stevehipwell:
The following chart has been around for a while as the author also created a Docker image.
https://github.com/komljen/helm-charts/tree/master/descheduler

seanmalloy:
> I'm interested in picking this up and am slightly curious around the Helm v3 requirement?

I'm fine with Helm v2 if you think that makes the most sense.

[...]

PR #298, opened 2020-05-27
Add Helm chart

stevehipwell:
This PR adds a native _Helm_ chart to this repository at `charts/descheduler` and uses [GitHub Actions](https://github.com/features/actions), specifically the [Helm actions](https://github.com/helm?q=action&type=&language=), to validate and release the chart (see the following rules).

- PRs with changes to `charts/**` will trigger the _lint-test_ action
- Tags with the pattern `chart-*` will trigger the _release_ action

**IMPORTANT** - For this PR to work the `helm/chart-releaser-action` [pre-requisites](https://github.com/helm/chart-releaser-action#pre-requisites) need to be completed.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,30 @@
+name: Release Charts
+
+on:
+  push:
+    tags:
+      - chart-*
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Configure Git
+        run: |
+          git config user.name "$GITHUB_ACTOR"
+          git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
+
+      - name: Fetch history
+        run: git fetch --prune --unshallow
+
+      - name: Add dependency chart repos
+        run: |
+          helm repo add stable https://kubernetes-charts.storage.googleapis.com/
+
+      - name: Run chart-releaser
+        uses: helm/chart-releaser-action@v1.0.0-rc.2
+        env:
+          CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
--- a/Makefile
+++ b/Makefile
@@ -41,6 +41,8 @@ IMAGE_GCLOUD:=$(REGISTRY)/descheduler:$(VERSION)
 # In the future binaries can be uploaded to
 # GCS bucket gs://k8s-staging-descheduler.
 
+HAS_HELM := $(shell which helm)
+
 all: build
 
 build:
@@ -62,7 +64,7 @@ push: push-container-to-gcloud
 clean:
 	rm -rf _output
 
-verify: verify-gofmt verify-vendor lint
+verify: verify-gofmt verify-vendor lint lint-chart
 
 verify-gofmt:
 	./hack/verify-gofmt.sh
@@ -86,3 +88,9 @@ ifndef HAS_GOLANGCI
 	curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b ./_output/bin ${GOLANGCI_VERSION}
 endif
 	./_output/bin/golangci-lint run
+
+lint-chart:
+ifndef HAS_HELM
+	curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3 && chmod 700 ./get_helm.sh && ./get_helm.sh
+endif
+	helm lint ./charts/descheduler
--- a/charts/descheduler/Chart.yaml
+++ b/charts/descheduler/Chart.yaml
@@ -1,0 +1,16 @@
+apiVersion: v1
+name: descheduler
+version: 1.0.0
+appVersion: 0.18.0
+description: Descheduler for Kubernetes is used to rebalance clusters by evicting pods that can potentially be scheduled on better nodes. In the current implementation, descheduler does not schedule replacement of evicted pods but relies on the default scheduler for that.
+keywords:
+- kubernetes
+- descheduler
+- kube-scheduler
+home: https://github.com/kubernetes-sigs/descheduler
+icon: https://kubernetes.io/images/favicon.png
+sources:
+- https://github.com/kubernetes-sigs/descheduler
+maintainers:
+- name: stevehipwell
+  email: steve.hipwell@github.com
--- a/charts/descheduler/README.md
+++ b/charts/descheduler/README.md
@@ -1,0 +1,59 @@
+# Descheduler for Kubernetes
+
+[Descheduler](https://github.com/kubernetes-sigs/descheduler/) for Kubernetes is used to rebalance clusters by evicting pods that can potentially be scheduled on better nodes. In the current implementation, descheduler does not schedule replacement of evicted pods but relies on the default scheduler for that.
+
+## TL;DR:
+
+```shell
+helm repo add descheduler https://kubernetes-sigs.github.io/descheduler/
+$ helm install descheduler/descheduler --name my-release
+```
+
+## Introduction
+
+This chart bootstraps a [desheduler](https://github.com/kubernetes-sigs/descheduler/) cron job on a [Kubernetes](http://kubernetes.io) cluster using the [Helm](https://helm.sh) package manager.
+
+## Prerequisites
+
+- Kubernetes 1.14+
+
+## Installing the Chart
+
+To install the chart with the release name `my-release`:
+
+```shell
+helm install --name my-release descheduler/descheduler
+```
+
+The command deploys _descheduler_ on the Kubernetes cluster in the default configuration. The [configuration](#configuration) section lists the parameters that can be configured during installation.
+
+> **Tip**: List all releases using `helm list`
+
+## Uninstalling the Chart
+
+To uninstall/delete the `my-release` deployment:
+
+```shell
+helm delete my-release
+```
+
+The command removes all the Kubernetes components associated with the chart and deletes the release.
+
+## Configuration
+
+The following table lists the configurable parameters of the _descheduler_ chart and their default values.
+
+| Parameter                      | Description                                                                                                           | Default                                                |
+| ------------------------------ | --------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------ |
+| `image.repository`             | Docker repository to use                                                                                              | `us.gcr.io/k8s-artifacts-prod/descheduler/descheduler` |
+| `image.tag`                    | Docker tag to use                                                                                                     | `v[chart appVersion]`                                              |
+| `image.pullPolicy`             | Docker image pull policy                                                                                              | `IfNotPresent`                                         |
+| `nameOverride`                 | String to partially override `descheduler.fullname` template (will prepend the release name)                          | `""`                                                   |
+| `fullnameOverride`             | String to fully override `descheduler.fullname` template                                                              | `""`                                                   |
+| `schedule`                     | The cron schedule to run the _descheduler_ job on                                                                     | `"*/2 * * * *"`                                        |
+| `cmdOptions`                   | The options to pass to the _descheduler_ command                                                                      | _see values.yaml_                                      |
+| `deschedulerPolicy.strategies` | The _descheduler_ strategies to apply                                                                                 | _see values.yaml_                                      |
+| `priorityClassName`            | The name of the priority class to add to pods                                                                         | `system-cluster-critical`                              |
+| `rbac.create`                  | If `true`, create & use RBAC resources                                                                                | `true`                                                 |
+| `serviceAccount.create`        | If `true`, create a service account for the cron job                                                                  | `true`                                                 |
+| `serviceAccount.name`          | The name of the service account to use, if not set and create is true a name is generated using the fullname template | `nil`                                                  |
--- a/charts/descheduler/templates/clusterrolebinding.yaml
+++ b/charts/descheduler/templates/clusterrolebinding.yaml
@@ -1,0 +1,16 @@
+{{- if .Values.rbac.create -}}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: {{ template "descheduler.fullname" . }}
+  labels:
+    {{- include "descheduler.labels" . | nindent 4 }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: {{ template "descheduler.fullname" . }}
+subjects:
+  - kind: ServiceAccount
+    name: {{ template "descheduler.serviceAccountName" . }}
+    namespace: {{ .Release.Namespace }}
+{{- end -}}
--- a/charts/descheduler/templates/cronjob.yaml
+++ b/charts/descheduler/templates/cronjob.yaml
@@ -1,0 +1,53 @@
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: {{ template "descheduler.fullname" . }}
+  labels:
+    {{- include "descheduler.labels" . | nindent 4 }}
+spec:
+  schedule: {{ .Values.schedule | quote }}
+  concurrencyPolicy: "Forbid"
+  jobTemplate:
+    spec:
+      template:
+        metadata:
+          name: {{ template "descheduler.fullname" . }}
+          annotations:
+            checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
+            {{- if .Values.podAnnotations }}
+            {{- .Values.podAnnotations | toYaml | nindent 12 }}
+            {{- end }}
+          labels:
+            app.kubernetes.io/name: {{ include "descheduler.name" . }}
+            app.kubernetes.io/instance: {{ .Release.Name }}
+            {{- if .Values.podLabels }}
+            {{- .Values.podLabels | toYaml | nindent 12 }}
+            {{- end }}
+        spec:
+          {{- if .Values.priorityClassName }}
+          priorityClassName: {{ .Values.priorityClassName }}
+          {{- end }}
+          serviceAccountName: {{ template "descheduler.serviceAccountName" . }}
+          restartPolicy: "Never"
+          containers:
+            - name: {{ .Chart.Name }}
+              image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default (printf "v%s" .Chart.AppVersion) }}"
+              imagePullPolicy: {{ .Values.image.pullPolicy }}
+              command:
+                - "/bin/descheduler"
+              args:
+                - "--policy-config-file"
+                - "/policy-dir/policy.yaml"
+                {{- range $key, $value := .Values.cmdOptions }}
+                - {{ printf "--%s" $key | quote }}
+                {{- if $value }}
+                - {{ $value | quote }}
+                {{- end }}
+                {{- end }}
+              volumeMounts:
+                - mountPath: /policy-dir
+                  name: policy-volume
+          volumes:
+          - name: policy-volume
+            configMap:
+              name: {{ template "descheduler.fullname" . }}
--- a/charts/descheduler/values.yaml
+++ b/charts/descheduler/values.yaml
@@ -1,0 +1,59 @@
+# Default values for descheduler.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+image:
+  repository: us.gcr.io/k8s-artifacts-prod/descheduler/descheduler
+  # Overrides the image tag whose default is the chart version
+  tag: ""
+  pullPolicy: IfNotPresent
+
+nameOverride: ""
+fullnameOverride: ""
+
+schedule: "*/2 * * * *"
+
+cmdOptions:
+  v: 3
+  # evict-local-storage-pods:
+  # max-pods-to-evict-per-node: 10
+  # node-selector: "key1=value1,key2=value2"
+
+deschedulerPolicy:
+  strategies:
+    RemoveDuplicates:
+      enabled: true
+    RemovePodsViolatingNodeTaints:
+      enabled: true
+    RemovePodsViolatingNodeAffinity:
+      enabled: true
+      params:
+         nodeAffinityType:
+         - requiredDuringSchedulingIgnoredDuringExecution
+    RemovePodsViolatingInterPodAntiAffinity:
+      enabled: true
+    LowNodeUtilization:
+      enabled: true
+      params:
+        nodeResourceUtilizationThresholds:
+          thresholds:
+            cpu: 20
+            memory: 20
+            pods: 20
+          targetThresholds:
+            cpu: 50
+            memory: 50
+            pods: 50
+
+priorityClassName: system-cluster-critical
+
+rbac:
+  # Specifies whether RBAC resources should be created
+  create: true
+
+serviceAccount:
+  # Specifies whether a ServiceAccount should be created
+  create: true
+  # The name of the ServiceAccount to use.
+  # If not set and create is true, a name is generated using the fullname template
+  name:
--- a/charts/descheduler/.helmignore
+++ b/charts/descheduler/.helmignore
@@ -1,0 +1,22 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
+.bzrignore
+.hg/
+.hgignore
+.svn/
+# Common backup files
+*.swp
+*.bak
+*.tmp
+*~
+# Various IDEs
+.project
+.idea/
+*.tmproj
+.vscode/
--- a/charts/descheduler/templates/NOTES.txt
+++ b/charts/descheduler/templates/NOTES.txt
@@ -1,0 +1,1 @@
+Descheduler installed as a cron job.
--- a/charts/descheduler/templates/_helpers.tpl
+++ b/charts/descheduler/templates/_helpers.tpl
@@ -1,0 +1,56 @@
+{{/* vim: set filetype=mustache: */}}
+{{/*
+Expand the name of the chart.
+*/}}
+{{- define "descheduler.name" -}}
+{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
+
+{{/*
+Create a default fully qualified app name.
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+If release name contains chart name it will be used as a full name.
+*/}}
+{{- define "descheduler.fullname" -}}
+{{- if .Values.fullnameOverride -}}
+{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" -}}
+{{- else -}}
+{{- $name := default .Chart.Name .Values.nameOverride -}}
+{{- if contains $name .Release.Name -}}
+{{- .Release.Name | trunc 63 | trimSuffix "-" -}}
+{{- else -}}
+{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Create chart name and version as used by the chart label.
+*/}}
+{{- define "descheduler.chart" -}}
+{{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
+
+{{/*
+Common labels
+*/}}
+{{- define "descheduler.labels" -}}
+app.kubernetes.io/name: {{ include "descheduler.name" . }}
+helm.sh/chart: {{ include "descheduler.chart" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
+{{- if .Chart.AppVersion }}
+app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+{{- end }}
+app.kubernetes.io/managed-by: {{ .Release.Service }}
+{{- end -}}
+
+{{/*
+Create the name of the service account to use
+*/}}
+{{- define "descheduler.serviceAccountName" -}}
+{{- if .Values.serviceAccount.create -}}
+    {{ default (include "descheduler.fullname" .) .Values.serviceAccount.name }}
+{{- else -}}
+    {{ default "default" .Values.serviceAccount.name }}
+{{- end -}}
+{{- end -}}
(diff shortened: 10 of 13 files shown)


stevehipwell:
CLA signed.

stevehipwell:
I signed it

seanmalloy:
/hold

As noted by @stevehipwell this is currently a work in progress.

seanmalloy:
@stevehipwell thanks for submitting this pull request!

I'll review this PR in the next 24 hours. I'll also answer the questions from your https://github.com/kubernetes-sigs/descheduler/issues/248#issuecomment-634624573 here as part of my review

stevehipwell:
Thanks @seanmalloy I think I've figured out the answers myself and have come up with a flow that should work well.

I've updated the GitHub actions config (which I can't test in this repo) to the following logic:

- PRs with changes to `charts/**` will trigger the _lint-test_ action
- Tags with the pattern `chart-*` will trigger the _release_ action

For clarification if you make a chart change you need to make sure that the chart version has been bumped (SemVer 2).

### Release Process

**NOTE** - The first two steps are what I assume your current process is for framing, although the specifics don't really matter here.

- Merge PRs with new **descheduler** functionality
- Release new **descheduler** version with `v*` tag
- Create PR for chart to use new **descheduler** version Docker image
- Merge chart PR
- Release chart with `chart-*` tag (where _*_ is the chart version)

stevehipwell:
It looks like #296 has broken builds on [master](https://github.com/kubernetes-sigs/descheduler/commits/master)?

.github/workflows/lint-test.yaml
```diff
@@ -0,0 +1,33 @@
+name: Lint and Test Charts
```
seanmalloy, line 1: The descheduler project recently setup all CI tests using [k8s prow](https://github.com/kubernetes/test-infra/tree/master/prow). The Travis CI configuration will probably be removed soon. Right now I don't think we should start using GitHub actions to run CI tests for the helm chart.

My recommendation is to remove this file from this PR. We can figure out how to add helm chart linting through prow in a follow up PR.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,30 @@
+name: Release Charts
+
+on:
+  push:
+    tags:
+      - chart-*
```
seanmalloy, line 6: I'm assuming this controls which git tags to build. I don't think we will want to have separate tags that are only used for chart releases.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+      - name: Fetch history
+        run: git fetch --prune --unshallow
+
+      - name: Add dependency chart repos
+        run: |
+          helm repo add stable https://kubernetes-charts.storage.googleapis.com/
+
+      - name: Run chart-releaser
+        uses: helm/chart-releaser-action@v1.0.0-rc.2
```
seanmalloy, line 28: Where does this push the chart? A GitHub pages branch in the descheduler repo?

charts/descheduler/Chart.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
+name: descheduler
+version: 1.0.0
+appVersion: 0.18.0
```
seanmalloy, line 4: Should this be `v0.18.0` to match the git tag for descheduler?

charts/descheduler/README.md
```diff
@@ -0,0 +1,58 @@
+# Descheduler for Kubernetes
+
+[Descheduler](https://github.com/kubernetes-sigs/descheduler/) for Kubernetes is used to manage pod scheduling.
```
seanmalloy, line 3: How about something like this here ...

[Descheduler](https://github.com/kubernetes-sigs/descheduler/) for Kubernetes is used to rebalance clusters by evicting pods that can potentially be scheduled on better nodes. In the current implementation, descheduler does not schedule replacement of evicted pods but relies on the default scheduler for that.

charts/descheduler/README.md
```diff
@@ -0,0 +1,58 @@
...
+The command removes all the Kubernetes components associated with the chart and deletes the release.
+
+## Configuration
+
+The following table lists the configurable parameters of the _descheduler_ chart and their default values.
+
+| Parameter                      | Description                                                                                                           | Default                                                |
+| ------------------------------ | --------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------ |
+| `image.repository`             | Docker repository to use                                                                                              | `us.gcr.io/k8s-artifacts-prod/descheduler/descheduler` |
+| `image.tag`                    | Docker tag to use                                                                                                     | `0.18.0`                                               |
```
seanmalloy, line 49: This should be `v0.18.0`

charts/descheduler/templates/clusterrolebinding.yaml
```diff
@@ -0,0 +1,16 @@
...
+  labels:
+    {{- include "descheduler.labels" . | nindent 4 }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: {{ template "descheduler.fullname" . }}
+subjects:
+  - kind: ServiceAccount
+    name: {{ template "descheduler.serviceAccountName" . }}
+    namespace: {{ .Release.Namespace }}
```
seanmalloy, line 15: Where is `.Release.Namespace` set?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,30 @@
+name: Release Charts
```
seanmalloy, line 1: I think it will be ok to use GitHub actions to automate the build and deploy of the helm chart at release time. We will have to see what the other project maintainers think about this.

seanmalloy:
> It looks like #296 has broken builds on [master](https://github.com/kubernetes-sigs/descheduler/commits/master)?

@stevehipwell the master branch has been fixed.

seanmalloy:
@stevehipwell here is the documentation for the current release process let me know if you have any questions.

https://github.com/kubernetes-sigs/descheduler/blob/master/docs/release-guide.md#semi-automatic

.github/workflows/release.yaml
stevehipwell: If you plan on testing and validating the chart it needs to be released after the new Docker image has been released. As with the linting and testing above, the chart process should be seen as complimentary to the actual release and not directly part of it. For this reason I'd strongly recommend this tagging strategy; I've borrowed it from the guys over at [Flux](https://github.com/fluxcd) but I think it's a standard strategy when you have a chart directly in your repo.

.github/workflows/lint-test.yaml
stevehipwell: I'd strongly recommend against dropping the lint and test component of the Helm integration. As with the tagging strategy below I think it should be seen as a separate but complimentary process with a separate pipeline. In this scenario using GitHub Actions for both releases and testing would make sense.

.github/workflows/release.yaml
stevehipwell: Yes it does ([docs](https://github.com/helm/chart-releaser-action)).

charts/descheduler/Chart.yaml
stevehipwell: The convention in Helm charts is not to add the _v_ prefix to versions (see https://hub.helm.sh/).

charts/descheduler/README.md
stevehipwell: Will update PR.

charts/descheduler/README.md
stevehipwell: Will update PR.

charts/descheduler/templates/clusterrolebinding.yaml
stevehipwell: This is part of _Helm_.

.github/workflows/release.yaml
stevehipwell: Please see my comments on releasing above.

.github/workflows/release.yaml
seanmalloy: Looks good. Thanks.

charts/descheduler/Chart.yaml
seanmalloy: Looks good.

charts/descheduler/templates/clusterrolebinding.yaml
seanmalloy: Looks good.

seanmalloy:
@stevehipwell I think this generally looks good. I'll test out the helm chart in the next few days, and provide any additional feedback.

I'll need to sort out the usage of GitHub actions and the release process with the other maintainers.

stevehipwell:
Thanks @seanmalloy.

seanmalloy:
@stevehipwell I will finish my review in the next 48 hours. Sorry for the delay.

stevehipwell:
Thanks @seanmalloy

seanmalloy:
How I tested the helm chart locally on my laptop:
```
kind create cluster --config hack/kind_config.yaml
helm package charts/descheduler/
helm install descheduler descheduler-1.0.0.tgz
kubectl get cronjobs --all-namespaces
NAMESPACE   NAME          SCHEDULE      SUSPEND   ACTIVE   LAST SCHEDULE   AGE
default     descheduler   */2 * * * *   False     0        <none>          52s
```

.github/workflows/lint-test.yaml
seanmalloy: @stevehipwell thanks for your feedback and thoughts. My opinion is still that the descheduler project should not three separate CI systems(Travis, Prow, and GH Actions). We are trying to get all CI tests run through Prow, so that we have a single CI system.

I think this file should be removed from this PR.

If you want to you can update the `Makefile` in this PR to add some targets for helm chart linting and testing. But this is not required before merging in my opinion.

seanmalloy:
/hold cancel

@stevehipwell from my perspective the only thing that needs to be addressed this comment https://github.com/kubernetes-sigs/descheduler/pull/298/files#r439711929. Once this is addressed I'll add my LGTM to this PR. Thanks.

.github/workflows/lint-test.yaml
stevehipwell: @seanmalloy I'll comment out the `lint-test` GH Action, although we will still be using all three CI systems as GH Actions is still being used for releasing. In the scope of this chart I don't think that the automated tests would have too many benefits, so I agree they could be omitted for now.

stevehipwell:
@seanmalloy could you take a look again? I've made the final changes.

seanmalloy:
/lgtm
/cc @ingvagabund @damemi 

@damemi and @ingvagabund please take a look at this PR when you have some time. This adds an official helm chart. I tested the helm chart and it works as I would expect.

There are a few things the maintainers need to understand prior to merging this PR. I'll add a summary comment hopefully in the next 24 hours covering these details, but right now I need to get some sleep.

billimek:
@damemi here is an example from the helm organization which leverages github actions to lint, test, and release charts: https://github.com/helm/charts-repo-actions-demo with the associated action code broken out:

* https://github.com/helm/chart-testing-action
* https://github.com/helm/chart-releaser-action

I believe this is the implementation that @stevehipwell attempted to implement in the PR.  It stands to reason that we could extrapolate the excellent linting, testing, and releasing implementation details to make it work in Prow.

damemi:
@billimek thanks, I think that is what @stevehipwell was originally adding to this PR. Though I was more asking if there are any existing examples in the kubernetes org, that show how to adapt the helm tests to prow or set a precedent for using prow and github actions side-by-side

seanmalloy:
@damemi and @ingvagabund here is my summary comment that explains the next steps for this PR from my perspective.

High level steps ...
1. Finish reviewing this PR and merge it.
2. Update descheduler repos settings(enable GitHub pages and GitHub Actions)
3. Update release documentation with steps for releasing the helm chart
4. Figure out how to add helm CI tests to Prow

Regarding 1 and 3 I propose we just focus on the getting the helm chart reviewed and merged as part of this pull request. I will volunteer to sort out documenting the helm chart release process in a separate PR if you are ok with that.

Regarding 2 GitHub pages will be used to store the helm artifacts(tar files). GitHub actions will be used to automatically build and push the artifacts to the GitHub pages branch. GitHub pages will function as a web server for accessing the helm artifacts. A few options for storing helm chart artifacts were considered see the discussion in #248 for details. @damemi will need to do any of the descheduler GitHub repo settings changes.

Regarding 4 it looks like there are `helm lint` and `helm test` subcommands, so I think we can just create some new targets in the `Makefile` and just have Prow run these new make targets. Please keep in mind I'm not a helm or prow expert, so I could be wrong.

Last but not least we will add new helm chart specific tags in the descheduler repo that will control the release automation for helm charts. For example `chart-0.18.0`. Creating these tags in the descheduler repo will trigger the GitHub actions release(CD) automation.

Sorry for the long comment and the delayed response. I just want to make sure we all understand what the maintainers are signing up for by offering an official helm chart for descheduler. I do think that having an official helm chart will be good for end users to more easily install the descheduler.

ingvagabund:
Thanks guys for all the hard work bringing and discussing the integration with Helm charts. As @damemi mentioned, even I am not skilled in using the charts. Also, I don't want to diverge from the community practices in the testing area wrt. GitHub Actions. https://github.com/kubernetes-sigs/service-catalog has the charts already available. You might sync with the repo owners what are/were their thoughts.

[1] https://github.com/kubernetes-sigs/service-catalog/tree/master/charts

damemi:
I see that in the service-catalog repo @ingvagabund mentioned, they don't have the separate branches/tags for helm chart. If I'm understanding correctly, the reason for setting up `chart-xxx` tags is to limit releases of the helm chart to merges of those tags. But, we'll have to release a new helm chart with each release of the descheduler, so could we just base these lint/release actions on our normal `v0.xx` tags?

Regarding @seanmalloy's comments above,

> High level steps ...
> 
>     1. Finish reviewing this PR and merge it.
> 
>     2. Update descheduler repos settings(enable GitHub pages and GitHub Actions)
> 
>     3. Update release documentation with steps for releasing the helm chart
> 
>     4. Figure out how to add helm CI tests to Prow

Everything you said sounds good to me, except that I think we should make sure we have CI tests enabled before turning on the Github pages/Actions and begin publishing charts.

Also,

> Last but not least we will add new helm chart specific tags in the descheduler repo that will control the release automation for helm charts. For example chart-0.18.0. Creating these tags in the descheduler repo will trigger the GitHub actions release(CD) automation.

Like I said above, could these actions just be triggered by our normal release tags?

seanmalloy:
@damemi and @ingvagabund here is what I found out about the helm chart release process for [k8s service-catalog](https://github.com/kubernetes-sigs/service-catalog) ...

The released helm charts are deployed to [this](https://svc-catalog-charts.storage.googleapis.com/) Google Cloud storage bucket. The [publish-helm-chart.sh shell script](https://github.com/kubernetes-sigs/service-catalog/blob/303abfaa5386a505f35aebd3e99792ea0006b656/contrib/hack/ci/publish-helm-chart.sh) is used to build and push the released helm charts to the storage bucket using [Travis CI](https://github.com/kubernetes-sigs/service-catalog/blob/189eb535ff5c76d5244b72ae5188a2d7e0577639/.travis.yml#L147).

In my opinion we should use GitHub actions and GitHub pages to do the helm chart release automation. @damemi and @ingvagabund do you want to use GitHub actions, TravisCI, or Prow to execute the helm chart release automation? It's probably not impossible to make it work with Prow, but I'm not sure the exact steps to make it work.

@stevehipwell would you be able to update the `Makefile` to run `helm lint` as part of this PR? Let me know if you need any help and I can sort it out if needed. We are going to use k8s Prow to run `helm lint`, not GitHub Actions.

stevehipwell:
@damemi the reason why I'm recommending a separate tag for the _Helm_ chart release is that the chart is a composition of the _descheduler_ release and additional logic and as such needs to be treated as a separate concern. Some projects like to keep their chart next to their application for locality reasons (e.g. [Flux](https://github.com/fluxcd/flux)) while others have a separate repo for their charts (e.g. [AWS EKS](https://github.com/aws/eks-charts)); I understood that the former was required here. On versioning a _Helm_ chart the version wont always match your release version but there are some standard patterns that can be adopted with different degrees of flexibility. It's also worth noting that the chart version comes from the `chart.yaml` file and not the tag so even if you attempted to treat everything as a single component you'd need to commit the tag version into `chart.yaml` before being able to tag for a general release.

@seanmalloy which actions would you like to use, linting on it's own or linting and testing?

seanmalloy:
> @damemi the reason why I'm recommending a separate tag for the _Helm_ chart release is that the chart is a composition of the _descheduler_ release and additional logic and as such needs to be treated as a separate concern. Some projects like to keep their chart next to their application for locality reasons (e.g. [Flux](https://github.com/fluxcd/flux)) while others have a separate repo for their charts (e.g. [AWS EKS](https://github.com/aws/eks-charts)); I understood that the former was required here. On versioning a _Helm_ chart the version wont always match your release version but there are some standard patterns that can be adopted with different degrees of flexibility. It's also worth noting that the chart version comes from the `chart.yaml` file and not the tag so even if you attempted to treat everything as a single component you'd need to commit the tag version into `chart.yaml` before being able to tag for a general release.
> 
> @seanmalloy which actions would you like to use, linting on it's own or linting and testing?

If you could do linting and testing in the Makefile that would be great. I think at a minimum we need helm chart linting.

damemi:
@stevehipwell thank you for explaining, I just wanted to make sure I understood the approach for separate tags here.

> In my opinion we should use GitHub actions and GitHub pages to do the helm chart release automation.

@seanmalloy that seems the most straightforward here too. I agree that if we can add the `helm lint` tests to the Makefile in this PR, it will look good to merge

stevehipwell:
@seanmalloy could you take a look at the Makefile and Travis config?

Makefile
```diff
@@ -86,3 +86,6 @@ ifndef HAS_GOLANGCI
 	curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b ./_output/bin ${GOLANGCI_VERSION}
 endif
 	./_output/bin/golangci-lint run
+
+lint-chart:
```
seanmalloy, line 90: Please add the `lint-chart` target to the list of targets on line 65.

Makefile
```diff
@@ -86,3 +86,6 @@ ifndef HAS_GOLANGCI
...
+	helm lint ./charts/descheduler
```
seanmalloy, line 91: Does the `helm lint` command exit non-zero if it finds a linting error?

seanmalloy: Also you may need to add some code to download the helm binary. See an example of lines 85 to 87.

.travis.yml
```diff
@@ -22,7 +22,9 @@ before_script:
...
   - kind get kubeconfig > /tmp/admin.conf
+  - curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3 && chmod 700 ./get_helm.sh && ./get_helm.sh
 script:
-- mkdir -p ~/gopath/src/sigs.k8s.io/
-- mv ~/gopath/src/github.com/kubernetes-sigs/descheduler ~/gopath/src/sigs.k8s.io/.
-- make test-e2e
+  - mkdir -p ~/gopath/src/sigs.k8s.io/
+  - mv ~/gopath/src/github.com/kubernetes-sigs/descheduler ~/gopath/src/sigs.k8s.io/.
+  - make test-e2e
+  - make lint-chart
```
seanmalloy, line 30: The Travis CI file does not need to be updated. We will use Prow to run the tests, not Travis CI. Sorry if this was not clear in my previous comments.

Makefile
stevehipwell: Yes `helm lint` will exit with a non-zero code if it finds an error.

.travis.yml
stevehipwell: Sorry, I missed this as I was working on the old Travis CI config before I had to rebase my branch.

stevehipwell:
@seanmalloy I've made the changes you requested.

stevehipwell:
I'm not very familiar with Prow, but it should be possible to get chart testing integrated as has been done at [helm/charts](https://github.com/helm/charts/tree/master/test). I think it'd be worth getting this merged first as a baseline though.

damemi:
> I'm not very familiar with Prow, but it should be possible to get chart testing integrated as has been done at [helm/charts](https://github.com/helm/charts/tree/master/test). I think it'd be worth getting this merged first as a baseline though.

Yeah, we can get testing added for the helm branches similarly to how it was done in https://github.com/kubernetes/test-infra/pull/18076. Is that what you're referring to?

stevehipwell:
@damemi yes that's the one.

.github/workflows/lint-test.yaml
```diff
@@ -0,0 +1,33 @@
+# name: Lint and Test Charts
```
seanmalloy, line 1: @stevehipwell please delete this file as it is not needed.

.travis.yml
```diff
@@ -23,6 +23,6 @@ before_script:
   - kind load docker-image kubernetes/pause
   - kind get kubeconfig > /tmp/admin.conf
 script:
-- mkdir -p ~/gopath/src/sigs.k8s.io/
-- mv ~/gopath/src/github.com/kubernetes-sigs/descheduler ~/gopath/src/sigs.k8s.io/.
-- make test-e2e
+  - mkdir -p ~/gopath/src/sigs.k8s.io/
+  - mv ~/gopath/src/github.com/kubernetes-sigs/descheduler ~/gopath/src/sigs.k8s.io/.
+  - make test-e2e
```
seanmalloy, line 28: @stevehipwell please remove the leading spaces here. This is not required in the PR. Thanks!

seanmalloy:
@stevehipwell just two small changes needed. See https://github.com/kubernetes-sigs/descheduler/pull/298#discussion_r453155960 and https://github.com/kubernetes-sigs/descheduler/pull/298#discussion_r453156016.

The changes you made to the `Makefile` look great. Thanks!

stevehipwell:
@seanmalloy I've made the requested changes, could you re-add the /lgtm tag?

As a side note it might be worth getting something like [EditorConfig](https://editorconfig.org/) set up in the repo to make sure that files are formatted in a consistent way to avoid commits causing an issue like https://github.com/kubernetes-sigs/descheduler/pull/298#discussion_r453156016.

charts/descheduler/values.yaml
```diff
@@ -0,0 +1,51 @@
+# Default values for descheduler.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+image:
+  repository: us.gcr.io/k8s-artifacts-prod/descheduler/descheduler
+  tag: v0.18.0
```
ndegory, line 7:
```suggestion
  # Overrides the image tag whose default is the chart version
  tag: ""
```

charts/descheduler/templates/cronjob.yaml
```diff
@@ -0,0 +1,53 @@
...
+            {{- end }}
+        spec:
+          {{- if .Values.priorityClassName }}
+          priorityClassName: {{ .Values.priorityClassName }}
+          {{- end }}
+          serviceAccountName: {{ template "descheduler.serviceAccountName" . }}
+          restartPolicy: "Never"
+          containers:
+            - name: {{ .Chart.Name }}
+              image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
```
ndegory, line 34:
```suggestion
              image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
```

charts/descheduler/README.md
```diff
@@ -0,0 +1,59 @@
...
+The command removes all the Kubernetes components associated with the chart and deletes the release.
+
+## Configuration
+
+The following table lists the configurable parameters of the _descheduler_ chart and their default values.
+
+| Parameter                      | Description                                                                                                           | Default                                                |
+| ------------------------------ | --------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------ |
+| `image.repository`             | Docker repository to use                                                                                              | `us.gcr.io/k8s-artifacts-prod/descheduler/descheduler` |
+| `image.tag`                    | Docker tag to use                                                                                                     | `v0.18.0`                                              |
```
ndegory, line 49:
```suggestion
| `image.tag`                    | Docker tag to use                                                                                                     | `v[chart appVersion]`                                              |
```

charts/descheduler/values.yaml
```diff
@@ -0,0 +1,51 @@
...
+  v: 3
+  # evict-local-storage-pods:
+  # max-pods-to-evict-per-node: 10
+  # node-selector: "key1=value1,key2=value2"
+
+deschedulerPolicy:
+  strategies:
+    RemoveDuplicates:
+      enabled: true
+    RemovePodsViolatingInterPodAntiAffinity:
```
ndegory, line 25:
```suggestion
    RemovePodsViolatingNodeTaints:
      enabled: true
    RemovePodsViolatingNodeAffinity:
      enabled: true
      params:
         nodeAffinityType:
         - requiredDuringSchedulingIgnoredDuringExecution
    RemovePodsViolatingInterPodAntiAffinity:
```

seanmalloy:
@stevehipwell please take a look at the suggestions from @ndegory and then squash commits. After that I will LGTM. Thanks for your patience!

stevehipwell:
@seanmalloy I've added the changes @ndegory asked for. I can manually squash my code if you'd like but is it not safer to use [GitHub](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-request-merges#squash-and-merge-your-pull-request-commits) for this when the PR is merged?

damemi:
@stevehipwell we generally prefer to have the commits squashed, as merging is handled by bots in k8s repos (not by users), so we have no option to choose squash and merge (except for project admins)

stevehipwell:
Thanks @damemi, that makes sense. I've squashed the commits, so should be all good now.

seanmalloy:
/assign @damemi @ingvagabund 

I can take of updating `docs/release-guide.md` after the this PR is merged.

stevehipwell:
@damemi I've just realised that I haven't updated the action to apply the CR token fix, although I have tested it on my own chart repo. I'll make the change to the release.yaml file and push it up squashed. Once I've done this there is nothing to be done from the admin side of things.

stevehipwell:
@damemi / @seanmalloy I've updated the GH action config so there is no admin actions required, sorry I missed it off. I think the `/lgtm` flag needs to be re-applied but we should be good to go.

seanmalloy:
/lgtm

@damemi I believe the only thing that needs to be configured after merging this PR is to enable GitHub pages in the descheduler repo. As previously mentioned I'll take care of updating `docs/release-guide.md` after this PR is merged.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,30 @@
...
+        run: git fetch --prune --unshallow
+
+      - name: Add dependency chart repos
+        run: |
+          helm repo add stable https://kubernetes-charts.storage.googleapis.com/
+
+      - name: Run chart-releaser
+        uses: helm/chart-releaser-action@v1.0.0-rc.2
+        env:
+          CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
```
damemi, line 30: Just double checking, this is still needed with the CR_TOKEN bug fixed?

stevehipwell: @damemi `secrets.GITHUB_TOKEN` is the auto generated token that means you don't have to manually generate one.

stevehipwell:
@damemi could you hold off on the merge, it looks like the requested change to use chart version for the tag didn't prepend a `v` to match the tag (chart version doesn't usually have a v prefix). I will push a new fix!

damemi:
@stevehipwell no problem
/approve cancel
Can you confirm my question above re: which branch github pages will need to be based on?

stevehipwell:
@damemi I've fixed this and validated it locally and I think it looks good!

stevehipwell:
@damemi sorry I missed the GH pages q. The action should create the gh-pages branch automatically.

damemi:
> @damemi sorry I missed the GH pages q. The action should create the gh-pages branch automatically.

Awesome, thanks for clearing that up (and for your patience in helping us understand how this will work!). I think this is all set and I don't have any more questions. Thank you for the work 👍 
/approve

stevehipwell:
@seanmalloy @damemi it looks like the autoscaler team have done the exact same thing as we've done and had the same issue.

https://github.com/kubernetes/autoscaler/pull/3393

damemi:
Yeah, it seems they are hitting the same problem as us. I like their suggestion of following the PR workflow (as done by kubernetes/dashboard https://github.com/kubernetes/autoscaler/pull/3393#issuecomment-668081392) but I don't think it will work because autoscaler, like us, is using the [official helm chart releaser action](https://github.com/kubernetes/autoscaler/blob/master/.github/workflows/release.yaml#L16) whereas dashboard is using a [different one](https://github.com/wiremind/dashboard/blob/e2bccc8019d61e18214de1070c2fb0fd7d1645e9/.github/workflows/cd-helm-workflow.yml#L43).

The helm chart releaser handles all of the pushing itself and doesn't have any configuration for that which I could find. So again, I'm torn between using the official action (which is less configurable) vs a 3rd party action with more options

damemi:
As I check the settings in our repo I see that the CLA check branch protection has somehow been re-enabled in our repo. With this check we basically cannot use the helm chart releaser action because the push fails on that branch.

damemi:
I opened an issue in the chart-releaser-action repo to try to fix this https://github.com/helm/chart-releaser-action/issues/33

damemi:
We might want to go with something like https://github.com/kubernetes/autoscaler/pull/3403 using just hack scripts. Thought it would be great if the helm action could decouple the git push from the rest of its flow

stevehipwell:
>We might want to go with something like kubernetes/autoscaler#3403 using just hack scripts. Thought it would be great if the helm action could decouple the git push from the rest of its flow

I'd think that using something like https://github.com/stefanprodan/gh-actions/tree/master/helm-gh-pages?

I also opened an issue to see if the chart releaser could be modified to use a tag prefix.

https://github.com/helm/chart-releaser/issues/71

damemi:
Perhaps it would be worth trying to push for a k8s-wide helm-charts repo, as suggested in https://github.com/kubernetes/autoscaler/pull/3403#issuecomment-670029502

stevehipwell:
@damemi that would probably make sense. The action that Flux uses supports a separate GitHub project as the repo.

stevehipwell:
@damemi @seanmalloy have you seen that the `chart-releaser` has a new flag which would enable the removal of the `-chart` postfix from the chart name.

https://github.com/helm/chart-releaser/pull/74

damemi:
@stevehipwell thanks, I did see that. That will be good to have our chart name make more sense. Any idea when that will be available in a release we can use?

stevehipwell:
@damemi it looks like the _chart-releaser_ action needs to be updated to support the `--release-name-template` flag.

https://github.com/helm/chart-releaser-action/issues/37

PR #300, opened 2020-05-27
Add more verbose logging to IsEvictable checks

damemi:
This makes the `IsEvictable` check more verbose at `logLevel=4`, but breaking each check into its own conditional and logging errors for each one, finally outputting the full list if the pod does not also have the "evictable" annotation.

This gives output like this:
```
I0529 16:14:06.044444       1 pods.go:60] Pod dedicated-nodes-vkr8s in namespace test-ns is not evictable: Pod lacks an eviction annotation and fails the following checks: [pod is critical, pod does not have any ownerrefs]
```

### DIFF
--- a/pkg/descheduler/pod/pods.go
+++ b/pkg/descheduler/pod/pods.go
@@ -18,10 +18,13 @@ package pod
 
 import (
 	"context"
+	"fmt"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/util/errors"
 	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/klog"
 	"sigs.k8s.io/descheduler/pkg/utils"
 )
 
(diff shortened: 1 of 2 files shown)


damemi:
/hold
I haven't personally tested this yet to see what the error output looks like. If someone else gets to it before I do and it looks fine anyone can feel free to remove the hold

pkg/descheduler/pod/pods.go
```diff
@@ -31,8 +34,30 @@ const (
...
+	if !evictLocalStoragePods && IsPodWithLocalStorage(pod) {
+		checkErrs = append(checkErrs, fmt.Errorf("pod has local storage and descheduler is not configured with --evict-local-storage-pods"))
+	}
+
+	if IsMirrorPod(pod) {
+		checkErrs = append(checkErrs, fmt.Errorf("pod is a mirror pod"))
+	}
+
+	if len(checkErrs) > 0 && !HaveEvictAnnotation(pod) {
+		klog.V(4).Infof("Pod %s is not evictable: Pod lacks an eviction annotation and fails the following checks: %v", errors.NewAggregate(checkErrs).Error())
```
ingvagabund, line 60: `pkg/descheduler/pod/pods.go:60:3: Infof format %v reads arg #2, but call has 1 arg`

Missing pod name

damemi: Ah yes, that would help. Updated thanks

seanmalloy:
@damemi I think it would be good to note in the [Pod Evictions](https://github.com/kubernetes-sigs/descheduler#pod-evictions) section of the README that using log level 4 will log the reason descheduler is not evicting a pod. I believe this would reduce the number of questions end users have about why pods are not being evicted.

damemi:
@seanmalloy good idea, updated

seanmalloy:
/lgtm

The hold can be removed once someone has verified what the log output actually looks like.

PR #301, opened 2020-05-28
Add support for degradation

pmundt:
The default behaviour is to not deschedule a Pod if it cannot be placed
anywhere else, this is fine for the default case where nodes are more
or less homogeneous, but it does not suit the case of heterogeneous
clusters where specific nodes may have their own unique hardware
configurations that are not be reproduced anywhere else within the
cluster.

In the case of heterogeneous accelerators, for example, there may be
pods that have a hard dependency on a specific resource (e.g. including
container runtimes geared at a specific accelerator), where allowing
them to continue running would produce undesired and unpredictable
behaviour. Consider the case of a Pod with a hard dependency on a
USB-attached accelerator, which may disappear during the lifecycle of
the Pod.

This adds a new `--degradation-allowed` flag that specifically allows
Pods to indicate when they should be descheduled after a label they
depend on suddenly goes away.

### DIFF
--- a/pkg/descheduler/evictions/evictions.go
+++ b/pkg/descheduler/evictions/evictions.go
@@ -40,6 +40,7 @@ type PodEvictor struct {
 	client             clientset.Interface
 	policyGroupVersion string
 	dryRun             bool
+	DegradationAllowed bool
 	maxPodsToEvict     int
 	nodepodCount       nodePodEvictedCount
 }
@@ -48,6 +49,7 @@ func NewPodEvictor(
 	client clientset.Interface,
 	policyGroupVersion string,
 	dryRun bool,
+	degradationAllowed bool,
 	maxPodsToEvict int,
 	nodes []*v1.Node,
 ) *PodEvictor {
@@ -61,6 +63,7 @@ func NewPodEvictor(
 		client:             client,
 		policyGroupVersion: policyGroupVersion,
 		dryRun:             dryRun,
+		DegradationAllowed: degradationAllowed,
 		maxPodsToEvict:     maxPodsToEvict,
 		nodepodCount:       nodePodCount,
 	}
--- a/pkg/descheduler/strategies/node_affinity.go
+++ b/pkg/descheduler/strategies/node_affinity.go
@@ -45,7 +45,7 @@ func RemovePodsViolatingNodeAffinity(ctx context.Context, client clientset.Inter
 
 				for _, pod := range pods {
 					if pod.Spec.Affinity != nil && pod.Spec.Affinity.NodeAffinity != nil && pod.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution != nil {
-						if !nodeutil.PodFitsCurrentNode(pod, node) && nodeutil.PodFitsAnyNode(pod, nodes) {
+						if !nodeutil.PodFitsCurrentNode(pod, node) && (nodeutil.PodFitsAnyNode(pod, nodes) || podEvictor.DegradationAllowed == true) {
 							klog.V(1).Infof("Evicting pod: %v", pod.Name)
 							if _, err := podEvictor.EvictPod(ctx, pod, node); err != nil {
 								klog.Errorf("Error evicting pod: (%#v)", err)
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ pod can or can not be scheduled, are guided by its configurable policy which com
 rules, called predicates and priorities. The scheduler's decisions are influenced by its view of
 a Kubernetes cluster at that point of time when a new pod appears for scheduling.
 As Kubernetes clusters are very dynamic and their state changes over time, there may be desire
-to move already running pods to some other nodes for various reasons:
+to move already running pods to some other nodes, or to allow the pods to be terminated outright:
 
 * Some nodes are under or over utilized.
 * The original scheduling decision does not hold true any more, as taints or labels are added to
@@ -20,9 +20,12 @@ or removed from nodes, pod/node affinity requirements are not satisfied any more
 * New nodes are added to clusters.
 
 Consequently, there might be several pods scheduled on less desired nodes in a cluster.
-Descheduler, based on its policy, finds pods that can be moved and evicts them. Please
-note, in current implementation, descheduler does not schedule replacement of evicted pods
-but relies on the default scheduler for that.
+Descheduler, based on its policy, finds pods that can be moved and evicts them. By default,
+Descheduler aims to ensure that there is no service degradation across the cluster. In the case
+where a pod is no longer runnable on a node and no suitable movement candidate can be found,
+Descheduler can also optionally terminate the problematic pod, when service degradation is allowed.
+Please note, in its current implementation, descheduler does not schedule replacements of evicted
+or terminated pods but instead relies on the default scheduler for that.
 
 ## Quick Start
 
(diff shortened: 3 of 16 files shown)


pkg/descheduler/evictions/evictions.go
```diff
@@ -61,6 +63,7 @@ func NewPodEvictor(
 		client:             client,
 		policyGroupVersion: policyGroupVersion,
 		dryRun:             dryRun,
+		DegradationAllowed: degradationAllowed,
```
ingvagabund, line 66: `DegradationAllowed` field is not used anywhere inside PodEvictor methods. Is this PR still WIP?

pmundt: No, it's used by the eviction strategy, which is passed in the pod evictor. I opted to make this exportable so we could access the flag there, rather than having to modify every single eviction strategy callsite to add in another argument.

ingvagabund: I see: `podEvictor.DegradationAllowed`. It's not a good practice. 

> rather than having to modify every single eviction strategy callsite to add in another argument

However, it's the right thing to do from the maintainability perspective. If you need to access DegradationAllowed through `PodEvictor`, introducing new method is more practical than accessing `DegradationAllowed` directly. Also, `DegradationAllowed` is not related to evicting pod itself, rather to relaxing constraint of selecting pods to be evicted.

pmundt: I would argue that changing every function every time we add a new flag is worse from a maintainability point of view, but it's not my code base, so I'm happy to follow whatever convention people like. In terms of the degradation, you are correct, we are presently doing this through a global flag and applying it across all pods that satisfy the criteria. We could also consider pushing this down to the pod level through an additional annotation?

ingvagabund: > I would argue that changing every function every time we add a new flag

You are right. The `strategyFunction` data type is still evolving. I am fine with changing the signature [1] to:
```
type strategyFunction func(
	ctx context.Context,
	client clientset.Interface,
	strategy api.DeschedulerStrategy,
	nodes []*v1.Node,
	opts Options,
)
``` 
where `Options` can be defined as:
```
type Options struct {
	EvictLocalStoragePods bool
	DegradationAllowed bool
}
```
as long as `Options` contains fields generic for any strategy.

[1] https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/descheduler/descheduler.go#L63

> We could also consider pushing this down to the pod level through an additional annotation?

If I understand the intention correctly, `DegradationAllowed` is a strategy level configuration.  As such, the option needs to be passed into the strategy before it's ran.

pmundt: The problem with this is that the descheduler package would define the Options struct, while the strategies package would need to import it for the function definitions, and the descheduler package already imports the strategies package, so we would end up with a circular import. I've worked around this by placing the Options struct into the strategies package. Should we want to use this in e.g. the pod package for option evaluation within ListEvictablePodsOnNode(), however, we will again run into a circular import.

I will open a separate PR for the options refactoring, and once we're happy with that I'll then update this PR.

pmundt: Please refer to PR #305

seanmalloy:
One thing I noticed is that this PR does not update any of the end user documentation. I believe the new `--degradation-allowed` CLI option should be documented in the file `docs/user-guide.md`.

pmundt:
@seanmalloy good point, I've pushed out some documentation updates now.

pkg/descheduler/evictions/evictions.go
damemi: @pmundt if this is only used by the `nodeAffinity` (and, in your new strategy `nodeSelector`), does it have to be a global setting? It could just be a [`StrategyParam`] field for those right?

pmundt: As noted in the other PR (sorry for the run-around, I mention it here for posterity), we would also plan to evaluate it within `IsEvictable` in cases where we know the pod can be explicitly degraded. I'm therefore not sure that the `StrategyParam` would be sufficient, unless we're also able to access this from the pod package somehow. In terms of the global setting, I don't mind if we get rid of this and use a Pod annotation or similar, the important thing is simply that we have a mechanism to degrade node-local Pods - I'll defer to you on whichever option you find more palatable! If we leave it as an annotation, we could presumably also leave it as a `StrategyParam`, as the annotation could be tested independently within `IsEvictable`.

pkg/descheduler/strategies/node_affinity.go
```diff
@@ -45,7 +45,7 @@ func RemovePodsViolatingNodeAffinity(ctx context.Context, client clientset.Inter
 
 				for _, pod := range pods {
 					if pod.Spec.Affinity != nil && pod.Spec.Affinity.NodeAffinity != nil && pod.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution != nil {
-						if !nodeutil.PodFitsCurrentNode(pod, node) && nodeutil.PodFitsAnyNode(pod, nodes) {
+						if !nodeutil.PodFitsCurrentNode(pod, node) && (nodeutil.PodFitsAnyNode(pod, nodes) || podEvictor.DegradationAllowed == true) {
```
ingvagabund, line 48: You don't want to ignore `nodeutil.PodFitsAnyNode` since it contains must conditions for a pod getting descheduled. So even when you allow degradation your pod will stay in Pending state until a suitable node is found.

pkg/descheduler/evictions/evictions.go
damemi: > In terms of the global setting, I don't mind if we get rid of this and use a Pod annotation or similar, the important thing is simply that we have a mechanism to degrade specific node-local Pods

Is this something that the existing [`descheduler.alpha.kubernetes.io/evict`](https://github.com/kubernetes-sigs/descheduler/blob/bd412bf87f72132efad2a07833666463cb909969/pkg/descheduler/pod/pods.go#L32) annotation could solve? It is already checked in `IsEvictable` and effectively bypasses that check for specific pods.

> If we leave it as an annotation, we could presumably also leave it as a StrategyParam, as the annotation could be tested independently within IsEvictable.

With the above annotation^ I think this is what you want. In [the other PR thread](https://github.com/kubernetes-sigs/descheduler/pull/305#issuecomment-637463597) you mentioned wanting to specifically bypass DaemonSets too, which that annotation supports.

pmundt: > Is this something that the existing [`descheduler.alpha.kubernetes.io/evict`](https://github.com/kubernetes-sigs/descheduler/blob/bd412bf87f72132efad2a07833666463cb909969/pkg/descheduler/pod/pods.go#L32) annotation could solve? It is already checked in `IsEvictable` and effectively bypasses that check for specific pods.

You're right, for some reason when I first took a look at the evict annotation I missed this. I think this would do the job, yes.
 
> With the above annotation^ I think this is what you want. In [the other PR thread](https://github.com/kubernetes-sigs/descheduler/pull/305#issuecomment-637463597) you mentioned wanting to specifically bypass DaemonSets too, which that annotation supports.
> 
> Sorry if it seems like I'm being difficult, I just don't yet see the need to pass this information to every strategy.

No problem, it's not always obvious what the preferred direction is when twiddling in someone else's code base, and as it turns out, I misread what the evict annotation actually does, so I'm happy for another set of eyes while I come to grips with things.

If you're happy with the `StrategyParam` direction, I'll give this a go with the evict annotation and see how it goes.

pmundt:
Yes, that's basically correct. In our case we are running these Pods across multiple (heterogeneous) Edge gateways (in a vehicle fleet) in which we rely on feature discovery / labelling to determine the precise capabilities of the Gateway and to schedule the requisite Pods on it - in this case, DaemonSets for metric exporters and device plugins, and normal Deployments for Pods that are using a specific container runtime + inference model adapted for that specific accelerator. In the event when there is a hardware change at the Gateway level we basically want to immediately deschedule Pods that had a tight coupling to the hardware and schedule replacement Pods that use an alternative hardware configuration (we handle this separately, but I mention it so you get a better idea of the scenario).

The USB removal scenario is one example, the somewhat more frequent case we have to deal with is an accelerator suddenly becoming unavailable to the node because it's been tasked with a more "important" task (e.g. we may be running an inference model on a GPU for monitoring road conditions, but then someone in the backseat decides they want to play a game, so we need to free the resource up in a reasonable amount of time and shift on to something else that can take over the inference workload). I wrote a separate label monitor that can immediately trigger the descheduler cronjob on label changes (as well as notify our orchestrator), and this, together with the degradation mode seems to be working as we need it to.

pkg/descheduler/evictions/evictions.go
damemi: Yes let's do that -- I agree that there are cases when these affinity strategies may want to evict even if the pod can't fit anywhere else, so adding another Param for these strategies makes sense to me.

I'm also assuming that if you enable that Param the user must know what they are doing, because as @ingvagabund [pointed out](https://github.com/kubernetes-sigs/descheduler/pull/301#discussion_r433909302) if you're not properly handling such an evict-at-all-costs case you will end up with stuck Pending pods. I don't think that's the descheduler's concern though (as long as the possibility is clearly documented)

pmundt:
Based on discussion with @damemi, this PR has been superseded by #311.

PR #305, opened 2020-05-28
Refactor strategy options for future extensibility

pmundt:
At present each strategy function codies a list of flags that are passed
in, we generalize this instead by switching to an Options struct which
can be easily extended for future flags without needing to modify the
function prototype.

### DIFF
--- a/pkg/descheduler/descheduler.go
+++ b/pkg/descheduler/descheduler.go
@@ -60,7 +60,7 @@ func Run(rs *options.DeschedulerServer) error {
 	return RunDeschedulerStrategies(ctx, rs, deschedulerPolicy, evictionPolicyGroupVersion, stopChannel)
 }
 
-type strategyFunction func(ctx context.Context, client clientset.Interface, strategy api.DeschedulerStrategy, nodes []*v1.Node, evictLocalStoragePods bool, podEvictor *evictions.PodEvictor)
+type strategyFunction func(ctx context.Context, client clientset.Interface, strategy api.DeschedulerStrategy, nodes []*v1.Node, opts strategies.Options, podEvictor *evictions.PodEvictor)
 
 func RunDeschedulerStrategies(ctx context.Context, rs *options.DeschedulerServer, deschedulerPolicy *api.DeschedulerPolicy, evictionPolicyGroupVersion string, stopChannel chan struct{}) error {
 	sharedInformerFactory := informers.NewSharedInformerFactory(rs.Client, 0)
@@ -101,9 +101,13 @@ func RunDeschedulerStrategies(ctx context.Context, rs *options.DeschedulerServer
 			nodes,
 		)
 
+		opts := strategies.Options{
+			EvictLocalStoragePods: rs.EvictLocalStoragePods,
+		}
+
 		for name, f := range strategyFuncs {
 			if strategy := deschedulerPolicy.Strategies[api.StrategyName(name)]; strategy.Enabled {
-				f(ctx, rs.Client, strategy, nodes, rs.EvictLocalStoragePods, podEvictor)
+				f(ctx, rs.Client, strategy, nodes, opts, podEvictor)
 			}
 		}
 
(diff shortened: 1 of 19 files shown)


pmundt:
Note that this is a prerequisite for #301, based on the discussion there, as well as a pending PR for a new strategy.

pkg/descheduler/strategies/lownodeutilization.go
```diff
@@ -46,7 +46,10 @@ const (
 	MaxResourcePercentage = 100
 )
 
-func LowNodeUtilization(ctx context.Context, client clientset.Interface, strategy api.DeschedulerStrategy, nodes []*v1.Node, evictLocalStoragePods bool, podEvictor *evictions.PodEvictor) {
+func LowNodeUtilization(ctx context.Context, client clientset.Interface, strategy api.DeschedulerStrategy, nodes []*v1.Node, opts Options, podEvictor *evictions.PodEvictor) {
+	if !strategy.Enabled {
```
ingvagabund, line 50: This was removed by https://github.com/kubernetes-sigs/descheduler/pull/302

pkg/descheduler/descheduler.go
```diff
@@ -60,7 +60,7 @@ func Run(rs *options.DeschedulerServer) error {
 	return RunDeschedulerStrategies(ctx, rs, deschedulerPolicy, evictionPolicyGroupVersion, stopChannel)
 }
 
-type strategyFunction func(ctx context.Context, client clientset.Interface, strategy api.DeschedulerStrategy, nodes []*v1.Node, evictLocalStoragePods bool, podEvictor *evictions.PodEvictor)
+type strategyFunction func(ctx context.Context, client clientset.Interface, strategy api.DeschedulerStrategy, nodes []*v1.Node, opts strategies.Options, podEvictor *evictions.PodEvictor)
```
ingvagabund, line 63: Actually, there's nothing forbidding to move `strategyFunction` and `strategyFuncs` under `"sigs.k8s.io/descheduler/pkg/descheduler/strategies"`. @damemi WDYT?

seanmalloy:
> Note that this is a prerequisite for #301, based on the discussion there, as well as a pending PR for a new strategy.

@pmundt I'm curious can you provide any details on the new strategy that you will be proposing?

pmundt:
@seanmalloy The additional strategy is simply for `nodeSelector`-based label monitoring. The current solution for `nodeAffinity` only covers affinity cases and not the nodeSelector cases. The only reason I haven't opened a PR for this yet is because I will ultimately have to rebase it on both the options refactoring and the degradation PRs, and wanted to limit the amount of churn while we try to come up with a solution that everyone can agree on.

If you're curious about the strategy addition itself, you can view it on its topic branch here:

https://github.com/adaptant-labs/descheduler/commit/6f2e1bf82e0f387d78b6307aa1e34544ed695999

pkg/descheduler/descheduler.go
pmundt: If we're ok with that, I would also suggest pushing down the options into its own `sigs.k8s.io/descheduler/pkg/descheduler/strategies/options` package, as we should then also be able to use this from the pod package and mitigate the circular import problem. This would mean that we could then pass in the opts directly to `ListEvictablePodsOnNode`, instead of passing in a flag at a time, as it does now.

pkg/descheduler/strategies/lownodeutilization.go
pmundt: Thanks, fixed.

pkg/descheduler/descheduler.go
pmundt: I've now added a commit that does the Options push-down and cleans up the `ListEvictablePodOnNode` usage. While I'm not thrilled about adding another package, I think this is going in the right direction.

pmundt:
I actually originally had the degradationAllowed flag in podEvictor precisely for the reason that we could access it in other areas. In addition to testing for it within those 2 specific strategies, we would also plan to use it for descheduling `DaemonSets`, which would mean circumventing the logic in `IsEvictable()` where degradation is supported - we need this for device plugins that are managing devices that are removed from the node during run-time, for example.

pmundt:
Yes, I'm happy with exploring the `StrategyParameters` approach, so we can consider this closed.

PR #309, opened 2020-06-01
Make evictLocalStoragePods a property of PodEvictor

damemi:
`evictLocalStoragePods` is (effectively) only used by the functions `ListEvictablePodsOnANode` and `IsEvictable`, both of which make sense to be methods of `PodEvictor`.

In addition, passing this field to every strategy when it is really a component-level configuration option is cumbersome and unnecessary (this was discussed in https://github.com/kubernetes-sigs/descheduler/pull/305 and the PRs related to it).

So I think it makes sense to have this be a property of `PodEvictor`. Each strategy already accepts a `PodEvictor` if they want to access it directly and this minimizes their function signatures.

### DIFF
--- a/pkg/descheduler/strategies/duplicates.go
+++ b/pkg/descheduler/strategies/duplicates.go
@@ -42,12 +42,64 @@ func RemoveDuplicatePods(
 	client clientset.Interface,
 	strategy api.DeschedulerStrategy,
 	nodes []*v1.Node,
-	evictLocalStoragePods bool,
 	podEvictor *evictions.PodEvictor,
 ) {
 	for _, node := range nodes {
 		klog.V(1).Infof("Processing node: %#v", node.Name)
-		duplicatePods := listDuplicatePodsOnANode(ctx, client, node, strategy, evictLocalStoragePods)
+		pods, err := podutil.ListPodsOnANode(ctx, client, node, podEvictor.IsEvictable)
+		if err != nil {
+			klog.Errorf("error listing evictable pods on node %s: %+v", node.Name, err)
+			continue
+		}
+
+		duplicatePods := make([]*v1.Pod, 0, len(pods))
+		// Each pod has a list of owners and a list of containers, and each container has 1 image spec.
+		// For each pod, we go through all the OwnerRef/Image mappings and represent them as a "key" string.
+		// All of those mappings together makes a list of "key" strings that essentially represent that pod's uniqueness.
+		// This list of keys representing a single pod is then sorted alphabetically.
+		// If any other pod has a list that matches that pod's list, those pods are undeniably duplicates for the following reasons:
+		//   - The 2 pods have the exact same ownerrefs
+		//   - The 2 pods have the exact same container images
+		//
+		// duplicateKeysMap maps the first Namespace/Kind/Name/Image in a pod's list to a 2D-slice of all the other lists where that is the first key
+		// (Since we sort each pod's list, we only need to key the map on the first entry in each list. Any pod that doesn't have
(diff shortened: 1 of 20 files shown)


pkg/descheduler/strategies/duplicates.go
```diff
@@ -42,77 +42,71 @@ func RemoveDuplicatePods(
...
 ) {
 	for _, node := range nodes {
 		klog.V(1).Infof("Processing node: %#v", node.Name)
-		duplicatePods := listDuplicatePodsOnANode(ctx, client, node, strategy, evictLocalStoragePods)
-		for _, pod := range duplicatePods {
-			if _, err := podEvictor.EvictPod(ctx, pod, node); err != nil {
-				klog.Errorf("Error evicting pod: (%#v)", err)
-				break
-			}
+		pods, err := podEvictor.ListEvictablePodsOnNode(ctx, client, node)
```
damemi, line 49: this diff for this strategy looks a lot more complicated than it is. I just moved the private `listDuplicatePodsOnANode` logic (~30 lines) into the main public function. If anyone has any objections I'll revert it but didn't really see that it was necessary to break it up like that

ingvagabund:
> In addition, passing this field to every strategy when it is really a component-level configuration option is cumbersome and unnecessary (this was discussed in #305 and the PRs related to it).

Agree

> evictLocalStoragePods is (effectively) only used by the functions ListEvictablePodsOnANode and IsEvictable, both of which make sense to be methods of PodEvictor.

> So I think it makes sense to have this be a property of PodEvictor. Each strategy already accepts a PodEvictor if they want to access it directly and this minimizes their function signatures.

I don't have any strong opinion against. Though, `IsEvictable`/`ListEvictablePodsOnANode` are semantically different from `PodEvictor`. `PodEvictor` is a global instance that's evicting and counting pod eviction per node. `IsEvictable` is a decision making logic which is also called in `ListEvictablePodsOnANode`.

`ListEvictablePodsOnANode is a special case of `ListPodsOnANode`. So, unless there are objections, I rather prefer to change the signature of `ListPodsOnANode` to:
```
func ListPodsOnANode(ctx context.Context, client clientset.Interface, node *v1.Node, filter func(*v1.Pod) bool) ([]*v1.Pod, error)
```

and instead of calling `ListEvictablePodsOnNode` just call `ListPodsOnANode` with `IsEvictable` properly initialized. E.g.
```
ListPodsOnANode(ctx, client, node, podEvictor.IsEvictable).
```

damemi:
@ingvagabund that's a great idea, and it could actually allow strategies such as NodeAffinity to easily extend `IsEvictable` by filtering non-evictable pods from the list. See the changes I pushed in https://github.com/kubernetes-sigs/descheduler/pull/309/commits/d7273407bc904dd25d1a67d019a86b95e3b565e7 (specifically the [NodeAffinity changes](https://github.com/kubernetes-sigs/descheduler/pull/309/commits/d7273407bc904dd25d1a67d019a86b95e3b565e7#diff-90d9a54aed57e48a15b387763744077dR46-R50) that would take advantage of this new capability)

ingvagabund:
@damemi it's cunning!!! It will also allow the degradation to easily extend the filter.

Looks good. Once you squash, it's good to go. I don't see much to unit test here.

damemi:
> It will also allow the degradation to easily extend the filter.

@ingvagabund my thoughts exactly :) squashed

ingvagabund:
Travis e2e passed as well: https://travis-ci.org/github/kubernetes-sigs/descheduler/builds/695056597

PR #310, opened 2020-06-02
Skip evicting when no evictable pod found on node

lixiang233:
In lowNodeUtilization, only try to evict pods when we found evictable pods on node, otherwise, just skip and try next node.

### DIFF
--- a/pkg/descheduler/strategies/lownodeutilization_test.go
+++ b/pkg/descheduler/strategies/lownodeutilization_test.go
@@ -87,6 +87,62 @@ func TestLowNodeUtilization(t *testing.T) {
 		evictedPods         []string
 	}{
 		{
+			name: "no evictable pods",
+			thresholds: api.ResourceThresholds{
+				v1.ResourceCPU:  30,
+				v1.ResourcePods: 30,
+			},
+			targetThresholds: api.ResourceThresholds{
+				v1.ResourceCPU:  50,
+				v1.ResourcePods: 50,
+			},
+			nodes: map[string]*v1.Node{
+				n1NodeName: test.BuildTestNode(n1NodeName, 4000, 3000, 9, nil),
+				n2NodeName: test.BuildTestNode(n2NodeName, 4000, 3000, 10, nil),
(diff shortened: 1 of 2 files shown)


lixiang233:
> Looks good to me, @lixiang233 could you just add a simple test case?

done. @damemi

pkg/descheduler/strategies/lownodeutilization_test.go
```diff
@@ -86,6 +86,62 @@ func TestLowNodeUtilization(t *testing.T) {
 		expectedPodsEvicted int
 		evictedPods         []string
 	}{
+		{
+			name: "no evictable pods",
```
ingvagabund, line 90: The test will pass even without `if len(node.allPods) == len(nonRemovablePods) {` condition. The right test is to make sure pods get evicted when `len(node.allPods) != len(nonRemovablePods)`. Though, those cases are already covered. @damemi did you have any specific test in mind?

damemi: Ah, I didn't realize this was already covered. I thought this PR was improving the strategy, my mistake sorry.

@lixiang233 if that's the case and this test always passes (like @ingvagabund said) then we don't need to add a new test case. I apologize for the runaround

lixiang233: Yes, this PR just saves time and won't  effect the eviction result. I thought you just want to make sure the code won't throw any errors or panics when `if len(node.allPods) == len(nonRemovablePods)` is true. Current testcases only covered the `false` condition, so I think it's better to keep it to cover the other, that's what unit-tests for, right? But if you think this change is not complicated enough to have an individual testcase, I will remove it. @damemi  @ingvagabund

ingvagabund: It's true we don't test the case when `len(node.allPods) == len(nonRemovablePods)` so +1 for keeping the new test. I was just pointing out the new test can't tell the difference between short-cutting the loop and the case where no pods were evicted.

damemi: I see what you mean now, thanks @ingvagabund. Doesn't hurt to add a new test then

PR #311, opened 2020-06-03
Add a nodeSelector strategy

pmundt:
Presently there is only the node affinity strategy which checks node
label constraints, while in practice, pods may be constrained by either
of the affinity or nodeSelector terms. We therefore add a new strategy
that carries out label checking directly on the nodeSelector terms.

For anyone wishing a comprehensive eviction strategy based off of sudden
label non-existence, this should be used together with the nodeAffinity
strategy to ensure that both cases are caught.

PR #314, opened 2020-06-03
Embed StrategyParam structs for simpler config (and bump to v1alpha2)

damemi:
This changes the `StrategyParam` structs to be embedded, making their configuration less redundant. For example, a current configuration for the RemoveDuplicates strategy looks like:

```
kind: "DeschedulerPolicy"
strategies:
  "RemoveDuplicates":
     enabled: true
     params:
       removeDuplicates:
         excludeOwnerKinds:
         - "ReplicaSet"
```

Where this strategies config has `RemoveDuplicates: Params.RemoveDuplicates:...`. There doesn't seem to be a reason why it's necessary to specify `RemoveDuplicates` again.

The new config would look like:

```
kind: "DeschedulerPolicy"
strategies:
  "RemoveDuplicates":
     enabled: true
     params:
       excludeOwnerKinds:
       - "ReplicaSet"
```

/kind cleanup


damemi:
/hold
for review

seanmalloy:
@damemi these changes seem fine to me. Do you think this requires bumping the `apiVersion` to `descheduler/v1alpha2`?

damemi:
@seanmalloy good point, and based on the [api conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#alpha-beta-and-stable-versions) I don't think we *need* to since this is an alpha api, but it would be good to make the distinction for users. If we do that we also don't need to continue supporting v1alpha1, so this can just replace that in our 1.19 release.

In that case I'd like to hold this on merging https://github.com/kubernetes-sigs/descheduler/pull/311 first, since that also makes changes to the config API. Once that's in I'll update this to include those changes as well.

(Also note, the `,inline` keyword is apparently a no-op so to achieve the goal of this PR I have updated it to just make these structs embedded. The effect is still the same)

seanmalloy:
I think if we go forward with v1alpha2 we should rename the LowNodeUtilization strategy to LowNodeAllocation or something like that.

go.mod
```diff
@@ -11,4 +11,5 @@ require (
 	k8s.io/client-go v0.18.2
 	k8s.io/component-base v0.18.2
 	k8s.io/klog v1.0.0
+	k8s.io/kube-openapi v0.0.0-20200410145947-61e04a5be9a6 // indirect
```
ingvagabund, line 14: I don't think we need this.

damemi: This change comes from running `make gen` (the generator binaries have updated). Since we verify the vendor directory, it has to be committed to pass CI.

We discussed pinning these generators to a specific version before, so that these random vendor/go.mod updates don't happen, but haven't made any progress on that yet.

ingvagabund:
> In that case I'd like to hold this on merging #311 first, since that also makes changes to the config API. Once that's in I'll update this to include those changes as well.

/hold

damemi:
Added a 2nd commit bumping the API to v1alpha2. Still want to hold this until the nodeselector strategy merges, since that makes some api changes as well and then this PR can be updated and merged.

/kind api-change

PR #315, opened 2020-06-03
Remove redundant tests from Travis CI

damemi:
These tests are now covered by k8s test-infra prow tests.


seanmalloy:
> @seanmalloy will this affect our current k8s version test matrix that is in travis? It would be good to switch our main testing over to prow, but I don't want to lose the version compatibility tests that we currently run.

@damemi no it will not change the Travis CI test matrix. The e2e tests will still be run by Travis CI against each k8s version specified in the `.travis.yml` file.

PR #319, opened 2020-06-09
Remove redundant eviction log message and add "Reason" param to EvictPod

damemi:
Related to the changes in https://github.com/kubernetes-sigs/descheduler/pull/287, because the PodEvictor logs when a pod is successfully evicted (either in dry-run or normal mode) this message becomes redundant and confusing when in dry run. For example, it currently looks like:

```
I0528 09:23:48.664691       1 evictions.go:100] Evicted pod in dry run mode: "hello1-54d7977cd-l8wx7" in namespace "test"
I0528 09:23:48.664700       1 pod_antiaffinity.go:50] Evicted pod: "hello1-54d7977cd-l8wx7" because of existing anti-affinity
```

This extra message doesn't provide much information, though it may be reasonable to add an extra "reason" parameter to `podEvictor.EvictPod` to give context on what strategy specifically caused the eviction (since all eviction logs now just originate from `evictions.go`). This includes that change and the log output looks like:

```
evictions.go:148] Evicted pod: "dedicated-nodes-zf7kh" in namespace "mdame-test" (RemoveDuplicatePods)
```

### DIFF
--- a/pkg/descheduler/evictions/evictions.go
+++ b/pkg/descheduler/evictions/evictions.go
@@ -19,6 +19,7 @@ package evictions
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	v1 "k8s.io/api/core/v1"
 	policy "k8s.io/api/policy/v1beta1"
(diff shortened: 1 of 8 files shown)


seanmalloy:
It looks like the Travis CI e2e tests did not get triggered.

Issue #294, opened 2020-05-26
Sorting pods before evicting may be helpful


lixiang233:
In README, the `Pod Evictions` chapter writes `Best efforts pods are evicted before burstable and guaranteed pods`. But I notice that only `LowNodeUtilization` implements it. I guess that's probably because sorting pods does not affect whether pods are evicted in other strategies. I think sorting pods before evicting may be helpful because we can evict high priority pods first to give them higher possibility of being perfertly scheduled.

Here I devide all these 7 strategies into two types: `single-pod-related-strategies` and `multi-pods-related-strategies`. In `single-pod-related-strategies`, pods will always be evicted if they violate the policies, so we should evict high priority pods first to give them higher possibility of being perfertly scheduled. But in `multi-pods-related-strategies`, pods may not need to be evicted if some other pods on the same node have been evicted, so here we should  evict low priority pods first to try to avoid evicting high priority pods.

`single-pod-related-strategies ` includes:
* `RemoveDuplicates`
* `RemovePodsViolatingNodeAffinity`
* `RemovePodsViolatingNodeTaints`
* `RemovePodsHavingTooManyRestarts`
* `PodLifeTime`

`multi-pods-related-strategies` includes:
* `LowNodeUtilization`
* `RemovePodsViolatingInterPodAntiAffinity`.

Sorting pods in `single-pod-related-strategies` may be optional, but I think at least in `multi-pods-related-strategies` we should sort them by their priority before eviction starts.

ingvagabund:
Strategies like `PodLifeTime` are iterating through every node sequentially. Once a node is cleared, other node is processed. Given one can set maximum number of eviction in total, you would need to first go through every node, collect all the pods that can be evicted, sort all the pods based on priority and then start eviction. Also, together with max evictions per node, you would need to check if the upper bound is not reached (trivial by checking pod.NodeName). So the transition looks to be quite trivial. Also, given `LowNodeUtilization` has the code for classifying the pods, it's more or less just about making the code exported and writing a bunch of unit tests to validate the new constraint.

[...]

PR #322, opened 2020-06-10
Move sortPodsBasedOnPriority to pod util

lixiang233:
As #294  mentioned, we may need to sort pods in other strategies, so I moved this func to pod util.

### DIFF
--- a/pkg/descheduler/pod/pods.go
+++ b/pkg/descheduler/pod/pods.go
@@ -23,6 +23,7 @@ import (
 	"k8s.io/apimachinery/pkg/fields"
 	clientset "k8s.io/client-go/kubernetes"
 	"sigs.k8s.io/descheduler/pkg/utils"
+	"sort"
 )
 
 // ListPodsOnANode lists all of the pods on a node
@@ -67,4 +68,28 @@ func IsBurstablePod(pod *v1.Pod) bool {
 
 func IsGuaranteedPod(pod *v1.Pod) bool {
 	return utils.GetPodQOS(pod) == v1.PodQOSGuaranteed
+}
+
+// SortPodsBasedOnPriorityLowToHigh sorts pods based on their priorities from low to high.
+// If pods have same priorities, they will be sorted by QoS in the following order:
+// BestEffort, Burstable, Guaranteed
+func SortPodsBasedOnPriorityLowToHigh(pods []*v1.Pod) {
+	sort.Slice(pods, func(i, j int) bool {
+		if pods[i].Spec.Priority == nil && pods[j].Spec.Priority != nil {
+			return true
+		}
+		if pods[j].Spec.Priority == nil && pods[i].Spec.Priority != nil {
+			return false
+		}
+		if (pods[j].Spec.Priority == nil && pods[i].Spec.Priority == nil) || (*pods[i].Spec.Priority == *pods[j].Spec.Priority) {
+			if IsBestEffortPod(pods[i]) {
+				return true
+			}
+			if IsBurstablePod(pods[i]) && IsGuaranteedPod(pods[j]) {
+				return true
+			}
+			return false
+		}
+		return *pods[i].Spec.Priority < *pods[j].Spec.Priority
+	})
 }
--- a/pkg/descheduler/pod/pods_test.go
+++ b/pkg/descheduler/pod/pods_test.go
@@ -19,6 +19,7 @@ package pod
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"strings"
 	"testing"
 
@@ -27,6 +28,11 @@ import (
 	"k8s.io/client-go/kubernetes/fake"
 	core "k8s.io/client-go/testing"
 	"sigs.k8s.io/descheduler/test"
+)
+
+var (
+	lowPriority  = int32(0)
+	highPriority = int32(10000)
 )
 
 func TestListPodsOnANode(t *testing.T) {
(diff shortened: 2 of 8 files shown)


pkg/descheduler/pod/pods_test.go
```diff
@@ -29,6 +30,31 @@ import (
...
+func makeBurstablePod(pod *v1.Pod) {
+	pod.Spec.Containers[0].Resources.Limits = nil
+	pod.Spec.Containers[0].Resources.Limits = nil
+}
+
+func makeGuaranteedPod(pod *v1.Pod) {
+	pod.Spec.Containers[0].Resources.Limits[v1.ResourceCPU] = pod.Spec.Containers[0].Resources.Requests[v1.ResourceCPU]
+	pod.Spec.Containers[0].Resources.Limits[v1.ResourceMemory] = pod.Spec.Containers[0].Resources.Requests[v1.ResourceMemory]
+}
+
```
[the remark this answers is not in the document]
seanmalloy: @lixiang233 yes, in my opinion I think it would be good to have `makeGuaranteedPod`, `makeBurstablePod` and `makeBestEffortPod` in `test/test_utils.go`

seanmalloy:
/ok-to-test
/kind cleanup

@lixiang233 I think the code change look pretty good. I just have two questions.

Are you planning to update the `RemovePodsViolatingInterPodAntiAffinity` strategy to sort pods by priority in a follow up pull request?

You mentioned in #294 that the documentation in the [Pod Evictions](https://github.com/kubernetes-sigs/descheduler#pod-evictions) section of the README is not quite accurate because only the `LowNodeUtilization` strategy sorts pods. Can you update the README in this PR too?

lixiang233:
Changed test util and README.

I'm willing to work on sorting pods in `RemovePodsViolatingInterPodAntiAffinity` strategy. I'll probably update it and create a PR in the next week. @seanmalloy

pkg/descheduler/pod/pods.go
```diff
@@ -68,3 +69,27 @@ func IsBurstablePod(pod *v1.Pod) bool {
 func IsGuaranteedPod(pod *v1.Pod) bool {
 	return utils.GetPodQOS(pod) == v1.PodQOSGuaranteed
 }
+
+// SortPodsBasedOnPriorityLowToHigh sorts pods based on their priorities from low to high.
+// If pods have same priorities, they will be sorted by QoS in the following order:
+// BestEffort, Burstable, Guaranteed
+func SortPodsBasedOnPriorityLowToHigh(pods []*v1.Pod) {
```
ingvagabund, line 76: What about to have:
```
func PodPrioritySortLowToHigh(i, j int) bool {
	if pods[i].Spec.Priority == nil && pods[j].Spec.Priority != nil {
		return true
	}
	if pods[j].Spec.Priority == nil && pods[i].Spec.Priority != nil {
		return false
	}
	if (pods[j].Spec.Priority == nil && pods[i].Spec.Priority == nil) || (*pods[i].Spec.Priority == *pods[j].Spec.Priority) {
		if IsBestEffortPod(pods[i]) {
			return true
		}
		if IsBurstablePod(pods[i]) && IsGuaranteedPod(pods[j]) {
			return true
		}
		return false
	}
	return *pods[i].Spec.Priority < *pods[j].Spec.Priority
}
```
instead and replace all invocations of `SortPodsBasedOnPriorityLowToHigh` with `sort.Slice(pods, PodPrioritySortLowToHigh`? So the implementation is minimal.

lixiang233: This implementation looks like [less func](https://github.com/kubernetes/kubernetes/blob/e7ca64fbe16d0c4b6c7b36aecde9cd75042b2828/pkg/scheduler/framework/v1alpha1/interface.go#L214) of QueueSortPlugin, as  func in `Slice(slice interface{}, less func(i, j int)` only receives two int params, we may implement and use it like:

```
func PodComparePriorityLowToHigh(i, j pod) bool {...}

sort.Slice(pods, func(i, j int) bool {
        return PodComparePriorityLowToHigh(pod[i], pod[j])
}
```

That looks a little complex, besides I think `pod util` should provide a `sort func` and not only a `less func`, maybe we can provide both and refactor `sort func` like:
```
func PodComparePriorityLowToHigh(i, j pod) bool {...}

func SortPodsBasedOnPriorityLowToHigh(pods []*v1.Pod) {
        sort.Slice(pods, func(i, j int) bool {
                return PodComparePriorityLowToHigh(pod[i], pod[j])
        }
}
```
What do you think @ingvagabund ?

ingvagabund: Sorry about that. Yeah, you are right. Too much lambda calculus in my head :) We can elaborate on that later once we have more callers of the same "less" implementation for sorting.

pkg/descheduler/strategies/lownodeutilization_test.go
```diff
@@ -52,23 +50,6 @@ func setHighPriority(pod *v1.Pod)        { pod.Spec.Priority = &highPriority }
 func setLowPriority(pod *v1.Pod)         { pod.Spec.Priority = &lowPriority }
 func setNodeUnschedulable(node *v1.Node) { node.Spec.Unschedulable = true }
```
ingvagabund, line 51: There's nothing specific to LowNodeUtilization strategy about these helper functions. You can move all of them under `test/test_utils.go` as well.

lixiang233:
Updated, I think `low/high priority` should be defined by user, so only provide `SetPodPriority` in `testutil`, I'll squash these 2 commits later if it's ok. @ingvagabund

ingvagabund:
@lixiang233 can you put all the helpers function moved to the same commit?  Two commits are more preferable here: 1 for moving the helper functions, 1 for moving the sort function.

lixiang233:
Completed. @ingvagabund

ingvagabund:
@lixiang233 just a nit. Can you switch order of the commits so the helpers are not copied into pkg/descheduler/pod/pods_test.go and then removed in the next? Othwerwise, lgtm,

lixiang233:
switched. @ingvagabund

ingvagabund:
@lixiang233 perfect, thanks!!!
/lgtm

lixiang233:
I'm planning to work on `RemovePodsViolatingInterPodAntiAffinity` strategy and will commit recently. I think we can divide the whole process into two parts, one is for `multi-pods-related-strategies` the other is for `single-pod-related-strategies` because these two types of strategies use different sorting functions. So I prefer to complete the work of `multi-pods-related-strategies` in this PR and open another for future work of `single-pod-related-strategies`. 

What do you think @damemi  ?

damemi:
@lixiang233 that sounds good to me, feel free to add the changes for multi-pods in a new commit. Thanks!

lixiang233:
Changed `RemovePodsViolatingInterPodAntiAffinity` strategy and added a simple testcase for it. Please take a look @seanmalloy @ingvagabund @damemi

seanmalloy:
This looks good to me. Travis CI still seems to be flaky, so the e2e tests did not run. Someone should run the e2e tests on their laptop before this is merged.

seanmalloy:
/lgtm

I was able to run the e2e tests on my laptop using k8s 1.18.2 ...
```
make test-e2e
./test/run-e2e-tests.sh
=== RUN   TestE2E
--- PASS: TestE2E (31.29s)
    e2e_test.go:261: Eviction of pods starting
    e2e_test.go:261: Eviction of pods starting
=== RUN   TestDeschedulingInterval
--- PASS: TestDeschedulingInterval (0.02s)
PASS
ok  	sigs.k8s.io/descheduler/test/e2e	31.509s
```

PR #327, opened 2020-06-21
Update klog to v2

farah:
Updates klog to v2

Fixes #286

PR #328, opened 2020-06-23
Update To Go 1.14.4

seanmalloy:
The kubernetes project has been updated to use Go 1.14.4. See below pull
request.

https://github.com/kubernetes/kubernetes/pull/88638

After making the updates to Go 1.14 "make gen" no longer worked. The
file hack/tools.go had to be created to get "make gen" working with Go
1.14.


hack/tools.go
```diff
@@ -0,0 +1,22 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// This package imports things required by build scripts, to force `go mod` to see them as dependencies
+package tools
+
+import _ "k8s.io/code-generator"
```
seanmalloy, line 22: Without adding this file `make gen` fails with the below error ...
```
make gen
+ go build -o /Users/myuser/projects/go/src/sigs.k8s.io/descheduler/_output/bin/conversion-gen k8s.io/code-generator/cmd/conversion-gen
can't load package: package k8s.io/code-generator/cmd/conversion-gen: cannot find package "." in:
	/Users/myuser/projects/go/src/sigs.k8s.io/descheduler/vendor/k8s.io/code-generator/cmd/conversion-gen
make: *** [gen] Error 1
```

seanmalloy:
/hold

Two things ...

The Prow jobs need to be updated to use Go 1.14, so that `make verify-vendor` will pass. Also, let's hold on this until #327 is merged and I will rebase this PR.

hack/tools.go
damemi: I'm fine with this, like we talked about before this also lets us pin a code generator version so we don't get the random go.mod updates

seanmalloy:
/hold cancel
/assign @ingvagabund @damemi 

I'll back port the Go 1.14 update to the release-1.17 and release-1.18 branches after this PR is merged other wise the Prow CI tests will fail for those branches. Also, Go 1.13 will unsupported once Go 1.15 is released.

ingvagabund:
Preferable I'd rather wait until k8s v1.19 is cut and vendored. Though, given Prow jobs have been bumped to 1.14 and I think go1.14 is backward compatible with go1.13, I don't see a big deal of postponing the PR to get merged.

> I'll back port the Go 1.14 update to the release-1.17 and release-1.18

We might update the Prow jobs to still spin go1.13 for the branches.

> Also, Go 1.13 will be unsupported once Go 1.15 is released.

By k8s or Go upstream?

seanmalloy:
> Preferable I'd rather wait until k8s v1.19 is cut and vendored. Though, given Prow jobs have been bumped to 1.14 and I think go1.14 is backward compatible with go1.13, I don't see a big deal of postponing the PR to get merged.
> 
> > I'll back port the Go 1.14 update to the release-1.17 and release-1.18
> 
> We might update the Prow jobs to still spin go1.13 for the branches.
> 
> > Also, Go 1.13 will be unsupported once Go 1.15 is released.
> 
> By k8s or Go upstream?

By Go upstream.

seanmalloy:
> We might update the Prow jobs to still spin go1.13 for the branches.

Yeah ... we should probably do this. I guess my assumption was that the k/k repo would also be updating the Go version for 1.18 and 1.17 release banches, but maybe my assumption is wrong.

seanmalloy:
> opened a PR to test-infra trying to update to 1.14 here: https://github.com/kubernetes/test-infra/pull/18054

@damemi and @ingvagabund If we are going to hold on merging this PR to upgrade to Go 1.14 until k8s 1.19 then we should probably revert the test-infra PR.

damemi:
I think we could upgrade to go 1.14 on our master branch, since this *is* technically the 1.19 branch until we cut that release. Wdyt?

seanmalloy:
> I think we could upgrade to go 1.14 on our master branch, since this _is_ technically the 1.19 branch until we cut that release. Wdyt?

@damemi I'm good with having it merged.

ingvagabund:
> I think we could upgrade to go 1.14 on our master branch, since this is technically the 1.19 branch until we cut that release. Wdyt?
/lgtm

damemi:
> > We might update the Prow jobs to still spin go1.13 for the branches.
> 
> Yeah ... we should probably do this. I guess my assumption was that the k/k repo would also be updating the Go version for 1.18 and 1.17 release banches, but maybe my assumption is wrong.

I think spinning 1.13 for those branches is probably the best approach. By the time go 1.15 is released, we will probably not need to support tests for 1.18 and 1.17 (and thus go 1.13) anymore anyway, right? (I'm completely guessing, because the go release cycle seems longer than k8s but I don't know exactly what it is)

seanmalloy:
> > > We might update the Prow jobs to still spin go1.13 for the branches.
> > 
> > 
> > Yeah ... we should probably do this. I guess my assumption was that the k/k repo would also be updating the Go version for 1.18 and 1.17 release banches, but maybe my assumption is wrong.
> 
> I think spinning 1.13 for those branches is probably the best approach. By the time go 1.15 is released, we will probably not need to support tests for 1.18 and 1.17 (and thus go 1.13) anymore anyway, right? (I'm completely guessing, because the go release cycle seems longer than k8s but I don't know exactly what it is)

@damemi we can keep the release-1.17 and release-1.18 branches for descheduler using Go 1.13. I think the Go 1.15 release will be released in July or August. I think this PR can be merged. Someone will need to figure how to at least get the decheduler Prow jobs for the release-1.18 branch using Go 1.13.

I'll keep track of the Go version used in the k/k repo, and if k8s 1.18 gets updated to use Go 1.14 then I'll get the descheduler release-1.18 branch updated too.

seanmalloy:
> I'm completely guessing, because the go release cycle seems longer than k8s but I don't know exactly what it is)

Go releases twice a year roughly in February and August, https://golang.org/doc/devel/release.html.

PR #332, opened 2020-06-29
Add maxPodsToEvictPerNode to LowNodeUtilization testcase struct

lixiang233:
As mentioned in [comment](https://github.com/kubernetes-sigs/descheduler/blob/7ece10a643906874bd3377a36fe5d8641db95804/pkg/descheduler/strategies/lownodeutilization_test.go#L57), `expectedPodsEvicted` represents both expected result and evictor's limitation, it may make testcases pass unexpectedly, so I added another param `maxPodsToEvictPerNode` here. And I found two testcases are affected by this change, I gave an limitation to `without priorities`, and changed another's expected result.

### DIFF
--- a/pkg/descheduler/strategies/lownodeutilization_test.go
+++ b/pkg/descheduler/strategies/lownodeutilization_test.go
@@ -54,11 +54,9 @@ func TestLowNodeUtilization(t *testing.T) {
 		thresholds, targetThresholds api.ResourceThresholds
 		nodes                        map[string]*v1.Node
 		pods                         map[string]*v1.PodList
-		// TODO: divide expectedPodsEvicted into two params like other tests
-		// expectedPodsEvicted should be the result num of pods that this testCase expected but now it represents both
-		// MaxNoOfPodsToEvictPerNode and the test's expected result
-		expectedPodsEvicted int
-		evictedPods         []string
+		maxPodsToEvictPerNode        int
+		expectedPodsEvicted          int
+		evictedPods                  []string
 	}{
 		{
 			name: "no evictable pods",
@@ -114,7 +112,8 @@ func TestLowNodeUtilization(t *testing.T) {
 				},
 				n3NodeName: {},
 			},
-			expectedPodsEvicted: 0,
+			maxPodsToEvictPerNode: 0,
+			expectedPodsEvicted:   0,
 		},
 		{
 			name: "without priorities",
@@ -172,7 +171,8 @@ func TestLowNodeUtilization(t *testing.T) {
 				},
 				n3NodeName: {},
 			},
-			expectedPodsEvicted: 3,
+			maxPodsToEvictPerNode: 0,
+			expectedPodsEvicted:   4,
 		},
 		{
 			name: "without priorities stop when cpu capacity is depleted",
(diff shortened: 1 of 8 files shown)


ingvagabund:
If that's the case, it should be changed all over the place. Including renaming `maxPodsToEvict` field in `PodEvictor` to reflect it is per node constraint.

lixiang233:
> If that's the case, it should be changed all over the place. Including renaming `maxPodsToEvict` field in `PodEvictor` to reflect it is per node constraint.

Yeah, I think we should rename it to `maxPodsToEvictPerNode` all over the place. Do you think if it's ok to do this in another PR? Because this PR is for adding missing param to the struct. @ingvagabund

seanmalloy:
> > If that's the case, it should be changed all over the place. Including renaming `maxPodsToEvict` field in `PodEvictor` to reflect it is per node constraint.
> 
> Yeah, I think we should rename it to `maxPodsToEvictPerNode` all over the place. Do you think if it's ok to do this in another PR? Because this PR is for adding missing param to the struct. @ingvagabund 

From my perspective it would be better to clean it up in a single PR. The PR can contain multiple well thought out commits making it easy to review. I'll defer to @ingvagabund and @damemi for the final decision on this.

damemi:
@seanmalloy agree, distinct commits in this PR would make sense and we can review the whole change in context

@lixiang233 does that sound good?

lixiang233:
I've created a new commit for this change, please take a look. @seanmalloy  @damemi  @ingvagabund

pkg/descheduler/strategies/lownodeutilization_test.go
```diff
@@ -172,10 +171,11 @@ func TestLowNodeUtilization(t *testing.T) {
 				},
 				n3NodeName: {},
 			},
-			expectedPodsEvicted: 3,
+			maxPodsToEvictPerNode: 3,
```
ingvagabund, line 174: `maxPodsToEvictPerNode: 0` to disable the check for the maximum. It has no use here as the tests are checking if the strategy properly removes all pods that are violating the utilization constraints.

pkg/descheduler/strategies/lownodeutilization_test.go
```diff
@@ -54,14 +54,12 @@ func TestLowNodeUtilization(t *testing.T) {
...
-		// MaxNoOfPodsToEvictPerNode and the test's expected result
-		expectedPodsEvicted int
-		evictedPods         []string
+		maxPodsToEvictPerNode        int
+		expectedPodsEvicted          int
+		evictedPods                  []string
 	}{
 		{
-			name: "no evictable pods",
+			name: "no evictable pods -- Max Evict Pods Num 0",
```
ingvagabund, line 62: Also, no need to mention `.. Max Evict Pods Num ...`. `0` means the test for maximum number of pods to evict per node is not checked.

pkg/descheduler/strategies/lownodeutilization_test.go
lixiang233: I found 2 testcases which didn't evict all pods because of limitation on `maxEvictPodsNum`, I thought someone might want to test this, so I just remained one testcase unchanged and modified another testcase's limitation. And I do agree that testing of `maxEvictPodsNum` shouldn't be done here. I'll commit later

lixiang233:
I edited the first commit to remove the limitation on `maxPodsToEvictPerNode` and clean unused comments. @ingvagabund

PR #334, opened 2020-07-02
Adding replica check when evicting managed pods from node

### LINKED ISSUE #263: RemoveDuplicates should average out based on node count
When using `RemoveDuplicates`, if I have 5 nodes and a ReplicaSet of 10, 2 pods should be on each. descheduler instead continuously evicts half of them even though it will end up with the same distribution.

hanumanthan:
When pods managed by `deployment/replicaset/replicationcontroller` have replica count greater than available nodes, even after descheduler evicting these pods, they will end up in the same node.
This would just run in a loop and is ineffective.

This PR checks for the owners replication factor and decides whether to consider the pod as duplicate or not.

Fixes #263

### DIFF
--- a/pkg/descheduler/pod/pods.go
+++ b/pkg/descheduler/pod/pods.go
@@ -18,13 +18,14 @@ package pod
 
 import (
 	"context"
-	"fmt"
+	"sort"
+
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/klog/v2"
 	"sigs.k8s.io/descheduler/pkg/utils"
-	"sort"
 )
 
 // ListPodsOnANode lists all of the pods on a node
@@ -70,7 +71,9 @@ func GetPodOwnerReplicationCount(ctx context.Context, client clientset.Interface
 		}
 		return int(owner.Status.Replicas), nil
 	default:
-		return 0, fmt.Errorf("pod owned by owner %s kind %s non managed", ownerRef.Name, ownerRef.Kind)
+		klog.Infof("pod is non managed by RS or RC - owner name %s kind %s", ownerRef.Name, ownerRef.Kind)
+		// Returning default value as 1 as its a single pod non managed by a rc or rs
+		return 1, nil
 	}
 }
 
--- a/pkg/descheduler/pod/pods.go
+++ b/pkg/descheduler/pod/pods.go
@@ -18,6 +18,7 @@ package pod
 
 import (
 	"context"
+	"fmt"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
@@ -53,7 +54,7 @@ func ListPodsOnANode(ctx context.Context, client clientset.Interface, node *v1.N
 	return pods, nil
 }
 
-// GetPodOwnerReplicationCount returns the owner's replica count of the pod based on the owner reference
+// GetPodOwnerReplicationCount returns the pod owner's replica count
 func GetPodOwnerReplicationCount(ctx context.Context, client clientset.Interface, ownerRef metav1.OwnerReference) (int, error) {
 	switch ownerRef.Kind {
 	case "ReplicaSet":
--- a/pkg/descheduler/pod/pods.go
+++ b/pkg/descheduler/pod/pods.go
@@ -70,7 +70,7 @@ func GetPodOwnerReplicationCount(ctx context.Context, client clientset.Interface
 		}
 		return int(owner.Status.Replicas), nil
 	default:
-		klog.Infof("pod is non managed by RS or RC - owner name %s kind %s", ownerRef.Name, ownerRef.Kind)
+		klog.V(4).Infof("pod is non managed by RS or RC - owner name %s kind %s", ownerRef.Name, ownerRef.Kind)
 		// Returning default value as 1 as its a single pod non managed by a rc or rs
 		return 1, nil
 	}
--- a/pkg/descheduler/pod/pods.go
+++ b/pkg/descheduler/pod/pods.go
@@ -53,6 +53,26 @@ func ListPodsOnANode(ctx context.Context, client clientset.Interface, node *v1.N
 	return pods, nil
 }
 
+// GetPodOwnerReplicationCount returns the owner's replica count of the pod based on the owner reference
+func GetPodOwnerReplicationCount(ctx context.Context, client clientset.Interface, ownerRef metav1.OwnerReference) (int, error) {
+	switch ownerRef.Kind {
+	case "ReplicaSet":
+		owner, err := client.AppsV1().ReplicaSets(v1.NamespaceAll).Get(ctx, ownerRef.Name, metav1.GetOptions{})
+		if err != nil {
+			return 0, err
+		}
+		return int(owner.Status.Replicas), nil
+	case "ReplicationController":
+		owner, err := client.CoreV1().ReplicationControllers(v1.NamespaceAll).Get(ctx, ownerRef.Name, metav1.GetOptions{})
+		if err != nil {
+			return 0, err
+		}
+		return int(owner.Status.Replicas), nil
+	default:
+		return 0, fmt.Errorf("pod owned by owner %s kind %s non managed", ownerRef.Name, ownerRef.Kind)
+	}
+}
+
 // OwnerRef returns the ownerRefList for the pod.
 func OwnerRef(pod *v1.Pod) []metav1.OwnerReference {
 	return pod.ObjectMeta.GetOwnerReferences()
--- a/pkg/descheduler/pod/pods.go
+++ b/pkg/descheduler/pod/pods.go
@@ -18,14 +18,13 @@ package pod
 
 import (
 	"context"
-	"sort"
-
+	"fmt"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	clientset "k8s.io/client-go/kubernetes"
-	"k8s.io/klog/v2"
 	"sigs.k8s.io/descheduler/pkg/utils"
+	"sort"
 )
 
 // ListPodsOnANode lists all of the pods on a node
@@ -71,9 +70,7 @@ func GetPodOwnerReplicationCount(ctx context.Context, client clientset.Interface
 		}
 		return int(owner.Status.Replicas), nil
 	default:
-		klog.Infof("pod is non managed by RS or RC - owner name %s kind %s", ownerRef.Name, ownerRef.Kind)
-		// Returning default value as 1 as its a single pod non managed by a rc or rs
-		return 1, nil
+		return 0, fmt.Errorf("pod owned by owner %s kind %s non managed", ownerRef.Name, ownerRef.Kind)
 	}
 }
 
--- a/pkg/descheduler/pod/pods.go
+++ b/pkg/descheduler/pod/pods.go
@@ -18,11 +18,11 @@ package pod
 
 import (
 	"context"
-	"fmt"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/klog/v2"
 	"sigs.k8s.io/descheduler/pkg/utils"
 	"sort"
 )
@@ -70,7 +70,9 @@ func GetPodOwnerReplicationCount(ctx context.Context, client clientset.Interface
 		}
 		return int(owner.Status.Replicas), nil
 	default:
-		return 0, fmt.Errorf("pod owned by owner %s kind %s non managed", ownerRef.Name, ownerRef.Kind)
+		klog.Infof("pod is non managed by RS or RC - owner name %s kind %s", ownerRef.Name, ownerRef.Kind)
+		// Returning default value as 1 as its a single pod non managed by a rc or rs
+		return 1, nil
 	}
 }
 
--- a/pkg/descheduler/pod/pods_test.go
+++ b/pkg/descheduler/pod/pods_test.go
@@ -80,7 +80,7 @@ func TestGetPodOwnerReplicaSetReplicaCount(t *testing.T) {
 	expectedReplicaCount := 3
 
 	rs := test.BuildTestReplicaSet(replicasetName, 3)
-	ownerRef1 := test.GetReplicaSetOwnerRefList()
+	ownerRef1 := test.GetReplicaSetOwnerRefList("replicaset-1")
 	pod.ObjectMeta.OwnerReferences = ownerRef1
 
 	fakeClient := &fake.Clientset{}
--- a/pkg/descheduler/pod/pods_test.go
+++ b/pkg/descheduler/pod/pods_test.go
@@ -75,12 +75,11 @@ func TestListPodsOnANode(t *testing.T) {
 }
 
 func TestGetPodOwnerReplicaSetReplicaCount(t *testing.T) {
-	t.SkipNow()
 	pod := test.BuildTestPod("pod1", 100, 0, "n1", nil)
 	replicasetName := "replicaset-1"
 	expectedReplicaCount := 3
 
-	test.BuildTestReplicaSet(replicasetName, 3)
+	rs := test.BuildTestReplicaSet(replicasetName, 3)
 	ownerRef1 := test.GetReplicaSetOwnerRefList()
 	pod.ObjectMeta.OwnerReferences = ownerRef1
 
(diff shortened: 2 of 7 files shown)


pkg/descheduler/pod/pods_test.go
```diff
@@ -74,6 +74,58 @@ func TestListPodsOnANode(t *testing.T) {
...
+	pod.ObjectMeta.OwnerReferences = ownerRef1
+
+	fakeClient := &fake.Clientset{}
+	fakeClient.Fake.AddReactor("get", "replicasets", func(action core.Action) (bool, runtime.Object, error) {
+		name := action.(core.GetAction)
+		fieldString := name.GetName()
+		if strings.Contains(fieldString, replicasetName) {
+			// how to return int as runtime.Object ? pod.DeepCopy is just a placeholder
+			return true, pod.DeepCopy(), nil
+		}
```
hanumanthan, line 94: Need help here. the reactor action expects an type that implements `DeepCopy`.
To return RS or RC, where do I define and implement the `DeepCopy` methods

pkg/descheduler/strategies/duplicates_test.go
```diff
@@ -193,6 +221,13 @@ func TestFindDuplicatePods(t *testing.T) {
 		fakeClient.Fake.AddReactor("get", "nodes", func(action core.Action) (bool, runtime.Object, error) {
 			return true, node, nil
 		})
+		// //TODO how to pass replicaas as int in reactor action
+		// fakeClient.Fake.AddReactor("get", "replicationcontrollers", func(action core.Action) (bool, runtime.Object, error) {
+		// 	return true, node, nil
+		// })
+		// fakeClient.Fake.AddReactor("get", "replicaset", func(action core.Action) (bool, runtime.Object, error) {
+		// 	return true, nil, nil
+		// })
```
hanumanthan, line 230: Need help here. the reactor action expects an type that implements `DeepCopy`.
To return RS or RC, where do I define and implement the `DeepCopy` methods

pkg/descheduler/pod/pods_test.go
lixiang233: I think just rerutn the RS you created before using `test.BuildTestReplicaSet(replicasetName, 3)`, because `apps/v1/ReplicaSet` implements `runtime.Object`.

hanumanthan: ReplicationController hasnt implemented DeepCopy. Should I add it to this [file](https://github.com/kubernetes-sigs/descheduler/blob/61eef936183ecca959e34a4ba7940efb41b24a16/vendor/k8s.io/api/apps/v1/zz_generated.deepcopy.go)

`fakeClient.Fake.AddReactor("get", "replicationcontrollers", func(action core.Action) (bool, runtime.Object, error) {` - Is this syntax **equivalent of faking** `client.AppsV1().ReplicaSets(v1.NamespaceAll).Get(ctx, ownerRef.Name, metav1.GetOptions{})`

lixiang233: RC is under core/v1 and here is its [DeepCopy](https://github.com/kubernetes-sigs/descheduler/blob/61eef936183ecca959e34a4ba7940efb41b24a16/vendor/k8s.io/api/core/v1/zz_generated.deepcopy.go#L4260), besides, we shouldn't change autogenerated and vendor files. You can reference to [this project ](https://github.com/kubernetes/code-generator) which is used to generate these files.

pkg/descheduler/strategies/duplicates.go
```diff
@@ -67,11 +69,30 @@ func RemoveDuplicatePods(
 		// If any of the existing lists for that first key matches the current pod's list, the current pod is a duplicate.
 		// If not, then we add this pod's list to the list of lists for that key.
 		duplicateKeysMap := map[string][][]string{}
+	PodLoop:
 		for _, pod := range pods {
 			ownerRefList := podutil.OwnerRef(pod)
 			if hasExcludedOwnerRefKind(ownerRefList, strategy) {
 				continue
 			}
+			for _, ownerRef := range pod.OwnerReferences {
```
damemi, line 78: Could this just be done in the ownerRef loop that's already going through them below this? (L97)

hanumanthan: I have extracted that into a separate method for better readability.

damemi: My point is that the changes in https://github.com/kubernetes-sigs/descheduler/pull/334/files#diff-f2a21c2960284e0b92eea0f1afd7ea40R125-R138 could be done in the ownerref loop that is already there (https://github.com/kubernetes-sigs/descheduler/pull/334/files/5e089eeeededd3ccfc2f389d9c14e5be9eeb6a5d#diff-f2a21c2960284e0b92eea0f1afd7ea40R97), otherwise you're looping through the ownerrefs twice in a row. Breaking it into a separate function doesn't address that

test/test_utils.go
```diff
@@ -72,10 +73,45 @@ func GetNormalPodOwnerRefList() []metav1.OwnerReference {
...
+			Replicas: int32(replicas),
+		},
+	}
+	return rs
+}
+
 // GetReplicaSetOwnerRefList returns the ownerRef needed for replicaset pod.
-func GetReplicaSetOwnerRefList() []metav1.OwnerReference {
+func GetReplicaSetOwnerRefList(rsName string) []metav1.OwnerReference {
+	ownerRefList := make([]metav1.OwnerReference, 0)
```
hanumanthan, line 106: This change passes the name of replicaset so that i can re-use this method to create ownerrefs of different RS

pkg/descheduler/strategies/duplicates.go
hanumanthan: @damemi Fixed this

pkg/descheduler/pod/pods.go
```diff
@@ -53,6 +54,28 @@ func ListPodsOnANode(ctx context.Context, client clientset.Interface, node *v1.N
...
+		}
+		return int(owner.Status.Replicas), nil
+	case "ReplicationController":
+		owner, err := client.CoreV1().ReplicationControllers(v1.NamespaceAll).Get(ctx, ownerRef.Name, metav1.GetOptions{})
+		if err != nil {
+			return 0, err
+		}
+		return int(owner.Status.Replicas), nil
+	default:
+		klog.Infof("pod is non managed by RS or RC - owner name %s kind %s", ownerRef.Name, ownerRef.Kind)
```
seanmalloy, line 73: I wonder if this should be a debug message. Maybe change this to level 4.

```
klog.V(4).Infof("pod is non managed by RS or RC - owner name %s kind %s", ownerRef.Name, ownerRef.Kind)
```

pkg/descheduler/strategies/duplicates.go
```diff
@@ -67,13 +69,30 @@ func RemoveDuplicatePods(
...
 			}
+
 			podContainerKeys := make([]string, 0, len(ownerRefList)*len(pod.Spec.Containers))
 			for _, ownerRef := range ownerRefList {
+				var err error
+				replicaCount, ok := ownerReplicationCounter[ownerRef.Name]
+				if !ok {
+					replicaCount, err = podutil.GetPodOwnerReplicationCount(ctx, client, ownerRef)
+					if err != nil {
+						klog.Errorf("Error retreiving owner replica count %s", ownerRef.Name)
```
seanmalloy, line 86: Can you add the pod name and namespace to this log message?

pkg/descheduler/strategies/duplicates.go
```diff
@@ -67,13 +69,30 @@ func RemoveDuplicatePods(
...
+						continue
+					}
+					ownerReplicationCounter[ownerRef.Name] = replicaCount
+				}
+				// If required replicas of the managing owner is greater than available nodes, pods will be duplicated
+				// and shouldnt not be considered for eviction
+				if replicaCount > totalNodes {
+					continue PodLoop
```
seanmalloy, line 94: Can you add a debug log message above this line?

pkg/descheduler/pod/pods.go
```diff
@@ -53,6 +54,28 @@ func ListPodsOnANode(ctx context.Context, client clientset.Interface, node *v1.N
 	return pods, nil
 }
 
+// GetPodOwnerReplicationCount returns the pod owner's replica count
+func GetPodOwnerReplicationCount(ctx context.Context, client clientset.Interface, ownerRef metav1.OwnerReference) (int, error) {
+	switch ownerRef.Kind {
+	case "ReplicaSet":
+		owner, err := client.AppsV1().ReplicaSets(v1.NamespaceAll).Get(ctx, ownerRef.Name, metav1.GetOptions{})
```
lixiang233, line 61: Maybe better to use pod's namespace here?

pkg/descheduler/strategies/duplicates.go
```diff
@@ -67,13 +69,31 @@ func RemoveDuplicatePods(
...
+					replicaCount, err = podutil.GetPodOwnerReplicationCount(ctx, client, ownerRef)
+					if err != nil {
+						klog.Errorf("Error retreiving replica count for owner %s of %s pod in %s namespace", ownerRef.Name, pod.Name, pod.Namespace)
+						continue
+					}
+					ownerReplicationCounter[ownerRef.Name] = replicaCount
+				}
+				// If required replicas of the managing owner is greater than available nodes, pods will be duplicated
+				// and shouldnt not be considered for eviction
+				if replicaCount > totalNodes {
```
lixiang233, line 93: Evict duplicated pods only when their owner's `replicaCount` less than `node num` may make descheduler do nothing in some cases. Assume we have 3 nodes and 1 rs and its all 9 replicas lie on the same node, then each node are expected to have 3 replicas for best distribution, so descheduler should evict 6 duplicated pods, but as we only evict duplicated pods when  `replicaCount` less than `node num`, no pod will be evicted.

I think we should calculate the count of duplicated pods that every node expected to have(by `replicaCount` /`node num`), and evicte all duplicated pods above that count. This is exactly what  #263 wants to solve.

What do you think? @damemi  @ingvagabund  @seanmalloy

ingvagabund:
@hanumanthan can you please rebase your changes on top of the master branch? It's not easy to see what are your changes and what's already in the master branch when checking all changes at once.

pkg/descheduler/strategies/duplicates.go
ingvagabund: I agree with @lixiang233. The check makes the strategy unnecessary. The idea is to evict the duplicate pods to allow uniform distribution of pods among available nodes. Which will not be true if the check gets merged.

ingvagabund:
> When pods managed by deployment/replicaset/replicationcontroller have replica count greater than available nodes, even after descheduler evicting these pods, they will end up in the same node.
This would just run in a loop and is ineffective.

Not necessarily. There's no guarantee any evicted pods gets scheduled to its original node (unless there's only one node available and it is exactly the node from which the pod was evicted).

pkg/descheduler/pod/pods.go
seanmalloy: > Maybe better to use pod's namespace here?

Yeah ... I think this makes sense to me. Use the pods namespace instead of `v1.NamespacesAll`. @hanumanthan can you make this change?

pkg/descheduler/pod/pods.go
```diff
@@ -53,6 +54,28 @@ func ListPodsOnANode(ctx context.Context, client clientset.Interface, node *v1.N
...
+		}
+		return int(owner.Status.Replicas), nil
+	case "ReplicationController":
+		owner, err := client.CoreV1().ReplicationControllers(v1.NamespaceAll).Get(ctx, ownerRef.Name, metav1.GetOptions{})
+		if err != nil {
+			return 0, err
+		}
+		return int(owner.Status.Replicas), nil
+	default:
+		klog.V(4).Infof("pod is non managed by RS or RC - owner name %s kind %s", ownerRef.Name, ownerRef.Kind)
```
seanmalloy, line 73: Could the pod name and namespace be added to this log message?

pkg/descheduler/strategies/duplicates.go
seanmalloy: @hanumanthan let us know if you have any questions about the change that @ingvagabund and @lixiang233 are requesting. We twill ry to answer any questions you may have. Thanks!

hanumanthan: I am thinking of approaching this way @lixiang233 @ingvagabund 
Group pods by owner references, and for each `ownerRef` check if `len(duplicated_pods) > replicaCount/totalNodes`.
If yes, then delete the extras. 

Have a doubt - as pod's ownerRef is a list, what if a pod is duplicate for one owner and non-duplicate of another owner - do we evict. 


Capturing the idea in the code below - its not optimized but just for conveying the thought
```
for _, pod := range duplicatePods {
	for _, ownerRef := range ownerRefList {
		if podList, ok := ownerRefPodsMap[ownerRef.Name]; ok {
			ownerRefPodsMap[ownerRef.Name] = append(podList, pod)
		} else {
			ownerRefPodsMap[ownerRef.Name] = []*v1.Pod{pod}
		}
	}
}
podsToNotDelete:= make([]*v1.Pod)
for owner, pods := range ownerRefPodsMap {
	if count, ok := ownerReplicationCounter[owner]; ok {
		if count/totalNodes < len(pods) {
			podsToNotDelete := pods[:len(pods)-(count/totalNodes)]
		}
	} 
}

for _, pods := range podsToNotDelete {
	// remove this pod from duplicate pods
}

for _, pod := range duplicatePods {
	if _, err := podEvictor.EvictPod(ctx, pod, node, "RemoveDuplicatePods"); err != nil {
		klog.Errorf("Error evicting pod: (%#v)", err)
		break
	}
}
```

ingvagabund: > `count/totalNodes < len(pods)`

`count` (as number of replicas of an owner) will always be greater or equal to `len(pods)` (Unless you hit the case when replicas are getting scaled down). On the other side, `len(duplicatePods)` should be equal to the number of replicas - 1 (let's assume a case with a single owner and all pods generated from the same template), i.e. count -1. Which turns the condition into `count/totalNodes < count - 1`. Which turns into `count > totalNodes/(totalNodes-1)`. `totalNodes/(totalNodes-1)` is not greater than 2. So for any RC with 3 or more replicas the condition will always be true. Please correct me if my understanding is incorrect.

> Have a doubt - as pod's ownerRef is a list, what if a pod is duplicate for one owner and non-duplicate of another owner - do we evict.

Yes. The goal is to eliminate as many duplicates as possible (with hope of keeping the workload properly balanced).

Though, you might check how many owners see a pod as duplicated, how many does not. Then based on majority decide if the pod is to be evicted or not.

ingvagabund: I suggest to do the following:
- get all the duplicated pods
- evict only so many pods so the number of duplicated pods on a node is not greater than `owner.replicas`/`total nodes`
- optionally: Check how many owners see a pod as duplicated, how many does not. Then based on majority decide if the pod is to be evicted or not.

damemi: > > Have a doubt - as pod's ownerRef is a list, what if a pod is duplicate for one owner and non-duplicate of another owner - do we evict.
>
> Yes. The goal is to eliminate as many duplicates as possible (with hope of keeping the workload properly balanced).

Agree with this, our goal is not to be stringent in deciding what is a duplicate. If it matches in any way, it's a dupe to someone

hanumanthan: yes count here is `owner.totalReplicas`. 
`len(pods)`  is the length of duplicates pods in the node grouped by owner ref
```
for owner, pods := range ownerRefPodsMap {
		if ownerReplicaCount/totalNodes < len(pods) {
```
> evict only so many pods so the number of duplicated pods on a node is not greater than owner.replicas/total nodes
I was attempting this but my code example is just complicated to convey that. Let me refine it
1. Group duplicated pods by ownerRef - this will be of form `owner - list of duplicated pods`
2. eg) if replica count is 9 and 3 nodes - each node should have max 3 pods. Eliminate anything above - Pods to be eliminated - `len(duplicatedPods grouped by owner) - (replicas/totalNodes)+1` (+1 is because, this node has already one pod which wasnt considered as duplicate)

farah:
Isn't 15 commits for this PR  a little excessive? Probably squashing would be good for history.

farah:
Isn't 15 commits for this PR a little excessive? Probably squashing would be better for history.

seanmalloy:
Greetings @hanumanthan we just completed the descheduler v0.19.0(k8s v1.19) release cycle. We are starting to work on the features for the descheduler v0.20.0(k8s v1.20.0) release. Are you planning on continuing to work on this bug fix?

hanumanthan:
> Greetings @hanumanthan we just completed the descheduler v0.19.0(k8s v1.19) release cycle. We are starting to work on the features for the descheduler v0.20.0(k8s v1.20.0) release. Are you planning on continuing to work on this bug fix?

Hi @seanmalloy Unfortunately I wouldn't be able to pick this up for few weeks. If anyone is interested to proceed, happy to handover

farah:
@ingvagabund @seanmalloy Happy to take it, if you guys don't mind.

seanmalloy:
> @ingvagabund @seanmalloy Happy to take it, if you guys don't mind.

@farah please go ahead if you think you have time. Thanks!

farah:
Ok guys, it looks like I'm unable to edit this PR as I am not a maintainer. So I've created PR #398, I think we can close this one.

I have rebased the PR and squashed the commits, haven't addressed the reviews here yet though.

PR #336, opened 2020-07-07
avoid appending list multiple times in RemoveDuplicates

lixiang233:
Currently, appending `podContainerKeys` to `duplicateKeysMap[podContainerKeys[0]]` is inside the for loop, so it may append same keys multiple times.

### DIFF
--- a/pkg/descheduler/strategies/duplicates.go
+++ b/pkg/descheduler/strategies/duplicates.go
@@ -86,11 +86,15 @@ func RemoveDuplicatePods(
 
 			// If there have been any other pods with the same first "key", look through all the lists to see if any match
 			if existing, ok := duplicateKeysMap[podContainerKeys[0]]; ok {
+				matched := false
 				for _, keys := range existing {
 					if reflect.DeepEqual(keys, podContainerKeys) {
+						matched = true
 						duplicatePods = append(duplicatePods, pod)
 						break
 					}
+				}
+				if !matched {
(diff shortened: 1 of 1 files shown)


pkg/descheduler/strategies/duplicates.go
```diff
@@ -86,11 +86,15 @@ func RemoveDuplicatePods(
...
 			if existing, ok := duplicateKeysMap[podContainerKeys[0]]; ok {
+				matched := false
 				for _, keys := range existing {
 					if reflect.DeepEqual(keys, podContainerKeys) {
+						matched = true
 						duplicatePods = append(duplicatePods, pod)
 						break
 					}
+				}
+				if !matched {
```
ingvagabund, line 97: It's not a bug. Even though the list of keys is duplicated, it will never get read since `if reflect.DeepEqual(keys, podContainerKeys) {` has `break` keyword so the duplicated lists are never checked against.

The code change avoids duplication. Which will have almost zero impact on the performance unless the list of pods is quite large and there's many duplicates. I will leave this for @damemi to decide.

lixiang233: Yes, this change improves the performance obviously only in some extreme situations which may never happen in real world. 

I think the original design was: append `podContainerKeys` to the list only one time if there's no same keys in it, but the implementation has defect that will cause this operation happens multiple times, so I treat this as a bug.

seanmalloy: /assign @damemi

Issue #251, opened 2020-03-11
Descheduling for a specific namespace


semin-park:
I'm trying to create a descheduler for a specific namespace.

I changed the `clusterrolebinding` to a `rolebinding` and gave it a namespace.

But now the descheduling pod's log indicates an error saying

> E0311 09:30:04.764541       1 reflector.go:156] pkg/mod/k8s.io/client-go@v0.17.0/tools/cache/reflector.go:108: Failed to list *v1.Node: nodes is forbidden: User "system:serviceaccount:kube-system:descheduler-serviceaccount" cannot list resource "nodes" in API group "" at the cluster scope
E0311 09:30:04.839594       1 server.go:46] nodes is forbidden: User "system:serviceaccount:kube-system:descheduler-serviceaccount" cannot list resource "nodes" in API group "" at the cluster scope

What I would like to do, is to be able to create a descheduler for a specific namespace. Is this possible?

seanmalloy:
/kind feature

@semin-park it is currently not possible to do this with descheduler. There is no way to restrict it to a namespace. PR #202 is open for adding a pod label selector option. Would the pod label selector feature meet your requirements? Or do you still want to be able to target a specific namespace?

younseunghyun:
Hi
Operation strategy of Kubernetes can be divided into cluster per environment(prod, dev), namespace per environment.

As I know, many people use as namespace per environment.

At that point of strategy, developer want to use separate descheduler policies for each namespace.

So if we can restrict the namespace, it would be helpful to use.

Thx

semin-park:
I agree with @younseunghyun 

It would be nice if we could target a specific namespace. Currently the only way to go seems to be to create multiple clusters, but that will likely increase the external fragmentation at the node level and also increase maintaining costs.

In our case we have a `dev` and `prod` namespace, and it'd be great if we could have deschedulers with potentially different policies for each namespace :)

seanmalloy:
@younseunghyun and @semin-park thanks for your feedback. I think I just have some more questions.

1. is it acceptable to simply added a CLI option to limit descheduler to one or more namespaces?
2. Do you want to be able to specify more than one namespace?

younseunghyun:
Usage case of the descheduler used to be as cron job.

So, I think we can specify which namespaces to apply descheduler.

seanmalloy:
I think that adding a CLI option to restrict descheduler operation to one or more namespaces is a reasonable feature.

[...]

PR #338, opened 2020-07-11
Filter pods by namespaces

ingvagabund:
Fixes: #251

E2e for the exclude namespaces part are deleting all pods from all namespaces but `default` which are passing [IsEvictable](https://github.com/kubernetes-sigs/descheduler/blob/05c69ee26a9b878f93efe29815ae7b457dc7d26e/pkg/descheduler/evictions/evictions.go#L81). It can be tricky when the test is run over a cluster with pods that are not supposed to be evicted during the test. Though, I don't see any other way how to do it now. On the other hand, running the descheduler is always destructive so it needs to be run under certain assumptions/conditions.

With introduction of included namespaces one can configure the same strategy with with different conditions for eviction. E.g. evict pods in namespace N1 that are older than 1h but in namespace N2 only pods older than 2h.

The following strategies are subject to namespace filtering now:
- RemovePodsViolatingInterPodAntiAffinity
- RemovePodsViolatingNodeAffinity
- RemovePodsViolatingNodeTaints
- RemovePodsHavingTooManyRestarts
- PodLifeTime

Note: the policy config has backward incompatible changes for which we need to bump the version to v1alpha2

TODO:
- [x] add namespace filters
- [x] add validation of params
- [x] e2e for namespace filtering
- [x] docs

Built on top of https://github.com/kubernetes-sigs/descheduler/pull/343


ingvagabund:
@seanmalloy I diverged a bit from the proposal renaming `includeNamespaces` to just `include`. The same for `excludeNamespaces` as the keyword `Namespaces` is already mentioned in the tree:
```
namespaces:
  includeNamespaces:
  - ns1
  - ns2
  excludeNamespaces:
  - ns3
  - ns4
```

ingvagabund:
https://travis-ci.org/github/kubernetes-sigs/descheduler/builds/707654262 going green as well

ingvagabund:
@seanmalloy @damemi @lixiang233 PTAL

ingvagabund:
We need to have `make test-e2e` run through Prow before this gets merged: https://github.com/kubernetes/test-infra/pull/18292

damemi:
Maybe a close/reopen will pull in the new test
/close

damemi:
Opened https://github.com/kubernetes-sigs/descheduler/pull/340 to try and fix the e2e

pkg/api/types.go
```diff
@@ -44,13 +44,21 @@ type DeschedulerStrategy struct {
...
+}
+
 // Only one of its members may be specified
 type StrategyParameters struct {
 	NodeResourceUtilizationThresholds *NodeResourceUtilizationThresholds
 	NodeAffinityType                  []string
 	PodsHavingTooManyRestarts         *PodsHavingTooManyRestarts
 	MaxPodLifeTimeSeconds             *uint
 	RemoveDuplicates                  *RemoveDuplicates
+	Namespaces                        Namespaces
```
lixiang233, line 61: As `Namespaces` is not a new strategy, it's a bit inappropriate to place it inside `StrategyParameters`. Besides, as the comment writes, we should only specify one of these params, but in this case we may use both `Namespaces` and `RemoveDuplicates`(or other strategies). I think we should place `Namespaces` inside [DeschedulerStrategy](https://github.com/kubernetes-sigs/descheduler/blob/07f476dfc4d40207312b7d9e99c0f2c3a7394faf/pkg/api/types.go#L36).

pkg/descheduler/pod/pods.go
```diff
@@ -18,21 +18,91 @@ package pod
...
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	clientset "k8s.io/client-go/kubernetes"
 	"sigs.k8s.io/descheduler/pkg/utils"
-	"sort"
 )
 
+type Options struct {
+	filter             func(pod *v1.Pod) bool
```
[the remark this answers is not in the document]
ingvagabund: You can always create function which will call all filter functions:
```
func(pod *v1.Pod) bool {
	return filter1(pod) && ... filtern(pod)
}
```
and pass into `WithFilter`.

pkg/api/types.go
ingvagabund: `Namespaces` is just another parameter for a strategy so it belongs to `StrategyParameters`. Also, it's up to every strategy to decide which parameters it will read.

damemi: > Also, it's up to every strategy to decide which parameters it will read.

I agree with this, but I think that actually makes a case to put `Namespaces` in each strategy's struct (eg, [`RemoveDuplicates.Namespaces`](https://github.com/kubernetes-sigs/descheduler/blob/07f476dfc4d40207312b7d9e99c0f2c3a7394faf/pkg/api/types.go#L70-L72) for the strategies that will implement it.

For Namespaces, I think that's every strategy right? I know for pod selector we discussed that some strategies shouldn't implement it, in which case it shouldn't be available as an api option.

Also, I don't think we do anything to strictly enforce that only one of these StrategyParameters may be specified besides convention. Tbh in the next version of this api I would prefer to see `StrategyParameters` replaced with an interface or `object.Runtime` to get rid of this big list (similar to how the scheduler configures [`PluginArgs`](https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/apis/config/types.go#L236)

ingvagabund: Yeah, `// Only one of its members may be specified` comment might be no longer valid here.

pkg/api/types.go
```diff
@@ -44,13 +44,21 @@ type DeschedulerStrategy struct {
 	Params *StrategyParameters
 }
 
+// Namespaces carries a list of included/excluded namespaces
+// for which a given strategy is applicable
+type Namespaces struct {
+	Include []string
+	Exclude []string
```
damemi, line 51: Should note that these can't both be set (is that right? How do you interpret setting both include and exclude?)

pkg/api/v1alpha1/types.go
```diff
@@ -44,13 +44,22 @@ type DeschedulerStrategy struct {
 	Params *StrategyParameters `json:"params,omitempty"`
 }
 
+// Namespaces carries a list of included/excluded namespaces
+// for which a given strategy is applicable.
+// Included namespace takes preference over excluded namespaces (they are ignored).
```
damemi, line 49: again, probably just clearer to note that only one of these can be set and return an error rather than taking implicit preference

pkg/descheduler/pod/pods_test.go
```diff
@@ -67,7 +67,7 @@ func TestListPodsOnANode(t *testing.T) {
 			}
 			return true, nil, fmt.Errorf("Failed to list: %v", list)
 		})
-		pods, _ := ListPodsOnANode(context.TODO(), fakeClient, testCase.node, nil)
+		pods, _ := ListPodsOnANode(context.TODO(), fakeClient, testCase.node)
```
damemi, line 70: +1 nice, we can get rid of these `nil`s

pkg/descheduler/strategies/node_affinity.go
```diff
@@ -43,11 +43,18 @@ func RemovePodsViolatingNodeAffinity(ctx context.Context, client clientset.Inter
...
+				pods, err := podutil.ListPodsOnANode(
+					ctx,
+					client,
+					node,
+					podutil.WithFilter(func(pod *v1.Pod) bool {
+						return podEvictor.IsEvictable(pod) &&
+							!nodeutil.PodFitsCurrentNode(pod, node) &&
+							nodeutil.PodFitsAnyNode(pod, nodes)
+					}),
+					podutil.WithNamespaces(strategy.Params.Namespaces.Include),
```
damemi, line 55: Ah, I see now how the strategies can easily control whether they implement the include/exclude. Cool!

pkg/api/v1alpha1/types.go
ingvagabund: Yeah, I was thinking about introducing validation for strategy params

pkg/api/types.go
ingvagabund: > I agree with this, but I think that actually makes a case to put Namespaces in each strategy's struct (eg, RemoveDuplicates.Namespaces for the strategies that will implement it.

Done.

pkg/api/types.go
ingvagabund: Added validation in every strategy for which the namespace filtering applies

ingvagabund:
Allowing multiple configurations for the same strategy breaks the current v1alpha1 API. The same holds for moving `Namespaces` field into structs of individual strategies (`PodsHavingTooManyRestarts` nor `PodLifeTime` have their params as a dedicated struct, just direct fields in `StrategyParameters`). Let's discuss how to proceed here later once we (re)start discussion about v1alpha2 version of the policy.

pkg/api/types.go
```diff
@@ -44,13 +44,22 @@ type DeschedulerStrategy struct {
...
+
 // Only one of its members may be specified
 type StrategyParameters struct {
 	NodeResourceUtilizationThresholds *NodeResourceUtilizationThresholds
-	NodeAffinityType                  []string
-	PodsHavingTooManyRestarts         *PodsHavingTooManyRestarts
-	MaxPodLifeTimeSeconds             *uint
+	NodeAffinity                      *NodeAffinity
+	PodsHavingTooManyRestarts         []PodsHavingTooManyRestarts
+	PodLifeTime                       []PodLifeTime
```
damemi, line 59: @ingvagabund pointed out to me that this change (as I recommended in https://github.com/kubernetes-sigs/descheduler/pull/338#discussion_r454429921, adding `Namespaces` to each strategy's struct) would break the current config API since `NodeAffinityType` and `MaxPodLifeTimeSeconds` don't currently have their own structs.

So to avoid changing the API right at the end of a release, I take back what I said and think adding `Namespaces` to `StrategyParameters`, to get Namespace filtering working as planned for 1.19, is the simplest approach for now.

There have been a lot of ideas for how to change the api, so I would prefer to gather those up and target a rehaul in 1.20, where perhaps `Namespaces` can be set on each strategy's struct (or something similar)

test/e2e/e2e_test.go
```diff
@@ -177,6 +178,145 @@ func TestLowNodeUtilization(t *testing.T) {
...
+	}
+
+	t.Logf("Existing pods: %v", getPodNames(podList.Items))
+	t.Logf("set the strategy to delete pods from %v namespace", rc.Namespace)
+	runPodLifetimeStrategy(ctx, clientSet, nodeInformer, deschedulerapi.Namespaces{
+		Include: []string{rc.Namespace},
+	})
+
+	t.Logf("Waiting 10s")
+	time.Sleep(10 * time.Second)
```
damemi, line 279: duplicate `time.Sleep` (you also call it at the end of `runPodLifetimeStrategy`)

btw `Sleep`s can introduce flakes if for whatever reason the test doesn't run as smoothly as you assume

test/e2e/e2e_test.go
```diff
@@ -177,6 +178,145 @@ func TestLowNodeUtilization(t *testing.T) {
...
+
+	nodeInformer := sharedInformerFactory.Core().V1().Nodes()
+
+	testNamespace := &v1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: "e2e-" + strings.ToLower(t.Name())}}
+	if _, err := clientSet.CoreV1().Namespaces().Create(ctx, testNamespace, metav1.CreateOptions{}); err != nil {
+		t.Fatalf("Unable to create ns %v", testNamespace.Name)
+	}
+	defer clientSet.CoreV1().Namespaces().Delete(ctx, testNamespace.Name, metav1.DeleteOptions{})
+
+	rc := RcByNameContainer("test-rc-toomanyrestarts", testNamespace.Name, 5, map[string]string{"test": "toomanyrestarts"}, nil)
```
damemi, line 248: should this be `podlifetime` instead of `toomanyrestarts`?

test/e2e/e2e_test.go
```diff
@@ -177,6 +178,145 @@ func TestLowNodeUtilization(t *testing.T) {
...
+		}
+	}
+
+	t.Logf("set the strategy to delete pods from namespaces except the %v namespace", rc.Namespace)
+	runPodLifetimeStrategy(ctx, clientSet, nodeInformer, deschedulerapi.Namespaces{
+		Exclude: []string{rc.Namespace},
+	})
+
+	t.Logf("Waiting 10s")
+	time.Sleep(10 * time.Second)
```
damemi, line 302: another duplicate `Sleep`, though I'm not sure why you need to wait after running the strategy? Just to make sure the pods are deleted?

test/e2e/e2e_test.go
```diff
@@ -177,6 +178,145 @@ func TestLowNodeUtilization(t *testing.T) {
...
+	t.Logf("Existing pods: %v", getPodNames(podList.Items))
+
+	podNamesExclude := make(map[string]struct{})
+	// validate all pod were deleted
+	for _, pod := range podList.Items {
+		podNamesExclude[pod.Name] = struct{}{}
+		if _, exists := podNamesInclude[pod.Name]; exists {
+			t.Fatalf("Pod %q is expected to be deleted", pod.Name)
+		}
+	}
```
damemi, line 294: I don't understand your approach here, from what I can tell you're only creating 1 RC so all the pods should be in the same (included/excluded) namespace, right?

damemi: Oh, I see you're getting the new pods created by the RC after being deleted the first time, got it

damemi: I think the CI failures you have right now are flakes, because the error [lists 6 pods](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_descheduler/338/pull-descheduler-test-e2e-k8s-master-1-17/1285576269018173440#1:build-log.txt%3A273) and one of them is an previously existing pod, so maybe it's still terminating or something? So a wait around this until it passes might reduce that

test/e2e/e2e_test.go
ingvagabund: I removed the one from `runPodLifetimeStrategy`. Also, replaced the first Sleep(10) with wait.Pool since we are waiting until all pods get deleted.

test/e2e/e2e_test.go
ingvagabund: Right. Done.

test/e2e/e2e_test.go
ingvagabund: Here, I am waiting for some time (assuming the strategy finished running in that time and all pods that were supposed to be deleted got deleted). 10 seconds is just a guess. Though, at this point it's impossible to decide if a strategy run properly and did not deleted any pod. Or, if the strategy failed running for some reason but did not report any error.

test/e2e/e2e_test.go
ingvagabund: I increased the timeout to 20 seconds with pooling.

test/e2e/e2e_test.go
```diff
@@ -177,6 +179,174 @@ func TestLowNodeUtilization(t *testing.T) {
...
+	// All pods are supposed to be deleted, wait until all the old pods are deleted
+	if err := wait.PollImmediate(time.Second, 20*time.Second, func() (bool, error) {
+		podList, err := clientSet.CoreV1().Pods(rc.Namespace).List(ctx, metav1.ListOptions{LabelSelector: labels.SelectorFromSet(rc.Spec.Template.Labels).String()})
+		if err != nil {
+			return false, nil
+		}
+
+		includePodNames := getPodNames(podList.Items)
+		// validate all pod were deleted
+		if len(intersectStrings(initialPodNames, includePodNames)) > 0 {
```
damemi, line 300: you could consider using [`sets.String`](https://godoc.org/k8s.io/apimachinery/pkg/util/sets#String)

test/e2e/e2e_test.go
```diff
@@ -177,6 +179,174 @@ func TestLowNodeUtilization(t *testing.T) {
...
+			t.Logf("Waiting until %v pods get deleted", intersectStrings(initialPodNames, includePodNames))
+			// check if there's at least one pod not in Terminating state
+			for _, pod := range podList.Items {
+				if pod.DeletionTimestamp == nil {
+					t.Logf("Pod %v not in terminating state", pod.Name)
+					return false, nil
+				}
+			}
+			t.Logf("All %v pods are terminating", intersectStrings(initialPodNames, includePodNames))
+			return true, nil
```
damemi, line 310: don't think you need this, you have a return at the end of the function

test/e2e/e2e_test.go
```diff
@@ -177,6 +179,174 @@ func TestLowNodeUtilization(t *testing.T) {
...
+			return false, nil
+		}
+
+		includePodNames := getPodNames(podList.Items)
+		// validate all pod were deleted
+		if len(intersectStrings(initialPodNames, includePodNames)) > 0 {
+			t.Logf("Waiting until %v pods get deleted", intersectStrings(initialPodNames, includePodNames))
+			// check if there's at least one pod not in Terminating state
+			for _, pod := range podList.Items {
+				if pod.DeletionTimestamp == nil {
```
damemi, line 304: This is including the newly created pods, so I think you just want to iterate over `initialPods`? The errors in CI are failing against a pod that's not in your intersect list: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_descheduler/338/pull-descheduler-test-e2e-k8s-master-1-16/1285935763120197633#1:build-log.txt%3A272

```
     e2e_test.go:301: Waiting until [test-rc-podlifetime-9wqm7 test-rc-podlifetime-fvqt2 test-rc-podlifetime-hprj9 test-rc-podlifetime-q59m7 test-rc-podlifetime-vrdmk] pods get deleted
    e2e_test.go:305: Pod test-rc-podlifetime-8lh7l not in terminating state 
```

test/e2e/e2e_test.go
damemi: Ah, this is the exclude test. So you're testing that none of these pods get deleted. I guess there isn't really a condition we can Poll for then

test/e2e/e2e_test.go
ingvagabund: I split it into two tests, each one running the strategy over separate namespace.

test/e2e/e2e_test.go
ingvagabund: Removed

test/e2e/e2e_test.go
ingvagabund: As long as the slices are not too long, the naive implementation is fine.

test/e2e/e2e_test.go
damemi: ah good idea

aviorma:
where can i find this feature ? it is not exist in master at all

PR #340, opened 2020-07-13
Update e2e script to use Kind setup

damemi:
Trying to fix e2es after they were added to prow in https://github.com/kubernetes/test-infra/pull/18292

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -19,6 +19,7 @@ package e2e
 import (
 	"context"
 	"math"
+	"os"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


damemi:
/priority critical-urgent

merging is broken on fixing these tests

damemi:
Ok, I think I have this figured out now. I moved the `kind` setup we used with travis into the e2e runner script (guarded by a conditional `KIND_E2E` variable) in https://github.com/kubernetes-sigs/descheduler/pull/340/commits/f5e9f07321c0b6a8d2bc879f04f14b2cc05882aa#diff-2b318554e6bdadccef0f06e3f0a58555. If we merge https://github.com/kubernetes/test-infra/pull/18298, this should pass and unblock our e2es

seanmalloy:
/lgtm

This code change assumes that Linux is used and also assumes that the user can run sudo. This might break some people's local development environments. But it does fix the e2e tests in Prow and TravisCI can now be turned off too.

damemi:
> /lgtm
> 
> This code change assumes that Linux is used and also assumes that the user can run sudo. This might break some people's local development environments. But it does fix the e2e tests in Prow and TravisCI can now be turned off too.

The kind changes are optional, and only enabled by setting the `KIND_E2E` flag. So if anyone is currently just running `make test-e2e` they shouldn't see any changes

PR #343, opened 2020-07-14
Clean e2e test so it's easier to extend it

ingvagabund:
SSIA

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -20,12 +20,14 @@ import (
 	"context"
 	"math"
 	"os"
+	"strings"
 	"testing"
 	"time"
 
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/informers"
 	coreinformers "k8s.io/client-go/informers/core/v1"
 	clientset "k8s.io/client-go/kubernetes"
@@ -175,6 +177,7 @@ func TestE2E(t *testing.T) {
 		t.Errorf("Error creating deployment %v", err)
 	}
 	evictPods(ctx, t, clientSet, nodeInformer, nodes, rc)
+	deleteRC(ctx, t, clientSet, rc)
 
 	rc.Spec.Template.Annotations = map[string]string{"descheduler.alpha.kubernetes.io/evict": "true"}
 	rc.Spec.Replicas = func(i int32) *int32 { return &i }(15)
@@ -192,6 +195,7 @@ func TestE2E(t *testing.T) {
 		t.Errorf("Error creating deployment %v", err)
 	}
 	evictPods(ctx, t, clientSet, nodeInformer, nodes, rc)
+	deleteRC(ctx, t, clientSet, rc)
 }
 
 func TestDeschedulingInterval(t *testing.T) {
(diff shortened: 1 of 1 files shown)


test/e2e/e2e_test.go
```diff
@@ -99,85 +99,121 @@ func RcByNameContainer(name string, replicas int32, labels map[string]string, gr
...
+	rc := RcByNameContainer("test-rc-node-utilization", int32(15), map[string]string{"test": "node-utilization"}, nil)
+	if _, err := clientSet.CoreV1().ReplicationControllers(rc.Namespace).Create(ctx, rc, metav1.CreateOptions{}); err != nil {
+		t.Errorf("Error creating deployment %v", err)
+	}
+
+	evictPods(ctx, t, clientSet, nodeInformer, nodes, rc)
+	deleteRC(ctx, t, clientSet, rc)
+
+	// wait until rc is deleted
+	time.Sleep(5 * time.Second)
```
damemi, line 184: Instead of time.Sleep maybe this would be better with a wait.Poll that actually checks for when the RC is deleted?

test/e2e/e2e_test.go
```diff
@@ -187,11 +223,16 @@ func TestE2E(t *testing.T) {
...
+
+	if _, err := clientSet.CoreV1().ReplicationControllers(rc.Namespace).Create(ctx, rc, metav1.CreateOptions{}); err != nil {
 		t.Errorf("Error creating deployment %v", err)
 	}
+
 	evictPods(ctx, t, clientSet, nodeInformer, nodes, rc)
+	deleteRC(ctx, t, clientSet, rc)
+
+	// wait until rc is deleted
+	time.Sleep(5 * time.Second)
```
damemi, line 235: same as above

test/e2e/e2e_test.go
```diff
@@ -80,7 +79,8 @@ func RcByNameContainer(name string, replicas int32, labels map[string]string, gr
 			APIVersion: "v1",
 		},
 		ObjectMeta: metav1.ObjectMeta{
-			Name: name,
+			Name:      name,
+			Namespace: "default",
```
damemi, line 83: might be better to create/use/destroy a namespace just for the e2e

test/e2e/e2e_test.go
```diff
@@ -99,85 +99,121 @@ func RcByNameContainer(name string, replicas int32, labels map[string]string, gr
...
-	}
-
-	podEvictor := evictions.NewPodEvictor(
-		clientset,
-		evictionPolicyGroupVersion,
-		false,
-		0,
 		nodes,
-		false,
+		evictions.NewPodEvictor(
```
damemi, line 134: could you pass the podEvictor created in `evictPods` to this function and use that?

ingvagabund:
@damemi comments addressed PTAL

ingvagabund:
@seanmalloy @lixiang233 PTAL

test/e2e/e2e_test.go
```diff
@@ -98,86 +100,119 @@ func RcByNameContainer(name string, replicas int32, labels map[string]string, gr
...
 		nodes = append(nodes, &node)
 	}
-	sharedInformerFactory := informers.NewSharedInformerFactory(clientSet, 0)
-	nodeInformer := sharedInformerFactory.Core().V1().Nodes()
+
+	testNamespace := &v1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: "e2e-" + strings.ToLower(t.Name())}}
+	if _, err := clientSet.CoreV1().Namespaces().Create(ctx, testNamespace, metav1.CreateOptions{}); err != nil {
+		t.Fatalf("Unable to create ns %v", testNamespace.Name)
+	}
+	defer clientSet.CoreV1().Namespaces().Delete(ctx, testNamespace.Name, metav1.DeleteOptions{})
```
lixiang233, line 169: I think it's better to check namespace too given we have checked if pods/rc are successfully deleted.

ingvagabund: Failing to delete a namespace is not a blocker for running other tests. So right now it's the best-effort.
On the other hand keeping a RC not deleted might influence tests where strategies like `LowNodeUtilization` are ran. So the test has to fail in case a RC is not deleted.

lixiang233:
/lgtm
Namespace won't block other tests, so it's ok not to check it after deleted.

besides, these commits are so clear that I can easily understand the whole process of this PR :thumbsup: .

PR #354, opened 2020-07-21
Bump Helm chart to version 0.18.1

damemi:
This still represents Descheduler v0.18.0, but this is to make a small change
in order to try triggering the Helm chart-releaser action.


seanmalloy:
/lgtm

Let's try is and see if it works now. :-)

PR #360, opened 2020-07-24
Add NPD+CA autohealing use case to user guide

dharmab:
Add documentation on combining descheduler, NPD and CA to autoheal Node problems. This is a real-world use case in production at my organization. We have even written some custom NPD plugins specific to our internal infrastructure and observed those problems be autohealed with this setup.


dharmab:
Pushed new commit to fix broken link

seanmalloy:
/lgtm

@dharmab thanks for contributing this documentation! I'm glad to see that someone is seeing value in running the descheduler in production.

PR #361, opened 2020-07-28
remove unnecessary line feed  in log messages

jjmengze:
If we fix these spaces, there will be no misspell errors in go report

### DIFF
--- a/pkg/descheduler/strategies/lownodeutilization_test.go
+++ b/pkg/descheduler/strategies/lownodeutilization_test.go
@@ -561,11 +561,11 @@ func TestValidateStrategyConfig(t *testing.T) {
 
 		if validateErr == nil || testCase.errInfo == nil {
 			if validateErr != testCase.errInfo {
-				t.Errorf("expected validity of strategy config: thresholds %#v targetThresholds %#v\nto be %v but got %v instead",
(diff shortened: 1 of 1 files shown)


pkg/descheduler/strategies/lownodeutilization_test.go
```diff
@@ -561,11 +561,11 @@ func TestValidateStrategyConfig(t *testing.T) {
 
 		if validateErr == nil || testCase.errInfo == nil {
 			if validateErr != testCase.errInfo {
-				t.Errorf("expected validity of strategy config: thresholds %#v targetThresholds %#v\nto be %v but got %v instead",
+				t.Errorf("expected validity of strategy config: thresholds %#v targetThresholds %#v\n to be %v but got %v instead",
```
seanmalloy, line 564: Do you think the `\n` is needed? Since you are changing this error message maybe it would be a good idea to remove the `\n` too.

jjmengze: It seems to me `\n` is not needed, because the error message is not huge that it needs to be displayed in a new line.
so let's remove it

damemi:
@jjmengze looks good, small change but could you update the PR title to be a bit more descriptive please?

jjmengze:
@damemi 

Thank you very much for your generous reminder that I have revised the title.

PR #364, opened 2020-07-29
Allow custom priority threshold

### LINKED ISSUE #329: Allow Ignoring Pods For Eviction

I would like to be able to tell the descheduler to ignore certain pods for eviction. Currently the descheduler provides several options for ignoring pods, but none of the current options fit my use case.

**Option 1:**
I would like to be able to set the annotation `descheduler.alpha.kubernetes.io/evict: "false"` on a pod and have descheduler always ignore this pod. Descheduler should never evict this pod. Users would need to add the annotation `descheduler.alpha.kubernetes.io/evict: "true"` to inform the descheduler to always evict a pod. This would be a breaking change.

**Option 2:**
I would like descheduler to have an option to specify a PriorityClass threshold that allows excluding pods for eviction. This is similar to ignoring system and node critical pods, but the end user can specify a PriorityClass threshold instead of only allowing the hard coded system and node critical  PriorityClasses.

* Marking the pods as a critical pod
* Assigning local storage to the pod
* Setting a very restrictive PDB

Marking the pod as a [critical pod ](https://kubernetes.io/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods/)is the closest option that fits my use case, but the pods I want descheduler to ignore are not considered system or node critical. It is just a pod running a business application therefore I do not want to mark it as system or node critical.

descheduler version: v0.18.0

**Additional context**

The k8s cluster autoscaler currently supports a similar annotation for pods `"cluster-autoscaler.kubernetes.io/safe-to-evict": "false"`. See the [CA FAQ](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-types-of-pods-can-prevent-ca-from-removing-a-node) for details.

lixiang233:
Fixes #329 
Introduce a new configuration for each strategies which allows users to define a PriorityClass which is used by PodEvictor to check if a pod is evictable.

TODO:
- [x] Add PriorityClass to `StrategyParameters` and read and set it to PodEvictor
- [x] e2e tests for priority class
- [x] README
- [x] e2e tests for priority

### DIFF
--- a/pkg/descheduler/strategies/duplicates.go
+++ b/pkg/descheduler/strategies/duplicates.go
@@ -18,6 +18,7 @@ package strategies
 
 import (
 	"context"
+	"fmt"
 	"reflect"
 	"sort"
 	"strings"
@@ -34,6 +35,17 @@ import (
 	"sigs.k8s.io/descheduler/pkg/utils"
 )
 
+func validateRemoveDuplicatePodsParams(params *api.StrategyParameters) error {
+	if params == nil {
+		return nil
+	}
+	if params.ThresholdPriority != nil && params.ThresholdPriorityClassName != "" {
+		return fmt.Errorf("only one of thresholdPriority and thresholdPriorityClassName can be set")
+	}
+
+	return nil
+}
+
 // RemoveDuplicatePods removes the duplicate pods on node. This strategy evicts all duplicate pods on node.
 // A pod is said to be a duplicate of other if both of them are from same creator, kind and are within the same
 // namespace, and have at least one container with the same image.
@@ -45,10 +57,20 @@ func RemoveDuplicatePods(
 	nodes []*v1.Node,
 	podEvictor *evictions.PodEvictor,
 ) {
+	if err := validateRemoveDuplicatePodsParams(strategy.Params); err != nil {
+		klog.V(1).Info(err)
+		return
+	}
+	thresholdPriority, err := utils.GetPriorityFromStrategyParams(ctx, client, strategy.Params)
+	if err != nil {
+		klog.V(1).Infof("failed to get threshold priority from strategy's params: %#v", err)
+		return
+	}
+
 	for _, node := range nodes {
 		klog.V(1).Infof("Processing node: %#v", node.Name)
 		pods, err := podutil.ListPodsOnANode(ctx, client, node, podutil.WithFilter(func(pod *v1.Pod) bool {
-			return podEvictor.IsEvictable(pod, utils.SystemCriticalPriority)
+			return podEvictor.IsEvictable(pod, thresholdPriority)
 		}))
 		if err != nil {
 			klog.Errorf("error listing evictable pods on node %s: %+v", node.Name, err)
--- a/pkg/descheduler/strategies/duplicates.go
+++ b/pkg/descheduler/strategies/duplicates.go
@@ -31,6 +31,7 @@ import (
 	"sigs.k8s.io/descheduler/pkg/api"
 	"sigs.k8s.io/descheduler/pkg/descheduler/evictions"
 	podutil "sigs.k8s.io/descheduler/pkg/descheduler/pod"
+	"sigs.k8s.io/descheduler/pkg/utils"
 )
 
 // RemoveDuplicatePods removes the duplicate pods on node. This strategy evicts all duplicate pods on node.
@@ -46,7 +47,9 @@ func RemoveDuplicatePods(
 ) {
 	for _, node := range nodes {
 		klog.V(1).Infof("Processing node: %#v", node.Name)
-		pods, err := podutil.ListPodsOnANode(ctx, client, node, podutil.WithFilter(podEvictor.IsEvictable))
+		pods, err := podutil.ListPodsOnANode(ctx, client, node, podutil.WithFilter(func(pod *v1.Pod) bool {
+			return podEvictor.IsEvictable(pod, utils.SystemCriticalPriority)
+		}))
 		if err != nil {
 			klog.Errorf("error listing evictable pods on node %s: %+v", node.Name, err)
 			continue
--- a/pkg/descheduler/strategies/lownodeutilization.go
+++ b/pkg/descheduler/strategies/lownodeutilization.go
@@ -50,13 +50,28 @@ const (
 	MaxResourcePercentage = 100
 )
 
+func validateLowNodeUtilizationParams(params *api.StrategyParameters) error {
+	if params == nil || params.NodeResourceUtilizationThresholds == nil {
+		return fmt.Errorf("NodeResourceUtilizationThresholds not set")
+	}
+	if params.ThresholdPriority != nil && params.ThresholdPriorityClassName != "" {
+		return fmt.Errorf("only one of thresholdPriority and thresholdPriorityClassName can be set")
+	}
+
+	return nil
+}
+
 // LowNodeUtilization evicts pods from overutilized nodes to underutilized nodes. Note that CPU/Memory requests are used
 // to calculate nodes' utilization and not the actual resource usage.
 func LowNodeUtilization(ctx context.Context, client clientset.Interface, strategy api.DeschedulerStrategy, nodes []*v1.Node, podEvictor *evictions.PodEvictor) {
-	// todo: move to config validation?
 	// TODO: May be create a struct for the strategy as well, so that we don't have to pass along the all the params?
-	if strategy.Params == nil || strategy.Params.NodeResourceUtilizationThresholds == nil {
-		klog.V(1).Infof("NodeResourceUtilizationThresholds not set")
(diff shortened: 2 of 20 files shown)


lixiang233:
/cc @ingvagabund  @seanmalloy 
/assign @damemi 
Please take a look, this pr is ready for review and will be ready for merge once I have updated README.

pkg/descheduler/strategies/duplicates.go
```diff
@@ -44,6 +44,17 @@ func RemoveDuplicatePods(
...
+	var priorityClass string
+	if strategy.Params != nil {
+		priorityClass = strategy.Params.ThresholdPriorityClassName
+	}
+	priority, err := podEvictor.SetPriorityThresholdFromPriorityClass(ctx, client, priorityClass)
+	if err != nil {
+		klog.V(1).Infof("Error setting priority threshold, %#v", err)
+		return
+	}
+	klog.V(1).Infof("Running RemoveDuplicatePods with priority threshold %d", priority)
```
damemi, line 56: I think we should only log the threshold if it's being set to something that's not default.

pkg/descheduler/strategies/lownodeutilization.go
```diff
@@ -53,6 +53,17 @@ const (
...
 func LowNodeUtilization(ctx context.Context, client clientset.Interface, strategy api.DeschedulerStrategy, nodes []*v1.Node, podEvictor *evictions.PodEvictor) {
+	var priorityClass string
+	if strategy.Params != nil {
+		priorityClass = strategy.Params.ThresholdPriorityClassName
+	}
+	priority, err := podEvictor.SetPriorityThresholdFromPriorityClass(ctx, client, priorityClass)
+	if err != nil {
+		klog.V(1).Infof("Error setting priority threshold, %#v", err)
+		return
+	}
```
damemi, line 64: You could simplify this a bit like:
```go
if strategy.Params != nil {
  priority, err := podEvictor.SetPriorityThresholdFromPriorityClass(ctx, client, strategy.Params.ThresholdPriorityClassName)
  if err != nil {
    klog.V(1).Errorf("Error setting priority threshold: %+v", err)
    return
  }
  klog.V(1).Infof("Running LowNodeUtilization with priority threshold %d", priority)
}
```

Or just pass `*strategyParams` to `podEvictor.SetPriorityThresholdFromPriorityClass` and do the nil check there (and the log message below this if it's successfully set). That would clean up some of the copy paste in each strategy

pkg/descheduler/strategies/pod_lifetime.go
```diff
@@ -49,6 +49,13 @@ func PodLifeTime(ctx context.Context, client clientset.Interface, strategy api.D
 		klog.V(1).Info(err)
 		return
 	}
+	priorityClass := strategy.Params.ThresholdPriorityClassName
+	priority, err := podEvictor.SetPriorityThresholdFromPriorityClass(ctx, client, priorityClass)
```
damemi, line 53: any reason to not pass `strategy.Params.ThresholdPriorityClassName` directly to the function?

damemi:
btw when your PR is ready for review in the future just remove the [WIP] and we'll know it's ready :)
We can /hold if you're waiting for pieces like README updates that will depend on feedback
/retitle Allow custom priority threshold
/hold

lixiang233:
> We could also consider just adding this as a `filter` function 

@damemi  I prefer this approach, as `IsPodEvictableBasedOnPriority` will be called for every pod no matter it's in `IsEvictable` or the filter function, just moving `IsPodEvictableBasedOnPriority` out of `IsEvictable` won't affect the performance. and if we add it to `EvictPod`, we may do some useless calculations because some pods may not be evictable and we could filter out them before running the strategy.

At first I've thought of many ways to do this, given we'll set priority threshold every time a strategy starts and our strategies are running one by one, I think it doesn't matter if one strategy may affect others, besides I think we should do all `IsEvictable` checks together in one function, so I choose to make the threshold a param of `PodEvictor`. That's not a very good implementation, I'll reimplement it using the methods mentioned above and add another configuration to set the threshold value directly.

ingvagabund:
I second what @damemi says.
For strategies which are not cluster wide (RemovePodsViolatingInterPodAntiAffinity, RemovePodsViolatingNodeAffinity, RemovePodsViolatingNodeTaints, ...), you can extend the filters so you don't bother with pods what have higher priority class than ThresholdPriorityClassName. In which case the pod evictor does not have to check the priority class at all.
For the remaining strategies (RemoveDuplicates, LowNodeUtilization), you can always check a pod's class right before you pass the pod to the evictor. So the pod evictor can stay intact.

lixiang233:
As `All types of pods with the annotation descheduler.alpha.kubernetes.io/evict are evicted.` and this annotation is checked in `IsEvictable`, if I add the priority check as an individual filter function, and use them like:
```
podutil.WithFilter(func(pod *v1.Pod) bool {
      return podEvictor.IsEvictable(pod) && isPodEvictableByPriority(pod)
})
```
Although a pod has the annotation, if its priority is above priority threshold, it won't be evictable.

So if we want this annotation still work for priority filtering, I have some solutions:
1.add priority check to `IsEvictable` and refactor `IsEvictable` to have a parameter.
2.check this annotation again in `isPodEvictableByPriority`
3.make annotation check another filter func and call them like:
```
podutil.WithFilter(func(pod *v1.Pod) bool {
      return (podEvictor.IsEvictable(pod) && isPodEvictableByPriority(pod)) || isPodEvictableByAnnotation(pod)
})
```
I prefer 1, what do you think @damemi  @ingvagabund ?

lixiang233:
I have implemented this
> 1.add priority check to `IsEvictable` and refactor `IsEvictable` to have a parameter.

and added `ThresholdPriority` to strategy's params.

pkg/utils/pod.go
```diff
@@ -112,15 +112,12 @@ func IsCriticalPod(pod *v1.Pod) bool {
 	if IsMirrorPod(pod) {
 		return true
 	}
-	if pod.Spec.Priority != nil && IsCriticalPodBasedOnPriority(*pod.Spec.Priority) {
```
ingvagabund, line 115: This needs to stay here. You should never be allowed to evict critical pods.

damemi: Isn't this check basically replaced by IsPodEvictableBasedOnPriority?

lixiang233:
I've moved `IsPodEvictableBasedOnPriority` to the `evictions` package and added e2e testcase for `ThresholdPriority`.

please take a look @damemi  @ingvagabund

pkg/utils/pod.go
ingvagabund: Given the way it's implemented:

```
func IsPodEvictableBasedOnPriority(pod *v1.Pod, priority int32) bool {
	return pod.Spec.Priority == nil || *pod.Spec.Priority < priority
}
```

a user can choose any int32 value. Even the one greater than the system critical priority. Thus, if improperly configured the descheduler can evict even the critical pods. `return pod.Spec.Priority == nil || (*pod.Spec.Priority < priority && *pod.Spec.Priority < SystemCriticalPriority)` would do a bit better.

My point is `IsEvictable` function is supposed to be static. The hardline for deciding when a pod is evictable or not without any degree of flexibility. If `IsEvictable` decides a pod is not evictable, it is not. It's a safe guard condition that can not be changed from the outside (except for setting some true/false flags).

damemi: Good points, I did notice this change at first too but thought it was just being covered by the new code.

In that case, we should probably keep it the same as it was right?

damemi: @lixiang233 what do you think?

lixiang233: Agree, if someone wants to evict system-critical pods, he can add `descheduler.alpha.kubernetes.io/evict` annotation. I'll add `*pod.Spec.Priority < SystemCriticalPriority` to `IsPodEvictableBasedOnPriority`.

lixiang233:
I've completed the change mentioned before and added a validation to the threshold. @damemi  @ingvagabund

pkg/utils/priority.go
```diff
@@ -33,3 +39,37 @@ func PodMatchesTermsNamespaceAndSelector(pod *v1.Pod, namespaces sets.String, se
...
+// GetPriorityFromPriorityClass gets priority from the given priority class.
+// If no priority class is provided, it will return SystemCriticalPriority by default.
+func GetPriorityFromPriorityClass(ctx context.Context, client clientset.Interface, name string) (int32, error) {
+	if name != "" {
+		priorityClass, err := client.SchedulingV1().PriorityClasses().Get(ctx, name, metav1.GetOptions{})
+		if err != nil {
+			return 0, err
+		} else {
+			return priorityClass.Value, nil
+		}
```
seanmalloy, line 52:
```suggestion
		}
		return priorityClass.Value, nil
```

lixiang233: done.

PR #372, opened 2020-08-13
Redefine IsEvictable to be customizable for a particular strategy

ingvagabund:
Use WithXXX methods to extend the list of constraints to also
provide a reasonable error message.

### DIFF
--- a/pkg/descheduler/evictions/evictions.go
+++ b/pkg/descheduler/evictions/evictions.go
@@ -77,41 +77,6 @@ func NewPodEvictor(
 	}
 }
 
-// IsEvictable checks if a pod is evictable or not.
-func (pe *PodEvictor) IsEvictable(pod *v1.Pod, thresholdPriority int32) bool {
-	checkErrs := []error{}
-	if IsCriticalPod(pod) {
-		checkErrs = append(checkErrs, fmt.Errorf("pod is critical"))
-	}
-
-	if !IsPodEvictableBasedOnPriority(pod, thresholdPriority) {
-		checkErrs = append(checkErrs, fmt.Errorf("pod is not evictable due to its priority"))
-	}
-
-	ownerRefList := podutil.OwnerRef(pod)
-	if IsDaemonsetPod(ownerRefList) {
-		checkErrs = append(checkErrs, fmt.Errorf("pod is a DaemonSet pod"))
-	}
-
-	if len(ownerRefList) == 0 {
-		checkErrs = append(checkErrs, fmt.Errorf("pod does not have any ownerrefs"))
-	}
-
-	if !pe.evictLocalStoragePods && IsPodWithLocalStorage(pod) {
-		checkErrs = append(checkErrs, fmt.Errorf("pod has local storage and descheduler is not configured with --evict-local-storage-pods"))
-	}
-
-	if IsMirrorPod(pod) {
-		checkErrs = append(checkErrs, fmt.Errorf("pod is a mirror pod"))
-	}
-
-	if len(checkErrs) > 0 && !HaveEvictAnnotation(pod) {
-		klog.V(4).Infof("Pod %s in namespace %s is not evictable: Pod lacks an eviction annotation and fails the following checks: %v", pod.Name, pod.Namespace, errors.NewAggregate(checkErrs).Error())
-		return false
-	}
-	return true
-}
-
 // NodeEvicted gives a number of pods evicted for node
 func (pe *PodEvictor) NodeEvicted(node *v1.Node) int {
 	return pe.nodepodCount[node]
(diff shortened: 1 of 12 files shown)


pkg/descheduler/evictions/evictions.go
```diff
@@ -187,6 +152,66 @@ func evictPod(ctx context.Context, client clientset.Interface, pod *v1.Pod, poli
 	return err
 }
 
+type constraint func(pod *v1.Pod) error
+
+type evictable struct {
+	constraints []constraint
+}
+
+func (pe *PodEvictor) Evictable() *evictable {
```
lixiang233, line 161: I think we should add some comments to these public functions.

pkg/descheduler/evictions/evictions.go
```diff
@@ -187,6 +152,66 @@ func evictPod(ctx context.Context, client clientset.Interface, pod *v1.Pod, poli
...
+	}
+
+	if len(checkErrs) > 0 && !HaveEvictAnnotation(pod) {
+		klog.V(4).Infof("Pod %s in namespace %s is not evictable: Pod lacks an eviction annotation and fails the following checks: %v", pod.Name, pod.Namespace, errors.NewAggregate(checkErrs).Error())
+		return false
+	}
+	return true
+}
+
+func (ev *evictable) WithPriorityClassThreshold(priority int32) {
```
lixiang233, line 206: Maybe better to name it `WithPriorityThreshold` given its param is priority not priority class.

pkg/descheduler/evictions/evictions.go
```diff
@@ -187,6 +152,66 @@ func evictPod(ctx context.Context, client clientset.Interface, pod *v1.Pod, poli
...
 
+type constraint func(pod *v1.Pod) error
+
+type evictable struct {
+	constraints []constraint
+}
+
+func (pe *PodEvictor) Evictable() *evictable {
+	ev := &evictable{}
+	if !pe.evictLocalStoragePods {
```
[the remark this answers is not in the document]
damemi: that's not a bad idea, but have we seen any requests for this functionality yet? If it currently works for everyone as a cluster-wide option it might overcomplicate the api, and we need to consider that it would be a change from current behavior

pkg/descheduler/evictions/evictions.go
```diff
@@ -187,6 +152,66 @@ func evictPod(ctx context.Context, client clientset.Interface, pod *v1.Pod, poli
 	return err
 }
 
+type constraint func(pod *v1.Pod) error
+
+type evictable struct {
```
damemi, line 157: It seems odd that this is a private type but all the functions that work with it are public. Might be better as a public type, and then if we add more options to it in the future, it could be declared as a literal

I'm not sure on the name, I think something like `EvictOptions` might be clearer. wdyt?

pkg/descheduler/evictions/evictions.go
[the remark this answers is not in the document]
lixiang233: yeah, we may consider this change when we really need it.

pkg/descheduler/evictions/evictions.go
```diff
@@ -223,5 +248,5 @@ func IsPodWithLocalStorage(pod *v1.Pod) bool {
 
 // IsPodEvictableBasedOnPriority checks if the given pod is evictable based on priority resolved from pod Spec.
 func IsPodEvictableBasedOnPriority(pod *v1.Pod, priority int32) bool {
-	return pod.Spec.Priority == nil || (*pod.Spec.Priority < priority && *pod.Spec.Priority < utils.SystemCriticalPriority)
+	return pod.Spec.Priority == nil || *pod.Spec.Priority < priority
```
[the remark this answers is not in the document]
lixiang233: This check has been moved back to `IsCriticalPod` so that it'll be called every time we run `IsEvictable`, needn't keep it here.

pkg/descheduler/evictions/evictions.go
ingvagabund: That was the idea. Keep the actual implementation hidden and just rely on a single `IsEvictable` invocation. I have created `Options` struct instead of having `WithPriorityThreshold` as a method.

pkg/descheduler/evictions/evictions.go
```diff
@@ -187,6 +152,66 @@ func evictPod(ctx context.Context, client clientset.Interface, pod *v1.Pod, poli
...
+
+	if len(ownerRefList) == 0 {
+		checkErrs = append(checkErrs, fmt.Errorf("pod does not have any ownerrefs"))
+	}
+
+	if IsMirrorPod(pod) {
+		checkErrs = append(checkErrs, fmt.Errorf("pod is a mirror pod"))
+	}
+
+	for _, c := range append(ev.constraints) {
```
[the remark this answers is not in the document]
ingvagabund: Removed

pkg/descheduler/evictions/evictions.go
ingvagabund: Done

pkg/descheduler/strategies/duplicates.go
```diff
@@ -67,11 +67,12 @@ func RemoveDuplicatePods(
 		return
 	}
 
+	evictable := podEvictor.Evictable()
+	evictable.WithPriorityClassThreshold(thresholdPriority)
```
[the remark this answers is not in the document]
ingvagabund: Turned into `evictable := podEvictor.Evictable(evictions.WithPriorityThreshold(thresholdPriority))`

pkg/descheduler/evictions/evictions.go
```diff
@@ -187,6 +152,83 @@ func evictPod(ctx context.Context, client clientset.Interface, pod *v1.Pod, poli
 	return err
 }
 
+type Options struct {
+	priority *int32
+}
+
+// WithFilter sets a pod filter.
+// The filter function should return true if the pod should be returned from ListPodsOnANode
+func WithPriorityThreshold(priority int32) func(opts *Options) {
```
damemi, line 161: comments are for `WithFilter`

ingvagabund: Done

pkg/descheduler/evictions/evictions.go
ingvagabund: Done

lixiang233 (commented):
/lgtm

PR #380, opened 2020-08-19
Deprecate node-selector, max-pods-to-evict-per-node and evict-local-storage-pods flags and promote then to policy v1alpha1 fields

ingvagabund:
TBD
/hold


ingvagabund:
Keeping `descheduling-interval` flag as it is since it's not related to a particular strategy but to the overall controller.

seanmalloy:
@ingvagabund please update the output from `descheduler --help` here: https://github.com/kubernetes-sigs/descheduler/blob/master/docs/user-guide.md#cli-options. Other than that this looks good. Thanks!

Issue #381, opened 2020-08-20
Node not indentified as underutilized


auduny:
**What version of descheduler are you using?**

descheduler version: 0.18.0

**Does this issue reproduce with the latest release?**

Yes

**Which descheduler CLI options are you using?**

            - "--policy-config-file"
            - "/policy-dir/policy.yaml"
            - "--v"
            - "3"
            - "--node-selector"
            - "node-role.kubernetes.io/production=true"

**Please provide a copy of your descheduler policy config file**
```json
apiVersion: "descheduler/v1alpha1"
kind: "DeschedulerPolicy"
strategies:
  "RemoveDuplicates":
     enabled: true
  "RemovePodsViolatingInterPodAntiAffinity":
     enabled: true
  "LowNodeUtilization":
     enabled: true
     params:
       nodeResourceUtilizationThresholds:
         thresholds:
           "cpu" : 30
           "memory": 30
           "pods": 30
         targetThresholds:
           "cpu" : 50
           "memory": 50
           "pods": 70
```

**What k8s version are you using (`kubectl version`)?**

<details><summary><code>kubectl version</code> Output</summary><br><pre>
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.8", GitCommit:"9f2892aab98fe339f3bd70e3c470144299398ace", GitTreeState:"clean", BuildDate:"2020-08-14T11:09:22Z", GoVersion:"go1.14.7", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.7", GitCommit:"be3d344ed06bff7a4fc60656200a93c74f31f9a4", GitTreeState:"clean", BuildDate:"2020-02-11T19:24:46Z", GoVersion:"go1.13.6", Compiler:"gc", Platform:"linux/amd64"}
</pre></details>

**What did you do?**

Run the job once

**What did you expect to see?**

Atleast one node (k8s-node-vm-01.int.vgnett.no) is registred as underutilized.

**What did you see instead?**
```
I0820 10:43:23.889948       1 lownodeutilization.go:157] Node "k8s-node-vm-01.int.vgnett.no" is appropriately utilized with usage: api.ResourceThresholds{"cpu":6.962025316455696, "memory":7.106983524831375, "pods":6.363636363636363}
[...]

PR #384, opened 2020-08-21
Refactor low node utilization

ingvagabund:
Each time a pod's resource consumption is translated into a fraction (number in (0; 1) interval), some precision is lost. Instead, summing all resource consumption absolutely allows to compute the resource usage precisely (up to rounding losses when using Quantity type).

Storing resources consumption into a map of  Quantities also allows to simplify arguments of functions used in the code.

### DIFF
--- a/pkg/descheduler/strategies/lownodeutilization.go
+++ b/pkg/descheduler/strategies/lownodeutilization.go
@@ -102,7 +102,11 @@ func LowNodeUtilization(ctx context.Context, client clientset.Interface, strateg
 		getNodeUsage(ctx, client, nodes, thresholds, targetThresholds),
 		// The node has to be schedulable (to be able to move workload there)
 		func(node *v1.Node, usage NodeUsage) bool {
-			return !nodeutil.IsNodeUnschedulable(node) && isNodeWithLowUtilization(usage)
+			if nodeutil.IsNodeUnschedulable(node) {
+				klog.V(2).InfoS("Node is unschedulable, thus not considered as underutilized", "node", klog.KObj(node))
+				return false
+			}
+			return isNodeWithLowUtilization(usage)
 		},
 		func(node *v1.Node, usage NodeUsage) bool {
 			return isNodeAboveTargetUtilization(usage)
--- a/pkg/descheduler/strategies/lownodeutilization.go
+++ b/pkg/descheduler/strategies/lownodeutilization.go
@@ -33,11 +33,14 @@ import (
 	"sigs.k8s.io/descheduler/pkg/utils"
 )
 
-// NodeUsageMap stores a node's info, pods on it and its resource usage
-type NodeUsageMap struct {
+// NodeUsage stores a node's info, pods on it, thresholds and its resource usage
+type NodeUsage struct {
 	node    *v1.Node
-	usage   api.ResourceThresholds
+	usage   map[v1.ResourceName]*resource.Quantity
 	allPods []*v1.Pod
+
+	lowResourceThreshold  map[v1.ResourceName]*resource.Quantity
+	highResourceThreshold map[v1.ResourceName]*resource.Quantity
 }
 
 // NodePodsMap is a set of (node, pods) pairs
@@ -95,7 +98,16 @@ func LowNodeUtilization(ctx context.Context, client clientset.Interface, strateg
 		targetThresholds[v1.ResourceMemory] = MaxResourcePercentage
 	}
 
-	lowNodes, targetNodes := classifyNodes(ctx, client, nodes, thresholds, targetThresholds)
+	lowNodes, targetNodes := classifyNodes(
+		getNodeUsage(ctx, client, nodes, thresholds, targetThresholds),
+		// The node has to be schedulable (to be able to move workload there)
+		func(node *v1.Node, usage NodeUsage) bool {
+			return !nodeutil.IsNodeUnschedulable(node) && isNodeWithLowUtilization(usage)
+		},
+		func(node *v1.Node, usage NodeUsage) bool {
+			return isNodeAboveTargetUtilization(usage)
+		},
+	)
 
 	klog.V(1).Infof("Criteria for a node under utilization: CPU: %v, Mem: %v, Pods: %v",
 		thresholds[v1.ResourceCPU], thresholds[v1.ResourceMemory], thresholds[v1.ResourcePods])
@@ -131,7 +143,6 @@ func LowNodeUtilization(ctx context.Context, client clientset.Interface, strateg
 		ctx,
 		targetNodes,
 		lowNodes,
-		targetThresholds,
 		podEvictor,
 		evictable.IsEvictable)
 
@@ -180,10 +191,13 @@ func validateThresholds(thresholds api.ResourceThresholds) error {
 	return nil
 }
 
-// classifyNodes classifies the nodes into low-utilization or high-utilization nodes. If a node lies between
-// low and high thresholds, it is simply ignored.
-func classifyNodes(ctx context.Context, client clientset.Interface, nodes []*v1.Node, thresholds api.ResourceThresholds, targetThresholds api.ResourceThresholds) ([]NodeUsageMap, []NodeUsageMap) {
-	lowNodes, targetNodes := []NodeUsageMap{}, []NodeUsageMap{}
+func getNodeUsage(
+	ctx context.Context,
+	client clientset.Interface,
+	nodes []*v1.Node,
+	lowThreshold, highThreshold api.ResourceThresholds,
+) []NodeUsage {
+	nodeUsageList := []NodeUsage{}
 
 	for _, node := range nodes {
 		pods, err := podutil.ListPodsOnANode(ctx, client, node)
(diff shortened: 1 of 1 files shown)


ingvagabund:
@seanmalloy @damemi @lixiang233 PTAL
- changing the utilization computation from fractions (in percentages) to resources that are left instead
- refactoring the code to make it more generic (so it's more reusable when implementing HighNodeUtilization strategy)

pkg/descheduler/strategies/lownodeutilization.go
```diff
@@ -181,37 +192,75 @@ func validateThresholds(thresholds api.ResourceThresholds) error {
...
+	for _, nodeUsage := range nodeUsages {
+		if lowThresholdFilter(nodeUsage.node, nodeUsage) {
+			klog.V(2).Infof("Node %#v is under utilized with usage: %v", nodeUsage.node.Name, nodeUsage)
+			lowNodes = append(lowNodes, nodeUsage)
+		} else if highThresholdFilter(nodeUsage.node, nodeUsage) {
+			klog.V(2).Infof("Node %#v is over utilized with usage: %v", nodeUsage.node.Name, nodeUsage)
+			highNodes = append(highNodes, nodeUsage)
 		} else {
-			klog.V(2).Infof("Node %#v is appropriately utilized with usage: %#v", node.Name, usage)
+			klog.V(2).Infof("Node %#v is appropriately utilized with usage: %v", nodeUsage.node.Name, nodeUsage)
```
lixiang233, line 251: As mentioned in #381 , can we print a log message when an `underutilized` node is `unschedulable`?

pkg/descheduler/strategies/lownodeutilization.go
```diff
@@ -271,50 +316,57 @@ func evictPodsFromTargetNodes(
...
+				totalAvailableUsage[v1.ResourceCPU].Sub(cpuQuantity)
+
+				memoryQuantity := utils.GetResourceRequestQuantity(pod, v1.ResourceMemory)
+				nodeUsage.usage[v1.ResourceMemory].Sub(memoryQuantity)
+				totalAvailableUsage[v1.ResourceMemory].Sub(memoryQuantity)
 
-				// update remaining memory
-				*totalMem -= float64(mUsage)
-				nodeUsage[v1.ResourceMemory] -= api.Percentage(float64(mUsage) / float64(nodeCapacity.Memory().Value()) * 100)
+				nodeUsage.usage[v1.ResourcePods].Sub(*resource.NewQuantity(1, resource.DecimalSI))
```
lixiang233, line 365: You forgot to sub `totalAvailableUsage[v1.ResourcePods]` here?

ingvagabund: Nice catch, thanks!!!

seanmalloy:
I don't see anything wrong with these changes. @lixiang233 please review again when you have some time. Thanks!

pkg/descheduler/strategies/lownodeutilization.go
ingvagabund: Addressed in classifyNodes invocation in the low threshold filter.

seanmalloy:
/lgtm

@lixiang233 please review one more time when you get a chance. Thanks!

lixiang233:
@seanmalloy  this change looks pretty good to me.

pkg/descheduler/strategies/lownodeutilization.go
```diff
@@ -95,8 +98,20 @@ func LowNodeUtilization(ctx context.Context, client clientset.Interface, strateg
...
 	}
 
-	npm := createNodePodsMap(ctx, client, nodes)
-	lowNodes, targetNodes := classifyNodes(npm, thresholds, targetThresholds)
+	lowNodes, targetNodes := classifyNodes(
+		getNodeUsage(ctx, client, nodes, thresholds, targetThresholds),
+		// The node has to be schedulable (to be able to move workload there)
+		func(node *v1.Node, usage NodeUsage) bool {
+			if nodeutil.IsNodeUnschedulable(node) {
+				klog.V(2).Infof("Node %#v is unschedulable, thus not considered as underutilized", node.Name)
```
seanmalloy, line 106: @ingvagabund can you change this log statement to use `InfoS` instead of `Infof`?

seanmalloy (changes requested):

pkg/descheduler/strategies/lownodeutilization.go
seanmalloy: I noticed this block of commented code here:
https://github.com/kubernetes-sigs/descheduler/blob/08b2dffa4224768e07bc568adeb43704de5473d5/pkg/descheduler/node/node.go#L90-L94

I did not research any further to see where the `IsReady` function is called from.

ingvagabund: Through https://github.com/kubernetes-sigs/descheduler/blob/08b2dffa4224768e07bc568adeb43704de5473d5/pkg/descheduler/descheduler.go#L98. Though, that code is supposed to be copy paste from the main and changed only when really necessary.

ingvagabund: Done

pkg/descheduler/strategies/lownodeutilization.go
```diff
@@ -95,8 +98,20 @@ func LowNodeUtilization(ctx context.Context, client clientset.Interface, strateg
...
 	}
 
-	npm := createNodePodsMap(ctx, client, nodes)
-	lowNodes, targetNodes := classifyNodes(npm, thresholds, targetThresholds)
+	lowNodes, targetNodes := classifyNodes(
+		getNodeUsage(ctx, client, nodes, thresholds, targetThresholds),
+		// The node has to be schedulable (to be able to move workload there)
+		func(node *v1.Node, usage NodeUsage) bool {
+			if nodeutil.IsNodeUnschedulable(node) {
+				klog.V(2).InfoS("Node is unschedulable, thus not considered as underutilized", "node", node.Name)
```
seanmalloy, line 106:
```suggestion
				klog.V(2).InfoS("Node is unschedulable, thus not considered as underutilized", "node", klog.KObj(node))
```

ingvagabund: In case of nodes `klog.KObj(node)` is equivalent to `node.Name`

ingvagabund:
@seanmalloy done PTAL

damemi (commented):
@ingvagabund could you add some more description to the commits/PR on what the goal is with this refactor, for future reference?

ingvagabund:
> @ingvagabund could you add some more description to the commits/PR on what the goal is with this refactor, for future reference?

Description updated. Lemme know whether it's sufficient.

pkg/descheduler/strategies/lownodeutilization.go
```diff
@@ -324,70 +379,45 @@ func evictPods(
...
-			if name == v1.ResourceCPU || name == v1.ResourceMemory || name == v1.ResourcePods {
-				ti += value
-			}
-		}
-		for name, value := range nodes[j].usage {
-			if name == v1.ResourceCPU || name == v1.ResourceMemory || name == v1.ResourcePods {
-				tj += value
-			}
-		}
+		ti := nodes[i].usage[v1.ResourceMemory].Value() + nodes[i].usage[v1.ResourceCPU].MilliValue() + nodes[i].usage[v1.ResourcePods].Value()
```
damemi, line 384: do these still make sense to add together now that you're storing absolute values instead of percentages? Like if I have a node with `500 pods, 1 mem, 1 cpu` that will be ranked higher usage than `10 pods, 100 mem, 100 cpu` right? But idk if that's a reasonable comparison

damemi (commented):
This is definitely one of the more complex strategies and I'm thankful for the improvements to quality and readability @ingvagabund. Just a couple comments to make sure my understanding is correct

pkg/descheduler/strategies/lownodeutilization.go
```diff
@@ -95,8 +98,20 @@ func LowNodeUtilization(ctx context.Context, client clientset.Interface, strateg
...
+		func(node *v1.Node, usage NodeUsage) bool {
+			if nodeutil.IsNodeUnschedulable(node) {
+				klog.V(2).InfoS("Node is unschedulable, thus not considered as underutilized", "node", klog.KObj(node))
+				return false
+			}
+			return isNodeWithLowUtilization(usage)
+		},
+		func(node *v1.Node, usage NodeUsage) bool {
+			return isNodeAboveTargetUtilization(usage)
+		},
```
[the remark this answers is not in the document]
ingvagabund: It's a preparation for HighNodeUtilization strategy where the upper/lower threshold conditions are different

pkg/descheduler/strategies/lownodeutilization.go
```diff
@@ -181,38 +196,75 @@ func validateThresholds(thresholds api.ResourceThresholds) error {
...
-		if !nodeutil.IsNodeUnschedulable(node) && isNodeWithLowUtilization(usage, thresholds) {
-			klog.V(2).InfoS("Node is underutilized", "node", klog.KObj(node), "usage", usage)
-			lowNodes = append(lowNodes, nuMap)
-		} else if isNodeAboveTargetUtilization(usage, targetThresholds) {
-			klog.V(2).InfoS("Node is overutilized", "node", klog.KObj(node), "usage", usage)
-			targetNodes = append(targetNodes, nuMap)
-		} else {
-			klog.V(2).InfoS("Node is appropriately utilized", "node", klog.KObj(node), "usage", usage)
+			lowResourceThreshold: map[v1.ResourceName]*resource.Quantity{
+				v1.ResourceCPU:    resource.NewMilliQuantity(int64(float64(lowThreshold[v1.ResourceCPU])*float64(nodeCapacity.Cpu().MilliValue())*0.01), resource.DecimalSI),
```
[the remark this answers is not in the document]
ingvagabund: Commented few lines up

pkg/descheduler/strategies/lownodeutilization.go
```diff
@@ -181,38 +196,75 @@ func validateThresholds(thresholds api.ResourceThresholds) error {
...
 }
 
 // evictPodsFromTargetNodes evicts pods based on priority, if all the pods on the node have priority, if not
 // evicts them based on QoS as fallback option.
 // TODO: @ravig Break this function into smaller functions.
 func evictPodsFromTargetNodes(
 	ctx context.Context,
-	targetNodes, lowNodes []NodeUsageMap,
+	targetNodes, lowNodes []NodeUsage,
 	targetThresholds api.ResourceThresholds,
```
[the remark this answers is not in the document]
ingvagabund: Done

pkg/descheduler/strategies/lownodeutilization.go
ingvagabund: I kept the original computation. Though, it might be better to rather carry addition over relative units. E.g. pods/totalpods + cpu/totalcpu + memory/totalmemory. So nodes with higher percentage of consumed resources have higher priority over those with lower percentage. As a feature we might allow to set weight for each resource. E.g. in case memory is more expensive than cpu.

seanmalloy:
@damemi please take a look when you have some time. I believe @ingvagabund has addressed all of your review feedback.

/lgtm

damemi (approved):
/approve
Thanks @ingvagabund !

PR #371, opened 2020-08-12
Update To Go 1.15.0

seanmalloy:
As part of the k8s 1.19 release cycle the Go version is being bumped to
1.15.0. Updating the descheduler to use Go 1.15 prior to the descheduler
v0.19.0 release.

See below issues for reference:
* https://github.com/kubernetes/release/issues/1421
* https://github.com/kubernetes/kubernetes/issues/93484

PR #385, opened 2020-08-25
LowNodeUtilization: use clientset in testing, drop all custom reactors

ingvagabund:
SSIA

### DIFF
--- a/pkg/descheduler/strategies/lownodeutilization_test.go
+++ b/pkg/descheduler/strategies/lownodeutilization_test.go
@@ -25,11 +25,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/api/policy/v1beta1"
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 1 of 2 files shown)


seanmalloy:
/lgtm
/assign @damemi 
/kind cleanup

This is required for #371

seanmalloy:
@damemi this PR will unblock the update to Go 1.15.

pkg/descheduler/strategies/lownodeutilization_test.go
```diff
@@ -803,29 +803,21 @@ func TestWithTaints(t *testing.T) {
 				objs = append(objs, pod)
 			}
 
-			fakePtr := newFake(objs...)
```
damemi, line 806: Is `newFake` called anywhere else? I don't see it used in this file anymore so we can remove it then right?

ingvagabund:
`newFake` function removed
/hold cancel

PR #398, opened 2020-09-11
Adding replica check when evicting managed pods from node

### LINKED ISSUE #263: RemoveDuplicates should average out based on node count

farah:
Overriding PR #334 

> When pods managed by deployment/replicaset/replicationcontroller have replica count greater than available nodes, even after descheduler evicting these pods, they will end up in the same node.
> This would just run in a loop and is ineffective.
> 
> This PR checks for the owners replication factor and decides whether to consider the pod as duplicate or not.
> 
> Fixes #263

### DIFF
--- a/pkg/descheduler/pod/pods.go
+++ b/pkg/descheduler/pod/pods.go
@@ -20,11 +20,14 @@ import (
 	"context"
 	"sort"
 
+	"fmt"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/klog/v2"
 	"sigs.k8s.io/descheduler/pkg/utils"
+	"sort"
 )
 
 type Options struct {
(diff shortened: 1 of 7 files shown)


seanmalloy:
@farah please let us know when this is ready for review. Thanks!

pkg/descheduler/strategies/duplicates.go
```diff
@@ -92,13 +94,31 @@ func RemoveDuplicatePods(
...
+				if !ok {
+					replicaCount, err = podutil.GetPodOwnerReplicationCount(ctx, client, ownerRef)
+					if err != nil {
+						klog.Errorf("Error retreiving replica count for owner %s of %s pod in %s namespace", ownerRef.Name, pod.Name, pod.Namespace)
+						continue
+					}
+					ownerReplicationCounter[ownerRef.Name] = replicaCount
+				}
+				// If required replicas of the managing owner is greater than available nodes, pods will be duplicated
+				// and shouldnt not be considered for eviction
```
ZongqiangZhang, line 117: `shouldnt` typo?

damemi:
@farah is this PR still a draft? Please switch it when ready for review, thanks!

farah:
@damemi I'll get to this very soon, thanks.

farah:
I'm still actively investigating this...

damemi:
Thanks for the updates @farah! It might be good to try to sync with @vadv because their PR (https://github.com/kubernetes-sigs/descheduler/pull/420) seems to be working on the same problem with a different approach. Perhaps they could be combined?

pkg/descheduler/pod/pods.go
```diff
@@ -130,6 +133,28 @@ func ListPodsOnANode(
...
+	default:
+		klog.V(4).Infof("pod is non managed by RS or RC - owner name %s kind %s", ownerRef.Name, ownerRef.Kind)
+		// Returning default value as 1 as its a single pod non managed by a rc or rs
+		return 1, nil
+	}
+}
+
 // OwnerRef returns the ownerRefList for the pod.
 func OwnerRef(pod *v1.Pod) []metav1.OwnerReference {
 	return pod.ObjectMeta.GetOwnerReferences()
```
farah, line 160: hi @ingvagabund  quick question, what does it mean for a pod to have multiple OwnerReferences? is this possible? This method  returns `[]OwnerReference`. Was wondering  if can  assume it will always have length 1, so to avoid a loop. From the test cases it seems there could only be 1.

farah:
@damemi yes I'm investigating both methods to see if that's possible/necessary.

pkg/descheduler/pod/pods.go
farah: nevermind i found the answer

damemi (commented):
Hi @farah, thanks for your work on this! However we have ended up going with the changes in https://github.com/kubernetes-sigs/descheduler/pull/463 to address this.

Thank you again for your efforts, but since this issue has been resolved I will close this PR.
/close

farah:
@damemi  no problem, i took embarrassingly long. This was tricky one & I have have had hard time find time to work on it consistently recently, will look forward to reading @ingvagabund  solution. I've give up my PR if i take too long in the future!! thanks

PR #406, opened 2020-09-18
Add Namespace filtering to RemoveDuplicates strategy

### LINKED ISSUE #405: Add namespace filtering to RemoveDuplicates strategy
Currently namespace filtering isn't available in RemoveDuplicates. I think this was intentionally chosen for reasons similar to LowNodeUtilization (where the strategy needs to consider the entire node to be effective), but looking more into RemoveDuplicates I think it does not depend on considering the entire node.

This is mainly because `namespace` is one of the fields we use to consider a pod's uniqueness:
https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/descheduler/strategies/duplicates.go#L106

So we have never considered pods in different namespaces to be duplicates, so there is no risk of making the strategy less effective. It seems reasonable then to want to run this strategy in only select namespaces (or exclude others).

Does it make sense then that we could add namespace filters to this strategy?

damemi:
Fixes #405 

This also adds a `parameters` table to each strategy for a quicker reference at what the params are available

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,16 @@ are ready again, this strategy could be enabled to evict those duplicate pods.
 It provides one optional parameter, `ExcludeOwnerKinds`, which is a list of OwnerRef `Kind`s. If a pod
 has any of these `Kind`s listed as an `OwnerRef`, that pod will not be considered for eviction.
 
+**Parameters:**
+
+|Name|Type|
+|---|---|
+|`excludeOwnerKinds`|list(string)|
+|`namespaces`|(see [namespace filtering](#namespace-filtering))|
+|`thresholdPriority`|int (see [priority filtering](#priority-filtering))|
+|`thresholdPriorityClassName`|string (see [priority filtering](#priority-filtering))|
+
+**Example:**
 ```
 apiVersion: "descheduler/v1alpha1"
 kind: "DeschedulerPolicy"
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -188,8 +206,17 @@ This strategy makes sure that pods violating interpod anti-affinity are removed
 if there is podA on a node and podB and podC (running on the same node) have anti-affinity rules which prohibit
 them to run on the same node, then podA will be evicted from the node so that podB and podC could run. This
 issue could happen, when the anti-affinity rules for podB and podC are created when they are already running on
-node. Currently, there are no parameters associated with this strategy. To disable this strategy, the
-policy should look like:
+node. Currently, there are no parameters associated with this strategy.
```
ingvagabund, line 209: `Currently, there are no parameters associated with this strategy.` not necessary here

ingvagabund: `To disable this strategy, the policy should look like` still applies here unless the example's `enabled` field is set to `true`

ingvagabund (changes requested):
+1 for parameter tables

README.md
damemi: updated

ingvagabund:
/lgtm
/hold

Holding for others to LGTM. Feel free to unhold.

lixiang233 (commented):
just one nit in README, otherwise LGTM

README.md
```diff
@@ -148,7 +158,15 @@ considered appropriately utilized and is not considered for eviction. The thresh
...
 These thresholds, `thresholds` and `targetThresholds`, could be tuned as per your cluster requirements.
-Here is an example of a policy for this strategy:
+
+**Parameters:**
+
+|Name|Type|
+|---|---|
+|`thresholds`|map(string:int)|
+|`targetThresholds`|map(string:int)|
+
```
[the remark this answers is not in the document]
damemi: Thanks for catching that! Updated

Issue #347, opened 2020-07-15
Add Support For Structured Logging


seanmalloy:
**Is your feature request related to a problem? Please describe.**

As a user of descheduler I'd like the descheduler to support generating structured logs, so that my log aggregation tools can more easily parse the descheduler logs. The k/k repo has recently started adding support for structured logs.

**Describe the solution you'd like**

Add a new `--logging-format` CLI option to descheduler to allow choosing between text and json log formats. The text log format should be the default.

Update klog usage in descheduler to use the new `InfoS` and `ErrorS` methods, https://github.com/kubernetes/enhancements/tree/master/keps/sig-instrumentation/1602-structured-logging#new-klog-methods.

**Describe alternatives you've considered**

Using the descheduler logs as is without any changes.

**What version of descheduler are you using?**

descheduler version: v0.18.0

**Additional context**

See structured logging KEP: https://github.com/kubernetes/enhancements/tree/master/keps/sig-instrumentation/1602-structured-logging.

PR to add the `--logging-format` option to kube-scheduler: https://github.com/kubernetes/kubernetes/pull/91522.

seanmalloy:
@farah would you be interested in working on this?

farah:
Of course, thanks sean. 

/assign

seanmalloy:
@farah thanks for your help with this. Let us know if you have any questions.

FYI @damemi and @ingvagabund I'm proposing adding structured logging support to descheduler. Please let me know if you have any concerns. Thanks!

ingvagabund:
Looks good. It will force us to write better log messages :) Also, opens a possibility to parse the logs properly and allows to build various models from the logs which can be used to see how descheduler behaves.

damemi:
+1 for improving logging!

seanmalloy:
@farah thanks for helping to get the initial structured logging changes completed in #376. Do you want to submit a PR to add the `--logging-format` CLI option?

Also, I think there are still some log messages that need to be converted to use structured logging(`InfoS` and `ErrorS`). Are you interesting in submitting some PR's to finish converting to structured logging?

We would like to complete the conversion to structured logging as part of the descheduler v0.20.0(k8s v1.20.0) release cycle. Let me know if you need any help with this effort.

See below for examples of log statements that still need to be converted to use `InfoS`. Note this list is not comprehensive.
```
$ ag Infof pkg/
pkg/descheduler/descheduler.go
[...]

PR #407, opened 2020-09-20
Flip Info/Infof/Error to InfoS/ErrorS

ingvagabund:
Towards https://github.com/kubernetes-sigs/descheduler/issues/347

### DIFF
--- a/pkg/descheduler/node/node.go
+++ b/pkg/descheduler/node/node.go
@@ -18,7 +18,8 @@ package node
 
 import (
 	"context"
-	"k8s.io/api/core/v1"
+
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	coreinformers "k8s.io/client-go/informers/core/v1"
@@ -80,16 +81,16 @@ func IsReady(node *v1.Node) bool {
 			klog.V(1).InfoS("Ignoring node", "node", klog.KObj(node), "condition", cond.Type, "status", cond.Status)
 			return false
 		} /*else if cond.Type == v1.NodeOutOfDisk && cond.Status != v1.ConditionFalse {
-			klog.V(4).InfoS("Ignoring node %v with %v condition status %v", node.Name, cond.Type, cond.Status)
+			klog.V(4).InfoS("Ignoring node with condition status", "node", klog.KObj(node.Name), "condition", cond.Type, "status", cond.Status)
 			return false
 		} else if cond.Type == v1.NodeNetworkUnavailable && cond.Status != v1.ConditionFalse {
-			klog.V(4).Infof("Ignoring node %v with %v condition status %v", node.Name, cond.Type, cond.Status)
+			klog.V(4).InfoS("Ignoring node with condition status", "node", klog.KObj(node.Name), "condition", cond.Type, "status", cond.Status)
 			return false
 		}*/
 	}
 	// Ignore nodes that are marked unschedulable
 	/*if node.Spec.Unschedulable {
-		klog.V(4).Infof("Ignoring node %v since it is unschedulable", node.Name)
+		klog.V(4).InfoS("Ignoring node since it is unschedulable", "node", klog.KObj(node.Name))
 		return false
 	}*/
 	return true
@@ -125,7 +126,7 @@ func PodFitsCurrentNode(pod *v1.Pod, node *v1.Node) bool {
 	ok, err := utils.PodMatchNodeSelector(pod, node)
 
 	if err != nil {
-		klog.Error(err)
+		klog.ErrorS(err, "Failed to match node selector")
 		return false
 	}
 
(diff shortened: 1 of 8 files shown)


pkg/descheduler/node/node.go
```diff
@@ -80,16 +81,16 @@ func IsReady(node *v1.Node) bool {
 			klog.V(1).InfoS("Ignoring node", "node", klog.KObj(node), "condition", cond.Type, "status", cond.Status)
 			return false
 		} /*else if cond.Type == v1.NodeOutOfDisk && cond.Status != v1.ConditionFalse {
-			klog.V(4).InfoS("Ignoring node %v with %v condition status %v", node.Name, cond.Type, cond.Status)
+			klog.V(4).InfoS("Ignoring node with condition status", "node", klog.KObj(node.Name), "condition", cond.Type, "status", cond.Status)
 			return false
 		} else if cond.Type == v1.NodeNetworkUnavailable && cond.Status != v1.ConditionFalse {
-			klog.V(4).Infof("Ignoring node %v with %v condition status %v", node.Name, cond.Type, cond.Status)
+		  klog.V(4).InfoS("Ignoring node with condition status", "node", klog.KObj(node.Name), "condition", cond.Type, "status", cond.Status)
```
lixiang233, line 87: There seems to be something wrong with the indentation.

pkg/descheduler/strategies/duplicates.go
```diff
@@ -58,7 +58,7 @@ func RemoveDuplicatePods(
 	podEvictor *evictions.PodEvictor,
 ) {
 	if err := validateRemoveDuplicatePodsParams(strategy.Params); err != nil {
-		klog.V(1).Info(err)
+		klog.V(1).InfoS("Invalid RemoveDuplicatePods parameters", "err", err)
```
lixiang233, line 61: Is this change duplicated with #408 ？

lixiang233 (commented):
This PR contains changes in #408 , given this PR has some additional changes, I think we should keep this PR and close #408, WDYT @farah ?

farah:
@lixiang233 Sure, makes sense to me.

farah:
Although I don't think this PR fixes #347 as there still the` --logging-format` flag for outputting json remaining.

pkg/descheduler/node/node.go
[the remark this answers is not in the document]
ingvagabund: Fixed, thanks!!!

pkg/descheduler/strategies/lownodeutilization.go
```diff
@@ -69,7 +69,7 @@ func validateLowNodeUtilizationParams(params *api.StrategyParameters) error {
 func LowNodeUtilization(ctx context.Context, client clientset.Interface, strategy api.DeschedulerStrategy, nodes []*v1.Node, podEvictor *evictions.PodEvictor) {
 	// TODO: May be create a struct for the strategy as well, so that we don't have to pass along the all the params?
 	if err := validateLowNodeUtilizationParams(strategy.Params); err != nil {
-		klog.V(1).Info(err)
+		klog.V(1).InfoS("Invalid LowNodeUtilization parameters", "err", err)
```
damemi, line 72: Just curious, is there a reason we log all these at `klog.V(1).InfoS` instead of `klog.ErrorS`?

farah: Good question, technically I think it should since it's an error. And the docs say `ErrorS` should be preferred anyway.

	// Info logs a non-error message with the given key/value pairs as context.
	//
	// The msg argument should be used to add some constant description to
	// the log line.  The key/value pairs can then be used to add additional
	// variable information.  The key/value pairs should alternate string
	// keys and arbitrary values.
	Info(msg string, keysAndValues ...interface{})

	// Error logs an error, with the given message and key/value pairs as context.
	// It functions similarly to calling Info with the "error" named value, but may
	// have unique behavior, and should be preferred for logging errors (see the
	// package documentations for more information).
	//
	// The msg field should be used to add context to any underlying error,
	// while the err field should be used to attach the actual error that
	// triggered this log line, if present.
	Error(err error, msg string, keysAndValues ...interface{})

ingvagabund: Historical one only? Error feels more intuitive here since the strategy can no longer continue executing.

ingvagabund: As a separete PR since there are more messages to flip: https://github.com/kubernetes-sigs/descheduler/pull/409

damemi: Is #409 replacing this PR or just building on this for now until it merges?

ingvagabund: Building on top. Those changes are semantically different. Though, I am fine just merging #409, which will merge this one as well

damemi: Well, we already have feedback for these changes here so to keep from moving discussion around too much we can merge this one, then #409. Just wasn't sure the intention

damemi (approved):
/approve

pkg/descheduler/strategies/lownodeutilization.go
farah: ^It's called 'stat padding'  :)

lixiang233 (approved):
/lgtm

PR #383, opened 2020-08-21
WIP - New Strategy RemovePodsViolatingTopologySpreadConstraint

seanmalloy:
This is a work in progress. It does not work yet. Fixes #146.

This PR has three commits:
* f79d007a94183e7fe20912a58f3de75825ee353b - original work by @krmayankk
* 4e08c15b7331a2a433cb70f3c8424f48197c6622 - documentation
* f4ccb5a781022fccd413ca719c3aaafd7668daec - incomplete code refactor

Task List:
- [x] Documentation
- [ ] Unit Tests
- [ ] New Code For Strategy

PR #413, opened 2020-09-30
Add PodTopologySpread strategy

### LINKED ISSUE #146: Policy for Balancing Pods across topology domains

damemi:
(This picks up from https://github.com/kubernetes-sigs/descheduler/pull/383)
Fixes #146 

This PR attempts to add a new re-balancing strategy for PodTopologySpreading, specifically by evicting pods which belong to topologies that violate any constraint's `MaxSkew` for that topology.

Because pods can only match topology constraints within the same namespace, I thought it made sense to write this strategy as a loop over all the namespaces in the cluster (we can obviously add filtering for namespaces). This lets us count all the constraints in the namespace, one at a time, and compare the counts of all matching pods (via labelSelector) in that topology. I think this also simplifies the data structures necessary to map `namespace -> topologyKey -> labelSelector -> topologyValue -> pods`.

In its current state it's a very "brute force" approach so any feedback would be very helpful. I've noted spots where it could be optimized and I'm sure there are others, but this seems like the general algorithm we'll need. This is definitely trickier than I thought so I'm really hoping to get more eyes on this attempt now.


damemi:
Note: As mentioned this needs some heavy review, and so it's still "wip-ish". I kept the original commits I picked for this but will squash them, as well as working on adding tests and updating docs.

README.md
```diff
@@ -293,6 +294,21 @@ strategies:
     enabled: true
 ````
 
+### RemovePodsViolatingTopologySpreadConstraint
+
+This strategy makes sure that pods violating [topology spread constraints](https://kubernetes.io/docs/concepts/workloads/pods/pod-topology-spread-constraints/)
+are evicted from nodes. This strategy requires k8s version 1.18 at a minimum. To disable this strategy, the
```
rhockenbury, line 300: 1.18? I thought this feature was introduced in 1.16?

seanmalloy: It is beta in 1.18 and enabled by default. It is GA in 1.19. Prior to 1.18 it was alpha and required a feature gate to enable it.

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -0,0 +1,182 @@
...
+	"sigs.k8s.io/descheduler/pkg/api"
+	"sigs.k8s.io/descheduler/pkg/descheduler/evictions"
+)
+
+// @seanmalloy notes:
+//
+// https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#topologyspreadconstraint-v1-core
+// https://kubernetes.io/docs/concepts/workloads/pods/pod-topology-spread-constraints/
+// https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/descheduler/strategies/pod_antiaffinity.go
+
```
seanmalloy, line 35: Feel free to remove my nots. I don't think they are super useful anymore. :-)

go.sum
```diff
@@ -125,6 +125,7 @@ github.com/gogo/protobuf v1.1.1/go.mod h1:r8qH/GZQm5c6nD/R0oafs1akxWv10x8SbQlK7a
 github.com/gogo/protobuf v1.2.1/go.mod h1:hp+jE20tsWTFYpLwKvXlhS1hjn+gTNwPg2I6zVXpSg4=
 github.com/gogo/protobuf v1.3.1 h1:DqDEcV5aeaTmdFBePNpYsp3FlcVH/2ISVVM9Qf8PSls=
 github.com/gogo/protobuf v1.3.1/go.mod h1:SlYgWuQ5SjCEi6WLHjHCa1yvBfUnHcTbrrZtXPKa29o=
+github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b h1:VKtxabqXZkF25pY9ekfRL6a582T4P37/31XEstQ5p58=
```
seanmalloy, line 128: It seems odd that this is being updated. Is this change needed?

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -0,0 +1,182 @@
...
+					}
+				}
+			}
+		}
+	}
+
+	for pod := range podsForEviction {
+		node, err := client.CoreV1().Nodes().Get(ctx, pod.Spec.NodeName, metav1.GetOptions{})
+		if err != nil {
+			klog.ErrorS(err, "couldn't get node", "node", pod.Spec.NodeName)
```
seanmalloy, line 175:
```suggestion
			klog.ErrorS(err, "Couldn't get node", "node", pod.Spec.NodeName)
```

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -0,0 +1,182 @@
...
+				// (this loop is where we count the number of pods per topologyValue that match this constraint's selector)
+				for _, pod := range namespacePods.Items {
+					if !evictable.IsEvictable(&pod) {
+						continue
+					}
+
+					// 4. if the pod matches this TopologySpreadConstraint LabelSelector
+					s, err := metav1.LabelSelectorAsSelector(constraint.LabelSelector)
+					if err != nil {
+						klog.ErrorS(err, "couldn't parse label selector as selector", "selector", constraint.LabelSelector)
```
seanmalloy, line 113:
```suggestion
						klog.ErrorS(err, "Couldn't parse label selector as selector", "selector", constraint.LabelSelector)
```

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -0,0 +1,182 @@
...
+	if err != nil {
+		klog.ErrorS(err, "couldn't list namespaces")
+		return
+	}
+	podsForEviction := make(map[*v1.Pod]struct{})
+	// 1. for each namespace...
+	for _, namespace := range namespaces.Items {
+		namespacePods, err := client.CoreV1().Pods(namespace.Name).List(ctx, metav1.ListOptions{})
+		if err != nil {
+			klog.ErrorS(err, "couldn't list pods in namespace", "namespace", namespace)
```
seanmalloy, line 83:
```suggestion
			klog.ErrorS(err, "Couldn't list pods in namespace", "namespace", namespace)
```

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -0,0 +1,182 @@
...
+	// iterate through all topoPairs for this topologyKey and diff currentPods -minPods <=maxSkew
+	// if diff > maxSkew, add this pod in the current bucket for eviction
+
+	evictable := podEvictor.Evictable()
+
+	namespacedTopologySpreadConstrainPods := make(map[string][]v1.Pod)
+	// First record all of the constraints by namespace
+	namespaces, err := client.CoreV1().Namespaces().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		klog.ErrorS(err, "couldn't list namespaces")
```
seanmalloy, line 75:
```suggestion
		klog.ErrorS(err, "Couldn't list namespaces")
```

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -0,0 +1,182 @@
...
+type topologyPair struct {
+	key   string
+	value string
+}
+
+// TODO: remove this type?
+type NamespacedTopologySpreadConstraint struct {
+	Namespace                 string
+	TopologySpreadConstraints []v1.TopologySpreadConstraint
+}
```
seanmalloy, line 46: Yes, it looks like this is not used. So I believe it can be deleted.

damemi:
/hold
thanks for the review comments so far, I'll update those. but I was thinking about the main part of the algorithm, which determines how many pods to evict from a topologyZone, and I think it could be improved over what I currently have.

Right now, when `zoneA` skews `zoneB` by `N` pods, I'm just evicting `N` pods from `zoneA`. Ideally, all these pods would go into the scheduling queue and get scheduled evenly by their topology requirements. But it might make more sense to evict something close to `N/2` pods from `zoneA`, which would ideally balance the zones without requiring so many evictions. You then need to consider how this affects future comparisons in this run (ie, then how does `zoneA` compare to`zoneC`). I was looking into this and it seems like maybe a weird partitioning problem. So I'm interested in what you guys think about this?

I'm pretty content with the algorithm for building the zone buckets, going namespace by namespace. I think this is just O(N) for pods in the cluster but in the worst case it could be N^2 if every pod has a constraint.

Huang-Wei:
@damemi if needed, i should be able to take a review review in the next couple of days.

seanmalloy:
> /hold
> thanks for the review comments so far, I'll update those. but I was thinking about the main part of the algorithm, which determines how many pods to evict from a topologyZone, and I think it could be improved over what I currently have.
> 
> Right now, when `zoneA` skews `zoneB` by `N` pods, I'm just evicting `N` pods from `zoneA`. Ideally, all these pods would go into the scheduling queue and get scheduled evenly by their topology requirements. But it might make more sense to evict something close to `N/2` pods from `zoneA`, which would ideally balance the zones without requiring so many evictions. You then need to consider how this affects future comparisons in this run (ie, then how does `zoneA` compare to`zoneC`). I was looking into this and it seems like maybe a weird partitioning problem. So I'm interested in what you guys think about this?
> 
> I'm pretty content with the algorithm for building the zone buckets, going namespace by namespace. I think this is just O(N) for pods in the cluster but in the worst case it could be N^2 if every pod has a constraint.

I did look through the code and based on my knowledge what you have looks correct to me. I think it would be great if @ingvagabund and @Huang-Wei could review when they have some time.

/cc @Huang-Wei @ingvagabund

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -0,0 +1,242 @@
...
+						continue
+					}
+					// 6. create a topoPair with key as this TopologySpreadConstraint
+					topoPair := topologyPair{key: constraint.TopologyKey, value: nodeValue}
+					// 7. add the pod with key as this topoPair
+					constraintTopologies[topoPair] = append(constraintTopologies[topoPair], &pod)
+					sumPods++
+				}
+
+				// To actually determine how many pods need to be moved, we sort the topologies in ascending length
```
damemi, line 129: I added what I think is a better algorithm for determining how many pods to evict from topologies here (in https://github.com/kubernetes-sigs/descheduler/pull/413/commits/a9976c1b005ed1e1d8c1a532a088fd034f1c86a2)

I put a lot of description in the comments, but basically what I'm doing is splitting the topologies into 2 lists, balanced around the "ideal average" (ie, the size if all pods were perfectly spread). Mathematically, it seems like all topologies need to fall within `maxSkew` of this average. So I built around those that are "borrowing" from the average to be above it, and those that are "lending" from the average to be below it.

So I compare the smallest topology to the largest topology (these are the most likely to be skewed, and the largest topology will have the most "borrowed" that it is able to pay back). If the largest topology has enough to satisfy the `maxSkew` of the smallest topology, we move on to check the same with the next largest "small" topology. If it does not, we move on to check with the next *smallest* "large" topology.

I've tried this with a couple manual examples but would like some help verifying that my assumptions are right. This would run in linear time and simplify the data structures needed to keep track of how many pods have been moved.

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -0,0 +1,242 @@
...
+				// Then, split the topologies in half around the ideal average (ie, perfect spreading)
+				// Sort the list of topologies with length <= the average in ascending order, and all > the average descending.
+				// For example, the list of topologies with these sizes (avg=5):
+				//  2
```
ingvagabund, line 133: Can you use list notation here? [2, 5, 3, 8, 5, 7]

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -0,0 +1,242 @@
...
+				//  5
+				//  3
+				//  8
+				//  5
+				//  7
+				// Would end up in 2 lists like:
```
ingvagabund, line 139: Can you use list notation here? [2, 3, 5, 5] and [8, 7] to avoid quick confusion. Those reading lists by lines (including me) see 4 lists actually.

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -0,0 +1,242 @@
...
+				//  5
+				// We then start at [0,0] and compare the 2 topology sizes.
+				// If the diff of the size of the topologies is more than the maxSkew, we will move up to half that skew,
+				// or the available pods from the "aboveAvg" topology, whichever is lower.
+				// (Note, we will only move as many pods from a topology as possible without bringing it below the ideal average.)
+				// If the diff is within the skew, we move to the next "belowAvg" topology.
+				// If the "aboveAvg" topology can't give any more without falling below the average, we move to the next "aboveAvg"
+				//
+				// Following this, the above topologies end up "sorted" as:
+				//  5  5
```
ingvagabund, line 152: Can you use list notation here? [5, 5, 5, 5] and [5, 5]

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -0,0 +1,242 @@
...
+// splitTopologiesAroundAverage sorts and splits the list of topologies based on their size
+// It returns 2 lists: topologies <= average (ascending) and topologies > average (descending)
+func splitTopologiesAroundAverage(constraintTopologies map[topologyPair][]*v1.Pod, idealAvg int) ([]topology, []topology) {
+	sortedTopologies := make([]topology, 0, len(constraintTopologies))
+	// sort the topologies and return 2 lists: those <= the average and those > the average (> list inverted)
+	for pair, list := range constraintTopologies {
+		sortedTopologies = append(sortedTopologies, topology{pair: pair, pods: list})
+	}
+
+	// create an ascending slice of all topologies
```
ingvagabund, line 220: "... all topology values". Topology keys is constant here as `splitTopologiesAroundAverage` is called within a single constraint. Each constraint has only a single topology key.

ingvagabund: Also, worth renaming `constraintTopologies` to `constraintTopologyKeyValues` or similar to stress the topology key is constant.

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -0,0 +1,242 @@
...
+	sort.Slice(sortedTopologies, func(i, j int) bool {
+		return len(sortedTopologies[i].pods) < len(sortedTopologies[j].pods)
+	})
+	belowAvg := make([]topology, 0)
+	aboveAvg := make([]topology, 0)
+
+	// separate those that are <= the average from those > the average
+	for _, topology := range sortedTopologies {
+		if len(topology.pods) <= idealAvg {
+			belowAvg = append(belowAvg, topology)
```
ingvagabund, line 230: `belowOrEqualAvg`

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -0,0 +1,242 @@
...
+				// j is the index for aboveAvg
+				i, j := 0, 0
+				for i < len(belowAvg) && j < len(aboveAvg) {
+					// if j has no more to give without falling below the ideal average, move to next aboveAvg
+					if len(aboveAvg[j].pods)-idealAvg <= 0 {
+						j++
+					}
+
+					// skew = diff - maxSkew
+					skew := math.Abs(float64(len(belowAvg[i].pods)-len(aboveAvg[j].pods))) - float64(constraint.MaxSkew)
```
ingvagabund, line 170: Skew of a given topology key is the difference between the maximum and minimum number of instances for two topology key values (can be the same values). What's your definition of skew on this line?

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -0,0 +1,242 @@
...
+				// (assuming even distribution by the scheduler of the evicted pods)
+				idealAvg := int(math.Ceil(sumPods / float64(len(constraintTopologies))))
+				belowAvg, aboveAvg := splitTopologiesAroundAverage(constraintTopologies, idealAvg)
+
+				// i is the index for belowAvg
+				// j is the index for aboveAvg
+				i, j := 0, 0
+				for i < len(belowAvg) && j < len(aboveAvg) {
+					// if j has no more to give without falling below the ideal average, move to next aboveAvg
+					if len(aboveAvg[j].pods)-idealAvg <= 0 {
```
ingvagabund, line 165: `if len(aboveAvg[j].pods) <= idealAvg {` looks more friendly here saying "once # of `aboveAvg` pods is less or equal the average, leave it be!!!"

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -0,0 +1,242 @@
...
+						j++
+					}
+
+					// skew = diff - maxSkew
+					skew := math.Abs(float64(len(belowAvg[i].pods)-len(aboveAvg[j].pods))) - float64(constraint.MaxSkew)
+
+					// if i and j are within the maxSkew of each other, move to next belowAvg
+					if int32(skew) <= constraint.MaxSkew {
```
ingvagabund, line 173: `skew` already has `constraint.MaxSkew` substracted from itself. Can you more elaborate on what's the test checking to reflect the condition?

pkg/descheduler/strategies/topologyspreadconstraint.go
damemi: Yeah, I wasn't sure on the terminology so I probably used "topology" to ambiguously refer to either an entire topology (all its pairs) and a single topology zone

pkg/descheduler/strategies/topologyspreadconstraint.go
damemi: > the difference between the maximum and minimum number of instances for two topology key values

that's is the `maxSkew`, here I'm measuring the actual skew between the 2 zones. I subtract the `maxSkew` from the difference because 2 zones only need to be within the `maxSkew` of each other.

So here I am measuring the actual skew from an acceptable value

pkg/descheduler/strategies/topologyspreadconstraint.go
damemi: You're right, don't know why I did it that weird way

pkg/descheduler/strategies/topologyspreadconstraint.go
damemi: Good catch, I'm doubly-accounting for the `maxSkew` from what I described above (in https://github.com/kubernetes-sigs/descheduler/pull/413#discussion_r499789794).

In this case, if I wanted to do what I said there, this check should be `if int32(skew) <= 0`. But removing that subtraction and keeping this check makes more sense imo

go.sum
damemi: I think it's leftover from one of the old commits I picked. Will double check this when I eventually squash but I imagine it will disappear then

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -0,0 +1,242 @@
...
+					s, err := metav1.LabelSelectorAsSelector(constraint.LabelSelector)
+					if err != nil {
+						klog.ErrorS(err, "couldn't parse label selector as selector", "selector", constraint.LabelSelector)
+					}
+					if !s.Matches(labels.Set(pod.Labels)) {
+						continue
+					}
+
+					// 5. If the pod's node matches this constraint's topologyKey, create a topoPair and add the pod
+					node, err := client.CoreV1().Nodes().Get(ctx, pod.Spec.NodeName, metav1.GetOptions{})
```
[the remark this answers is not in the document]
damemi: updated to build a cached `nodeMap` for this based on the `nodes` param passed to the strategy

damemi:
Thanks for the review so far. I added another commit in db12bba8666ae6135e909bcbe4f9249aecbfb02a that does the same algorithm with only 1 sorted list of topology zones. This is probably more efficient but I'm open to whichever style others prefer for readability

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -0,0 +1,219 @@
...
+					// if i and j are within the maxSkew of each other, move to next belowOrEqualAvg
+					if int32(skew) <= constraint.MaxSkew {
+						i++
+						continue
+					}
+
+					// the most that can be given from aboveAvg is up to halfway,
+					// or how many it has remaining without falling below the average
+					movePods := int(math.Min(math.Ceil(skew/2), float64(len(sortedTopologies[j].pods)-idealAvg)))
+					if movePods == 0 {
```
Huang-Wei, line 174: `movePods <= 0`? As `len(sortedTopologies[j].pods)-idealAvg` can be negative.

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -0,0 +1,219 @@
...
+						i++
+						continue
+					}
+
+					// remove pods from aboveAvg and add them to the list of pods to be evicted
+					// also (just for tracking), add them to the list of pods in belowOrEqualAvg[i]
+					aboveToEvict := sortedTopologies[j].pods[movePods-1:]
+					for _, pod := range aboveToEvict {
+						podsForEviction[pod] = struct{}{}
+					}
```
Huang-Wei, line 184: Don't repeat storing a temporary pointer (`pod`) as the temporary pointer in range loop points to the same address. So this logic would end up with storing only one single entry in `podsForEviction` - which is the last Pod in `aboveToEvict`.

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -0,0 +1,219 @@
...
+					// or how many it has remaining without falling below the average
+					movePods := int(math.Min(math.Ceil(skew/2), float64(len(sortedTopologies[j].pods)-idealAvg)))
+					if movePods == 0 {
+						i++
+						continue
+					}
+
+					// remove pods from aboveAvg and add them to the list of pods to be evicted
+					// also (just for tracking), add them to the list of pods in belowOrEqualAvg[i]
+					aboveToEvict := sortedTopologies[j].pods[movePods-1:]
```
Huang-Wei, line 181: Shouldn't this be either:

- `aboveToEvict := sortedTopologies[j].pods[:movePods]` - the first `movePods` pods
- or, `aboveToEvict := sortedTopologies[j].pods[len(sortedTopologies[j].pods) - movePods:]` - the last `movePods` pods

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -0,0 +1,219 @@
...
+
+					// the most that can be given from aboveAvg is up to halfway,
+					// or how many it has remaining without falling below the average
+					movePods := int(math.Min(math.Ceil(skew/2), float64(len(sortedTopologies[j].pods)-idealAvg)))
+					if movePods == 0 {
+						i++
+						continue
+					}
+
+					// remove pods from aboveAvg and add them to the list of pods to be evicted
```
Huang-Wei, line 179: The comment needs update.

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -0,0 +1,219 @@
...
+					// also (just for tracking), add them to the list of pods in belowOrEqualAvg[i]
+					aboveToEvict := sortedTopologies[j].pods[movePods-1:]
+					for _, pod := range aboveToEvict {
+						podsForEviction[pod] = struct{}{}
+					}
+					sortedTopologies[j].pods = sortedTopologies[j].pods[:movePods]
+					sortedTopologies[i].pods = append(sortedTopologies[i].pods, aboveToEvict...)
```
Huang-Wei, line 186: What do these 2 lines mean?

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -0,0 +1,219 @@
...
+				namespaceTopologySpreadConstrainPods = append(namespaceTopologySpreadConstrainPods, pod)
+			}
+		}
+		if len(namespaceTopologySpreadConstrainPods) == 0 {
+			return
+		}
+
+		// 2. for each topologySpreadConstraint in that namespace
+		for _, constrainedPod := range namespaceTopologySpreadConstrainPods {
+			// find pods to evict one constraint at a time
```
Huang-Wei, line 96: It doesn't look right - we should check all (hard) constraints in a pod at a time. For instance, for a Pod that has 2 (hard) constraints, evicting Pods individually based on the 1st and 2nd constraint doesn't make sense.

pkg/descheduler/strategies/topologyspreadconstraint_test.go
```diff
@@ -0,0 +1,103 @@
...
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes/fake"
+	core "k8s.io/client-go/testing"
+	"sigs.k8s.io/descheduler/pkg/descheduler/evictions"
+	"sigs.k8s.io/descheduler/test"
+)
+
+func TestTopologySpreadConstraint(t *testing.T) {
```
Huang-Wei, line 17: We need more cases by mixing the following conditions:

- multiple constraints
- qualified topology domains that have 0 matching pods
- pod with topologySpreadConstraint matches or not match itself on its labelSelector
- pod has nodeSelector/nodeAffinity

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -0,0 +1,220 @@
...
+		for _, pod := range namespacePods.Items {
+			if pod.Spec.TopologySpreadConstraints != nil {
+				namespacedTopologySpreadConstrainPods[pod.Namespace] = append(namespacedTopologySpreadConstrainPods[pod.Namespace], pod)
+			}
+		}
+		if len(namespacedTopologySpreadConstrainPods[namespace.Name]) == 0 {
+			continue
+		}
+
+		// 2. for each topologySpreadConstraint in that namespace
```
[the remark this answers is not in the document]
damemi: Yeah that's a good question. I kind of intentionally just went with checking all constraints here because I wasn't sure how we should handle hard vs soft. Hard is obviously a requirement but maybe a strategy parameter to consider soft constraints too? (ie `softConstraints: true`)

Then, should the 2 be handled any differently? It seems like hard constraints should be evaluated first but I don't think it makes a difference

pkg/descheduler/strategies/topologyspreadconstraint.go
damemi: `sortedTopologies[j].pods = sortedTopologies[j].pods[:movePods]` - shortens the list of available pods to be moved from the "higher" domain

`sortedTopologies[i].pods = append(sortedTopologies[i].pods, aboveToEvict...)` adds the evicted pods to the "lower" domain

I meant this to keep track of what pods can still be moved (so we don't mark the same pod twice) and track how many still need to be moved by updating the smaller domain's list with a simulated ideal movement

pkg/descheduler/strategies/topologyspreadconstraint.go
damemi: Is a DeepCopy() good?

pkg/descheduler/strategies/topologyspreadconstraint.go
damemi: Yes it should, thanks for catching! I'll go with the second one (makes more sense to me to think of this as chopping off the end of the list)

pkg/descheduler/strategies/topologyspreadconstraint.go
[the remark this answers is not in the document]
Huang-Wei: I'd suggest starting with hard constraints. The semantics of soft constraints is a bit complex, you have to strictly follow up with the upstream code to make it compliant, which I think is a good-to-have for descheduelr, at least for now.

pkg/descheduler/strategies/topologyspreadconstraint.go
Huang-Wei: Just `for i := range aboveToEvict` and use the value: aboveToEvict[i]

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -0,0 +1,293 @@
...
+			}
+		}
+		if len(namespaceTopologySpreadConstrainPods) == 0 {
+			return
+		}
+
+		// 2. for each topologySpreadConstraint in that namespace
+		for _, constrainedPod := range namespaceTopologySpreadConstrainPods {
+			// find pods to evict one constraint at a time
+			for _, constraint := range constrainedPod.Spec.TopologySpreadConstraints {
```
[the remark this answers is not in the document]
Huang-Wei: So I think an ideal approach is to iterate over pods and gets a merged slice with tuple `{[]constraints, nodeSelector/nodeAffinity}` as its element. And later we process each tuple instead of each constraint.

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -0,0 +1,293 @@
...
+	podsForEviction := make(map[*v1.Pod]struct{})
+	// 1. for each namespace...
+	for _, namespace := range namespaces.Items {
+		if (!includedNamespaces.Has(namespace.Name) || excludedNamespaces.Has(namespace.Name)) && (includedNamespaces.Len()+excludedNamespaces.Len() > 0) {
+			continue
+		}
+		namespacePods, err := client.CoreV1().Pods(namespace.Name).List(ctx, metav1.ListOptions{})
+		if err != nil {
+			klog.ErrorS(err, "Couldn't list pods in namespace", "namespace", namespace)
+			return
```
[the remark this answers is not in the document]
damemi: Yup, good catch

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -0,0 +1,293 @@
...
+				// TODO(@damemi): add support for soft constraints
+				if constraint.WhenUnsatisfiable != v1.DoNotSchedule {
+					continue
+				}
+				constraintTopologies := make(map[topologyPair][]*v1.Pod)
+				// pre-populate the topologyPair map with all the topologies available from the nodeMap
+				// (we can't just build it from existing pods' nodes because a topology may have 0 pods)
+				for _, node := range nodeMap {
+					if val, ok := node.Labels[constraint.TopologyKey]; ok {
+						constraintTopologies[topologyPair{key: constraint.TopologyKey, value: val}] = make([]*v1.Pod, 0, len(namespacePods.Items))
```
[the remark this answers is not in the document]
damemi: true, since we're adding to these lists later on

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -0,0 +1,293 @@
...
+			for _, constraint := range constrainedPod.Spec.TopologySpreadConstraints {
+				// Only deal with hard topology constraints
+				// TODO(@damemi): add support for soft constraints
+				if constraint.WhenUnsatisfiable != v1.DoNotSchedule {
+					continue
+				}
+				constraintTopologies := make(map[topologyPair][]*v1.Pod)
+				// pre-populate the topologyPair map with all the topologies available from the nodeMap
+				// (we can't just build it from existing pods' nodes because a topology may have 0 pods)
+				for _, node := range nodeMap {
```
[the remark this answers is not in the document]
damemi: > This `for` loop is expensive. I'd suggest putting it in an outer loop right after we have calculated all constraints within the current namespace.

The problem is this loop is meant to identify any topologyPairs/domains that have 0 pods in them. The only way for us to know that a label on a pair is a topologyKey (and not just a label) is to check against each constraint's key. So even if it's moved outside the loop, it will still end up being O(M*N) where we have to check each node for each constraint.

> And, IMO even we should move it to the most outer loop after we have all constraints over all eligible namespaces - i.e., constraintTopologies is a map keyed with namespace, and values with `map[topologyPair][]*v1.Pod` - which would be the most efficient way.

I'm not sure the efficiency that's gained here, as we will still be looping over each namespace->constraint

> BTW: you can move the logic of building nodeMap below calculating constraints, as if len(constraints) == 0, we can return earlier.

This loop is ranging over `constraints`, so if `len(constraints) == 0` we don't have to worry about that

pkg/descheduler/strategies/topologyspreadconstraint.go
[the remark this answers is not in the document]
damemi: Good point, but I think we could also just cover this with a `map[v1.TopologySpreadConstraint]struct{}` (see new commit)

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -0,0 +1,295 @@
...
+					constraintTopologies[topologyPair{key: constraint.TopologyKey, value: val}] = make([]*v1.Pod, 0)
+				}
+			}
+
+			// 3. for each evictable pod in that namespace
+			// (this loop is where we count the number of pods per topologyValue that match this constraint's selector)
+			var sumPods float64
+			for i := range namespacePods.Items {
+				if !evictable.IsEvictable(&namespacePods.Items[i]) {
+					continue
```
lixiang233, line 152: Unevictable pods are ignored while their constraints not, then pod-num in some topology domains may be less than it actually is. Suppose we have two domains A(1 pod) and B(3 Pods), maxSkew is 1, 1 pod on domain B should be evicted, but if all these pods except one pod on B are unevictable, no pod will be evicted because maxSkew hasn't be reached.
I think we can treat unevictable pods like pods with nodeAffinity, only ignore them when we try to evict them.

lixiang233 (commented):
I notice that PodTopologySpread plugin supports [Cluster-level default constraints](https://kubernetes.io/docs/concepts/workloads/pods/pod-topology-spread-constraints/#cluster-level-default-constraints), should we support this in this strategy too?

pkg/descheduler/strategies/topologyspreadconstraint.go
damemi: I don't think the descheduler should ever consider unevictable pods as evictable, even in this case. Though it is worth noting in documentation that unevictable pods may hurt the effectiveness of this strategy, that's about the best we can do.

imo they are unevictable for a reason and we provide various knobs for the user to adjust that criteria

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -0,0 +1,295 @@
...
+				}
+				nodeValue, ok := node.Labels[constraint.TopologyKey]
+				if !ok {
+					continue
+				}
+				// 6. create a topoPair with key as this TopologySpreadConstraint
+				topoPair := topologyPair{key: constraint.TopologyKey, value: nodeValue}
+				// 7. add the pod with key as this topoPair
+				// (If the pod has nodeAffinity/nodeSelector, push it onto the front of the list
+				// so it is considered last for eviction)
```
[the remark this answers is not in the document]
damemi: That's a good point. I think in this case then the ordering to consider for eviction should be: normal pods from lowest-to-highest priority followed by pods with affinity/Selectors (also sorted from lowest-to-highest). Wdyt?

damemi (commented):
At this point I'm pretty satisfied with the use case coverage in this strategy. Through feedback we have identified some edge cases and limitations of the implementation, but for a basic "balance the topologies" approach I think this works well.

I will try to add some more test cases. I also made some updates to the balancing algorithm to improve its effectiveness and readability. Please review some more and let me know if there are any updates (or edge cases you think I should add to the unit tests).

pkg/descheduler/strategies/topologyspreadconstraint.go
damemi: @lixiang233 revisiting this, after running the strategy in more live-cluster scenarios I apologize and take back what I said. I agree with you that we shouldn't be excluding unevictable pods at this point, but rather when we try to evict them. That's much more intuitive and you were absolutely correct!

damemi: However we should sort non-evictable pods to be considered last for eviction, so I think it could be helpful to make the [`evictable`](https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/descheduler/evictions/evictions.go#L170) type exported so that I can pass it to my sort functions here. @ingvagabund wdyt?

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -0,0 +1,336 @@
...
+				return true
+			} else if !hasSelectorOrAffinity(*list[i]) && hasSelectorOrAffinity(*list[j]) {
+				// if just one has selectors/affinity
+				// j should come first
+				return false
+			} else {
+				// if neither has selectors/affinity, fall back to just priority
+				return comparePodsByPriority(list[i], list[j])
+			}
+		})
```
damemi, line 303: @lixiang233 or @ingvagabund mind double checking this sorting code to make sure it's right? The goal is:

1. Put non-evictable pods first
2. Put pods with affinity or selector next
3. Order remaining pods by priority

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -0,0 +1,336 @@
...
+		// followed by the highest priority,
+		// followed by the lowest priority pods with affinity or nodeSelector,
+		// followed by the highest priority pods with affinity or nodeSelector
+		sort.Slice(list, func(i, j int) bool {
+			// any non-evictable pods should be considered last (ie, first in the list)
+			if !isEvictable(list[i]) || !isEvictable(list[j]) {
+				// false - i is the only non-evictable, so return true to put it first
+				// true - j is non-evictable, so return false to put j before i
+				// if true and both and non-evictable, order doesn't matter
+				return !isEvictable(list[j])
```
lixiang233, line 285: Return `!isEvictable(list[j])` will make `evictable pod` comes before `non-evictable` pod: if j is evictable, then i is non-evictable, `!isEvictable(list[j])` is false, then j is before i. return `isEvictable(list[j])` should be correct.

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -0,0 +1,336 @@
...
+		// remove pods from the higher topology and add them to the list of pods to be evicted
+		// also (just for tracking), add them to the list of pods in the lower topology
+		aboveToEvict := sortedDomains[j].pods[len(sortedDomains[j].pods)-movePods:]
+		for i := range aboveToEvict {
+			// if the pod has a hard nodeAffinity or nodeSelector, don't bother evicting it as it will just end up back on the same node
+			// however we still account for it "being evicted" so the algorithm can complete
+			if aboveToEvict[i].Spec.NodeSelector != nil ||
+				(aboveToEvict[i].Spec.Affinity != nil &&
+					aboveToEvict[i].Spec.Affinity.NodeAffinity != nil &&
+					aboveToEvict[i].Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution != nil) {
```
lixiang233, line 259: I think we should check if pod is evictable here, given we added non-evictable pods to the list.

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -0,0 +1,336 @@
...
+	// j is the index for aboveAvg
+	i := 0
+	j := len(sortedDomains) - 1
+	for i < j {
+		// if j has no more to give without falling below the ideal average, move to next aboveAvg
+		if float64(len(sortedDomains[j].pods)) < idealAvg {
+			j--
+		}
+
+		// skew = diff - maxSkew
```
ingvagabund, line 225: The comment does not correspond to the line below. I.e. `maxSkew` is not used to compute `skew`.

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -0,0 +1,336 @@
...
+			continue
+		}
+
+		// the most that can be given from aboveAvg is:
+		// 1. up to half the distance between them
+		// 2. how many it has remaining without falling below the average, or
+		// 3. how many can be added without bringing the smaller domain above the average,
+		// whichever is less
+		// (This is the basic principle of keeping all sizes within ~skew of the average)
+		aboveAvg := math.Ceil(float64(len(sortedDomains[j].pods)) - idealAvg)
```
ingvagabund, line 240: If `idealAvg` is 4.5 and `len(sortedDomains[j].pods` is 5, `aboveAvg` is 1 (0.5 rounded up).
Following what's said in `// 2. how many it has remaining without falling below the average, or`, `1` is remaining though if you remove that one from `sortedDomains[j].pods`, you will get bellow `idealAvg`. Is not the computation  contradicting the comment? The same for `belowAvg` computation.

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -0,0 +1,336 @@
...
+		belowAvg := math.Ceil(idealAvg - float64(len(sortedDomains[i].pods)))
+		smallestDiff := math.Min(aboveAvg, belowAvg)
+		halfSkew := math.Floor(skew / 2)
+		movePods := int(math.Min(smallestDiff, halfSkew))
```
ingvagabund, line 244: You are comparing `smallestDiff` (which is a float rounded up to an integer) and `halfSkew`, which can be non-integer. Should not you round up `halfSkew` first and then take the minimum?

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -0,0 +1,336 @@
...
+		if movePods <= 0 {
+			i++
+			continue
+		}
+
+		// remove pods from the higher topology and add them to the list of pods to be evicted
+		// also (just for tracking), add them to the list of pods in the lower topology
+		aboveToEvict := sortedDomains[j].pods[len(sortedDomains[j].pods)-movePods:]
```
ingvagabund, line 252: In case `idealAvg` is 4.5, `aboveAvg` is `5` and selected as the smallest of `{aboveAvg, belowAvg, halfSkew}`, by removing the pod from `sortedDomains[j].pods`, the domain size will fall under `idealAvg`. Which contradicts again with `how many it has remaining without falling below the average`.

E.g. [3 5 5 5], `idealAvg` = 18/4 = `4.5`, `skew(i=0, j=3)` = `2`, `halfSkew` = `1`, `aboveAvg` = `1`, `belowAvg` = `2`.

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -0,0 +1,336 @@
...
+		for i := range aboveToEvict {
```
ingvagabund, line 253: Please use other index variable than `i` even though it's scoped only to this loop. E.g. `k`.

pkg/descheduler/strategies/topologyspreadconstraint.go
ingvagabund: > it will just end up back on the same node

Node selector can target a set of nodes. The same holds for node affinity. What about to check if the node selector or affinity can target at most one node and only then continue?

ingvagabund: > I think we should check if pod is evictable here, given we added non-evictable pods to the list.

Pods from `podsForEviction` are not evicted until all constraints are processed. You might be running the check multiple times compared to one if the pod participates in multiple constraints.

pkg/descheduler/strategies/topologyspreadconstraint.go
damemi: `halfSkew` is rounded down

pkg/descheduler/strategies/topologyspreadconstraint.go
damemi: True, in that case of [3 5 5 5] we should choose 1 to make the new sizes [4 5 5 4]

I think really the comment just needs clarification that no domain will fall *more than a whole pod* under the ideal average. Those fractions of a pod can mathematically add up and be allowed to move

pkg/descheduler/strategies/topologyspreadconstraint.go
damemi: > Node selector can target a set of nodes. The same holds for node affinity. What about to check if the node selector or affinity can target at most one node and only then continue?

That's a good point, I have actually been wondering about if it really makes sense to hard-exclude any pods with affinity/selector. My thinking was different though: the affinity conditions could change on the cluster making the pod perfectly evictable (see our own descheduling strategies). We could actually re-use some helpers from those strategies in informing this decision, but that's toeing the line of conflating 2 different strategies. I think it would be fine though wdyt?

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -0,0 +1,336 @@
...
+			}
+
+			// 3. for each evictable pod in that namespace
+			// (this loop is where we count the number of pods per topologyValue that match this constraint's selector)
+			var sumPods float64
+			for i := range namespacePods.Items {
+				// 4. if the pod matches this TopologySpreadConstraint LabelSelector
+				s, err := metav1.LabelSelectorAsSelector(constraint.LabelSelector)
+				if err != nil {
+					klog.ErrorS(err, "Couldn't parse label selector as selector", "selector", constraint.LabelSelector)
```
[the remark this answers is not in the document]
damemi: I would still log the error, because label selectors aren't always guaranteed to be validated by the apiserver (what we've run into in the scheduler, and still find areas they aren't parsed correctly)

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -0,0 +1,336 @@
...
+				constraintTopologies[topoPair] = append(constraintTopologies[topoPair], &namespacePods.Items[i])
+				sumPods++
+			}
+			balanceDomains(podsForEviction, constraint, constraintTopologies, sumPods, evictable.IsEvictable)
+		}
+	}
+
+	for pod := range podsForEviction {
+		node, ok := nodeMap[pod.Spec.NodeName]
+		if !ok {
```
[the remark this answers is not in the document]
damemi: This goes with what you mentioned above where I was checking `nodeMap` right? (https://github.com/kubernetes-sigs/descheduler/pull/413#discussion_r511934154). So any pods that get to this point must have a node in the node map

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -0,0 +1,336 @@
...
+			// any non-evictable pods should be considered last (ie, first in the list)
+			if !isEvictable(list[i]) || !isEvictable(list[j]) {
+				// false - i is the only non-evictable, so return true to put it first
+				// true - j is non-evictable, so return false to put j before i
+				// if true and both and non-evictable, order doesn't matter
+				return !isEvictable(list[j])
+			}
+
+			// if both pods have selectors/affinity, compare them by their priority
+			if hasSelectorOrAffinity(*list[i]) && hasSelectorOrAffinity(*list[j]) {
```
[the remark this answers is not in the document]
damemi: Ah thank you! Thought this looked like it could be simplified

pkg/descheduler/strategies/topologyspreadconstraint.go
damemi: @ingvagabund I added a helper here to count the # of nodes this pod could fit on (besides the one it's currently on). If that value is 0, it either only fits on this node or doesn't fit on any node, either way it's not PTS's job to evict it in that case.

I did add a TODO here, because I'm realizing since we don't order the pods with affinity based on how many other nodes they can fit on, there could be enough valid evictable pods with affinity that we don't get to (since this list is created with just `aboveToEvict := sortedDomains[j].pods[len(sortedDomains[j].pods)-movePods:]`, which is efficient but optimistic). 

The alternative would be to either add additional sorting for affinity pods based on other feasible nodes, or loop through all the pods in `sortedDomains[j]` and maintain a counter, and keep trying until count==movePods or k=0

seanmalloy:
Let's do this. In my opinion it will be good to get end user feedback to improve this feature in the future. The only way to get end user feedback it to ship the feature.

/lgtm

PR #418, opened 2020-10-04
charts/descheduler: add PodSecurityPolicy support

invidian:
This commit adds restrictive PodSecurityPolicy, which can be
optionally created, so descheduler can be deployed on clusters with
PodSecurityPolicy admission controller, but which do not ship default
policies.

Signed-off-by: Mateusz Gozdek <mgozdekof@gmail.com>

### DIFF
--- a/charts/descheduler/templates/clusterrole.yaml
+++ b/charts/descheduler/templates/clusterrole.yaml
@@ -21,4 +21,11 @@ rules:
 - apiGroups: ["scheduling.k8s.io"]
   resources: ["priorityclasses"]
   verbs: ["get", "watch", "list"]
+{{- if .Values.podSecurityPolicy.create }}
+- apiGroups: ['policy']
+  resources: ['podsecuritypolicies']
+  verbs:     ['use']
+  resourceNames:
+  - {{ template "descheduler.fullname" . }}
+{{- end }}
 {{- end -}}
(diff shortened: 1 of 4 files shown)


damemi (commented):
/ok-to-test

invidian:
>level=error msg="Timeout exceeded: try increasing it by passing --timeout option"

Seems linter timeout should be increased?

damemi:
> > level=error msg="Timeout exceeded: try increasing it by passing --timeout option"
> 
> Seems linter timeout should be increased?

That's a weird error, and it's coming from the golang linter, so let's try a retest first maybe just an infra issue
/retest

damemi (approved):
/approve
looks good to me, though I wonder if we should also provide this somehow for non-helm deployments (under our `/kubernetes` manifests)

charts/descheduler/templates/clusterrole.yaml
```diff
@@ -21,4 +21,11 @@ rules:
 - apiGroups: ["scheduling.k8s.io"]
   resources: ["priorityclasses"]
   verbs: ["get", "watch", "list"]
+{{- if .Values.rbac.podSecurityPolicy.create }}
```
seanmalloy, line 24: @invidian what do you think about instead using `.Values.podSecurityPolicy.create` instead of `.Values.rbac.podSecurityPolicy.create`. I feel like the `PodSecurityPolicy` is separate from RBAC settings.

seanmalloy (changes requested):

seanmalloy:
@invidian sorry for the delay on the review. Thanks for submitting the pull request!

I think the [helm chart configuration section](https://github.com/kubernetes-sigs/descheduler/blob/master/charts/descheduler/README.md#configuration) should also be updated as part of this PR.

charts/descheduler/templates/clusterrole.yaml
invidian: Sounds good. Some charts put it in `rbac`, but I agree it's rather separate from RBAC :+1:

PR #420, opened 2020-10-05
RemoveDuplicates behavior when the number of pods is greater than the number of nodes

vadv:
RemoveDuplicates currently works when the number of pods is less than the number of nodes. In the case of HPA, it is possible to achieve a case where the number of pods can exceed the number of nodes, in which case the cluster will never reach a balanced state and the pods will always move.

### DIFF
--- a/pkg/descheduler/strategies/duplicates.go
+++ b/pkg/descheduler/strategies/duplicates.go
@@ -61,12 +61,12 @@ func RemoveDuplicatePods(
 	podEvictor *evictions.PodEvictor,
 ) {
 	if err := validateRemoveDuplicatePodsParams(strategy.Params); err != nil {
-		klog.V(1).Info(err)
+		klog.ErrorS(err, "Invalid RemoveDuplicatePods parameters")
 		return
 	}
 	thresholdPriority, err := utils.GetPriorityFromStrategyParams(ctx, client, strategy.Params)
 	if err != nil {
-		klog.V(1).Infof("failed to get threshold priority from strategy's params: %#v", err)
+		klog.ErrorS(err, "Failed to get threshold priority from strategy's params")
 		return
 	}
 
@@ -80,13 +80,14 @@ func RemoveDuplicatePods(
 			includedNamespaces = strategy.Params.Namespaces.Include
 			excludedNamespaces = strategy.Params.Namespaces.Exclude
 		}
-		klog.V(1).Infof("Processing node: %#v", node.Name)
+		klog.V(1).InfoS("Processing node", "node", klog.KObj(node))
 		pods, err := podutil.ListPodsOnANode(ctx,
 			client,
 			node,
 			podutil.WithFilter(evictable.IsEvictable),
 			podutil.WithNamespaces(includedNamespaces),
-			podutil.WithoutNamespaces(excludedNamespaces))
+			podutil.WithoutNamespaces(excludedNamespaces),
+		)
 		if err != nil {
 			klog.ErrorS(err, "Error listing evictable pods on node", "node", klog.KObj(node))
 			continue
--- a/pkg/descheduler/strategies/duplicates.go
+++ b/pkg/descheduler/strategies/duplicates.go
@@ -71,15 +71,15 @@ func RemoveDuplicatePods(
 	}
 
 	evictable := podEvictor.Evictable(evictions.WithPriorityThreshold(thresholdPriority))
+	var includedNamespaces, excludedNamespaces []string
+	if strategy.Params != nil && strategy.Params.Namespaces != nil {
+		includedNamespaces = strategy.Params.Namespaces.Include
+		excludedNamespaces = strategy.Params.Namespaces.Exclude
+	}
 
 	listOfPodsOnNode := map[*v1.Node][]*v1.Pod{}
 	ownerKeyCountInCluster := map[string]int32{}
 	for _, node := range nodes {
-		var includedNamespaces, excludedNamespaces []string
-		if strategy.Params != nil && strategy.Params.Namespaces != nil {
-			includedNamespaces = strategy.Params.Namespaces.Include
-			excludedNamespaces = strategy.Params.Namespaces.Exclude
-		}
 		klog.V(1).InfoS("Processing node", "node", klog.KObj(node))
 		pods, err := podutil.ListPodsOnANode(ctx,
 			client,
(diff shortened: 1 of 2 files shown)


pkg/descheduler/strategies/duplicates.go
```diff
@@ -71,15 +70,16 @@ func RemoveDuplicatePods(
...
 	evictable := podEvictor.Evictable(evictions.WithPriorityThreshold(thresholdPriority))
 
+	listOfPodsOnNode := map[*v1.Node][]*v1.Pod{}
+	ownerKeyCountInCluster := map[string]int32{}
 	for _, node := range nodes {
+		var includedNamespaces, excludedNamespaces []string
+		if strategy.Params != nil && strategy.Params.Namespaces != nil {
+			includedNamespaces = strategy.Params.Namespaces.Include
+			excludedNamespaces = strategy.Params.Namespaces.Exclude
+		}
```
damemi, line 82: why move this to inside the node loop?

damemi (commented):
/ok-to-test
@vadv thanks for the PR! There is an open issue for this (https://github.com/kubernetes-sigs/descheduler/issues/263) and some other attempts at similarly fixing the issue already (https://github.com/kubernetes-sigs/descheduler/pull/398 cc @farah). Are the differences in your PR worth trying to combine these efforts, or supercede the existing work?

pkg/descheduler/strategies/duplicates.go
vadv: 👍

vadv:
> Are the differences in your PR worth trying to combine these efforts, or supercede the existing work?
This is a new implementation. In this pr, I do not use GetPodOwnerReplicationCount via the kube api (I make fewer api calls), but in reality, the scheduler will still most likely place the pod on the same node.

farah:
@damemi I will start the implementation of #398 tonight and fixing the review comments that were left from the previous owner of the issue. 

In the mean time happy to hear what other reviewers think of this particular implementation vs #398.

sorry for the delay, its been a while, i've been putting it off so I can properly understand how scheduling works to attempt it but I think i'd better started.

farah:
@vadv if you are interested in this issue (#263), happy for you to take over #398, which has had many a lot of discussion already and review comments already (#334). I have the metrics issue to look at anyway.

pkg/descheduler/strategies/duplicates.go
```diff
@@ -166,3 +145,20 @@ func hasExcludedOwnerRefKind(ownerRefs []metav1.OwnerReference, strategy api.Des
...
+			// Namespace/Kind/Name should be unique for the cluster.
+			// We also consider the image, as 2 pods could have the same owner but serve different purposes
+			// So any non-unique Namespace/Kind/Name/Image pattern is a duplicate pod.
+			s := strings.Join([]string{pod.ObjectMeta.Namespace, ownerRef.Kind, ownerRef.Name, container.Image}, "/")
+			podContainerKeys = append(podContainerKeys, s)
+		}
+	}
+	sort.Strings(podContainerKeys)
+	// Each pod has a list of owners and a list of containers, and each container has 1 image spec.
+	return podContainerKeys[0]
```
ingvagabund, line 163: Counter-example:
- pod P1 has two containers C1 and C2 (each with a different image)
- pod P2 has two containers C3 and C4 (each with a different image).
- containers C2 and C4 have the same image
- images of containers C1, C2 and C3 are unique
- sorting containers key gives the following order: C2>C1, C4>C3.
Then:
- getUniqueOwnerKeyOfPod(P1) = C2
- getUniqueOwnerKeyOfPod(P2) = C4

Both pods P1 and P2 are not duplicates (since C1 != C3). However, `getUniqueOwnerKeyOfPod` marks both pods as duplicates.

farah: @ingvagabund could you please elaborate on how you determined `getUniqueOwnerKeyOfPod` marks P1 & P2 as duplicates? 


> getUniqueOwnerKeyOfPod(P1) = C2
> getUniqueOwnerKeyOfPod(P2) = C4
> 

how does this assertion imply P1 and P2 are duplicates?

ingvagabund: > how does this assertion imply P1 and P2 are duplicates? wouldn't they both return the same value if it were?

`getUniqueOwnerKeyOfPod` takes the first item in sorted `podContainerKeys` list. Which means the pod container keys that make both pods `P1` and `P2` distinct may get ignored based on how `podContainerKeys` list is sorted.

ingvagabund: Assuming both pods `P1` and `P2` are in the same namespace `NS`, owned by the same owner of kind `K` and name `N`. 

For `P1`: `podContainerKeys` list has `[NS/K/N/C1image, NS/K/N/C2image]`.
For `P2`: `podContainerKeys` list has `[NS/K/N/C3image, NS/K/N/C4image]`.

> sorting containers key gives the following order: C2>C1, C4>C3.

After sorting, you will get:

For `P1`: `podContainerKeys` list has `[NS/K/N/C2image, NS/K/N/C1image]`.
For `P2`: `podContainerKeys` list has `[NS/K/N/C4image, NS/K/N/C3image]`.

Given `getUniqueOwnerKeyOfPod` picks the first item in the list, then:

`getUniqueOwnerKeyOfPod(P1)` = `NS/K/N/C2image`
`getUniqueOwnerKeyOfPod(P2)` = `NS/K/N/C4image`

Given both `C2image` and `C4image` images are identical, both pods `P1` and `P2` get the same unique key.

farah: Thanks that makes sense now

farah: @ingvagabund apologies for not getting back to you sooner. I have to admit i didn't understand it fully, but i finally found the real question i needed to ask.

If `C2` and `C4` have the same image, why do they have different image names?

```
getUniqueOwnerKeyOfPod(P1) = NS/K/N/C2image
getUniqueOwnerKeyOfPod(P2) = NS/K/N/C4image
```

From your example, two images can have different image names, but at the same time are identical. So what makes an image unique?

ingvagabund: > containers C2 and C4 have the same image

From that C2image = C4image. `C2image`, resp. `C4image` is a variable name. Though, their value is the same.

farah: Ah ok, thanks. let me test that.

damemi:
Hi @vadv, thank you for working on this issue. However we have gone with different changes in https://github.com/kubernetes-sigs/descheduler/pull/463, and since this issue is now resolved I am going to close this PR. Thanks again!

/close

Issue #432, opened 2020-10-28
 KubeJobCompletion Prometheus alert for desheduler jobs


KR411-prog:
We are receiving  KubeJobCompletion Prometheus alert for desheduler jobs with the below alert message,
kube-system/descheduler-1603424400 is taking more than 12 hours to complete.

The descheduler values file config is as shown below,
```
fullnameOverride: descheduler
nameOverride: descheduler
deschedulerPolicy:
  nodeSelector: kops.k8s.io/instancegroup=nodes
  evict-local-storage-pods: false
  maxNoOfPodsToEvictPerNode: 4
  strategies:
    RemoveDuplicates:
      enabled: true
    RemovePodsViolatingInterPodAntiAffinity:
      enabled: true
    RemovePodsViolatingNodeAffinity:
      enabled: true
    LowNodeUtilization:
      enabled: true
      params:
        numberOfNodes: 2
        nodeResourceUtilizationThresholds:
          # node is underutilized if all 3 metrics are below the threshold
          thresholds:
            "cpu" : 50
            "memory": 30
            "pods": 30
          # node is overutilized is any of these 3 metrics is above the target threshold
          targetThresholds:
            "cpu" : 80
            "memory": 70
            "pods": 50
    RemovePodsHavingTooManyRestarts:
      enabled: true
      params:
        podsHavingTooManyRestarts:
          podRestartThreshold: 100
          includingInitContainers: true
    PodLifeTime:
      enabled: false
rbac:
  create: true
schedule: "*/2 * * * *"
```

I am not sure if there is a way in the config to tune to delete the job if it takes more than 30 mins. I dont find that tuning configuration in the values file in this helm chart.
    chart: descheduler/descheduler-helm-chart
    version: "0.19.0"

Any help on how to avoid getting this alert? Is there any tuning that can be done in descheduler config?

seanmalloy:
@KR411-prog thanks for opening this issue. Please provide the below details and we will try to help.

**Please provide the full CronJob yaml with any sensitive info redacted.**
```
kubectl get cronjob -n mynamespace mycronjob -o yaml
```

**What k8s version are you using?**

**Please provide the pod log for long running descheduler CronJob pod with any sensitive info redacted.**
```
kubectl logs -n mynamespace mypod
```

seanmalloy:
[...]

PR #442, opened 2020-11-20
Set Container Resources In YAML Manifests

seanmalloy:
Prior to this commit the YAML manifests used to install the descheduler
Job and CronJob did not set container requests or limits. This is
considered an anti-pattern when deploying applications on k8s.

Set descheduler container resources to make it a burstable pod. This
will ensure a high quality experience for end users when deploying
descheduler into their clusters using the instructions from the README.

The values choosen for CPU/Memory are not based on any real data.

### DIFF
--- a/kubernetes/cronjob/cronjob.yaml
+++ b/kubernetes/cronjob/cronjob.yaml
@@ -27,6 +27,10 @@ spec:
               - "/policy-dir/policy.yaml"
               - "--v"
               - "3"
+            resources:
+              requests:
+                cpu: "500m"
+                memory: "256Mi"
           restartPolicy: "Never"
           serviceAccountName: descheduler-sa
           volumes:
--- a/kubernetes/job/job.yaml
+++ b/kubernetes/job/job.yaml
@@ -25,6 +25,10 @@ spec:
             - "/policy-dir/policy.yaml"
             - "--v"
             - "3"
+          resources:
+            requests:
+              cpu: "500m"
+              memory: "256Mi"
       restartPolicy: "Never"
       serviceAccountName: descheduler-sa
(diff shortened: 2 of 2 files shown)


seanmalloy:
This fixes the first part of #432. After this is merged I'll also update the helm chart to do the same thing.

/assign @damemi 
/cc @ingvagabund

ingvagabund:
Do we know if the resource constraints are sufficient enough/not sufficient?

damemi:
> Do we know if the resource constraints are sufficient enough/not sufficient?

Yeah, also curious about these numbers (especially the limits) and if we have any way to measure them.

seanmalloy:
All I did was test it on the kind cluster on my laptop, so this really doesn't say much. :-)

I believe there should be a way to allow people to override the requests/limits using kustomize. So I could set some sane defaults and then allow people to override it. Would this approach work for you?

In the short term I don't have time to extensively test the requests/limits a real world environment. Hopefully in the future I can do this, but not right now.

seanmalloy:
Or an even simpler solution would be to only set the requests and not set the limits. Let me know what you think.

damemi:
@seanmalloy  I was wondering if just setting the requests would be sufficient, I think for that we can definitely just go with reasonable values

seanmalloy:
@damemi and @ingvagabund I amended the commit to only sets request.

damemi (approved):
/approve
thanks @seanmalloy

seanmalloy:
Just need a LGTM from someone.

@ingvagabund or @lixiang233 please review when you have some time. Thanks!

/cc @lixiang233 @ingvagabund

PR #443, opened 2020-11-22
Set Container Resources In Helm Chart

seanmalloy:
Prior to this commit the helm chart used to install the descheduler
CronJob did not set container requests or limits. This is considered
an anti-pattern when deploying applications on k8s.

Set descheduler container resources to make it a burstable pod. This
will ensure a high quality experience for end users when deploying
descheduler into their clusters using the instructions from the README.

The default values choosen for CPU/Memory are not based on any real data.

### DIFF
--- a/charts/descheduler/README.md
+++ b/charts/descheduler/README.md
@@ -56,5 +56,7 @@ The following table lists the configurable parameters of the _descheduler_ chart
 | `priorityClassName`            | The name of the priority class to add to pods                                                                         | `system-cluster-critical`            |
 | `rbac.create`                  | If `true`, create & use RBAC resources                                                                                | `true`                               |
 | `podSecurityPolicy.create`     | If `true`, create PodSecurityPolicy                                                                                   | `true`                               |
+| `resources.cpuRequest`        | Descheduler container CPU request                                                                  | `500m`                               |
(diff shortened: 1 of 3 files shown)


seanmalloy:
This fixes the second part of #432.

/kind cleanup
/assign @damemi 
/cc @ingvagabund 

FYI ... @stevehipwell please take a look if you have some time. Thanks!

stevehipwell:
@seanmalloy it looks good in principal, I was actually waiting for everything to have settled down before putting in a PR for this as our Descheduler chart has the last best effort pod on our cluster.

I would strongly recommend using the the idiomatic way to set resources (see below). If you what to leave the defaults fully up to the user you can also comment out the `requests` block.

_values.yaml_
```yaml
resources:
  requests:
    cpu: 500m
    memory: 256Mi
  # limits:
  #   cpu: 100m
  #   memory: 128Mi
```

_cronjob.yaml_
```yaml
            resources:
              {{- toYaml .Values.resources | nindent 14 }}
```

seanmalloy:
I made the changes suggested on @stevehipwell. I amended the commit.

I have not tested it yet to verify it works.

/hold

seanmalloy:
Tested it. Works now.

/hold cancel

damemi (approved):
/approve

PR #448, opened 2020-12-01
Add more topology spread logs

damemi:
This adds a simple log message to indicate to users that the topology strategy is running (all other strategies log a message for each node they are processing, but since topology spread processes by namespace rather than node that would be a lot of messages).

Also added a log message when pods are skipped due to node selector/affinity

/priority low
/kind cleanup


damemi:
/cc @ingvagabund
decided to add this since we were talking about it earlier

PR #449, opened 2020-12-02
Enable Multi-Arch Container Image Builds

### LINKED ISSUE #445: ARM Container Image

I would like to run the descheduler on ARM(specifically ARM64) hardware. I'd like an official descheduler container image for this use case.

I would like the descheduler project to provide an official ARM64 container image in addition to the already provided x86_64 container image.

Not running descheduler on ARM.

descheduler version: v0.19.0

**Additional context**

The [k8s external-dns](https://github.com/kubernetes-sigs/external-dns) and [k8s cluster-proportional-autoscaler](https://github.com/kubernetes-sigs/cluster-proportional-autoscaler) projects already provide official ARM container images on k8s.gcr.io. These can be used as a reference when setting up the build automation for the descheduler.

seanmalloy:
Previous to this change official descheduler container images only
supported the AMD64 hardware architecture. This change enables
building official descheduler container images for multiple
architectures.

The initially supported architectures are AMD64 and ARM64.

Fixes #445

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,8 +15,9 @@ FROM golang:1.15.2
 
 WORKDIR /go/src/sigs.k8s.io/descheduler
 COPY . .
+ARG ARCH
 ARG VERSION
-RUN VERSION=${VERSION} make
+RUN VERSION=${VERSION} make build.$ARCH
 
 FROM scratch
 
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,7 @@
 VERSION?=v$(shell date +%Y%m%d)-$(shell git describe --tags --match "v*")
 BUILD=$(shell date +%FT%T%z)
 LDFLAG_LOCATION=sigs.k8s.io/descheduler/cmd/descheduler/app
+ARCHS = amd64 arm64
 
 LDFLAGS=-ldflags "-X ${LDFLAG_LOCATION}.version=${VERSION} -X ${LDFLAG_LOCATION}.buildDate=${BUILD}"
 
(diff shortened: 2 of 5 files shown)


seanmalloy:
The container build and push automation seems to be working fine, but I have not verified that the images actually work when run in a real cluster yet.

I'll remove the hold once I've verified that the container images are working in a kind cluster on AMD64 and ARM64.

/kind feature
/hold

seanmalloy:
I confirmed that everything is working as expected for the AMD64 image, so existing functionality is not broken. Testing on ARM64 is proving to be more difficult as it looks like kind does not run quite as easily on ARM64 as it does on AMD64.

If anyone reading this has a working k8s cluster on ARM64 and would be willing the test the container image that would be super helpful.

I'm removing the hold. I will try to get a k8s cluster running on ARM64, but this might take me some time. I understand the the project maintainers might be hesitant to merge this until the ARM64 image is confirmed working.

/assign @damemi 
/cc @ingvagabund

seanmalloy:
One note for @damemi is that when this merges the image promotion YAML file will need to look a little bit different. See below file for an example of what we will need to do.

https://github.com/kubernetes/k8s.io/blob/master/k8s.gcr.io/images/k8s-staging-cpa/images.yaml

seanmalloy:
Forget to remove the hold.

/hold cancel

damemi:
@seanmalloy will we need to update anything else to pass that `ARCH` arg, like during the automated builds? or do the names in that file you linked get parsed to that?

seanmalloy:
> @seanmalloy will we need to update anything else to pass that `ARCH` arg, like during the automated builds? or do the names in that file you linked get parsed to that?

All the changes required for the build automation pushing to the staging registry are included in this PR. Every time a PR is merged the automation will build three container images and push them to the staging registry. I tested this by running `REGISTRY=myawesomeprivategcr make push-all` which built and pushed the images to my private gcr.io registry.

When we do the promotion process for each descheduler release we will need to promote three container images instead of just one for each release.

End users don't need to know worry about which of the three images to pull. They will always reference the "main" image for example `k8s.gcr.io/descheduler/descheduler:v0.20.0`. If running on AMD64 it will pull the AMD64 image. If running on ARM64 it will pull the ARM64 image.

damemi (commented):
@seanmalloy thanks for clarifying, this lgtm
/approve

PR #451, opened 2020-12-02
bump k8s dependencies to 1.20-rc.0

damemi:


damemi (commented):
/cc @ingvagabund @seanmalloy 
just a pre-bump now that `rc.0` tags are available in all the dep repos, ptal

PR #453, opened 2020-12-03
Allow setting options successfulJobsHistoryLimit and failedJobsHistoryLimit for cronjob

dieterdemeyer:
Allow end users to optionally set the descheduler CronJob `.spec.successfulJobsHistoryLimit` and `.spec.failedJobsHistoryLimit` when installing using helm chart.

### DIFF
--- a/charts/descheduler/Chart.yaml
+++ b/charts/descheduler/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: descheduler
-version: 0.19.0
-appVersion: 0.19.0
+version: 0.20.0
+appVersion: 0.20.0
 description: Descheduler for Kubernetes is used to rebalance clusters by evicting pods that can potentially be scheduled on better nodes. In the current implementation, descheduler does not schedule replacement of evicted pods but relies on the default scheduler for that.
 keywords:
 - kubernetes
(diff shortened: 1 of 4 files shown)


charts/descheduler/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: descheduler
-version: 0.19.0
-appVersion: 0.19.0
+version: 0.20.0
+appVersion: 0.20.0
```
ingvagabund, line 4: descheduler 1.20 is not out yet, do you expect the PR to be merge after 1.20 is released?

dieterdemeyer: I was hoping that this could be merged and released in version `0.20.0`.
Should I not have modified this file ?

ingvagabund: That's still doable. It just feels strange to merge anything with 0.20 before the descheduler image for the release is published. I was just making sure :) No blocker here.

dieterdemeyer: If you want, I can revert the file to be more in line with other pull requests. It's a quick change..

ingvagabund: I will defer this to @damemi and @seanmalloy

damemi: Technically the master branch correlates to the upcoming release, so this is a change we'll have to make anyway before the release. I don't see a problem making that change now if @seanmalloy doesn't either

damemi (commented):
/approve
I'm fine with the changes here if it looks good to everyone else

seanmalloy (approved):

PR #454, opened 2020-12-04
Fix broken namespace logic in TopologySpreadConstraint

damemi:
This logic is broken, when passing in any list of excluded namespaces it will skip all namespaces since it just looks at the size of include+exclude together. This PR breaks them out into separate checks for sanity

/kind bug
/priority critical-urgent

### DIFF
--- a/pkg/descheduler/strategies/topologyspreadconstraint.go
+++ b/pkg/descheduler/strategies/topologyspreadconstraint.go
@@ -110,7 +110,8 @@ func RemovePodsViolatingTopologySpreadConstraint(
 	podsForEviction := make(map[*v1.Pod]struct{})
 	// 1. for each namespace...
 	for _, namespace := range namespaces.Items {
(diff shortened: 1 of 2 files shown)


damemi (commented):
/cc @ingvagabund @seanmalloy 
please take a look! I think I tried getting a little too fancy with this check when I first wrote it

seanmalloy:
This looks good to me. I added a hold just in case @ingvagabund or someone else wants to review before merging. Feel free to remove the hold when you are ready for merge.

/lgtm
/hold

damemi:
> This looks good to me. I added a hold just in case @ingvagabund or someone else wants to review before merging. Feel free to remove the hold when you are ready for merge.
> 
> /lgtm
> /hold

Jan is starting PTO today so may not see this. But since we were working on this earlier I'm comfortable removing the hold and letting this merge
/hold cancel

PR #455, opened 2020-12-08
Helm: allow the topologySpreadConstraint strategy to list namespaces

AmoVanB:
The `topologySpreadConstraint` strategy tries to list namespaces but the `ClusterRole` of the Helm chart does not allow it to. 

This PR adds this rule to the `ClusterRole` when the `topologySpreadConstraint` strategy is enabled.


damemi (approved):
/ok-to-test
/approve
thanks for catching this!

damemi:
@AmoVanB looks like there is a syntax error in your change, not sure what the fix is for helm here:
```
 helm lint ./charts/descheduler
==> Linting ./charts/descheduler
[ERROR] templates/: template: descheduler/templates/clusterrole.yaml:15:14: executing "descheduler/templates/clusterrole.yaml" at <.Values.deschedulerPolicy.strategies.RemovePodsViolatingTopologySpreadConstraint.enabled>: nil pointer evaluating interface {}.enabled 
```

charts/descheduler/templates/clusterrole.yaml
```diff
@@ -12,6 +12,11 @@ rules:
 - apiGroups: [""]
   resources: ["nodes"]
   verbs: ["get", "watch", "list"]
+{{- if .Values.deschedulerPolicy.strategies.RemovePodsViolatingTopologySpreadConstraint.enabled }}
```
seanmalloy, line 15: In my opinion the conditional is not needed. The namespace list permission can just always be included. It is already included in file `kubernetes/base/rbac.yaml`.

seanmalloy (changes requested):

charts/descheduler/templates/clusterrole.yaml
seanmalloy: https://github.com/kubernetes-sigs/descheduler/blob/70d1fadae750207acd6f7f3db07f868c2624b368/kubernetes/base/rbac.yaml#L13-L15

charts/descheduler/templates/clusterrole.yaml
```diff
@@ -12,6 +12,11 @@ rules:
...
+- apiGroups: [""]
+  resources: ["namespaces"]
+  verbs: ["list"]
```
seanmalloy, line 18:
```suggestion
  verbs: ["get", "list"]
```

charts/descheduler/templates/clusterrole.yaml
AmoVanB: sounds good to me, updating

seanmalloy (approved):

PR #457, opened 2020-12-08
Helm chart rename - Step #2

stevehipwell:
**IMPORTANT** - This needs to be merged after PR https://github.com/kubernetes-sigs/descheduler/pull/457 and before PR https://github.com/kubernetes-sigs/descheduler/pull/459.

This PR will release the v0.18 chart with the new _descheduler_ name.

PR #459, opened 2020-12-08
Helm chart rename - Step #3

stevehipwell:
IMPORTANT - This needs to be merged after PR #457 and PR #459.

This PR will release the v0.19 chart with the new _descheduler_ name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,43 +11,42 @@ a Kubernetes cluster at that point of time when a new pod appears for scheduling
 As Kubernetes clusters are very dynamic and their state changes over time, there may be desire
 to move already running pods to some other nodes for various reasons:
 
-* Some nodes are under or over utilized.
-* The original scheduling decision does not hold true any more, as taints or labels are added to
-or removed from nodes, pod/node affinity requirements are not satisfied any more.
-* Some nodes failed and their pods moved to other nodes.
-* New nodes are added to clusters.
+- Some nodes are under or over utilized.
+- The original scheduling decision does not hold true any more, as taints or labels are added to
+  or removed from nodes, pod/node affinity requirements are not satisfied any more.
+- Some nodes failed and their pods moved to other nodes.
+- New nodes are added to clusters.
 
 Consequently, there might be several pods scheduled on less desired nodes in a cluster.
 Descheduler, based on its policy, finds pods that can be moved and evicts them. Please
 note, in current implementation, descheduler does not schedule replacement of evicted pods
 but relies on the default scheduler for that.
 
-Table of Contents
-=================
+# Table of Contents
 
-  * [Quick Start](#quick-start)
-     * [Run As A Job](#run-as-a-job)
-     * [Run As A CronJob](#run-as-a-cronjob)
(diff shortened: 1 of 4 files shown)


damemi:
/hold
for steps 1 & 2

README.md
```diff
@@ -11,43 +11,42 @@ a Kubernetes cluster at that point of time when a new pod appears for scheduling
 As Kubernetes clusters are very dynamic and their state changes over time, there may be desire
 to move already running pods to some other nodes for various reasons:
 
-* Some nodes are under or over utilized.
-* The original scheduling decision does not hold true any more, as taints or labels are added to
-or removed from nodes, pod/node affinity requirements are not satisfied any more.
-* Some nodes failed and their pods moved to other nodes.
-* New nodes are added to clusters.
+- Some nodes are under or over utilized.
```
damemi, line 14: There's a lot of odd updates to the README.md in this PR, are these intentional or maybe an editor auto-formatting? Would like to undo these if possible

damemi (changes requested):
looks like steps 1 & 2 worked out, the index.yaml is getting updated and the 1.18 release still the right naming scheme, so I think we can keep going with this PR

@stevehipwell thanks again for your help so far, could you just revert the changes in this PR to README.md? It seems like they were unintentional

stevehipwell:
@damemi I made a change in the README but it looks like my formatter tidied it up too. I'm not by my machine at the moment so is it possible to take the change or do I need to update the PR? This is time sensitive as v0.19 will be missing from the hub so if I need to I can get back to my machine.

damemi:
@stevehipwell I suppose those changes aren't a big problem, and it does correct some formatting that was messed up (like extra backtics in code blocks)

Since this is time sensitive, I'm fine merging the changes in formatting now too.
/hold cancel

Once this merges, please check that everything looks good on your end too and let us know. Thank you!

stevehipwell:
@damemi this all looks good to me.

stevehipwell:
@damemi it looks like the GitHub release action has failed for the v0.19 release.

damemi:
@stevehipwell thanks, looks like a CI check was blocking it. Re-ran and it looks like it completed

stevehipwell:
@damemi I think you either need to delete the `descheduler-helm-chart-0.19.2` tag and then re run the action or manually modify the _index.yaml_ file on the `gh-pages` branch; as it looks like the failure wasn't cleaned up correctly so resulted in a NOP.

damemi:
@stevehipwell oh, that makes sense. I was able to delete the branch/release and re-run the job. The right version appears to be in index.yaml now: https://github.com/kubernetes-sigs/descheduler/blob/97a0ab9/index.yaml

and the release is back https://github.com/kubernetes-sigs/descheduler/releases/tag/descheduler-helm-chart-0.19.2

If you can double check that I got this right and I'm not missing anything else that would be great!

stevehipwell:
@damemi I can now see the v0.19 chart on artifacthub.io so it looks good to me.

PR #436, opened 2020-11-09
Rename Helm chart

stevehipwell:
Now that the [helm/chart-releaser](https://github.com/helm/chart-releaser/releases/tag/v1.1.0) and [helm/chart-releaser-action](https://github.com/helm/chart-releaser-action/releases/tag/v1.1.0) can support creating a custom tag name for helm chart release the _descheduler_ helm chart no longer needs to be named to avoid tag collisions.

This PR updates the CI tooling for helm, renames the _descheduler_ helm chart to `descheduler` from `descheduler-helm-chart` and updates the related documentation.

On applying this merge request the existing helm charts will not be modified or removed and only future releases will be impacted. My suggestion would be to use the [helm/chart-releaser](https://github.com/helm/chart-releaser) with the `--commit` flag to re-create the existing charts with the new name in parallel.

PR #460, opened 2020-12-09
Delete everything but index.yaml from gh-pages

damemi:
This deletes everything from the `gh-pages` branch except for *index.yaml* which is used to host our helm chart index, as recommended in https://github.com/kubernetes-sigs/descheduler/pull/436#issuecomment-729870148


damemi (commented):
/hold
this isn't a total blocker now, just a cleanup. I would also like some other reviewers to double check this change for my sanity :)

/kind cleanup
/cc @seanmalloy 
@stevehipwell

seanmalloy:
This looking correct to me.

1. Merging into the correct branch(gh-pages)
2. All files deleted except for index.yaml
3. No changes to index.yaml

/lgtm

stevehipwell:
Looks correct to me.

damemi:
/hold cancel
thanks for the review

PR #269, opened 2020-04-22
review duplicates strategies

safanaj:

PR #463, opened 2020-12-09
RemoveDuplicatePods: evict uniformly

ingvagabund:
Fixes: #263

Previous work:
- https://github.com/kubernetes-sigs/descheduler/pull/269
- https://github.com/kubernetes-sigs/descheduler/pull/398
- https://github.com/kubernetes-sigs/descheduler/pull/420

### DIFF
--- a/pkg/descheduler/strategies/duplicates.go
+++ b/pkg/descheduler/strategies/duplicates.go
@@ -19,6 +19,7 @@ package strategies
 import (
 	"context"
 	"fmt"
+	"math"
 	"reflect"
 	"sort"
 	"strings"
@@ -50,6 +51,11 @@ func validateRemoveDuplicatePodsParams(params *api.StrategyParameters) error {
 	return nil
 }
 
+type podOwner struct {
+	namespace, kind, name string
+	imagesHash            string
+}
+
 // RemoveDuplicatePods removes the duplicate pods on node. This strategy evicts all duplicate pods on node.
 // A pod is said to be a duplicate of other if both of them are from same creator, kind and are within the same
 // namespace, and have at least one container with the same image.
@@ -79,6 +85,11 @@ func RemoveDuplicatePods(
 
 	evictable := podEvictor.Evictable(evictions.WithPriorityThreshold(thresholdPriority))
 
+	duplicatePods := make(map[podOwner]map[string][]*v1.Pod)
+	ownerKeyOccurence := make(map[podOwner]int32)
+	nodeCount := 0
+	nodeMap := make(map[string]*v1.Node)
+
 	for _, node := range nodes {
 		klog.V(1).InfoS("Processing node", "node", klog.KObj(node))
 		pods, err := podutil.ListPodsOnANode(ctx,
@@ -92,8 +103,8 @@ func RemoveDuplicatePods(
 			klog.ErrorS(err, "Error listing evictable pods on node", "node", klog.KObj(node))
 			continue
 		}
-
-		duplicatePods := make([]*v1.Pod, 0, len(pods))
+		nodeMap[node.Name] = node
+		nodeCount++
 		// Each pod has a list of owners and a list of containers, and each container has 1 image spec.
 		// For each pod, we go through all the OwnerRef/Image mappings and represent them as a "key" string.
 		// All of those mappings together makes a list of "key" strings that essentially represent that pod's uniqueness.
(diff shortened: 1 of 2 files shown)


pkg/descheduler/strategies/duplicates.go
```diff
@@ -50,6 +51,10 @@ func validateRemoveDuplicatePodsParams(params *api.StrategyParameters) error {
 	return nil
 }
 
+type podOwner struct {
```
damemi, line 54: could you just use `metav1.OwnerReference` for this?

pkg/descheduler/strategies/duplicates.go
```diff
@@ -131,7 +143,14 @@ func RemoveDuplicatePods(
...
 					if reflect.DeepEqual(keys, podContainerKeys) {
 						matched = true
-						duplicatePods = append(duplicatePods, pod)
+						klog.V(3).InfoS("Duplicate found", "pod", klog.KObj(pod))
+						for _, ownerRef := range ownerRefList {
+							ownerKey := podOwner{pod.ObjectMeta.Namespace, ownerRef.Kind, ownerRef.Name}
+							if _, ok := duplicatePods[ownerKey]; !ok {
+								duplicatePods[ownerKey] = make(map[string][]*v1.Pod)
+							}
+							duplicatePods[ownerKey][node.Name] = append(duplicatePods[ownerKey][node.Name], pod)
```
damemi, line 152: will this work if 2 different pods have the same owner?

damemi (commented):
I like this approach, it is simple and solves the problem in a good way ("if any node has more than ceil(average), evict down to that"). 

I am a little concerned about keying your map on the owner though, because you could get multiple pods falling under the same owner. I wonder if that is necessary to track, or if you could just key it on something else. Maybe you have already considered that though and I'm overlooking it?

ingvagabund:
> I am a little concerned about keying your map on the owner though, because you could get multiple pods falling under the same owner. I wonder if that is necessary to track, or if you could just key it on something else. Maybe you have already considered that though and I'm overlooking it?

I considered this case as well. Unfortunately, it requires a different approach and cross-comparing list of duplicates to find the biggest intersection of duplicates to evict. On the other hand, even with the current implementation (without this PR) there's no guarantee of evicting more pods due to random ordering of pods in duplicates lists. This is again best-effort approach. For case of a pod (or a set of pods) being owned by two entities, we can implement the logic either after 1.20 or wait until it's requested.

pkg/descheduler/strategies/duplicates.go
```diff
@@ -79,8 +84,13 @@ func RemoveDuplicatePods(
...
 	evictable := podEvictor.Evictable(evictions.WithPriorityThreshold(thresholdPriority))
 
+	duplicatePods := make(map[podOwner]map[string][]*v1.Pod)
+	ownerKeyOccurence := make(map[podOwner]int32)
+	nodeCount := 0
+	nodeMap := make(map[string]*v1.Node)
+
 	for _, node := range nodes {
-		klog.V(1).InfoS("Processing node", "node", klog.KObj(node))
+		klog.V(0).InfoS("Processing node", "node", klog.KObj(node))
```
damemi, line 93: why change this?

pkg/descheduler/strategies/duplicates.go
damemi: To show what I mean, see the PR I opened here (picked your branch and added a test case that fails): https://github.com/kubernetes-sigs/descheduler/pull/467/commits/71a2a7e9e6556579cca4bf29b584797251db9c79

say we have 4 pods with the same ownerref, but only 2 of them are actually duplicates. with how you currently count occurrences, the average gets calculated out to 2 for this ownerKey, but it should be 1 because ownerKey counts are unique to the image

I think we can still fix up this PR, but we need to refactor how you're counting the occurrences of each, perhaps to just include the image? Something like that could fix it

pkg/descheduler/strategies/duplicates.go
ingvagabund: `metav1.OwnerReference` does not have a namespace field

pkg/descheduler/strategies/duplicates.go
ingvagabund: leftover from running unit tests, sorry, undone

pkg/descheduler/strategies/duplicates.go
ingvagabund: Nice catch!!! PR updated.

pkg/descheduler/strategies/duplicates.go
damemi: It does have a `uid` though, which should be similarly unique right?

damemi:
Adding the `imagesHash` to the ownerKey looks like it solves the issue I was worried about (and I see now the test case I showed was included 👍). With that I think this is good to merge.

As a separate todo, we should think about refactoring this strategy. It was pretty easy to understand before we had to add the edge case for different images/same owner which made it complicated (but that was mostly me, so I might have just written it poorly...)

/approve
thanks @ingvagabund !

seanmalloy (approved):

Evilweebl:
Hi I've just came up against this exact issue and I can confirm 0.20 fixes the issue when I have a higher replica count than nodes BUT it does not take into account nodes that are cordoned. If I have a replica set of 3, 2 active nodes and 1 cordoned node I still get the issue with a continuous loop destroying my pod.

seanmalloy:
> Hi I've just came up against this exact issue and I can confirm 0.20 fixes the issue when I have a higher replica count than nodes BUT it does not take into account nodes that are cordoned. If I have a replica set of 3, 2 active nodes and 1 cordoned node I still get the issue with a continuous loop destroying my pod.

@Evilweebl can you open a separate bug for the node cordon problem?

damemi:
Did we open an issue to address https://github.com/kubernetes-sigs/descheduler/pull/463#issuecomment-749550117 or do we still need one?

seanmalloy:
> Did we open an issue to address [#463 (comment)](https://github.com/kubernetes-sigs/descheduler/pull/463#issuecomment-749550117) or do we still need one?

I created #531 just now.

PR #465, opened 2020-12-09
Skip topology calculations if domains are already balanced

damemi:
This just adds a shortcut to skip doing all the sorting/balancing calculations if all domains in a topology for a constraint are already within the MaxSkew of each other

/kind cleanup


damemi:
/close
going to reopene, master-1-17 test was deleted

lixiang233 (commented):
looks good, it makes sense to avoid some useless calculations.
/lgtm

PR #468, opened 2020-12-10
Run as user 1000 so pod can be runAsNonRoot for security purposes

eatwithforks:
Can add this to the cronjob/job kubernetes manifests
```
securityContext:
  runAsNonRoot: true
  readOnlyRootFilesystem: true
```

/cc @seanmalloy

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,6 +23,8 @@ FROM scratch
 
 MAINTAINER Avesh Agarwal <avesh.ncsu@gmail.com>
 
+USER 1000
+
 COPY --from=0 /go/src/sigs.k8s.io/descheduler/_output/bin/descheduler /bin/descheduler
 
 CMD ["/bin/descheduler", "--help"]
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -15,6 +15,8 @@ FROM scratch
 
 MAINTAINER Avesh Agarwal <avesh.ncsu@gmail.com>
 
+USER 1000
+
 COPY _output/bin/descheduler /bin/descheduler
 
 CMD ["/bin/descheduler", "--help"]


damemi (commented):
/ok-to-test

seanmalloy:
I was able to build the container image on my laptop and run it in a kind cluster. Someone can update the manifests in the `kubernetes` folder and update the helm chart to drop all privileges after this PR is merged.

/lgtm
/assign @damemi

damemi:
> Someone can update the manifests in the `kubernetes` folder and update the helm chart to drop all privileges after this PR is merged.

Are you referring to the k8s rbac rules? Those aren't affected by the container user are they?

seanmalloy:
> > Someone can update the manifests in the `kubernetes` folder and update the helm chart to drop all privileges after this PR is merged.
> 
> Are you referring to the k8s rbac rules? Those aren't affected by the container user are they?

Correct, we do not need to change any k8s RBAC. What I meant to say was we can now add a `securityContext` to the Pod spec for the Job and CronJob in the `kubernetes` directory. This would then default to dropping all privileges(i.e. not running as root).

Something like ...
```
$ git diff
diff --git a/kubernetes/job/job.yaml b/kubernetes/job/job.yaml
index a478f4e5b..b58d30eec 100644
--- a/kubernetes/job/job.yaml
+++ b/kubernetes/job/job.yaml
@@ -29,6 +29,14 @@ spec:
             requests:
               cpu: "500m"
               memory: "256Mi"
+          securityContext:
+            allowPrivilegeEscalation: false
+            capabilities:
+              drop:
+                - ALL
+            privileged: false
+            readOnlyRootFilesystem: true
+            runAsNonRoot: true
       restartPolicy: "Never"
       serviceAccountName: descheduler-sa
       volumes:
```

We can also review if any changes are needed to the helm chart to make it drop additional privileges. The PodSecurityPolicy that we recent added looks pretty good, but by default the chart does not deploy it. Maybe we could consider a default least privileged deployment via helm after this PR merges.

damemi:
Ah I see @seanmalloy, thanks for clarifying I was just confused
/approve

Issue #469, opened 2020-12-10
NodeSelector may cause descheduler to abort when unnecessary


damemi:
**What version of descheduler are you using?**

descheduler version: latest

**Which descheduler CLI options are you using?**
--node-selector

**What did you do?**
When using `--node-selector` to restrict descheduler to just 1 node, [this section of code](https://github.com/kubernetes-sigs/descheduler/blob/4108362/pkg/descheduler/descheduler.go#L99-L110) prevents the descheduler from running with:

```
I1210 17:32:36.835916       1 descheduler.go:107] "The cluster size is 0 or 1 meaning eviction causes service disruption or degradation. So aborting.."
```

**What did you expect to see?**
Even if the cluster size is actually larger than 1, but we only want to restrict evictions to 1 node (such as only preventing duplicates from existing on 1 node) this failure still occurs. This seems to be due to `nodeSelector` being passed to the original node list that is used in this check.

**What did you see instead?**
I should still be able to run descheduler against a single node out of my cluster, because I may have other nodes available for evicted pods that will not cause degradation

lixiang233:
Yes, descheduler should work normally if we got more than one node in cluster available even if we only selected 1 node. I can help fix this, and will submit PR recently.
/assign

aveshagarwal:
node selector above is working as expected as it was initially designed to select smaller set of nodes from all nodes, and that means only those nodes are the intended nodes not only for eviction but also for where evicted pods will land. 
So when node selector selects only 1 node, that is like having 1-node cluster, and thats why it does not process further. 

To give some context, Descheduler was fundamentally designed to balance nodes in an imbalanced cluster, for example in case of LowNodeUtilization, it computes resources from all selected nodes, most likely all nodes, or could be smaller set of nodes as selected by node selector. This functionality with node selector was needed in multi-tenant scenarios for OpenShift.

If there is need to change the behavior of node selector, i.e., for having different behavior for different strategies,  that is fine, but looks like the above change could break LowNodeUtilization strategy.

damemi:
@aveshagarwal that is a good point, that having a node selector over only 1 node would hurt the performance of LowNodeUtilization specifically. However, I believe the error that's logged should be updated to reflect that, because running other basic strategies against only 1 node in the cluster will not inherently cause service disruption if there are other nodes in the cluster.

[...]

PR #470, opened 2020-12-11
Avoid unnecessary aborting

lixiang233:
fixes: #469

### DIFF
--- a/pkg/descheduler/descheduler.go
+++ b/pkg/descheduler/descheduler.go
@@ -96,15 +96,16 @@ func RunDeschedulerStrategies(ctx context.Context, rs *options.DeschedulerServer
 	}
 
 	wait.Until(func() {
-		nodes, err := nodeutil.ReadyNodes(ctx, rs.Client, nodeInformer, nodeSelector, stopChannel)
+		nodes, totalReadyNodeNum, err := nodeutil.ReadyNodes(ctx, rs.Client, nodeInformer, nodeSelector)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


aveshagarwal:
/hold

I will leave a comment in the issue to describe what was the intended design of having node selector initially.

aveshagarwal:
Feel free to cancel hold, as far as the intended design of node selector is clear and any change does not affect some of the strategies like LowNodeUtilization.

damemi (commented):
I agree that we should think more about this, because the points @aveshagarwal raised in the issue about this affecting certain strategies are valid.

What if we log a warningf that running LowNodeUtilization over 1 node-selector is not performant? We could also disable LNU if it's enabled.

Are there any other strategies we should take into consideration that would be hurt by this?

lixiang233:
I think we do not need this PR given #488  has a better way to do this.
/close

damemi:
@lixiang233 does #488 affect the questions here? I thought this PR was more about considering the difference between node selector scope for various strategies

lixiang233:
@damemi  I think we can list and filter nodes in each strategies when running descheduler in `Informed` mode, so every strategies can custom `nodeSelector` and handle boundary cases as needed. Or if we want to add this feature for `Default` mode too, I can reopen this PR and complete it as what was suggested in [this comment](https://github.com/kubernetes-sigs/descheduler/issues/469#issuecomment-763413379).

damemi:
Ah, that's fair enough. I think it would be worth it to fix this up in both approaches, but overall this isn't a critical bug

PR #434, opened 2020-11-04
Support extended resources in LowNodeUtilization

ZongqiangZhang:
Fixes #399 

Support extended/custom resources, so as to evict pods based on generic resource including but not limited to (cpu, memory, pods)

PR #473, opened 2020-12-16
Add a parameter for making the thresholds of the LowNodeUtilization strategy relative to average values

AmoVanB:
### Reason for this PR

Currently, the thresholds that can be set for the `LodNodeUtilization` strategy only allow to make sure that nodes aren't over or underutilized compared to thresholds relative to their capacities. However, sometimes, it's desirable to make sure all nodes are "similarly" utilized to ensure the proper balance of a cluster. While this is possible to implement using the current thresholds, the values needed to achieve such a balancing behavior would be dependent on the number of nodes, pods, and their memory/CPU requests, which is not very practical to configure, especially for dynamic clusters. 

### Feature of this PR

This PR enables the `descheduler` to achieve such a goal. We define an additional `useDeviationThresholds` boolean parameter (the name is maybe not the best, it can change). If `false`, the `descheduler` behaves as now. If `true`, the thresholds are considered as percentage deviations relative to the average utilizations among all nodes. 

For example. Considering only the memory metric. If nodes have utilizations of _[10%, 20%, 30%, 20%]_ and both `threshold` and `targetThreshold` are 5%. The average node utilization is _(10% + 20% + 30% + 20%) / 4 = 20%_. The first node has an utilization of 10%, which is lower than 20% (_average_) - 5% (_threshold_). The node is hence considered underutilized. The third node has an utilization of 30%, which is greater than 20% (_average_) + 5% (_targetThreshold_). The node is hence considered overutilized. Both other nodes have utilizations within the [20% - 5%, 20% + 5%] window and are hence considered appropriately utilized. 

Once nodes are labeled as _under-_ or _overutilized_, the strategy behaves exactly the same as before. 

### Code style comment

Since the only change compared to the original strategy is the way nodes are categorized as _over-_ and _underutilized_, I thought it is best to include that in the existing `LowNodeUtilization` strategy and not to create a new strategy. Another option would be to create two distinct strategies that share methods. I went for the first option as that was the easiest to implement. I would definitely not be against the second option.

### Test

We had this need/problem in our 200+ pods cluster. We deployed the `descheduler` from this PR with success in our test clusters. It passes our tests (that make sure that the correct pods are evicted when the cluster is unbalanced) and works as expected and without any detected issues for now.

### Vision

In general, the proposed feature goes into the direction of a more general intention we have with the `descheduler`: take the state of the cluster, smartly compute a new schedule for all the pods (not one by one but all at once to leverage the complete knowledge we have) according to a given strategy (ours: balanced utilization), and then implementing the transition from the current state to the desired state. This PR is our first (small) step towards that goal but is there such a plan for the `descheduler`? Overall, that  would turn the `descheduler` into a "cluster scheduler" in contrast to the current "per-pod scheduler" of Kubernetes. I know there is a plan to eventually incorporate scheduling in the tool and not to rely on the existing scheduler, but not sure if you're looking into such optimization-driven strategies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,10 +176,16 @@ can be configured for cpu, memory, and number of pods too in terms of percentage
 
 These thresholds, `thresholds` and `targetThresholds`, could be tuned as per your cluster requirements.
 
+Additionally, the strategy accepts a `useDeviationThresholds` parameter.
+If that parameter is set to `true`, the thresholds are considered as percentage deviations from mean resource usage.
+`thresholds` will be deducted from the mean among all nodes and `targetThresholds` will be added to the mean.
+A resource consumption above (resp. below) this window is considered as overutilization (resp. underutilization).
+
 **Parameters:**
 
 |Name|Type|
 |---|---|
+|`useDeviationThresholds`|bool|
 |`thresholds`|map(string:int)|
 |`targetThresholds`|map(string:int)|
 |`numberOfNodes`|int|
--- a/pkg/descheduler/strategies/lownodeutilization.go
+++ b/pkg/descheduler/strategies/lownodeutilization.go
@@ -78,43 +78,45 @@ func LowNodeUtilization(ctx context.Context, client clientset.Interface, strateg
 		return
 	}
 
-	thresholds := strategy.Params.NodeResourceUtilizationThresholds.Thresholds
-	targetThresholds := strategy.Params.NodeResourceUtilizationThresholds.TargetThresholds
-	if err := validateStrategyConfig(thresholds, targetThresholds); err != nil {
+	strategyConfig := strategy.Params.NodeResourceUtilizationThresholds
+	if err := validateStrategyConfig(strategyConfig); err != nil {
 		klog.ErrorS(err, "LowNodeUtilization config is not valid")
 		return
 	}
 	// check if Pods/CPU/Mem are set, if not, set them to 100
-	if _, ok := thresholds[v1.ResourcePods]; !ok {
-		thresholds[v1.ResourcePods] = MaxResourcePercentage
-		targetThresholds[v1.ResourcePods] = MaxResourcePercentage
-	}
-	if _, ok := thresholds[v1.ResourceCPU]; !ok {
-		thresholds[v1.ResourceCPU] = MaxResourcePercentage
-		targetThresholds[v1.ResourceCPU] = MaxResourcePercentage
-	}
-	if _, ok := thresholds[v1.ResourceMemory]; !ok {
-		thresholds[v1.ResourceMemory] = MaxResourcePercentage
-		targetThresholds[v1.ResourceMemory] = MaxResourcePercentage
+	// for % deviations, we'll handle that in the getNodeUsage function
+	if !strategyConfig.UseDeviationThresholds {
+		if _, ok := strategyConfig.Thresholds[v1.ResourcePods]; !ok {
+			strategyConfig.Thresholds[v1.ResourcePods] = MaxResourcePercentage
+			strategyConfig.TargetThresholds[v1.ResourcePods] = MaxResourcePercentage
+		}
+		if _, ok := strategyConfig.Thresholds[v1.ResourceCPU]; !ok {
+			strategyConfig.Thresholds[v1.ResourceCPU] = MaxResourcePercentage
+			strategyConfig.TargetThresholds[v1.ResourceCPU] = MaxResourcePercentage
+		}
+		if _, ok := strategyConfig.Thresholds[v1.ResourceMemory]; !ok {
+			strategyConfig.Thresholds[v1.ResourceMemory] = MaxResourcePercentage
+			strategyConfig.TargetThresholds[v1.ResourceMemory] = MaxResourcePercentage
+		}
 	}
 
 	lowNodes, targetNodes := classifyNodes(
-		getNodeUsage(ctx, client, nodes, thresholds, targetThresholds),
+		getNodeUsage(ctx, client, nodes, strategyConfig),
 		// The node has to be schedulable (to be able to move workload there)
-		func(node *v1.Node, usage NodeUsage) bool {
+		func(node *v1.Node, usage *NodeUsage) bool {
 			if nodeutil.IsNodeUnschedulable(node) {
 				klog.V(2).InfoS("Node is unschedulable, thus not considered as underutilized", "node", klog.KObj(node))
 				return false
 			}
 			return isNodeWithLowUtilization(usage)
 		},
-		func(node *v1.Node, usage NodeUsage) bool {
+		func(node *v1.Node, usage *NodeUsage) bool {
 			return isNodeAboveTargetUtilization(usage)
 		},
 	)
 
 	klog.V(1).InfoS("Criteria for a node under utilization",
-		"CPU", thresholds[v1.ResourceCPU], "Mem", thresholds[v1.ResourceMemory], "Pods", thresholds[v1.ResourcePods])
+		"CPU", strategyConfig.Thresholds[v1.ResourceCPU], "Mem", strategyConfig.Thresholds[v1.ResourceMemory], "Pods", strategyConfig.Thresholds[v1.ResourcePods])
 
 	if len(lowNodes) == 0 {
 		klog.V(1).InfoS("No node is underutilized, nothing to do here, you might tune your thresholds further")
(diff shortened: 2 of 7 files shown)


pkg/apis/componentconfig/v1alpha1/types.go
```diff
@@ -48,4 +49,8 @@ type DeschedulerConfiguration struct {
 
 	// EvictLocalStoragePods allows pods using local storage to be evicted.
 	EvictLocalStoragePods bool `json:"evictLocalStoragePods,omitempty"`
+
+	// Logging specifies the options of logging.
+	// Refer [Logs Options](https://github.com/kubernetes/component-base/blob/master/logs/options.go) for more information.
+	Logging componentbaseconfig.LoggingConfiguration `json:"logging,omitempty"`
```
AmoVanB, line 55: I was getting a complaint from `update-generated-conversions.sh` without that. Not sure if it's really needed though.

pkg/descheduler/strategies/lownodeutilization.go
```diff
@@ -215,30 +220,96 @@ func getNodeUsage(
...
+		if strategyConfig.UseDeviationThresholds {
+			// we'll have to compute the absolute thresholds later because
+			// we need to compute the averages over all nodes
+			totalCPUUsagePercent += float64(usage[v1.ResourceCPU].MilliValue()) / float64(nodeCapacity.Cpu().MilliValue()) * 100.0
+			totalMemoryUsagePercent += float64(usage[v1.ResourceMemory].Value()) / float64(nodeCapacity.Memory().Value()) * 100.0
+			totalPodsUsagePercent += float64(usage[v1.ResourcePods].Value()) / float64(nodeCapacity.Pods().Value()) * 100.0
+		} else {
+			lowResourceThreshold = map[v1.ResourceName]*resource.Quantity{
+				v1.ResourceCPU:    resource.NewMilliQuantity(int64(float64(strategyConfig.Thresholds[v1.ResourceCPU])*float64(nodeCapacity.Cpu().MilliValue())*0.01), resource.DecimalSI),
+				v1.ResourceMemory: resource.NewQuantity(int64(float64(strategyConfig.Thresholds[v1.ResourceMemory])*float64(nodeCapacity.Memory().Value())*0.01), resource.BinarySI),
```
AmoVanB, line 240: Is `resource.BinarySI` correct here? I took that from the existing code, but I'm surprised that this is different from `resource.DecimalSI` for the CPU and pods. Not sure about the purpose/meaning of that parameter though.

pkg/descheduler/strategies/lownodeutilization_test.go
```diff
@@ -103,7 +104,7 @@ func TestLowNodeUtilization(t *testing.T) {
 				},
 				n2NodeName: {
 					Items: []v1.Pod{
-						*test.BuildTestPod("p9", 400, 0, n1NodeName, test.SetRSOwnerRef),
+						*test.BuildTestPod("p9", 400, 0, n2NodeName, test.SetRSOwnerRef),
```
AmoVanB, line 107: Here and in the 3 following changes, I thought that was a typo as the pod is actually supposed to be on node 2. Not sure if that is either important or correct.

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -125,19 +127,16 @@ func RemovePodsViolatingTopologySpreadConstraint(
...
 				namespaceTopologySpreadConstraints[constraint] = struct{}{}
 			}
 		}
 		if len(namespaceTopologySpreadConstraints) == 0 {
+			klog.V(1).InfoS("No constraint to check, skipping namespace", "namespace", namespace.Name)
 			continue
 		}
 
 		// 2. for each topologySpreadConstraint in that namespace
+		klog.V(1).InfoS("Processing "+strconv.Itoa(len(namespaceTopologySpreadConstraints))+" constraints", "namespace", namespace.Name)
```
AmoVanB, line 139: Just added some logs which I think are very helpful for simple debug.

lixiang233 (commented):
Hi @AmoVanB. Thanks for your PR.

If we run `lowNodeUtilization` periodically in a dynamic cluster, we may need to change `Thresholds` or `TargetThresholds` once the total utilization of the cluster changes. So it makes sense to make `Thresholds` and `TargetThresholds` change with the average resources usage of the cluster.

/ok-to-test
/kind feature

pkg/descheduler/strategies/lownodeutilization.go
lixiang233: Yes, they're different because for CPU, 5 cores = 5 * 1000m, but for Memory, 5GiB = 5 * 1024 MiB.

README.md
```diff
@@ -176,10 +176,16 @@ can be configured for cpu, memory, and number of pods too in terms of percentage
...
+`thresholds` will be deducted from the mean among all nodes and `targetThresholds` will be added to the mean.
+A resource consumption above (resp. below) this window is considered as overutilization (resp. underutilization).
+
 **Parameters:**
 
 |Name|Type|
 |---|---|
+|`useDeviationThresholds`|bool|
 |`thresholds`|map(string:int)|
 |`targetThresholds`|map(string:int)|
```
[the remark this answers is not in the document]
ingvagabund: Yeah, we were discussing renaming the params few times back. Wrt. bumping version of the strategy config as well. E.g. reorganize the data type so we can allow to configure the same strategy multiple times with different configuration.

pkg/descheduler/strategies/lownodeutilization.go
```diff
@@ -78,43 +78,45 @@ func LowNodeUtilization(ctx context.Context, client clientset.Interface, strateg
 		return
 	}
 
-	thresholds := strategy.Params.NodeResourceUtilizationThresholds.Thresholds
-	targetThresholds := strategy.Params.NodeResourceUtilizationThresholds.TargetThresholds
-	if err := validateStrategyConfig(thresholds, targetThresholds); err != nil {
+	strategyConfig := strategy.Params.NodeResourceUtilizationThresholds
+	if err := validateStrategyConfig(strategyConfig); err != nil {
```
ingvagabund, line 82: Please put the changes generalizing `thresholds` and `targetThresholds` into `strategyConfig.thresholds` and `strategyConfig.targetThresholds` into a separate commit. It will be easier to follow the changes.

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -114,9 +115,10 @@ func RemovePodsViolatingTopologySpreadConstraint(
 			(len(excludedNamespaces) > 0 && excludedNamespaces.Has(namespace.Name)) {
 			continue
 		}
+		klog.V(1).InfoS("Processing namespace", "namespace", namespace.Name)
```
ingvagabund, line 118: Please put changes in TSC into its own PR as it's not related to the deviation feature.

pkg/descheduler/strategies/lownodeutilization_test.go
ingvagabund: Please put this changes into a separate PR.

ingvagabund (changes requested):
Some suggestions before a proper review of this PR. I like the idea in overall.

ingvagabund:
We should get https://github.com/kubernetes-sigs/descheduler/pull/434 merged first though.

ingvagabund:
@AmoVanB https://github.com/kubernetes-sigs/descheduler/pull/434 merged

AmoVanB:
I'm not working on that anymore. I tried to rebase but there are a couple of non-obvious conflicts. I unfortunately won't have time to work on rebasing and fixing that. Hopefully someone can take over! :)

matthieu-eck:
Thanks @AmoVanB, I will take over from here ;) @ingvagabund, Shall I open another PR with the required changes?

ingvagabund:
@matthieu-eck by all means. Thank you for continuing the work @AmoVanB started!!!

pkg/descheduler/strategies/lownodeutilization_test.go
matthieu-eck: Already fixed on master by this [commit](https://github.com/kubernetes-sigs/descheduler/commit/81b816d4a4ff6417009cd7bd384d0da87b802b44#diff-67abe73ca76878708a926cd707d8d6a205acc3706a83802459b0a5082234168fL166).

ingvagabund:
Closing in favor of https://github.com/kubernetes-sigs/descheduler/pull/751
/close

Issue #461, opened 2020-12-09
RemovePodsViolatingTopologySpreadConstraint Support Soft Constraints


seanmalloy:
**Is your feature request related to a problem? Please describe.**

As of the soon to be released descheduler version v0.20.0 the `RemovePodsViolatingTopologySpreadConstraint` strategy only supports evicting pods that are using [Pod Topology Spread](https://kubernetes.io/docs/concepts/workloads/pods/pod-topology-spread-constraints/) hard constraints.

It would be nice if descheduler also support evicting pods that use Pod Topology Spread soft constraints.

**Describe the solution you'd like**

Descheduler supports evicting pods that are using Pod Topology Spread soft constraints.

**Describe alternatives you've considered**

Pods using Pod Topology Spread soft constraints are not always balanced at runtime.

**What version of descheduler are you using?**

descheduler version: v0.20.0

**Additional context**

Slack thread: https://kubernetes.slack.com/archives/C09TP78DV/p1607421341263300
PR review comment in initial implementation: https://github.com/kubernetes-sigs/descheduler/pull/413#discussion_r499824758

AmoVanB:
Simply removing the following check works fine in our test clusters:

```diff
diff --git a/pkg/descheduler/strategies/topologyspreadconstraint.go b/pkg/descheduler/strategies/topologyspreadconstraint.go
index 8c5067311..f9c72bd4c 100644
--- a/pkg/descheduler/strategies/topologyspreadconstraint.go
+++ b/pkg/descheduler/strategies/topologyspreadconstraint.go
@@ -127,11 +127,6 @@ func RemovePodsViolatingTopologySpreadConstraint(
                namespaceTopologySpreadConstraints := make(map[v1.TopologySpreadConstraint]struct{})
                for _, pod := range namespacePods.Items {
                        for _, constraint := range pod.Spec.TopologySpreadConstraints {
-                               // Only deal with hard topology constraints
-                               // TODO(@damemi): add support for soft constraints
-                               if constraint.WhenUnsatisfiable != v1.DoNotSchedule {
-                                       continue
-                               }
                                namespaceTopologySpreadConstraints[constraint] = struct{}{}
                        }
                }
```

We however only have soft constraints, so nothing guaranteed regarding the cohabitation of both types of constraints

lixiang233:
We can add a parameter to let users decide whether to handle soft rules, the strategy may look like:
```yaml
apiVersion: "descheduler/v1alpha1"
[...]

Issue #431, opened 2020-10-23
Use componentconfig api


damemi:
In https://github.com/kubernetes-sigs/descheduler/pull/380, we deprecated some component-level flags in favor of new policy config fields. Though, looking at our [ComponentConfig API](https://github.com/kubernetes-sigs/descheduler/blob/6240aa68f77a3e4f48413eb4c12962807b0b61da/pkg/apis/componentconfig/v1alpha1/types.go) I think adding them to policy might have been a duplicate.

However we don't seem to be using the ComponentConfig API anywhere now. We should determine a plan for this api and decide if we want to move ahead with using it or get rid of it (before we add any more component-level fields to the Policy api, if that's what we want to do)

damemi:
Correction, we do use this API internally however I'm proposing that we expose it for external configuration as well

/cc @seanmalloy @ingvagabund 
@ravisantoshgudimetla or @aveshagarwal, could either of you provide some insight as to the original plan for the componentconfig api and if it was meant to be exposed?

ingvagabund:
> could either of you provide some insight as to the original plan for the componentconfig api and if it was meant to be exposed?

As I recall, by the time the descheduler was delivered to this world, component config was the default way of configuring new components. The component config got serialized as a json/yaml and served through a config map that got attached to a container as a volume and passed to the component through `--config` (or its variation) flag.

Checking history of `cmd/descheduler/app/options/` I don't see CC has ever been passed to the descheduler through a flag.

> In #380, we deprecated some component-level flags in favor of new policy config fields. Though, looking at our ComponentConfig API I think adding them to policy might have been a duplicate.

`DeschedulerConfiguration` type's fields are currently used as value holder for various descheduler options. Thus the reason why those fields are duplicated. Given all three `--node-selector`, `--max-pods-to-evict-per-node` and `evict-local-storage-pods` are deprecated, `NodeSelector`, `MaxNoOfPodsToEvictPerNode` and `EvictLocalStoragePods` fields are deprecated as well and will get removed eventually. So the duplication is only temporary for [12 months or 2 releases (whichever is longer) since the deprecation was announced](https://kubernetes.io/docs/reference/using-api/deprecation-policy/#deprecating-a-flag-or-cli).

> However we don't seem to be using the ComponentConfig API anywhere now. We should determine a plan for this api and decide if we want to move ahead with using it or get rid of it (before we add any more component-level fields to the Policy api, if that's what we want to do)

[...]

PR #474, opened 2020-12-18
Add a parameter to include soft topology spread constraints

lixiang233:
Fixes: #461

### DIFF
--- a/pkg/descheduler/strategies/topologyspreadconstraint.go
+++ b/pkg/descheduler/strategies/topologyspreadconstraint.go
@@ -125,9 +125,8 @@ func RemovePodsViolatingTopologySpreadConstraint(
 		namespaceTopologySpreadConstraints := make(map[v1.TopologySpreadConstraint]struct{})
 		for _, pod := range namespacePods.Items {
 			for _, constraint := range pod.Spec.TopologySpreadConstraints {
-				// Only deal with hard topology constraints
-				// TODO(@damemi): add support for soft constraints
-				if constraint.WhenUnsatisfiable != v1.DoNotSchedule {
+				// Ignore soft topology constraints if they are not included
+				if (strategy.Params == nil || !strategy.Params.IncludeSoftConstraints) && constraint.WhenUnsatisfiable != v1.DoNotSchedule {
 					continue
 				}
 				namespaceTopologySpreadConstraints[constraint] = struct{}{}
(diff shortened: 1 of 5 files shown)


pkg/apis/componentconfig/v1alpha1/types.go
```diff
@@ -48,4 +49,8 @@ type DeschedulerConfiguration struct {
 
 	// EvictLocalStoragePods allows pods using local storage to be evicted.
 	EvictLocalStoragePods bool `json:"evictLocalStoragePods,omitempty"`
+
+	// Logging specifies the options of logging.
+	// Refer [Logs Options](https://github.com/kubernetes/component-base/blob/master/logs/options.go) for more information.
+	Logging componentbaseconfig.LoggingConfiguration
```
lixiang233, line 55: It seems `make gen` hasn't been executed last time we change the component config api, so we didn't notice this missing member.

BTW, do we need to use versioned apis for component config? Because I noticed that flags were directly added to component config rather than being added to versioned config.

damemi: > BTW, do we need to use versioned apis for component config? Because I noticed that flags were directly added to component config rather than being added to versioned config.

since this componentconfig api is currently only used internally by the descheduler, I don't think we technically need to maintain any versioning for it. However it would be good to keep the versioning up to date if we plan to expose this api at any point (as suggested in https://github.com/kubernetes-sigs/descheduler/issues/431)

lixiang233:
Rebased, PTAL @damemi  @seanmalloy  @ingvagabund

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -76,7 +76,7 @@ func RemovePodsViolatingTopologySpreadConstraint(
 	nodes []*v1.Node,
 	podEvictor *evictions.PodEvictor,
 ) {
-	thresholdPriority, includedNamespaces, excludedNamespaces, err := validateAndParseTopologySpreadParams(ctx, client, strategy.Params)
+	includeSoftConstraints, thresholdPriority, includedNamespaces, excludedNamespaces, err := validateAndParseTopologySpreadParams(ctx, client, strategy.Params)
```
damemi, line 79: I don't think `includeSoftConstraints` needs to be validated by this function, as it just complicates the function signature (and isn't actually validated, since it's just a bool).

We can just use `params.IncludeSoftConstraints` on L129 and add a nil check in that if statement for `params` so we don't get a dereference panic. wdyt?

lixiang233: I originally wanted to check `params` only once so I put it in `validateAndParseTopologySpreadParams`.

It's more concise to do the nil check inside that if statement, I'll update soon.

damemi (commented):
/approve

Issue #479, opened 2021-01-03
helm chart nodeSelector in cronjob


matti:
https://github.com/kubernetes-sigs/descheduler/blob/master/charts/descheduler/templates/cronjob.yaml

damemi:
Hi @matti, could you please add some more detail as to what it is you are requesting? I am not sure from your title whether you are referring to the `--node-selector` option in descheduler, or adding nodeSelectors to the cronjob, or something else. Thanks!

matti:
in the cronjob.yaml

damemi:
Hi @matti 
It sounds like you're referring to adding nodeSelectors to the cronjob itself. I opened a PR that I think will add that here: https://github.com/kubernetes-sigs/descheduler/pull/480

PR #480, opened 2021-01-09
Add nodeSelector to cronJob helm chart

damemi:
Fixes https://github.com/kubernetes-sigs/descheduler/issues/479

### DIFF
--- a/charts/descheduler/templates/cronjob.yaml
+++ b/charts/descheduler/templates/cronjob.yaml
@@ -33,6 +33,10 @@ spec:
             {{- .Values.podLabels | toYaml | nindent 12 }}
             {{- end }}
         spec:
+          {{- with .Values.nodeSelector }}
+          nodeSelector:
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
           {{- if .Values.priorityClassName }}
           priorityClassName: {{ .Values.priorityClassName }}
           {{- end }}
--- a/charts/descheduler/values.yaml
+++ b/charts/descheduler/values.yaml
@@ -58,6 +58,9 @@ deschedulerPolicy:
 
 priorityClassName: system-cluster-critical
 
+nodeSelector: {}
+#  foo: bar
+
 rbac:
   # Specifies whether RBAC resources should be created
   create: true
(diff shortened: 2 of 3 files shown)


damemi (commented):
/cc @seanmalloy 
My first attempt at a helm chart change, haven't tried this out myself yet

seanmalloy:
This looks good to me.

@stevehipwell would you have time to review as well?

/kind feature

charts/descheduler/values.yaml
```diff
@@ -58,6 +58,9 @@ deschedulerPolicy:
 
 priorityClassName: system-cluster-critical
 
+#nodeSelector:
+#  foo: bar
```
stevehipwell, line 62: I'd suggest we use the idiomatic way that _Helm_ sets these up for new charts.

```suggestion
nodeSelector: {}
  # foo: bar
```

charts/descheduler/templates/cronjob.yaml
```diff
@@ -33,6 +33,9 @@ spec:
             {{- .Values.podLabels | toYaml | nindent 12 }}
             {{- end }}
         spec:
+          {{- if .Values.nodeSelector }}
+          nodeSelector: {{ .Values.nodeSelector }}
+          {{- end }}
```
stevehipwell, line 38: I don't think this would work. I'd suggest we use the idiomatic way that Helm sets these up for new charts.

```suggestion
          {{- with .Values.nodeSelector }}
          nodeSelector:
            {{- toYaml . | nindent 12 }}
          {{- end }}
```

stevehipwell (changes requested):
@seanmalloy I've added comments. Also the _Chart.yaml_ file will need to be updated if this needs to be published.

damemi:
Thanks for the review @stevehipwell , I made your changes. Please check if they look good :)

> Also the Chart.yaml file will need to be updated if this needs to be published.

I think this can wait to be published until the next release, so we don't need to update now

stevehipwell:
Looks good @damemi.

seanmalloy (changes requested):
Please also add this new parameter to the docs: https://github.com/kubernetes-sigs/descheduler/tree/master/charts/descheduler#configuration

damemi:
@seanmalloy good catch, updated

seanmalloy (approved):

Issue #96, opened 2018-05-23
filter pods with pvc volumes


caruccio:
https://github.com/kubernetes-incubator/descheduler/blob/af2198428e0f52c10d9bd58bc8f3173c80a45142/pkg/descheduler/pod/pods.go#L132

Hey!
It looks like PVCs should be marked as non-evictable, or there should be a user-defined flag to allow that behavior.
Greetings.

aveshagarwal:
@caruccio i think having a flag would be better as that could be enabled/disabled as needed.

kfirfer:
Would be great to see it

seanmalloy:
@kfirfer so you want an option to tell the descheduler to not evict pods that have a PVC associated with them?

/reopen

kfirfer:
@seanmalloy Yes, Its sounds great

seanmalloy:
If this is implemented it will probably work similar to `evictLocalStoragePods`. See below link for docs.

https://github.com/kubernetes-sigs/descheduler#policy-and-strategies

caruccio:
@seanmalloy that is exactly what I meant.

damemi:
> If this is implemented it will probably work similar to `evictLocalStoragePods`. See below link for docs for details.
> 
> https://github.com/kubernetes-sigs/descheduler#policy-and-strategies

This is pretty straightforward, but something to consider is how we default this flag to preserve current behavior. Since we currently evict PVC pods, maybe the flag should be more like `ignorePvcPods` with a default=false? open to other ideas like `preserve`, `dontEvict`, etc

/assign

damemi:
Opened #481 to add this

PR #481, opened 2021-01-11
Add option to ignore pods with PVCs from eviction

damemi:
Fixes https://github.com/kubernetes-sigs/descheduler/issues/96

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,7 @@ parameters associated with the strategies can be configured too. By default, all
 The policy also includes common configuration for all the strategies:
 - `nodeSelector` - limiting the nodes which are processed
 - `evictLocalStoragePods` - allowing to evict pods with local storage
+- `ignorePvcPods` - set whether PVC pods should be evicted or ignored (defaults to `false`)
 - `maxNoOfPodsToEvictPerNode` - maximum number of pods evicted from each node (summed through all strategies)
 
 ```yaml
@@ -119,6 +120,7 @@ kind: "DeschedulerPolicy"
 nodeSelector: prod=dev
 evictLocalStoragePods: true
 maxNoOfPodsToEvictPerNode: 40
+ignorePvcPods: false
 strategies:
   ...
 ```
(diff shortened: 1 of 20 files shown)


damemi:
/hold
`make gen` is currently broken due to the logging fields

damemi (commented):
/hold cancel
make gen is fixed now

cmd/descheduler/app/options/options.go
```diff
@@ -77,4 +77,5 @@ func (rs *DeschedulerServer) AddFlags(fs *pflag.FlagSet) {
 	fs.IntVar(&rs.MaxNoOfPodsToEvictPerNode, "max-pods-to-evict-per-node", rs.MaxNoOfPodsToEvictPerNode, "DEPRECATED: limits the maximum number of pods to be evicted per node by descheduler")
 	// evict-local-storage-pods allows eviction of pods that are using local storage. This is false by default.
 	fs.BoolVar(&rs.EvictLocalStoragePods, "evict-local-storage-pods", rs.EvictLocalStoragePods, "DEPRECATED: enables evicting pods using local storage by descheduler")
+
```
ingvagabund, line 80: You can drop the additional line

ingvagabund (changes requested):
Can you add a test case for the new flag? In any of the strategies you choose.

Wrt. `ignore` prefix, we already have `ExcludeOwnerKinds` and `Exclude` namespaces fields. Do you think it would make sense to align the new field with existing ones? I.e. `excludePVCPods`? Even `IgnorePVCPods` sounds ok to me.

lixiang233 (commented):
Besides, we should update the [user guide](https://github.com/kubernetes-sigs/descheduler/blob/master/docs/user-guide.md) to include the new flag

damemi:
Updated this to remove the extra line and added the flag to readme

ingvagabund:
Unit test please

damemi:
Sorry, forgot the unit test. Added to `PodLifeTime`

lixiang233 (approved):
/lgtm

jsravn:
Minor nit: could we rename the option from `ignorePvcPods` to `evictPvcPods` before this is released? I found the option name confusing - I took it to mean "ignore PVCs when considering a pod for eviction", rather than "ignore pods with PVCs when evicting".

damemi:
Hi @jsravn, thanks for the feedback. The motivation behind naming this `ignorePvcPods` was the the current default Descheduler behavior is to evict Pvc Pods. So, rather than add the new flag with a default value `true`, we added it so that it semantically defaults to `false`, preserving the existing default behavior without any change to behavior.

I hope this clears up the confusion. At this point, I would personally prefer not to make that change since it would be a full invert of the behavior. However if others disagree I am open to the consideration

PR #484, opened 2021-01-12
Fix TopologySpread bug that evicts non-evictable pods

damemi:
TopologySpreadConstraint does not properly check if a pod is evictable before calling `EvictPod` (due to the fact that it does not rely on `ListPodsOnNode`, which takes `IsEvictable` as a parameter and does the check there for most of the other strategies).

This breaks the check out explicitly. It could also be refactored to do this check in `balanceDomains`, when building the list of evictable pods.


damemi (commented):
/cc @ingvagabund 
ptal

lixiang233 (commented):
/lgtm

ingvagabund:
LGTM

PR #488, opened 2021-01-25
[WIP] Allow Node strategies to run with informers

damemi:
/kind feature

### DIFF
--- a/pkg/descheduler/strategies/node_affinity.go
+++ b/pkg/descheduler/strategies/node_affinity.go
@@ -21,9 +21,11 @@ import (
 	"fmt"
 
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/informers"
 	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog/v2"
-
 	"sigs.k8s.io/descheduler/pkg/api"
 	"sigs.k8s.io/descheduler/pkg/descheduler/evictions"
 	nodeutil "sigs.k8s.io/descheduler/pkg/descheduler/node"
@@ -32,6 +34,55 @@ import (
 )
 
 const NodeAffinityName = "RemovePodsViolatingNodeAffinity"
+
+// NewRemovePodsViolatingNodeAffinity returns a StrategyController to run Node affinity with informers
+func NewRemovePodsViolatingNodeAffinity(
+	ctx context.Context,
+	client clientset.Interface,
+	sharedInformerFactory informers.SharedInformerFactory,
+	strategy api.DeschedulerStrategy,
+	podEvictor *evictions.PodEvictor,
+	nodeSelector string,
+	stopChannel chan struct{},
+) *StrategyController {
+	c := &StrategyController{
+		ctx:                   ctx,
+		client:                client,
+		queue:                 workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter()),
+		sharedInformerFactory: sharedInformerFactory,
+		strategy:              strategy,
+		podEvictor:            podEvictor,
+		nodeSelector:          nodeSelector,
+		stopChannel:           stopChannel,
+	}
+	sharedInformerFactory.Core().V1().Nodes().Informer().AddEventHandler(nodeEventHandler(c))
+	return c
+}
+
+func nodeEventHandler(c *StrategyController) cache.ResourceEventHandler {
+	return cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			nodes, err := nodeutil.ReadyNodes(c.ctx, c.client, c.sharedInformerFactory.Core().V1().Nodes(), c.nodeSelector)
+			if err != nil {
+				klog.V(1).InfoS("Unable to get ready nodes", "err", err)
+				close(c.stopChannel)
+				return
+			}
+			c.nodes = nodes
+			c.queue.Add(workQueueKey)
+		},
+		UpdateFunc: func(old, new interface{}) {
+			nodes, err := nodeutil.ReadyNodes(c.ctx, c.client, c.sharedInformerFactory.Core().V1().Nodes(), c.nodeSelector)
+			if err != nil {
+				klog.V(1).InfoS("Unable to get ready nodes", "err", err)
+				close(c.stopChannel)
+				return
+			}
+			c.nodes = nodes
+			c.queue.Add(workQueueKey)
+		},
+	}
+}
 
 func validatePodsViolatingNodeAffinityParams(params *api.StrategyParameters) error {
 	if params == nil || len(params.NodeAffinityType) == 0 {
--- a/pkg/descheduler/strategies/node_affinity.go
+++ b/pkg/descheduler/strategies/node_affinity.go
@@ -54,6 +54,7 @@ func NewRemovePodsViolatingNodeAffinity(
 		podEvictor:            podEvictor,
 		nodeSelector:          nodeSelector,
 		stopChannel:           stopChannel,
+		name:                  NodeAffinityName,
 	}
 	sharedInformerFactory.Core().V1().Nodes().Informer().AddEventHandler(nodeEventHandler(c))
 	return c
--- a/pkg/descheduler/strategies/node_affinity.go
+++ b/pkg/descheduler/strategies/node_affinity.go
@@ -31,6 +31,8 @@ import (
 	"sigs.k8s.io/descheduler/pkg/utils"
 )
 
+const NodeAffinityName = "RemovePodsViolatingNodeAffinity"
+
 func validatePodsViolatingNodeAffinityParams(params *api.StrategyParameters) error {
 	if params == nil || len(params.NodeAffinityType) == 0 {
 		return fmt.Errorf("NodeAffinityType is empty")
--- a/pkg/descheduler/strategies/types.go
+++ b/pkg/descheduler/strategies/types.go
@@ -3,11 +3,30 @@ package strategies
 import (
 	"context"
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/informers"
 	clientset "k8s.io/client-go/kubernetes"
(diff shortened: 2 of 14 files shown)


pkg/descheduler/strategies/types.go
```diff
@@ -0,0 +1,34 @@
...
+	"context"
+	v1 "k8s.io/api/core/v1"
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/util/workqueue"
+	"sigs.k8s.io/descheduler/pkg/api"
+	"sigs.k8s.io/descheduler/pkg/descheduler/evictions"
+)
+
+// StrategyFunction defines the function signature for each strategy's main function
+type StrategyFunction func(
```
ingvagabund, line 36: I'd rather keep this type private until we discuss how to refactor the way strategies are initialized.

pkg/descheduler/strategies/types.go
```diff
@@ -39,16 +43,60 @@ type StrategyFunction func(
...
-	client clientset.Interface
-	queue  workqueue.RateLimitingInterface
-	f      StrategyFunction
+	ctx                   context.Context
+	client                clientset.Interface
+	queue                 workqueue.RateLimitingInterface
+	sharedInformerFactory informers.SharedInformerFactory
+	stopChannel           chan struct{}
+
+	f            StrategyFunction
```
ingvagabund, line 52: s/f/func to avoid one letter variable (at least two letters to make searching for it easy please :))

pkg/descheduler/strategies/node_affinity.go
```diff
@@ -33,6 +35,55 @@ import (
...
+			nodes, err := nodeutil.ReadyNodes(c.ctx, c.client, c.sharedInformerFactory.Core().V1().Nodes(), c.nodeSelector)
+			if err != nil {
+				klog.V(1).InfoS("Unable to get ready nodes", "err", err)
+				close(c.stopChannel)
+				return
+			}
+			c.nodes = nodes
+			c.queue.Add(workQueueKey)
+		},
+		UpdateFunc: func(old, new interface{}) {
```
ingvagabund, line 75: Can you put down a comment saying `DeleteFunc is not needed since pods are automatically evicted` (or similar)

ingvagabund (commented):
As long as an operator is used to deploy both instances of descheduler (in both modes) and assuming the same config file gets used (each mode will just filter out what's not relevant), this change should be seemingly transparent for a user. Without an operator, one will need to maintain both a CronJob and a Deployment.

In any case there will be two instances of a descheduler running which might interfere with each other. With strategies no longer running in sequence, we need to revisit the code for potential races. Also, improve some error message where an eviction failed (e.g. due to non-existing pod). We might also need to add a mechanism which will make sure both mode instances are ran over separate sets of pods (e.g. label selector filtering) to minimize interference.

damemi:
> Without an operator, one will need to maintain both a CronJob and a Deployment. In any case there will be two instances of a descheduler running which might interfere with each other.

@ingvagabund this does not require 2 descheduler instances. The informed strategies are spun off into separate, non-blocking goroutines while the main wait loop handles the iterative strategies. This is done with 1 descheduler, run as a deployment with `deschedulingInterval`.

ingvagabund:
> This is done with 1 descheduler, run as a deployment with `deschedulingInterval`.

I see. So you suggest to drop the cron job and move back to the original way of deploying the descheduler (or keep cron but  provide deployment as well as the main manifest). I recall it was more practical to use cron job than to have a descheduler instance stopped on `time.Sleep` for `deschedulingInterval`. Deployment makes more sense now.

Also, Kubelet reports node status every 10 seconds by default. There's gonna be a lot of "empty" iterations so it might make sense to add a check to `UpdateFunc` for each strategy to compare what actually changed so a strategy does not have to be run every 10 seconds going through all pods and nodes just because a node changed e.g. its labels.

lixiang233 (commented):
It seems every informed strategies has its own `StrategyController`, then we may run several strategies at the same time, this may cause some data synchronization problems. I think we can keep only one `StrategyController` and work queue. When we got an event, `EventHandler` should decide which strategy(s) should be called by the event's type and add the strategy's name to work queue, worker will process them one by one.

damemi:
> keep cron but provide deployment as well as the main manifest

Yeah, this is my thinking. There are advantages and disadvantages to either way of running the scheduler, so no reason to not provide them both.

> it might make sense to add a check to UpdateFunc for each strategy to compare what actually changed so a strategy does not have to be run every 10 seconds

That is a good idea. This is why I did not make the `eventHandler` a function of `StrategyController`, with the idea that each informed strategy may want to implement its own logic for filtering out noop events (or making modifications prior to operating on the event)

> It seems every informed strategies has its own StrategyController, then we may run several strategies at the same time, this may cause some data synchronization problems. I think we can keep only one StrategyController and work queue. When we got an event, EventHandler should decide which strategy(s) should be called by the event's type and add the strategy's name to work queue, worker will process them one by one.

This is interesting, and sort of goes along with what @ingvagabund suggested above.

I thought it made sense (at least from a code organization standpoint) to have individual strategies responsible for their own `EventHandler`s, but I see your point about data synchronization. A central event handler "registry" which then processes the informed strategies serially would be safer. I would probably like to add some way to pass the event itself to each strategy, where the strategy's code can decide whether to run or ignore it.

This brings up another data synchronization point, what if an event comes in right around the same time as a `deschedulingInterval`? We shouldn't run the periodic strategies at the same time as the informed strategies.

Maybe this could be solved with some kind of mutex? That way if the lock is held by either the `StrategyController` or the `wait` loop (whichever is triggered first) we don't hit multithreading data sync issues.

I think this design (of running informed and default strategies in the same descheduler instance) is critical for both usability and further development of reactive descheduling. So I would like to make sure that this can run smoothly.

pkg/descheduler/strategies/node_affinity.go
```diff
@@ -21,16 +21,70 @@ import (
...
+		name:                  NodeAffinityName,
+	}
+	sharedInformerFactory.Core().V1().Nodes().Informer().AddEventHandler(nodeEventHandler(c))
+	return c
+}
+
+func nodeEventHandler(c *StrategyController) cache.ResourceEventHandler {
+	return cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			nodes, err := nodeutil.ReadyNodes(c.ctx, c.client, c.sharedInformerFactory.Core().V1().Nodes(), c.nodeSelector)
```
lixiang233, line 66: If strategies run serially, the time between listing nodes and running strategy may be long, if another event comes during this period, we'll list nodes again, acturally we only need to do this once. So like what was disscussed in #469 , listing and filtering nodes could be done in each strategies, we can also custom `nodeSelector` for each strategies by doing so.

pkg/descheduler/strategies/node_affinity.go
```diff
@@ -21,16 +21,70 @@ import (
...
+		podEvictor:            podEvictor,
+		nodeSelector:          nodeSelector,
+		stopChannel:           stopChannel,
+		name:                  NodeAffinityName,
+	}
+	sharedInformerFactory.Core().V1().Nodes().Informer().AddEventHandler(nodeEventHandler(c))
+	return c
+}
+
+func nodeEventHandler(c *StrategyController) cache.ResourceEventHandler {
```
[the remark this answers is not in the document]
damemi: It is a common handler, but only between the node strategies (taints+affinity). I actually considered putting these strategies into their own subpackage (like `pkg/descheduler/strategies/node/`) but that seems a bit overcomplicated.

pkg/descheduler/strategies/node_affinity.go
damemi: > if another event comes during this period, we'll list nodes again, acturally we only need to do this once

I'm not sure, because the point of reacting to Node events as they happen is to get a real-time updated view of the cluster to operate on. So each event should trigger a re-list so the strategy doesn't have old data.

I do agree though with refactoring this a bit like was previously mentioned.

ingvagabund: All the "reactive" strategies listening to node informers need to take into account strategies like LowNodeUtilization. `LowNodeUtilization` needs to take into account entire cluster (or its reasonable subset). So if the strategies are ran in incorrect order, e.g. `LowNodeUtilization` followed by running `PodLifeTime` removing many old pods, running `LowNodeUtilization` before might be actually nullified and make the overall resource consumption worst. So far, the order of strategies was [kinda hardcoded](https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/descheduler/descheduler.go#L72-L81) (depending on the map iterator). We might compute some static impact/score of each strategy to overall resource utilization and run the strategies on order starting with strategies changing utilization the most to strategies changing it the least (just a though, it might not be possible).

damemi: I doubt that `LowNodeUtilization` will be able to easily convert to be reactive. That is the point of the design I proposed here, that we can begin to convert some of the strategies while keeping the hardcoded order of others.

I think @lixiang233's idea of a single strategy controller with a master registry of event handlers would solve any ordering issues. That with a mutex ensures we aren't running 2 strategies at the same time.

For example, with an interval of 60 minutes we could have:
```
0min - interval run of strategies
12 min - strategycontroller runs nodeaffinity
60 min - interval run
119 min - strategycontroller run (mutex lock)
121 min - interval (blocked waiting for mutex) runs
181 min - interval run
```

And since each periodic interval does a node re-list, those are working with an up-to-date list each time. Any results from NodeAffinity/NodeTaints won't interfere. So really, we are just triggering a descheduling run at dynamic intervals along with the hardcoded/cron intervals

JaneLiuL:
@damemi  very good PR~~ Very happy to see that descheduler support Default  and Informer mode.
 I can help to separate severl PR if you not mind :)

Dentrax:
Hey @damemi, According to issue https://github.com/kubernetes-sigs/descheduler/issues/696, we can get to this PR if you show us what kind of things needs to be done. 🙏 We can be whether work on this PR or create a new carry PR by _cherry-picking_ your commits. Can you please enlighten us to see our next step?

cc @developer-guy @eminaktas @yasintahaerol @necatican @f9n

PR #492, opened 2021-01-27
upgrade klog.InfoS level 

eatjeff:


damemi (commented):
Hi @eatjeff 
Is there a justification for this change? This is a pretty important failure message.

If anything, this should probably be changed to return an error since in the case where policyConfig is nil, the caller errors out anyway (https://github.com/kubernetes-sigs/descheduler/blob/241f1325c968889e6abaf250b66bf0522c517506/pkg/descheduler/descheduler.go#L46-L52)

/ok-to-test

seanmalloy:
@eatjeff inorder to move forward with this pull request you will need to sign the CLA. Let me know if you need any assistance signing the CLA.

Issue #466, opened 2020-12-09
Don't run image as root


eatwithforks:
Needs USER 1000 in Dockerfile so I can put in securityContext runAsNonRoot: true

seanmalloy:
@eatwithforks thanks for reporting this issue. Would you be willing to submit a PR updating `Dockerfile.dev` and `Dockerfile` with the required changes?

/kind bug
/remove-kind feature

eatwithforks:
https://github.com/kubernetes-sigs/descheduler/pull/468

seanmalloy:
Now that the container image is setup to allow not running as root I'd like to make sure we have the YAML manifests in the `kubernetes` directory and the helm chart setup properly to also not run as root.

I can take care of this if others do not have time.

/reopen

perosb:
I could not run latest master using the yaml files:
`Error: container has runAsNonRoot and image will run as root`

Assume its missing user in Dockerfile?
Adding the below makes it work tho.
```
securityContext:
  runAsUser: 1000
```

seanmalloy:
> I could not run latest master using the yaml files:
> `Error: container has runAsNonRoot and image will run as root`
> 
> Assume its missing user in Dockerfile for the built v0.20.0 image.
> Adding the below makes it work tho.
> 
> ```
> securityContext:
>   runAsUser: 1000
> ```

The descheduler v0.20.0 image is setup to run as root. The yaml manifests in the `kubernetes` directory on the master branch work with the latest development build of the descheduler.

Use the yaml manifests from the release-1.20 branch to install descheduler v0.20.0.

Thanks.

seanmalloy:
Running the descheduler container as non-root should work with the upcoming v0.21.0 release.

PR #495, opened 2021-02-10
Update Job and CronJob YAML to run as non-root

seanmalloy:
Part of #466.


ingvagabund:
user is set to 1000
/lgtm

damemi (commented):
/approve

Issue #348, opened 2020-07-15
Add Support For Prometheus Metrics


seanmalloy:
**Is your feature request related to a problem? Please describe.**

I'd like to monitor the descheduler using Prometheus metrics. Metrics should only be collected when run without the `--dry-run` CLI option.

**Describe the solution you'd like**

Add an optional configuration option(CLI option?) to enable exposing Prometheus metrics. At this point in time I have not given much thought to exactly which metrics and labels the descheduler would expose.

**Describe alternatives you've considered**

I have considered using the descheduler logs with log aggregation software for monitoring the descheduler. I also considered using the k8s events generated by the descheduler in combination with [eventrouter](https://github.com/heptiolabs/eventrouter). The eventrouter does provide some simple Prometheus metrics for events that it collects.

**What version of descheduler are you using?**

descheduler version: v0.18.0

**Additional context**

Currently the descheduler is run as a `Job` or `CronJob` this makes it difficult for Prometheus to scrape metrics from the descheduler pod because it is not long lived. One option is to have descheduler push metrics using the [Prometheus pushgateway](https://github.com/prometheus/pushgateway).

damemi:
> Currently the descheduler is run as a Job or CronJob this makes it difficult for Prometheus to scrape metrics from the descheduler pod because it is not long lived

The scheduler can run as a long lived pod with the `--descheduling-interval` flag. This is actually what we chose to prefer in the [OpenShift Descheduler Operator](https://github.com/openshift/cluster-kube-descheduler-operator) because it also allowed us to not yet depend on CronJobs (which are still beta)

ingvagabund:
@seanmalloy any particular metrics in mind? I can quickly think of:
- number of pods evicted per each strategy
- number of pods failed to be evicted per each strategy

seanmalloy:
> @seanmalloy any particular metrics in mind? I can quickly think of:
> 
> * number of pods evicted per each strategy
> * number of pods failed to be evicted per each strategy

@ingvagabund yep that is pretty much it. See below for some additional details. I'm open to suggestions.

### Metrics
| name  | type  | description  |
|---|---|---|
| build_info  |  gauge | constant 1  |
| pods_evicted_success   |  counter | total number of pods successfully evicted  |
|  pods_evicted_failed  |  counter | total number of pods failed eviction  |

The metric names can be changed. There is probably some k8s best practice for naming Prometheus metrics. We should follow those guidelines to determine the names of the metrics.

[...]

PR #505, opened 2021-02-22
Collect metrics

ingvagabund:
Fixes: #348

Have the descheduler collect metrics and serve them through `/metrics` endpoint on 10258 secure port.

How to test it? Have descheduler running with some descheduling internal (e.g. 1s) and have some pods evicted. Then just run `curl https://localhost:10258/metrics -k`.

Metrics bits mostly copy-pasted from kube-scheduler.


ingvagabund:
https://github.com/kubernetes-sigs/descheduler/issues/348#issuecomment-685787815

@seanmalloy I merged `pods_evicted_success` and `pods_evicted_failed` into `pods_evicted` as a labeled `CounterVec`. Though, not against having two metrics either.

pkg/descheduler/evictions/evictions.go
```diff
@@ -97,19 +98,21 @@ func (pe *PodEvictor) TotalEvicted() int {
 // EvictPod returns non-nil error only when evicting a pod on a node is not
 // possible (due to maxPodsToEvictPerNode constraint). Success is true when the pod
 // is evicted on the server side.
-func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod, node *v1.Node, reasons ...string) (bool, error) {
+func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod, node *v1.Node, strategy string, reasons ...string) (bool, error) {
```
damemi, line 101: Is this new `strategy` param used anywhere? It's kind of confusing to have to update all the strategies with lines like `podEvictor.EvictPod(ctx, pod, node, "NodeAffinity", "NodeAffinity")`

ingvagabund: I don't like it either. We might drop `reasons` field since I don't see it used for anything but passing a strategy name.

damemi: I think `reasons` is fine, but specifically the `strategy` param you added here... I don't see it used anywhere, and it seems like you're just using `reasons` anyway, right?

ingvagabund: Oh, I see. My bad. `metrics.PodsEvicted.With(map[string]string{"result": "maximum number reached", "strategy": reason, "namespace": pod.Namespace}).Inc()` is supposed to have `"strategy": strategy` pair in it.

damemi: Ah yeah, I agree then we could either drop the `...reasons` param all together, *or* add this strategy param, and just use that in each strategy (since reasons is optional, the strategy can just pass its name which will be used for metrics).

I can see a possible use case where one strategy may have different reasons for eviction, so I'm not sure it's worth getting rid of entirely yet. But it doesn't make much sense to keep if we're not using it. Wdyt?

ingvagabund: > I'm not sure it's worth getting rid of entirely yet

Yeah, not completely sure either. I'll keep it there for the moment. Better than drop it and re-introduce it later.

ingvagabund:
```
$ curl -s https://localhost:10258/metrics -k | grep descheduler
# HELP descheduler_build_info [ALPHA] Build info about descheduler, including Go version, Descheduler version, Git SHA, Git branch
# TYPE descheduler_build_info gauge
descheduler_build_info{DeschedulerVersion="v20210301-v0.20.0-38-gf83615418",GitBranch="collect-metrics",GitSha1="f83615418e729595e6a39d8632304c05442bc67a",GoVersion="go1.15"} 0
```

ingvagabund:
After some time running over vanilla OpenShift 4.8 with the example policy config:
```
$ curl -s https://localhost:10258/metrics -k | grep descheduler
# HELP descheduler_build_info [ALPHA] Build info about descheduler, including Go version, Descheduler version, Git SHA, Git branch
# TYPE descheduler_build_info gauge
descheduler_build_info{DeschedulerVersion="v20210301-v0.20.0-38-gf83615418",GitBranch="collect-metrics",GitSha1="f83615418e729595e6a39d8632304c05442bc67a",GoVersion="go1.15"} 0
# HELP descheduler_pods_evicted [ALPHA] Number of successfully evicted pods, by the result, by the strategy, by the namespace. 'failed' result means a pod could not be evicted
# TYPE descheduler_pods_evicted counter
descheduler_pods_evicted{namespace="openshift-marketplace",result="success",strategy="RemoveDuplicatePods"} 1
```

pkg/descheduler/evictions/evictions.go
```diff
@@ -97,19 +98,21 @@ func (pe *PodEvictor) TotalEvicted() int {
...
 	var reason string
 	if len(reasons) > 0 {
 		reason = " (" + strings.Join(reasons, ", ") + ")"
```
damemi, line 104: Continuing from what we're talking about above (https://github.com/kubernetes-sigs/descheduler/pull/505/files#r584629100), maybe we can change this line to:
```go
		reason = strategy + " (" + strings.Join(reasons, ", ") + ")"
```
then,  all the calls to `EvictPod` can be shortened to just 1 parameter (so, no longer passing in the optional `reasons` list, but still keeping that available):
```go
podEvictor.EvictPod(ctx, pod, nodeMap[nodeName], "RemoveDuplicatePods")
```

this way, we don't have to remove the `reasons` parameter but we clean up those weird-looking function calls. We can then start some work to actually use `reasons` for each strategy in a more informative way than how it is currently. What do you think?

ingvagabund: Good idea!!! Updated.

pkg/descheduler/evictions/evictions.go
```diff
@@ -97,19 +98,21 @@ func (pe *PodEvictor) TotalEvicted() int {
 // EvictPod returns non-nil error only when evicting a pod on a node is not
 // possible (due to maxPodsToEvictPerNode constraint). Success is true when the pod
 // is evicted on the server side.
-func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod, node *v1.Node, reasons ...string) (bool, error) {
+func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod, node *v1.Node, strategy string, reasons ...string) (bool, error) {
 	var reason string
 	if len(reasons) > 0 {
-		reason = " (" + strings.Join(reasons, ", ") + ")"
+		reason = strategy + " (" + strings.Join(reasons, ", ") + ")"
 	}
```
damemi, line 105: sorry, I just noticed the `if len(reasons) > 0` check here... maybe instead we should initialize `reason := strategy` (instead of nil) above this and then append the reasons inside this check

ingvagabund:
Not sure if `10258` port is allocated by a different component. The PR also pulls a lot of new dependencies (mainly apiserver bits)  to provide secure serving of the metrics. Right now, the metrics are served by default and there's no way to disable it. We might provide additional option/configuration to disable the metrics explicitly, wdyt?

ingvagabund:
@damemi @seanmalloy @lixiang233 PTAL

cmd/descheduler/app/options/options.go
```diff
@@ -19,20 +19,29 @@ package options
...
-
 	"k8s.io/component-base/logs"
+
 	"sigs.k8s.io/descheduler/pkg/apis/componentconfig"
 	"sigs.k8s.io/descheduler/pkg/apis/componentconfig/v1alpha1"
 	deschedulerscheme "sigs.k8s.io/descheduler/pkg/descheduler/scheme"
 )
 
+const (
+	DefaultDeschedulerPort = 10258
```
lixiang233, line 34: Can you mention in somewhere that port `10258` is used for metrics by default and users can use command line flag `--secure-port` to custom port number? It's not easy to get this info from the help command.

metrics/metrics.go
```diff
@@ -0,0 +1,87 @@
...
+
+// ActivePods returns the pending pods metrics with the label active
+// func ActivePods() metrics.GaugeMetric {
+// 	return PodsEvicted.With(metrics.Labels{"queue": "active"})
+// }
+//
+// // BackoffPods returns the pending pods metrics with the label backoff
+// func BackoffPods() metrics.GaugeMetric {
+// 	return PodsEvicted.With(metrics.Labels{"queue": "backoff"})
+// }
```
lixiang233, line 87: Are these comments useful?

lixiang233:
> We might provide additional option/configuration to disable the metrics explicitly, wdyt?

+1 for providing a way to disable metrics.

ingvagabund:
Let's merge https://github.com/kubernetes-sigs/descheduler/pull/513 first

metrics/metrics.go
ingvagabund: Leftover, removed.

cmd/descheduler/app/options/options.go
ingvagabund: Added some comments in `--disable-metrics` flag about it.

README.md
```diff
@@ -521,6 +521,16 @@ Setting `--v=4` or greater on the Descheduler will log all reasons why any pod i
 Pods subject to a Pod Disruption Budget(PDB) are not evicted if descheduling violates its PDB. The pods
 are evicted by using the eviction subresource to handle PDB.
 
+## Metrics
```
damemi, line 524: Can you update the table of contents too? (we should add a verify script for that...)
Other than that, lgtm
/approve

ingvagabund: Done

Issue #483, opened 2021-01-12
Add verify scripts for `make gen`


damemi:
Issues such as the one described in https://github.com/kubernetes-sigs/descheduler/pull/482 could be avoided if we add automatic verify scripts on PRs to ensure that `make gen` (and its subscripts) have been run. It should be simple enough as running `make gen` in a tmp dir and checking if any changes have been generated.

/kind cleanup
/help

pravarag:
@damemi I would like to work on this issue if it's not already been taken  :)

pravarag:
@damemi I've come across few doubts while trying to understand the issue. As you mentioned that we need this change to ensure `make gen` has run it's all subscripts proprerly and when I try to run `make gen` locally, it basically generates three binaries,
```
conversion-gen
deepcopy-gen
defaulter-gen
```
So, do we need to ensure that `make gen` has generated those three binaries in `/tmp` location properly or not?

damemi:
Those binaries are what actually run to generate the updated code. `make gen` conveniently downloads them (if they do not already exist) and stores them in the tmp directory which is [ignored by git](https://github.com/kubernetes-sigs/descheduler/blob/master/.gitignore#L1).

The goal here is to make sure that code which is about to be committed has already run these generators. What we want to check is that the diff in a PR doesn't change when you run `make gen` against it. 

So, since the binaries themselves are not committed we can't check that they are stored in `/tmp` (but if they aren't being downloaded, then the generators won't be running anyway, so you will still see a diff).

seanmalloy:
Not quite done yet.

/reopen

pravarag:
@seanmalloy @damemi is it okay to close this issue now? The last PR for `defaulters gen` has been merged today. Let me know if there is anything else I need to check w.r.t this issue.

seanmalloy:
I believe everything is done now. @pravarag thanks for your help with this!

/close

PR #507, opened 2021-02-25
Add verify scripts for make gen to run during PR

pravarag:
This PR fixes: https://github.com/kubernetes-sigs/descheduler/issues/483

As part of this change, we want to add steps to ensure that `make gen` has been run and there are no changes being generated during the PR. 
For now, I've tested the code in local for the steps which I've added in the Makefile and below is the output for same,
```
 ->  make verify-gen
./hack/verify-makegen.sh

Diff in the PR has changed
make: *** [Makefile:112: verify-gen] Error 1
```
Since it's reporting error, which means it's pointing that the generators have been downloaded at the latest in the `_output` directory in my local.

### DIFF
--- a/hack/verify-conversions.sh
+++ b/hack/verify-conversions.sh
@@ -1,0 +1,36 @@
+#!/bin/bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+source "$(dirname "${BASH_SOURCE}")/lib/init.sh"
+DESCHEDULER_ROOT=$(dirname "${BASH_SOURCE}")/..
+mkdir -p "${DESCHEDULER_ROOT}/_tmp"
+_tmpdir="$(mktemp -d "${DESCHEDULER_ROOT}/_tmp/kube-verify.XXXXXX")"
+
+_deschedulertmp="${_tmpdir}"
+mkdir -p "${_deschedulertmp}"
+
+git archive --format=tar --prefix=descheduler/ "$(git write-tree)" | (cd "${_deschedulertmp}" && tar xf -)
+_deschedulertmp="${_deschedulertmp}/descheduler"
+
+pushd "${_deschedulertmp}" > /dev/null 2>&1
+go build -o "${OS_OUTPUT_BINPATH}/conversion-gen" "k8s.io/code-generator/cmd/conversion-gen"
+
+${OS_OUTPUT_BINPATH}/conversion-gen \
+		--go-header-file "hack/boilerplate/boilerplate.go.txt" \
+		--input-dirs "./pkg/apis/componentconfig/v1alpha1,./pkg/api/v1alpha1" \
+		--output-file-base zz_generated.conversion
+popd > /dev/null 2>&1
+
+pushd "${DESCHEDULER_ROOT}" > /dev/null 2>&1
+if ! _out="$(diff -Naupr pkg/ "${_deschedulertmp}/pkg/")"; then
+    echo "Generated output differs:" >&2
+    echo "${_out}" >&2
+    echo "Generated conversions verify failed."
+    exit 1
+fi
+popd > /dev/null 2>&1
+
+echo "Generated conversions verified."
--- a/Makefile
+++ b/Makefile
@@ -94,7 +94,7 @@ push-all: image.amd64 image.arm image.arm64
 clean:
 	rm -rf _output
 
-verify: verify-gofmt verify-vendor lint lint-chart verify-spelling verify-toc
+verify: verify-gofmt verify-vendor lint lint-chart verify-spelling verify-toc verify-gen
 
 verify-spelling:
 	./hack/verify-spelling.sh
@@ -119,6 +119,9 @@ gen:
 	./hack/update-generated-deep-copies.sh
 	./hack/update-generated-defaulters.sh
 	./hack/update-toc.sh
+
+verify-gen:
+	./hack/verify-conversions.sh
 
 lint:
 ifndef HAS_GOLANGCI


pravarag:
I signed it

hack/verify-makegen.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+result=$(git diff ${OS_OUTPUT_BINPATH})
```
damemi, line 3: This script should do more than just check for a git diff, it should ideally actually run the generator scripts in a temp directory and compare that to the current output. The reason for this is because this script on its own isn't verifying `make gen`, so if I call it directly with `./hack/verify-makegen.sh` or `make verify-gen`, I won't really be checking that.

You can see how [`verify-vendor.sh`](https://github.com/kubernetes-sigs/descheduler/blob/6ecbc85/hack/verify-vendor.sh#L52-L58) works for an example.

Following that pattern, I think it would be better to have separate `./hack/verify-xx...` scripts for each of the generators (for the same reason: that I may want to just verify one of them).

hack/verify-makegen.sh
```diff
@@ -0,0 +1,10 @@
...
+
+if [[ -n "$result" ]]; then
+    echo "Diff in the PR has changed"
+    exit 1
+    fi
+echo "Diff for make gen hasn't changed in the PR, good to go"
+exit 0
```
damemi, line 10: nit: newline at end of file

damemi (changes requested):
thanks for the contribution @pravarag! This is a good start, some comments below

hack/verify-makegen.sh
pravarag: Hi @damemi thanks for pointing out the reference script `verify-vendor.sh`. I wanted to check, is this the line which suggests correct way of using `git diff` : [L64](https://github.com/kubernetes-sigs/descheduler/blob/6ecbc85/hack/verify-vendor.sh#L64) ? Also, in our case for the `git diff` to compare, do we need to take output generated by `./hack/updated-generated-xxx` scripts in the latest commit with the output generated in the `temp_` directory( created during PR check) ?

pravarag:
@damemi also, I've come across a doubt during local execution. When I run `make gen` in my local I see the below output,
```
->  make gen
./hack/update-generated-deep-copies.sh
./hack/update-generated-defaulters.sh
```
But, when I explicitly run this command [L4](https://github.com/kubernetes-sigs/descheduler/blob/master/hack/update-generated-conversions.sh#L4), I'm getting below error as well,
```
->  ./conversion-gen "k8s.io/code-generator/cmd/conversion-gen"
F0228 21:55:13.353319    8843 conversion.go:204] Failed loading boilerplate: open /Users/pravaragrawal/Work/go_work/src/k8s.io/code-generator/hack/boilerplate.go.txt: no such file or directory
goroutine 1 [running]:
```
Is this error something which is to be ignored at this point?

hack/verify-makegen.sh
damemi: I was referring more to how this script is pushing the existing changes onto a tmp directory, running the generator, and then comparing if any new changes occurred. So yes, the diff should be comparing the latest commit to a copy of the latest commit that has had the generator run. In this case, you are just getting that by doing `git diff` on the copy

damemi:
> "k8s.io/code-generator/cmd/conversion-gen"

@pravarag the command you linked to is a `go build` command that is attempting to compile the `conversion-gen` binary. It looks like you are running it as a command itself, which is trying to run `conversion-gen` with its own package as a parameter.

So yes, you can ignore this error. In general, trying to run these bash scripts line-by-line may not work very well (due to imports of other init files setting variables etc). You should just be running `make gen`, or the individual gen targets/scripts

damemi:
@pravarag you might also be able to take a look at the PR I just opened to add a similar verify script for the table of contents: https://github.com/kubernetes-sigs/descheduler/pull/516

That could help you on these other generators. We could also probably move some of the redundant code to a lib file so it isn't run over and over, but that's not vital.

A problem I found is that the files from `_output` (ie, our generator binaries) aren't copied to the temp directory with `git archive` (since they are .gitignore'd). So I had to manually copy that directory like https://github.com/kubernetes-sigs/descheduler/pull/516/files#diff-d257d22e73af313e0d4311fccbdd5054c2affab762a1d2965fd68217f0331963R36

Hope this helps

pravarag:
> A problem I found is that the files from _output (ie, our generator binaries) aren't copied to the temp directory with git archive (since they are .gitignore'd). So I had to manually copy that directory like https://github.com/kubernetes-sigs/descheduler/pull/516/files#diff-d257d22e73af313e0d4311fccbdd5054c2affab762a1d2965fd68217f0331963R36

Thanks @damemi , I'll check the changes done in mentioned PR.

pravarag:
@damemi I've made some changes and pushed the latest code taking reference from the script you suggested in previous comments. But, I keep getting `bad substitution` error on this [line](https://github.com/kubernetes-sigs/descheduler/pull/507/files#diff-7318ba5bb4b500e18e147fd0db7ccb39d26c47690dd9ac70ba5d4bb7e6fcc917R26) and I'm guessing it's something to do with how `diff _output` is being used. So, one doubt here - since in this case as when we run the generators, I don't see a file other than the ones generated in `_output` directory which may get changed unlike README.md in for this [change](https://github.com/kubernetes-sigs/descheduler/pull/516/files#diff-d257d22e73af313e0d4311fccbdd5054c2affab762a1d2965fd68217f0331963R45). Hence, which two files I must be comparing [here](https://github.com/kubernetes-sigs/descheduler/pull/507/commits/7abc780be29dcaa383301e14c138bb52aff9a787#diff-7318ba5bb4b500e18e147fd0db7ccb39d26c47690dd9ac70ba5d4bb7e6fcc917R26) to check the `diff` output and proceed based on that?

damemi:
@pravarag the script shouldn't be running `diff _output`, because `_output` is just the temporary directory where the generator binaries are downloaded to.

If you see [this line in `verify-vendor.sh`](https://github.com/kubernetes-sigs/descheduler/blob/6b8d4cd5a7c943d059f702a0cf93ec7a5334c96a/hack/verify-vendor.sh#L64), it is calling:
```
diff -Naupr --ignore-matching-lines='^\s*\"GoVersion\":' go.mod "${_deschedulertmp}/go.mod"
```
which is essentially
```
diff go.mod {tmpdir}/go.mod
```
Because when we update the vendor, we expect changes to `go.mod`.

If you are running the generators for this PR without any significant changes to the code (for example, modifying the API) then you will not see any generated changes besides.

As an example, if you check out [this commit from my fork](https://github.com/damemi/descheduler/commit/c169c8f0d7115cd39247c91d29b1103c60b044f9) and run `make gen`, you'll see that it adds the following to `pkg/api/v1alpha1/zz_generated.conversion.go`:
```diff
diff --git a/pkg/api/v1alpha1/zz_generated.conversion.go b/pkg/api/v1alpha1/zz_generated.conversion.go
index 727b9c776..8f56bf309 100644
--- a/pkg/api/v1alpha1/zz_generated.conversion.go
+++ b/pkg/api/v1alpha1/zz_generated.conversion.go
@@ -151,6 +151,7 @@ func autoConvert_v1alpha1_DeschedulerStrategy_To_api_DeschedulerStrategy(in *Des
        out.Enabled = in.Enabled
        out.Weight = in.Weight
        out.Params = (*api.StrategyParameters)(unsafe.Pointer(in.Params))
+       out.Foo = in.Foo
        return nil
 }
 
@@ -163,6 +164,7 @@ func autoConvert_api_DeschedulerStrategy_To_v1alpha1_DeschedulerStrategy(in *api
        out.Enabled = in.Enabled
        out.Weight = in.Weight
        out.Params = (*StrategyParameters)(unsafe.Pointer(in.Params))
+       out.Foo = in.Foo
        return nil
 }
```

So, if you imagine a PR that just has the above commit (and did not run `make gen`), then your verify script would do a couple things when it runs in CI:
1. Copy the current entire descheduler code directory to a `{tmpdir}`
2. `pushd` into the `{tmpdir}` (should be something like `_tmp/kube-vendor.ZhSdEN/descheduler` if you are running locally)
3. Run `make gen` (or each individual script) inside the `{tmpdir}`
4. `popd` out to the real directory
5. Compare the current directory to the tmpdir with something like `diff -Naupr pkg/ {deschedulertmp}/descheduler/pkg/` (any generated changes will just be under `pkg/`)

pravarag:
> So, if you imagine a PR that just has the above commit (and did not run `make gen`), then your verify script would do a couple things when it runs in CI:

@damemi thanks for suggesting the changes. I've updated the code with the latest push and have also given a run in my local with below output for each of the scripts:

hack/verify-conversions.sh
```
$  ./hack/verify-conversions.sh
Generated conversions verified.
```

hack/verify-defaulters.sh
```
$ ./hack/verify-defaulters.sh
Generated Defaulters verified.
```

hack/verify-deep-copies.sh
```
$ ./hack/verify-deep-copies.sh
Generated Deep Copes verified.
```

damemi:
This is looking good @pravarag! We are making progress. Though, it seems there is an issue here that I didn't foresee.

I checked out these scripts and made a small change to the API that should trigger a failure in `verify-conversions.sh` (but not the other scripts):
```diff
$ git diff HEAD~1
diff --git a/pkg/api/types.go b/pkg/api/types.go
index e545d8861..6b4d9ba52 100644
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -40,6 +40,9 @@ type DeschedulerPolicy struct {
 
        // MaxNoOfPodsToEvictPerNode restricts maximum of pods to be evicted per node.
        MaxNoOfPodsToEvictPerNode *int
+
+       // foo does foo
+       Foo bool
 }
 
 type StrategyName string
diff --git a/pkg/api/v1alpha1/types.go b/pkg/api/v1alpha1/types.go
index 3585addab..9b4607aa4 100644
--- a/pkg/api/v1alpha1/types.go
+++ b/pkg/api/v1alpha1/types.go
@@ -40,6 +40,9 @@ type DeschedulerPolicy struct {
 
        // MaxNoOfPodsToEvictPerNode restricts maximum of pods to be evicted per node.
        MaxNoOfPodsToEvictPerNode *int `json:"maxNoOfPodsToEvictPerNode,omitempty"`
+
+       // foo does foo
+       Foo bool `json:"foo,omitempty"`
 }
 
 type StrategyName string
```

However, all 3 failed with the same error:
```
$ make verify-gen
ls: _output/bin/golangci-lint: No such file or directory
./hack/verify-conversions.sh
Generated output differs:
diff -Naupr pkg/api/v1alpha1/zz_generated.conversion.go ./hack/../_tmp/kube-verify.PDbhgu/descheduler/pkg/api/v1alpha1/zz_generated.conversion.go
--- pkg/api/v1alpha1/zz_generated.conversion.go	2021-03-10 13:13:29.000000000 -0500
+++ ./hack/../_tmp/kube-verify.PDbhgu/descheduler/pkg/api/v1alpha1/zz_generated.conversion.go	2021-03-10 13:13:20.000000000 -0500
@@ -125,7 +125,6 @@ func autoConvert_v1alpha1_DeschedulerPol
 	out.EvictLocalStoragePods = (*bool)(unsafe.Pointer(in.EvictLocalStoragePods))
 	out.IgnorePVCPods = (*bool)(unsafe.Pointer(in.IgnorePVCPods))
 	out.MaxNoOfPodsToEvictPerNode = (*int)(unsafe.Pointer(in.MaxNoOfPodsToEvictPerNode))
-	out.Foo = in.Foo
 	return nil
 }
 
@@ -140,7 +139,6 @@ func autoConvert_api_DeschedulerPolicy_T
 	out.EvictLocalStoragePods = (*bool)(unsafe.Pointer(in.EvictLocalStoragePods))
 	out.IgnorePVCPods = (*bool)(unsafe.Pointer(in.IgnorePVCPods))
 	out.MaxNoOfPodsToEvictPerNode = (*int)(unsafe.Pointer(in.MaxNoOfPodsToEvictPerNode))
-	out.Foo = in.Foo
 	return nil
 }
 
Generated conversions verify failed.
Generated conversions verified.
./hack/verify-deep-copies.sh
Generated deep-copies output differs:
diff -Naupr pkg/api/v1alpha1/zz_generated.conversion.go ./hack/../_tmp/kube-verify.kIyT6j/descheduler/pkg/api/v1alpha1/zz_generated.conversion.go
--- pkg/api/v1alpha1/zz_generated.conversion.go	2021-03-10 13:13:29.000000000 -0500
+++ ./hack/../_tmp/kube-verify.kIyT6j/descheduler/pkg/api/v1alpha1/zz_generated.conversion.go	2021-03-10 13:13:29.000000000 -0500
@@ -125,7 +125,6 @@ func autoConvert_v1alpha1_DeschedulerPol
 	out.EvictLocalStoragePods = (*bool)(unsafe.Pointer(in.EvictLocalStoragePods))
 	out.IgnorePVCPods = (*bool)(unsafe.Pointer(in.IgnorePVCPods))
 	out.MaxNoOfPodsToEvictPerNode = (*int)(unsafe.Pointer(in.MaxNoOfPodsToEvictPerNode))
-	out.Foo = in.Foo
 	return nil
 }
 
@@ -140,7 +139,6 @@ func autoConvert_api_DeschedulerPolicy_T
 	out.EvictLocalStoragePods = (*bool)(unsafe.Pointer(in.EvictLocalStoragePods))
 	out.IgnorePVCPods = (*bool)(unsafe.Pointer(in.IgnorePVCPods))
 	out.MaxNoOfPodsToEvictPerNode = (*int)(unsafe.Pointer(in.MaxNoOfPodsToEvictPerNode))
-	out.Foo = in.Foo
 	return nil
 }
 
Generated deep-copies verify failed.
Generated deep-copies verified.
./hack/verify-defaulters.sh
Generated defaulters output differs:
diff -Naupr pkg/api/v1alpha1/zz_generated.conversion.go ./hack/../_tmp/kube-verify.Nx2CoQ/descheduler/pkg/api/v1alpha1/zz_generated.conversion.go
--- pkg/api/v1alpha1/zz_generated.conversion.go	2021-03-10 13:13:29.000000000 -0500
+++ ./hack/../_tmp/kube-verify.Nx2CoQ/descheduler/pkg/api/v1alpha1/zz_generated.conversion.go	2021-03-10 13:13:38.000000000 -0500
@@ -125,7 +125,6 @@ func autoConvert_v1alpha1_DeschedulerPol
 	out.EvictLocalStoragePods = (*bool)(unsafe.Pointer(in.EvictLocalStoragePods))
 	out.IgnorePVCPods = (*bool)(unsafe.Pointer(in.IgnorePVCPods))
 	out.MaxNoOfPodsToEvictPerNode = (*int)(unsafe.Pointer(in.MaxNoOfPodsToEvictPerNode))
-	out.Foo = in.Foo
 	return nil
 }
 
@@ -140,7 +139,6 @@ func autoConvert_api_DeschedulerPolicy_T
 	out.EvictLocalStoragePods = (*bool)(unsafe.Pointer(in.EvictLocalStoragePods))
 	out.IgnorePVCPods = (*bool)(unsafe.Pointer(in.IgnorePVCPods))
 	out.MaxNoOfPodsToEvictPerNode = (*int)(unsafe.Pointer(in.MaxNoOfPodsToEvictPerNode))
-	out.Foo = in.Foo
 	return nil
 }
 
Generated defaulters verify failed.
Generated Defaulters verified.
```

After running this, I noticed that the generators had actually made changes to my local directory (rather than the `_tmp` dir). This is odd, because your scripts are all properly pushing the copy into the tmp dir.

Then I realized, that in our scripts (like `./hack/update-generated-conversions.sh`), they are coded to run the generators at the project root (`${PRJ_PREFIX}`), not any temp directory. See:
```bash
${OS_OUTPUT_BINPATH}/conversion-gen \
    --go-header-file "hack/boilerplate/boilerplate.go.txt" \
    --input-dirs "${PRJ_PREFIX}/pkg/apis/componentconfig/v1alpha1,${PRJ_PREFIX}/pkg/api/v1alpha1" \
    --output-file-base zz_generated.conversion
```

So, what you have right now is completely correct in terms of:
1. Copy everything into tmp dir
2. Run generators
3. Compare against project root

Except in 2, we are running the generators against the project root. This isn't ideal because `verify` scripts shouldn't affect working project files.

So, what we need to do now is play with the calls to the hack scripts within your verify scripts. Specifically, we need to find a way to pass the new "PRJ_PREFIX" (`deschedulertmp`) to these binaries to fill in those parameters. I'll also play around with this, but if you find a solution before me feel free to post it :)

Thanks again for the work you're doing on this so far, this is great progress!

pravarag:
> So, what we need to do now is play with the calls to the hack scripts within your verify scripts. Specifically, we need to find a way to pass the new "PRJ_PREFIX" (deschedulertmp) to these binaries to fill in those parameters. I'll also play around with this, but if you find a solution before me feel free to post it :)

Hi @damemi thanks again for your review and as suggested, I've tried to find a way around it but I'm not sure whether that's a valid one or not.
So, what I've thought is to add an `export PRJ_PREFIX=$_deschedulertmp` here after [this](https://github.com/kubernetes-sigs/descheduler/pull/507/files#diff-7318ba5bb4b500e18e147fd0db7ccb39d26c47690dd9ac70ba5d4bb7e6fcc917R17) line and then maybe we might be able to change the actual value inside generator script for conversions to that of `_deschedulertmp`. And then one we have calculated the diff, we can do an `unset PRJ_PREFIX` here on [this](https://github.com/kubernetes-sigs/descheduler/pull/507/files#diff-7318ba5bb4b500e18e147fd0db7ccb39d26c47690dd9ac70ba5d4bb7e6fcc917R30) line. So this may sound not-so-best solution around it and I'm still trying to test it in local if that actually works. But since, we might not want to make any modifications in the generator script so might have to try this way 🤔

damemi:
Hi @pravarag 
> So, what I've thought is to add an export PRJ_PREFIX=$_deschedulertmp here after this line and then maybe we might be able to change the actual value inside generator script for conversions to that of _deschedulertmp. 

This is actually along the lines of how I was thinking we could fix this, so I think we're on the right track. However, I checked out your updated scripts and ran them with a fake change (same as above https://github.com/kubernetes-sigs/descheduler/pull/507#issuecomment-795883719) but it looks like they are still doing the same thing (modifying the root files, and then all scripts are failing with a diff).

It's important that when testing these, we not only test the "passing" case, but also the "failure" case. So when you test it locally, you should try making an API change (like above) and verify that the script fails as expected. In other words, when only `conversions` should fail, `make verify-gen` should only print a failure for that script and not modify any files in the root project tree.

I did find that manually copying the generator scripts over and replacing the `input-dir` with a leading `./` prefix worked (see this commit to see what I mean: https://github.com/damemi/descheduler/commit/bfd9658359ddc2fef758bf07ceda6ff18bdabdfd). 

```
$ make verify-gen
ls: _output/bin/golangci-lint: No such file or directory
./hack/verify-conversions.sh
PRJ_PREFIX=: ,./hack/../_tmp/kube-verify.rtXrac/descheduler
Generated output differs:
diff -Naupr pkg/api/v1alpha1/zz_generated.conversion.go ./hack/../_tmp/kube-verify.rtXrac/descheduler/pkg/api/v1alpha1/zz_generated.conversion.go
--- pkg/api/v1alpha1/zz_generated.conversion.go	2021-03-15 16:17:34.000000000 -0400
+++ ./hack/../_tmp/kube-verify.rtXrac/descheduler/pkg/api/v1alpha1/zz_generated.conversion.go	2021-03-15 16:17:48.000000000 -0400
@@ -125,6 +125,7 @@ func autoConvert_v1alpha1_DeschedulerPol
 	out.EvictLocalStoragePods = (*bool)(unsafe.Pointer(in.EvictLocalStoragePods))
 	out.IgnorePVCPods = (*bool)(unsafe.Pointer(in.IgnorePVCPods))
 	out.MaxNoOfPodsToEvictPerNode = (*int)(unsafe.Pointer(in.MaxNoOfPodsToEvictPerNode))
+	out.Foo = in.Foo
 	return nil
 }
 
@@ -139,6 +140,7 @@ func autoConvert_api_DeschedulerPolicy_T
 	out.EvictLocalStoragePods = (*bool)(unsafe.Pointer(in.EvictLocalStoragePods))
 	out.IgnorePVCPods = (*bool)(unsafe.Pointer(in.IgnorePVCPods))
 	out.MaxNoOfPodsToEvictPerNode = (*int)(unsafe.Pointer(in.MaxNoOfPodsToEvictPerNode))
+	out.Foo = in.Foo
 	return nil
 }
 
Generated conversions verify failed.
Generated conversions verified.
./hack/verify-deep-copies.sh
Generated deep-copies verified.
./hack/verify-defaulters.sh
Generated Defaulters verified.
```

However trying to set `PRJ_PREFIX="."` like how you are currently doing it did not. This is interesting, and we may want to look at that more.

For now, I will let you decide if you would like to implement these changes as I did in the commit I shared or continue investigating better ways we can effectively call the scripts directly. @ingvagabund and @seanmalloy wdyt?

pravarag:
Hi @damemi thanks for correcting me in above comment and I'll surely test by making a fake API change further. Also, I can continue to investigate a little more in finding a different way if we are not planning to merge this change urgently 🤔 Otherwise, we'll go with how you have suggested by copying the generator scripts manually.

seanmalloy:
> Also, I can continue to investigate a little more in finding a different way if we are not planning to merge this change urgently

@pravarag there is no strict timeline for getting this merged. Feel free to investigate further and come up with the best solution.

hack/verify-conversions.sh
```diff
@@ -0,0 +1,39 @@
...
+git archive --format=tar --prefix=descheduler/ "$(git write-tree)" | (cd "${_deschedulertmp}" && tar xf -)
+_deschedulertmp="${_deschedulertmp}/descheduler"
+
+pushd "${_deschedulertmp}" > /dev/null 2>&1
+export PRJ_PREFIX="${_deschedulertmp}"
+echo "PRJ_PREFIX=: ","${PRJ_PREFIX}"
+hack/update-generated-conversions.sh
+popd > /dev/null 2>&1
+
+ret=0
```
seanmalloy, line 23: Where it the codes does this get set to something other than 0?

seanmalloy:
I attempted to find an example of another k8s project that is doing a similar type of verification. I found the below example from the kind project.

Does the verification:
https://github.com/kubernetes-sigs/kind/blob/64ac2e7f5d47e391432aa13898b78aaa0b98e7a4/hack/make-rules/verify/generated.sh#L56-L74

Does the code generation:
https://github.com/kubernetes-sigs/kind/blob/64ac2e7f5d47e391432aa13898b78aaa0b98e7a4/hack/make-rules/update/generated.sh#L34-L37

Not saying we should do it exactly like this, but seeing another example might help.

pravarag:
@damemi @seanmalloy I tried to look for another way in order to pass `${PRF_PREFIX}` value to generators scripts, but couldn't find any. The main problem which I've realized:
The [init.sh](https://github.com/kubernetes-sigs/descheduler/blob/master/hack/lib/init.sh#L46) script which will run at the beginning of running generators inside generator script. And since it'll set `PRJ_PREFIX` value back to Descheduler root, that's why I think exporting `PRJ_PREFIX` didnt' work like @damemi mentioned [here](https://github.com/kubernetes-sigs/descheduler/pull/507#issuecomment-799724783).

I even tried to check references as suggested by @seanmalloy above pointing to `kind` project but that also didn't work somehow in our case. Even though they have similar structure but they are calculating `REPO_ROOT` [here](https://github.com/kubernetes-sigs/kind/blob/64ac2e7f5d47e391432aa13898b78aaa0b98e7a4/hack/make-rules/update/generated.sh#L20) before running generators and I feel that could be the difference. So for implementing something like that in our case, it might need a change in generator script or the `init.sh` script on how we have defined the `PRJ_PREFIX` currently.

There was one more way I was trying to implement (using `sed` to modify `PRJ_PREFIX` value) which is based on this SO example I found [here](https://stackoverflow.com/a/41640759/4120230) but that also didn't work somehow.

For now, I've pushed the changes in all three verify scripts similar to how @damemi suggested in his last comment by copying generators in verify scripts. Let me know if we are good to go or not 🤞🏾

hack/verify-conversions.sh
pravarag: I've removed `ret=0` with latest push, was using mainly to collect last status.

seanmalloy:
> Let me know if we are good to go or not.

@pravarag looks like one of the CI checks is failing. I can review again once the CI check is passing. Let me know if you need any help getting the CI check to pass.

Thanks!

pravarag:
@damemi @seanmalloy for the above CI check error, I can see that after I've added [this](https://github.com/kubernetes-sigs/descheduler/pull/507/files#diff-00adf27c4e4a9b010885188ccc9539b5d6e5e175fedeaf0f761e8453c3a137b8R20) change, the `verify-defaulters.sh` script is throwing below error,
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x68 pc=0x12cbce7]

goroutine 1 [running]:
k8s.io/gengo/examples/defaulter-gen/generators.getManualDefaultingFunctions(0xc004b14a10, 0x0, 0xc00ab92ab0)
	/Users/pravaragrawal/Work/go_work/src/github.com/pravarag/descheduler/_tmp/kube-verify.MMMDRZ/descheduler/vendor/k8s.io/gengo/examples/defaulter-gen/generators/defaulter.go:148 +0xa7
k8s.io/gengo/examples/defaulter-gen/generators.Packages(0xc004b14a10, 0xc00011cf00, 0x13719ab, 0x6, 0xc004b14a10)
	/Users/pravaragrawal/Work/go_work/src/github.com/pravarag/descheduler/_tmp/kube-verify.MMMDRZ/descheduler/vendor/k8s.io/gengo/examples/defaulter-gen/generators/defaulter.go:274 +0x75f
k8s.io/gengo/args.(*GeneratorArgs).Execute(0xc00011cf00, 0xc00007d050, 0x13719ab, 0x6, 0x139ef48, 0x0, 0x0)
	/Users/pravaragrawal/Work/go_work/src/github.com/pravarag/descheduler/_tmp/kube-verify.MMMDRZ/descheduler/vendor/k8s.io/gengo/args/args.go:206 +0x1a9
main.main()
	/Users/pravaragrawal/Work/go_work/src/github.com/pravarag/descheduler/_tmp/kube-verify.MMMDRZ/descheduler/vendor/k8s.io/code-generator/cmd/defaulter-gen/main.go:76 +0x385
```
I'm wondering if it could be due to the passing of `PRJ_PREFIX="."` instead of previous value is making this happen. The other two scripts did work fine. And I think for the mentioned path which would be something like `"./pkg/apis/componentconfig/v1alpha1,./pkg/api/v1alpha1"` , the generator code is not able to obtain the required package from the downloaded ones. Please correct me if I'm wrong here in my understanding.

damemi:
```
./hack/verify-conversions.sh
Generated output differs:
diff -Naupr pkg/api/v1alpha1/zz_generated.conversion.go ./hack/../_tmp/kube-verify.Tunn7Y/descheduler/pkg/api/v1alpha1/zz_generated.conversion.go
--- pkg/api/v1alpha1/zz_generated.conversion.go	2021-03-31 09:20:26.448978691 +0000
+++ ./hack/../_tmp/kube-verify.Tunn7Y/descheduler/pkg/api/v1alpha1/zz_generated.conversion.go	2021-03-31 09:22:10.661940342 +0000
@@ -123,6 +123,7 @@ func autoConvert_v1alpha1_DeschedulerPol
 	out.Strategies = *(*api.StrategyList)(unsafe.Pointer(&in.Strategies))
 	out.NodeSelector = (*string)(unsafe.Pointer(in.NodeSelector))
 	out.EvictLocalStoragePods = (*bool)(unsafe.Pointer(in.EvictLocalStoragePods))
+	out.EvictSystemCriticalPods = (*bool)(unsafe.Pointer(in.EvictSystemCriticalPods))
 	out.IgnorePVCPods = (*bool)(unsafe.Pointer(in.IgnorePVCPods))
 	out.MaxNoOfPodsToEvictPerNode = (*int)(unsafe.Pointer(in.MaxNoOfPodsToEvictPerNode))
 	return nil
@@ -137,6 +138,7 @@ func autoConvert_api_DeschedulerPolicy_T
 	out.Strategies = *(*StrategyList)(unsafe.Pointer(&in.Strategies))
 	out.NodeSelector = (*string)(unsafe.Pointer(in.NodeSelector))
 	out.EvictLocalStoragePods = (*bool)(unsafe.Pointer(in.EvictLocalStoragePods))
+	out.EvictSystemCriticalPods = (*bool)(unsafe.Pointer(in.EvictSystemCriticalPods))
 	out.IgnorePVCPods = (*bool)(unsafe.Pointer(in.IgnorePVCPods))
 	out.MaxNoOfPodsToEvictPerNode = (*int)(unsafe.Pointer(in.MaxNoOfPodsToEvictPerNode))
 	return nil
Generated conversions verify failed.
Generated conversions verified.
./hack/verify-deep-copies.sh
Generated deep-copies output differs:
diff -Naupr pkg/api/v1alpha1/zz_generated.deepcopy.go ./hack/../_tmp/kube-verify.O6DRe4/descheduler/pkg/api/v1alpha1/zz_generated.deepcopy.go
--- pkg/api/v1alpha1/zz_generated.deepcopy.go	2021-03-31 09:20:26.448978691 +0000
+++ ./hack/../_tmp/kube-verify.O6DRe4/descheduler/pkg/api/v1alpha1/zz_generated.deepcopy.go	2021-03-31 09:22:19.018658887 +0000
@@ -46,6 +46,11 @@ func (in *DeschedulerPolicy) DeepCopyInt
 		*out = new(bool)
 		**out = **in
 	}
+	if in.EvictSystemCriticalPods != nil {
+		in, out := &in.EvictSystemCriticalPods, &out.EvictSystemCriticalPods
+		*out = new(bool)
+		**out = **in
+	}
 	if in.IgnorePVCPods != nil {
 		in, out := &in.IgnorePVCPods, &out.IgnorePVCPods
 		*out = new(bool)
diff -Naupr pkg/api/zz_generated.deepcopy.go ./hack/../_tmp/kube-verify.O6DRe4/descheduler/pkg/api/zz_generated.deepcopy.go
--- pkg/api/zz_generated.deepcopy.go	2021-03-31 09:20:26.448978691 +0000
+++ ./hack/../_tmp/kube-verify.O6DRe4/descheduler/pkg/api/zz_generated.deepcopy.go	2021-03-31 09:22:19.010658199 +0000
@@ -46,6 +46,11 @@ func (in *DeschedulerPolicy) DeepCopyInt
 		*out = new(bool)
 		**out = **in
 	}
+	if in.EvictSystemCriticalPods != nil {
+		in, out := &in.EvictSystemCriticalPods, &out.EvictSystemCriticalPods
+		*out = new(bool)
+		**out = **in
+	}
 	if in.IgnorePVCPods != nil {
 		in, out := &in.IgnorePVCPods, &out.IgnorePVCPods
 		*out = new(bool)
Generated deep-copies verify failed.
Generated deep-copies verified.
./hack/verify-defaulters.sh
```

It looks like you might just need to run `make gen`. I think https://github.com/kubernetes-sigs/descheduler/pull/523 made these changes and forgot to run the generator, so now we are catching it in this PR (meaning the test works! Woo!)

You can just add that change in a new commit, since it's separate from the real work here

seanmalloy:
> It looks like you might just need to run `make gen`. I think #523 made these changes and forgot to run the generator, so now we are catching it in this PR (meaning the test works! Woo!)
> 
> You can just add that change in a new commit, since it's separate from the real work here

@pravarag I submitted PR https://github.com/kubernetes-sigs/descheduler/pull/541 to fix the generated code on the master branch. We can do the final review on your PR once then generated code is fixed on master. Thanks!

pravarag:
> @pravarag I submitted PR #541 to fix the generated code on the master branch. We can do the final review on your PR once then generated code is fixed on master. Thanks!

Thanks @seanmalloy for the `make gen` changes and I could see that the PR has been merged. I did a rebase and ran again but it still seems to be failing 😕

damemi:
Yeah, the issue I pointed out was separate and was fixed by merging the `make gen` changes. The seg fault is still there (I was hoping it would also be resolved by `make gen` but it seems it wasn't unfortunately)

pravarag:
> Yeah, the issue I pointed out was separate and was fixed by merging the make gen changes. The seg fault is still there (I was hoping it would also be resolved by make gen but it seems it wasn't unfortunately)

I was thinking if it could be coming due to me running a lower version of go (1.15) in my local. But I upgraded it to 1.16.2 and still the same result. I think it's not able to download or may be refer to some packages under `_tmp` directory while running `verify-defaulters.sh` with the path=`./hack`. It's strange that it's coming for `default generators` only 😕

seanmalloy:
> > Yeah, the issue I pointed out was separate and was fixed by merging the make gen changes. The seg fault is still there (I was hoping it would also be resolved by make gen but it seems it wasn't unfortunately)
> 
> @damemi I was thinking if it could be coming due to me running a lower version of go (1.15) in my local. But I upgraded it to 1.16.2 and still the same result. I think it's not able to download or may be refer to some packages under `_tmp` directory while running `verify-defaulters.sh` with the path=`./hack`. It's strange that it's coming for `default generators` only 😕
> 
> Update: in my local I tried to build the changes within `$GOPATH` thinking if that could be the reason pointing to the SIGSEV error but unfortunately no luck there as well. @seanmalloy @damemi could it be related to upcoming 1.21 changes or shall I wait until 1.21 changes are merged if any?

@pravarag I did not look into the seg fault error in detail. It sound like only the `verify-defaulters.sh` script is causing the seg fault. One simple idea would be to do multiple smaller PRs. You could submit a separate PR to only add the `hack/verify-conversions.sh` script. Then if that works do another PR to add `hack/verify-deep-copies.sh`.

pravarag:
> @pravarag I did not look into the seg fault error in detail. It sound like only the verify-defaulters.sh script is causing the seg fault. One simple idea would be to do multiple smaller PRs. You could submit a separate PR to only add the hack/verify-conversions.sh script. Then if that works do another PR to add hack/verify-deep-copies.sh.

Thanks @seanmalloy , as you suggested I've shortened this PR to include changes for `verify-conversions.sh` only. And if this goes well, I'll create two other PRs for `verify-deep-copies.sh` & `verify-defaulters.sh`.

damemi:
Hi @pravarag, thanks for working on this so far. I agree that we can just merge the verify checks for the 2 scripts we know are working and continue to debug why `defaulters` fails like that. Thanks for sticking with this 🙂 
/approve

pravarag:
Thanks @damemi , it was all your help :) and it's great to see my first PR coming towards a closure. Also, this PR only consists changes for `verify-conversions` so do you want me to include changes for `verify-deep-copies` as well along with this PR and take `verify-defaulters` separately? Or like how @seanmalloy mentioned above, three PRs for respective three generators?

seanmalloy:
> three PRs for respective three generators

I think we can do three separate PRs. One for each generator.

seanmalloy:
So the automated CI checks all work. But I get the below error when running `make verify-gen` from my laptop. I get the same error on Mac OSX and Linux.

```
$ make verify-gen
./hack/verify-conversions.sh
mktemp: failed to create directory via template ‘./hack/../_tmp/kube-verify.XXXXXX’: No such file or directory
make: *** [verify-gen] Error 1
```

@pravarag were you able to successfully run `make verify-gen` from your local computer? Am I doing something wrong?

pravarag:
> @pravarag were you able to successfully run make verify-gen from your local computer? Am I doing something wrong?

@seanmalloy here's an output from my local,
```
-bash>descheduler(add-verify-scripts)$
 ->  pwd
/Users/pravaragrawal/Work/go_work/src/github.com/pravarag/descheduler
-bash>descheduler(add-verify-scripts)$
 ->  make verify-gen
./hack/verify-conversions.sh
Generated conversions verified.
-bash>descheduler(add-verify-scripts)$
 ->  ls -latr
total 156
drwxr-xr-x  4 pravaragrawal staff   128 Feb 15 19:57 .github
-rw-r--r--  1 pravaragrawal staff  1591 Feb 15 19:57 CONTRIBUTING.md
-rw-r--r--  1 pravaragrawal staff 11357 Feb 15 19:57 LICENSE
-rw-r--r--  1 pravaragrawal staff   617 Feb 15 19:57 SECURITY_CONTACTS
drwxr-xr-x  3 pravaragrawal staff    96 Feb 15 19:57 charts
-rw-r--r--  1 pravaragrawal staff   886 Feb 15 19:57 cloudbuild.yaml
drwxr-xr-x  3 pravaragrawal staff    96 Feb 15 19:57 cmd
-rw-r--r--  1 pravaragrawal staff   148 Feb 15 19:57 code-of-conduct.md
drwxr-xr-x  6 pravaragrawal staff   192 Feb 15 19:57 examples
drwxr-xr-x  5 pravaragrawal staff   160 Feb 15 19:57 kubernetes
drwxr-xr-x  3 pravaragrawal staff    96 Feb 15 19:58 _output
drwxr-xr-x  6 pravaragrawal staff   192 Mar  8 20:22 ..
-rw-r--r--  1 pravaragrawal staff   200 Mar 14 12:38 .golangci.yml
-rw-r--r--  1 pravaragrawal staff    65 Apr  4 10:16 .gitignore
-rw-r--r--  1 pravaragrawal staff   949 Apr  4 10:16 Dockerfile
-rw-r--r--  1 pravaragrawal staff   777 Apr  4 10:16 Dockerfile.dev
-rw-r--r--  1 pravaragrawal staff   219 Apr  4 10:16 OWNERS
drwxr-xr-x  6 pravaragrawal staff   192 Apr  4 10:16 docs
-rw-r--r--  1 pravaragrawal staff   432 Apr  4 10:16 go.mod
-rw-r--r--  1 pravaragrawal staff 65877 Apr  4 10:16 go.sum
drwxr-xr-x  3 pravaragrawal staff    96 Apr  4 10:16 metrics
drwxr-xr-x  7 pravaragrawal staff   224 Apr  4 10:16 pkg
drwxr-xr-x 12 pravaragrawal staff   384 Apr  4 10:16 vendor
-rw-r--r--  1 pravaragrawal staff  4338 Apr  7 08:58 Makefile
drwxr-xr-x 30 pravaragrawal staff   960 Apr  7 08:58 .
-rw-r--r--  1 pravaragrawal staff 25846 Apr  7 08:58 README.md
drwxr-xr-x 19 pravaragrawal staff   608 Apr  7 08:58 hack
drwxr-xr-x  6 pravaragrawal staff   192 Apr  7 08:58 test
drwxr-xr-x 70 pravaragrawal staff  2240 Apr  7 08:58 _tmp
drwxr-xr-x 16 pravaragrawal staff   512 Apr  7 08:58 .git
```
I'm not getting the error which you mentioned above with my current clone of the repository. But, I'm not sure if it's not coming since I already have the working-repository setup. I'll try to check by doing a fresh clone as well.

pravarag:
Okay, I was getting the similar error like @seanmalloy mentioned [above](https://github.com/kubernetes-sigs/descheduler/pull/507#issuecomment-814571398) in my local(OSX) when I did a fresh clone of the fork:
```
-bash>descheduler((HEAD detached at origin/add-verify-scripts))$
 ->  make verify-gen
./hack/verify-conversions.sh
mktemp: failed to create directory via template ‘./hack/../_tmp/kube-verify.XXXXXX’: No such file or directory
make: *** [Makefile:124: verify-gen] Error 1
```
And I found the culprit, it's a missing `mkdir` command before this [line](https://github.com/kubernetes-sigs/descheduler/pull/507/files#diff-7318ba5bb4b500e18e147fd0db7ccb39d26c47690dd9ac70ba5d4bb7e6fcc917R9). After adding the same, it works fine without any error,
```
-bash>descheduler((HEAD detached at origin/add-verify-scripts))$
 ->  make verify-gen
./hack/verify-conversions.sh
Generated conversions verified.
```
I'll make changes and push the code to working branch then.

damemi:
Awesome work @pravarag! Could you please squash this down to 1 commit?
/hold
for squash

pravarag:
> Awesome work @pravarag! Could you please squash this down to 1 commit?

Squashed the commits with the latest push @damemi 🙂

damemi (commented):
/hold cancel

hack/verify-conversions.sh
```diff
@@ -0,0 +1,41 @@
...
+		--go-header-file "hack/boilerplate/boilerplate.go.txt" \
+		--input-dirs "./pkg/apis/componentconfig/v1alpha1,./pkg/api/v1alpha1" \
+		--output-file-base zz_generated.conversion
+popd > /dev/null 2>&1
+
+pushd "${DESCHEDULER_ROOT}" > /dev/null 2>&1
+if ! _out="$(diff -Naupr pkg/ "${_deschedulertmp}/pkg/")"; then
+    echo "Generated output differs:" >&2
+    echo "${_out}" >&2
+    echo "Generated conversions verify failed."
```
seanmalloy, line 31:
```suggestion
    echo "Generated conversions verify failed."
    exit 1
```

hack/verify-conversions.sh
```diff
@@ -0,0 +1,41 @@
...
+fi
+popd > /dev/null 2>&1
+ret=$?
+unset PRJ_PREFIX
+
+if [[ ${ret} -ne 0 ]]; then
+    exit ${ret}
+fi
+
```
seanmalloy, line 40:
```suggestion
```

seanmalloy (changes requested):
The script always exits zero. See inline comments for suggested fix.

hack/verify-conversions.sh
pravarag: Thanks @seanmalloy, I've made the change with latest push.

pravarag:
Updated the review comments.

pravarag:
Thanks @seanmalloy @damemi  🙂

PR #508, opened 2021-02-26
Move Inactive Maintainers to Emeritus Status

seanmalloy:
For roughly the past year damemi has been the only active approver for
the descheduler. Therefore move the inactive approvers to emeritus
status. This will help clarify to contributors who should be assigned to
pull requests.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,7 @@ RUN VERSION=${VERSION} make build.$ARCH
 
 FROM scratch
 
-MAINTAINER Avesh Agarwal <avesh.ncsu@gmail.com>
+MAINTAINER Kubernetes SIG Scheduling <kubernetes-sig-scheduling@googlegroups.com>
 
 USER 1000
 
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -13,7 +13,7 @@
 # limitations under the License.
 FROM scratch
 
-MAINTAINER Avesh Agarwal <avesh.ncsu@gmail.com>
+MAINTAINER Kubernetes SIG Scheduling <kubernetes-sig-scheduling@googlegroups.com>
 
 USER 1000
 
--- a/OWNERS
+++ b/OWNERS
@@ -1,8 +1,7 @@
 approvers:
-- aveshagarwal
-- k82cn
-- ravisantoshgudimetla
 - damemi
+- ingvagabund
+- seanmalloy
 reviewers:
 - aveshagarwal
 - k82cn
@@ -11,3 +10,7 @@ reviewers:
 - seanmalloy
 - ingvagabund
 - lixiang233
+emeritus_approvers:
+- aveshagarwal
+- k82cn
+- ravisantoshgudimetla


seanmalloy:
I did not move @aveshagarwal to emeritus status because I think it is wise to have at least 2 approvers. But maybe @ingvagabund should be added as an approver?

/assign @damemi @ravisantoshgudimetla @k82cn 
/cc @ingvagabund

ingvagabund:
I still prefer to have approvers across different companies so we don't monopolize. Though, having two approvers will allow to distribute the responsibility so @damemi does not have to carry all the burden. Unless there are volunteers, I can take it.

seanmalloy:
> I still prefer to have approvers across different companies so we don't monopolize. Though, having two approvers will allow to distribute the responsibility so @damemi does not have to carry all the burden. Unless there are volunteers, I can take it.

I'm willing to volunteer to be an approver. Although I'm not confident that I meet the criteria to be an approver.

damemi (commented):
Ravi and Klaus's contributions to the descheduler are remarkable, and I think a move to emeritus status is a great way to recognize that work. It also makes it clearer to new contributors (and ci-bot) as to who may actively review their PRs, and cuts down on notification spam for them 🙂 If they are okay with this move, then it's fine with me.

Though, I think it would be best to add new approvers before removing any. Jan and Sean, I think either (or both) of you would be excellent additions. I do understand Jan's point about company diversity, however I would like our priority to be having the most active and knowledgeable contributors as approvers, to which I think you both fit the bill.

seanmalloy:
I made some updated.

I'm proposing moving @ravisantoshgudimetla @k82cn and @aveshagarwal to emeritus status. I've added myself and @ingvagabund as new approvers.

ingvagabund:
Worth updating https://github.com/kubernetes-sigs/descheduler/blob/master/Dockerfile#L24

ravisantoshgudimetla:
Thank you for bringing this up @seanmalloy 

I agree with @damemi's assessment both @seanmalloy and @ingvagabund deserve to become approvers and +1 to making me an emeritus approver.

seanmalloy:
> Worth updating https://github.com/kubernetes-sigs/descheduler/blob/master/Dockerfile#L24

Fixed in commit 5f4dfbc9221154527bf4bff2713e8ee7ed463c65

seanmalloy:
@aveshagarwal and @k82cn any concerns with this proposed change?

seanmalloy:
@k82cn I believe you are the last person we need sign off from.

/assign @k82cn

seanmalloy:
> @k82cn I believe you are the last person we need sign off from.
> 
> /assign @k82cn

@k82cn we are planning to merge this PR on Monday March 22nd if we do not here back from you. Thanks!

damemi:
/approve
@aveshagarwal @ravisantoshgudimetla @k82cn @seanmalloy @ingvagabund thank you everyone for all of your hard work and dedication to this project 🙂

PR #490, opened 2021-01-25
Add default logging-format value

damemi:
Fixes https://github.com/kubernetes-sigs/descheduler/issues/487
/kind bug

PR #513, opened 2021-03-03
Default to text logging

ingvagabund:
So we can end when s.Validate()'s err is not nil


lixiang233 (commented):
This PR is duplicated with #490, both look good to me.

damemi (commented):
Like @lixiang233 said I did already open a PR to do this just by setting a default value in the `StringVar` flag itself which I *think* is a better way than setting it on the object, or at least clearer, but they're functionally the same. Do you have a preference for this way @ingvagabund ?

ingvagabund:
No preference here. +1 for merging https://github.com/kubernetes-sigs/descheduler/pull/490 as it was created first.

PR #516, opened 2021-03-05
Add table of contents generator/verify scripts

damemi:
This adds scripts to generate and verify that the README table of contents is updated
/kind feature


damemi (commented):
/cc @ingvagabund

hack/update-toc.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+source "$(dirname "${BASH_SOURCE}")/lib/init.sh"
+
+# forked from github.com/ekalinin/github-markdown-toc/gh-md-toc
+wget -O ${OS_OUTPUT_BINPATH}/gh-md-toc https://raw.githubusercontent.com/damemi/github-markdown-toc/hide-footer/gh-md-toc
```
seanmalloy, line 5: Should the `gh-md-toc` script just be added to the descheduler repo instead of downloading it from another repo?

ingvagabund: What about to duplicate https://github.com/kubernetes-sigs/metrics-server/blob/master/Makefile#L177-L189 ? Looks quite shorter.

damemi: Oh sweet, I did not even know there was a k8s toc generator already. I'll update to use that instead.

We can vendor it like we do with our other generators so it doesn't have to be downloaded each time.

damemi:
/close
keeping this PR as an example though

PR #521, opened 2021-03-09
Enable ARM32v7 Container Image Builds

### LINKED ISSUE #475: ARM (not ARM64) image support
I have a Raspberry Pi-based (i.e. arm) cluster I would like to be able to use descheduler on.

I would like arm supported to be added to the existing multi-arch build. Since building an arm image for descheduler is trivial, this should be easy to add.

I could just continue building my own multi-arch version of descheduler, but it seems wasteful when I'm almost certainly not the only one wanting to use it on arm.

descheduler version: 0.20.0

seanmalloy:
Add ARM32v7 in addition to the currently supported architectures. This
will allow running descheduler on Raspberry Pi v3 devices.

Fixes #475

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ BRANCH?=$(shell git branch --show-current)
 SHA1?=$(shell git rev-parse HEAD)
 BUILD=$(shell date +%FT%T%z)
 LDFLAG_LOCATION=sigs.k8s.io/descheduler/pkg/version
-ARCHS = amd64 arm64
+ARCHS = amd64 arm arm64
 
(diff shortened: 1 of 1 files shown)


seanmalloy:
/kind feature
/hold

This should work, but it doesn't. The docker manifest cannot be pushed. I think it's close, but not quite close enough.

seanmalloy:
Looks like the docker version on my work laptop is too old to get the ARMv7 docker manifest to work.

Using a newer version of docker works correctly:
image

damemi (commented):
/approve

seanmalloy:
Just need to test that I can pull the images from my test registry and update the end user docs.

seanmalloy:
I confirmed that the AMD64 image works. I have not way to test the ARM64 or ARMv7 images. I'll add the end user docs for the ARMv7 image to the end user guide closer to the v0.21.0 release.

/hold cancel

lixiang233 (commented):
/lgtm

PR #522, opened 2021-03-10
Added evictSystemCriticalPods flag to allow for eviction of pods of any priority #378

### LINKED ISSUE #378: Permit descheduling of critical pods
We scale down our dev and test GKE clusters to 0 nodes for roughly 8 hours a day - over a month this is roughly a 50% cost saving.

The problem that we have is when we scale the clusters back up, kube-dns often gets multiple instances scheduled to the same node which can often prevent our other services being scheduled properly. This is because we use the smallest possible nodes, again, for cost saving reasons.

This is a known issue: https://github.com/kubernetes/kubernetes/pull/52193 and was at one point solved with antiAffinity rules being added, however, the rules caused performance issues and were eventually removed.

Since the changes have been removed from kubernetes core and we can't modify the configuration of kube-system pods in GKE ourselves, the only other option seems to be to use descheduler, however, in the descheduler docs under "Pod Evictions" it says that critical pods can never be evicted: https://github.com/kubernetes-sigs/descheduler#pod-evictions

I can understand the reason why this is a sensible default but was wondering if the project would be open to a patch to allow this to be configurable?

RyanDevlin:
Fixes #378

The `evictSystemCriticalPods` flag disables priority checking by the descheduler. When this flag is true, pods of any priority are evicted, including system pods like `kube-dns`. Daemonsets, Mirror Pods, Static Pods, and pods without owner references are not evicted when this flag is true. If `thresholdPriority` or `thresholdPriorityClassName` are set, and `evictSystemCriticalPods` is true, then the threshold priority filtering will be disabled.


damemi (commented):
/ok-to-test

README.md
```diff
@@ -113,7 +113,8 @@ parameters associated with the strategies can be configured too. By default, all
 
 The policy also includes common configuration for all the strategies:
 - `nodeSelector` - limiting the nodes which are processed
-- `evictLocalStoragePods` - allowing to evict pods with local storage
+- `evictLocalStoragePods` - allows eviction of pods with local storage
+- `evictSystemCriticalPods` - allows eviction of pods with any priority, including system pods like kube-dns
```
damemi, line 117: It would be good to add a big, bold warning that this option should be used very carefully (and maybe a warning in the logs too)

pkg/descheduler/evictions/evictions.go
```diff
@@ -266,6 +281,11 @@ func IsMirrorPod(pod *v1.Pod) bool {
 	return utils.IsMirrorPod(pod)
 }
 
+// IsStaticPod checks whether the pod is a static pod.
+func IsStaticPod(pod *v1.Pod) bool {
+	return utils.IsStaticPod(pod)
```
damemi, line 286: I know in this PR it's just following the pattern that's already set, but is there any reason we have these single-line wrappers in this package? It seems like we could get rid of them all (`IsMirrorPod`, `IsStaticPod`, `IsPriorityPod`, etc) and just call directly to `utils.IsXXX()`

pkg/utils/pod.go
```diff
@@ -94,6 +94,11 @@ func IsStaticPod(pod *v1.Pod) bool {
 	return err == nil && source != "api"
 }
 
+// IsPriorityPod returns true if the pod is a priority pod.
+func IsPriorityPod(pod *v1.Pod) bool {
```
damemi, line 98: Maybe `IsCriticalPriorityPod` just to be clear what kind of priority we're checking

damemi (commented):
Generally looks good to me, some feedback (and could you also squash your commits?)

I was a little confused at first by how you broke the checks out from `IsCriticalPod` but I think I see now. This allows critical pods (by priority) to be evicted, but still prevents static and mirror pods. Sgtm

pkg/utils/pod.go
RyanDevlin: Good call here, I'll edit this.

README.md
RyanDevlin: Sure I can add that. Also yes you correctly saw how `IsCriticalPod` needed to be split into individual functions. Also the fact that the only place that called it was evictions.go made the decision easier.

PR #528, opened 2021-03-15
add StrategyOption structure to add allow strategies to be variadic

dbainbri-ciena:
Adds a StrategyOption, based on the GRPC client DialOption structure along with an implementation of a `WithWaitGroup` option.

The general StrategyOption capability extends the existing strategies to be variadic and allows for additional options to be added in the future without having to extend the strategy function signature.

The WithWaitGroup StrategyOption provides for a strategy to inform the calling loop when it has completed its processing. This can be important if a given strategy cannot run to completion in a single call and wants to allow further strategies to be run. This can be relevant if a strategy needs to update k8s resources which may require a retry loop based on resource version checks. As part of this implementation an additional command line parameter `--descheduling-run-timeout` was added to prevent any strategy from indefinitely blocking the descheduler.

### DIFF
--- a/cmd/descheduler/app/options/options.go
+++ b/cmd/descheduler/app/options/options.go
@@ -18,6 +18,8 @@ limitations under the License.
 package options
 
 import (
+	"time"
+
 	"github.com/spf13/pflag"
 
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
@@ -31,7 +33,8 @@ import (
 )
 
 const (
-	DefaultDeschedulerPort = 10258
+	DefaultDeschedulerPort        = 10258
+	DefaultDeschedulingRunTimeout = 1 * time.Minute
 )
 
 // DeschedulerServer configuration
@@ -50,6 +53,7 @@ func NewDeschedulerServer() (*DeschedulerServer, error) {
 	if err != nil {
 		return nil, err
 	}
+	cfg.DeschedulingRunTimeout = DefaultDeschedulingRunTimeout
 
 	secureServing := apiserveroptions.NewSecureServingOptions().WithLoopback()
 	secureServing.BindPort = DefaultDeschedulerPort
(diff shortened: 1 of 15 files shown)


dbainbri-ciena:
taking this back of CLA issues as we can't sign the CLA at this point.

damemi (commented):
@dbainbri-ciena thanks for suggesting these changes. I'm sorry that it won't be possible to contribute them without accepting the CLA terms (all k8s contributions need to be licensed for the community).

If you don't mind though, could you explain the advantage you see from this approach of passing options vs using the options struct we already pass to each strategy? This is currently how we configure specific options for different strategies. Adding in an options function pattern seems redundant to that, but perhaps there are advantages I am missing.

Thanks again for taking the time to try this!

dbainbri-ciena:
Maybe I am misunderstanding something. The function signature for a strategy is

`type strategyFunction func(ctx context.Context, client clientset.Interface, strategy api.DeschedulerStrategy, nodes []*v1.Node, podEvictor *evictions.PodEvictor`

My assumption is that your comment is referencing the `api.DeschedulerStrategy` parameter which seems to have two "main" parts: StrategyParameters and "global config settings" (i.e. enabled and weight).

The parameters being managed through the `api.DeschedulerStrategy` seem to be organized around the individual strategy where the "wait group" implementation is organized around the caller of each strategy, i.e. transference of proxy configuration file data to the execution of the given strategy, but that is a fairly minimal difference at some level and may not justify the addition of the StrategyOption.

I envisioned the StrategyOption more around parameters which I more dynamic for each run of the descheduler, again referencing the WaitGroup example. This specific example is effectively a parameter that is shared between calls to the strategies. 

So I guess my argument for such an additional is twofold: (1) the types of options are different, shared via strategy specific, and (2) the ability to add options without modifying the base "type" of the `api.DeschedulerStrategy` thus separation of concerns from strategy definition and the options to those strategies. I can see and understand an opposing viewpoint that these are minor issues that can be addressed by the current implementation.

damemi:
@dbainbri-ciena interesting, thanks for the explanation. I wasn't trying to oppose this design, just curious what the motivation was beyond the option you added here. I see your decision now, thanks for sharing! It would be great if we could work to implement something like this, but I'm glad we were able to discuss it nonetheless.

dbainbri-ciena:
No worries and didn't mean to suggest you were opposing the idea, although I would argue that opposition-based push back can be a good tool to get a better idea in the end. 

I will be working on the CLA issues internally, but until they are settled had to revoke the request sorry.

PR #535, opened 2021-03-28
E2e refactor

ingvagabund:
- `TestLowNodeUtilization`: normalize nodes before running the strategy
- `TestEvictAnnotation`: replace LowNodeUtilization strategy with PodLifetime
- wait for all pods to disappear before deleting RC


test/e2e/e2e_test.go
```diff
@@ -167,24 +170,144 @@ func TestLowNodeUtilization(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("Unable to create load balancing pods: %v", err)
+	}
+	defer cleanUp()
+
+	t.Log("Creating pods each consuming 10% of node's allocatable")
+	nodeCpu := workerNodes[0].Status.Allocatable[v1.ResourceCPU]
+	tenthOfCpu := int64(float64((&nodeCpu).MilliValue()) * 0.1)
+
+	t.Log("Creating pods all bound to a single node")
```
ingvagabund, line 164: @damemi @seanmalloy I did not like the way TestLowNodeUtilization was tested. The threshold of deciding when a node is overutilized and underutilized was quite thin. Now, the idea is to resource balance all worker nodes first (all eating 50% of resources) and then overutilizing the first worker node. Once LowNodeUtilization is ran, the test verifies at least one pod was evicted from the overutilized node. Open to more improvements.

test/test_utils.go
```diff
@@ -173,3 +184,227 @@ func SetPodPriority(pod *v1.Pod, priority int32) {
 func SetNodeUnschedulable(node *v1.Node) {
 	node.Spec.Unschedulable = true
 }
+
+var balancePodLabel = map[string]string{"podname": "priority-balanced-memory"}
+
+// track min memory limit based on crio minimum. pods cannot set a limit lower than this
+// see: https://github.com/cri-o/cri-o/blob/29805b13e9a43d9d22628553db337ce1c1bec0a8/internal/config/cgmgr/cgmgr.go#L23
+// see: https://bugzilla.redhat.com/show_bug.cgi?id=1595256
+var crioMinMemLimit = 12 * 1024 * 1024
```
ingvagabund, line 193: @damemi I borrowed bits from k/k's e2e's for scheduler priorities

ingvagabund:
@damemi @seanmalloy @lixiang233 not perfect though, more utils might be moved over the place.

test/e2e/e2e_test.go
```diff
@@ -240,24 +210,145 @@ func TestLowNodeUtilization(t *testing.T) {
...
-		t.Errorf("Error creating deployment %v", err)
+		t.Fatalf("Error creating RC %v", err)
+	}
+	defer deleteRC(ctx, t, clientSet, rc)
+	waitForRCPodsRunning(ctx, t, clientSet, rc)
+
+	// Run LowNodeUtilization strategy
+	evictionPolicyGroupVersion, err := eutils.SupportEviction(clientSet)
+	if err != nil || len(evictionPolicyGroupVersion) == 0 {
+		klog.Fatalf("%v", err)
```
seanmalloy, line 303: Should this be `t.Fatalf()` instead of using `klog`?

Looks like this is one of three places where `klog` is used in the e2e tests. Not sure why it is necessary use `klog` in the e2e tests.

ingvagabund: > Should this be t.Fatalf() instead of using klog?

Yes, thanks for noticing. Done.

seanmalloy:
This looks good to me. Let's wait for @damemi or @lixiang233 to review before merging.

test/e2e/e2e_test.go
damemi: Making sure I understand this, you're doing:
1. Balance nodes at (at least) 50% usage using `createBalancedPodForNodes`
2. Create 5 more pods on 1 node, bringing that node to 100% usage
3. Run LNU with `lowThreshold: 70%` and `highThreshold: 80%`

My only feedback is maybe we should be careful about trying to fill the node up to 100%, if it's running on a high-usage cluster then this test might balance the pods at higher than 50%, then the test will fail when you can't create all 5 pods.

test/e2e/e2e_test.go
```diff
@@ -240,24 +210,145 @@ func TestLowNodeUtilization(t *testing.T) {
...
+		evictionPolicyGroupVersion,
+		false,
+		0,
+		nodes,
+		true,
+		false,
+		false,
+	)
+
+	podsOnleastUtilizedNode, err := podutil.ListPodsOnANode(ctx, clientSet, workerNodes[0], podutil.WithFilter(podEvictor.Evictable().IsEvictable))
```
damemi, line 316: isn't `workerNodes[0]` the most utilized node? That's the node you added those 5 pods on above

test/e2e/e2e_test.go
```diff
@@ -891,47 +1042,226 @@ func deleteRC(ctx context.Context, t *testing.T, clientSet clientset.Interface,
...
+		v1.ResourceCPU:    resource.MustParse("100m"),
+	},
+	Requests: v1.ResourceList{
+		v1.ResourceMemory: resource.MustParse("100Mi"),
+		v1.ResourceCPU:    resource.MustParse("100m"),
+	},
+}
+
+// createBalancedPodForNodes creates a pod per node that asks for enough resources to make all nodes have the same mem/cpu usage ratio.
+func createBalancedPodForNodes(
```
damemi, line 1064: Maybe this could be moved to `component-helpers`, since it currently lives in https://github.com/kubernetes/kubernetes/blob/84483a5/test/e2e/scheduling/priorities.go#L478 and is used in a couple places. Could be a good external scheduling test util function to have

test/e2e/e2e_test.go
ingvagabund: Right, I forgot to rename `podsOnleastUtilizedNode` to `podsOnMostUtilizedNode`

test/e2e/e2e_test.go
ingvagabund: That's a good point. Decreased to 4 pods.

test/e2e/e2e_test.go
ingvagabund: It turned out this `createBalancedPodForNodes` is not the exact copy of the one in https://github.com/kubernetes/kubernetes/blob/84483a5/test/e2e/scheduling/priorities.go#L478. Priorities take [non-zero requests](https://github.com/kubernetes/kubernetes/blob/84483a5/test/e2e/scheduling/priorities.go#L626) whereas descheduler takes even zero requests. Also, `schedutil.GetNonzeroRequests` is still in [k/k repo](https://github.com/kubernetes/kubernetes/blob/84483a5aacba8e535c9e185e9b0c01a5248b5086/pkg/scheduler/util/non_zero.go#L42) including [DefaultMilliCPURequest and DefaultMemoryRequest](https://github.com/kubernetes/kubernetes/blob/84483a5aacba8e535c9e185e9b0c01a5248b5086/pkg/scheduler/util/non_zero.go#L35-L37) consts which can not be separated from the scheduling framework :-|.

test/e2e/e2e_test.go
```diff
@@ -46,6 +48,8 @@ import (
 	nodeutil "sigs.k8s.io/descheduler/pkg/descheduler/node"
 	podutil "sigs.k8s.io/descheduler/pkg/descheduler/pod"
 	"sigs.k8s.io/descheduler/pkg/descheduler/strategies"
+
+	"sigs.k8s.io/descheduler/pkg/utils"
```
lixiang233, line 52: Can you put this line together with these `descheduler` imports above?

ingvagabund: Done

test/e2e/e2e_test.go
```diff
@@ -240,24 +209,145 @@ func TestLowNodeUtilization(t *testing.T) {
...
+
+	podsOnMosttUtilizedNode, err = podutil.ListPodsOnANode(ctx, clientSet, workerNodes[0], podutil.WithFilter(podEvictor.Evictable().IsEvictable))
+	if err != nil {
+		t.Errorf("Error listing pods on a node %v", err)
+	}
+	podsAfter := len(podsOnMosttUtilizedNode)
+
+	if podsBefore > podsAfter {
+		t.Fatalf("No pod has been evicted from %v node", workerNodes[0].Name)
+	}
```
lixiang233, line 350: I think `podsAfter` is supposed to be less than `podsBefore` given some pods will be evicted from `MosttUtilizedNode` by `LNU`.

lixiang233: Can you add some logs to show the actual value of `podsBefore` and `podsAfter`? I think it'll be helpful if this testcase failed.

ingvagabund: Oh man, nice catch!!! Thanks.

It actually takes a while before the evicted pod disappears from the node. Added `waitForTerminatingPodsToDisappear` check before checking the numbers. The test logs look like this now:

test/e2e/e2e_test.go
damemi: Ah, it's more complicated than I thought then. That's fine, this can be a TODO at a later time as we improve the exportability of some of these functions. Maybe we can add a comment about that function in k/k just as a reference of what this is based off of

ingvagabund:
```
// TODO(jchaloup): The function is updated version of what under https://github.com/kubernetes/kubernetes/blob/84483a5/test/e2e/scheduling/priorities.go#L478.
//                 Import it once the function is moved under k8s.io/components-helper repo and modified to work for both priority and predicates cases.
```
added to the code

damemi: Thanks!

PR #543, opened 2021-04-05
e2e_test: fix goroutine leak

gaurav1086:
If the goroutine running the descheduler strategies times out, it will still write to the channel without a reader thus causing a goroutine leak. To fix this, use a buffered channel (c).


damemi (commented):
/ok-to-test
/approve
thanks @gaurav1086 !

lixiang233 (commented):
/lgtm

PR #546, opened 2021-04-08
Add diagram of strategies

ingvagabund:
Provide diagram with various strategies and the way they operate over pods.

All except LNU (it does not fit properly into the diagram).


ingvagabund:
@damemi @seanmalloy @lixiang233 the diagram is not necessarily a final one. Happy to improve or add more information into it. The goal is to make a high level diagram to show how each strategy works in general. LNU is not included since it does not fit there.

damemi (commented):
Looks nice, do you want to also add some for the individual strategy sections (or maybe just add one under LNU to explain that a bit more?)

I really like how this fits into the readme though, which is currently a lot of text. Breaking it up with some diagrams will make it more digestable

seanmalloy:
I think the diagram looks very good. Thanks!

I'll let you work out with @damemi if you want to add more diagrams in this PR or a separate PR.

ingvagabund:
> Looks nice, do you want to also add some for the individual strategy sections (or maybe just add one under LNU to explain that a bit more?)

I'd like to avoid creating a single diagram for each strategy. Some strategies are quite straightforward to understand  (e.g. PodLifeTime, RemovePodsHavingTooManyRestarts, RemovePodsViolatingNodeTaints) and a diagram with a single node and one or two pods would be too simple. Unless each diagram gets more detailed.

Providing more diagrams later sounds better. `LowNodeUtilization` requires different approach, `PodLifeTime`'s diagram might contain warnings about setting the TTL too low (e.g. in case of VMs), `RemovePodsViolatingTopologySpreadConstraint` might include soft constraints, etc.

damemi:
@ingvagabund sounds good to me, yeah everything doesn't need a diagram and this is a great summary of all the strategies working together.
/lgtm

PR #548, opened 2021-04-10
Updating policy api version used in pod evictor

lx1036:
fix eviction APIVersion typo in test case

### DIFF
--- a/pkg/descheduler/strategies/duplicates_test.go
+++ b/pkg/descheduler/strategies/duplicates_test.go
@@ -21,6 +21,7 @@ import (
 	"testing"
 
 	v1 "k8s.io/api/core/v1"
+	policyv1 "k8s.io/api/policy/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -402,7 +403,7 @@ func TestRemoveDuplicatesUniformly(t *testing.T) {
 			})
 			podEvictor := evictions.NewPodEvictor(
 				fakeClient,
-				"v1",
+				policyv1.SchemeGroupVersion.String(),
 				false,
 				testCase.maxPodsToEvictPerNode,
 				testCase.nodes,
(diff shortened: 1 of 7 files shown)


damemi (commented):
/ok-to-test
These could actually probably reference the [Policy GroupVersion constant](https://github.com/kubernetes/api/blob/v0.21.0/policy/v1beta1/register.go#L29) directly, and then parse it with [GroupVersion.String()](https://pkg.go.dev/k8s.io/apimachinery/pkg/runtime/schema#GroupVersion.String)

damemi:
Just a nit but could you also retitle the PR to better describe that this is actually updating the Policy API version used in PodEvictor?

lx1036:
> Just a nit but could you also retitle the PR to better describe that this is actually updating the Policy API version used in PodEvictor?

thanks, i have retitled the PR,  and updated the commit content and message.

pkg/descheduler/strategies/duplicates_test.go
```diff
@@ -21,6 +21,7 @@ import (
 	"testing"
 
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/api/policy/v1beta1"
```
seanmalloy, line 24: How about ...
```suggestion
	policyv1 "k8s.io/api/policy/v1"
```

pkg/descheduler/strategies/duplicates_test.go
```diff
@@ -402,7 +403,7 @@ func TestRemoveDuplicatesUniformly(t *testing.T) {
 			})
 			podEvictor := evictions.NewPodEvictor(
 				fakeClient,
-				"v1",
+				v1beta1.SchemeGroupVersion.String(),
```
seanmalloy, line 406: How about ...
```suggestion
				policyv1.SchemeGroupVersion.String(),
```

seanmalloy:
@damemi please take another look at this PR when you get a chance. Thanks!

/lgtm

seanmalloy (approved):

pkg/descheduler/strategies/duplicates_test.go
lx1036: resolve it. thanks

pkg/descheduler/strategies/duplicates_test.go
lx1036: resolve it. thanks

lx1036:
@seanmalloy  have used "k8s.io/api/policy/v1". thanks.

damemi (commented):
/approve

PR #553, opened 2021-04-23
[wip]Add same workload metric

ravisantoshgudimetla:
A metric to count the number of duplicate pods running on a node. We can convert this to an alert later for certain components based on namespace label.

@ingvagabund - Not sure if this the correct place to inject this metric. PTAL

### DIFF
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -48,9 +48,23 @@ var (
 		},
 	)
 
+	DuplicatePods = metrics.NewGaugeVec(
+		&metrics.GaugeOpts{
+			Subsystem:      DeschedulerSubsystem,
+			Name:           "pods_duplicate",
+			Help:           "Number of duplicate pods on a node",
+			StabilityLevel: metrics.ALPHA,
+		},
+		// name is the name of the workload
+		// kind is the type of the workload
+		// namespace is the namespace in which workload is running
+		// node is the node on which the workload pods are running
+		[]string{"name", "kind", "namespace", "node"})
+
 	metricsList = []metrics.Registerable{
 		PodsEvicted,
 		buildInfo,
+		DuplicatePods,
 	}
 )
 
--- a/pkg/descheduler/evictions/evictions.go
+++ b/pkg/descheduler/evictions/evictions.go
@@ -114,7 +114,6 @@ func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod, node *v1.Node, 
 		metrics.PodsEvicted.With(map[string]string{"result": "maximum number reached", "strategy": strategy, "namespace": pod.Namespace}).Inc()
 		return false, fmt.Errorf("Maximum number %v of evicted pods per %q node reached", pe.maxPodsToEvictPerNode, node.Name)
 	}
-
 	err := evictPod(ctx, pe.client, pod, pe.policyGroupVersion, pe.dryRun)
 	if err != nil {
 		// err is used only for logging purposes
@@ -122,7 +121,6 @@ func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod, node *v1.Node, 
 		metrics.PodsEvicted.With(map[string]string{"result": "error", "strategy": strategy, "namespace": pod.Namespace}).Inc()
 		return false, nil
 	}
-
 	pe.nodepodCount[node]++
 	if pe.dryRun {
 		klog.V(1).InfoS("Evicted pod in dry run mode", "pod", klog.KObj(pod), "reason", reason)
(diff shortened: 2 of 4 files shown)


pkg/descheduler/strategies/duplicates.go
```diff
@@ -167,6 +168,9 @@ func RemoveDuplicatePods(
 								duplicatePods[ownerKey] = make(map[string][]*v1.Pod)
 							}
 							duplicatePods[ownerKey][node.Name] = append(duplicatePods[ownerKey][node.Name], pod)
+							// We would like to see the same copy
+							metrics.DuplicatePods.With(map[string]string{"workloadName": ownerKey.name,
```
damemi, line 172: does this need to reset or account for when duplicate pods are removed? Or is it just infinitely increasing?

metrics/metrics.go
```diff
@@ -48,6 +48,19 @@ var (
 		},
 	)
 
+	DuplicatePods = metrics.NewCounterVec(
+		&metrics.CounterOpts{
+			Subsystem:      DeschedulerSubsystem,
+			Name:           "pods_duplicate",
+			Help:           "Number of duplicate pods on a node",
```
ingvagabund, line 55: You mean duplicates as they are defined by RemoveDuplicates strategy? Or, is the metrics supposed to include duplicates of other strategies (in the future) as well?

pkg/descheduler/strategies/duplicates.go
ingvagabund: What's the workload in this context? Might still be more user friendly to use owner/kind directly. E.g. in case of a deployment, one gets RS.

metrics/metrics.go
ravisantoshgudimetla: Ya, I want to include statefulsets in future.

pkg/descheduler/strategies/duplicates.go
ravisantoshgudimetla: @damemi - You're right. I have used a CounterVec and couldn't find decrement method associated with it. I changed it to Guage now and we should be good.

ravisantoshgudimetla (commented):
@ingvagabund and @damemi - I wish to test this on a cluster before it gets merged. Let me know if you have any questions around code.

pkg/descheduler/strategies/duplicates.go
ingvagabund: I am not in a favor of prefixing the names with `workload` keyword and inventing new names where we can still use the existing `kind` and `name`. Unless there's clear benefit to consumers of the new metric.

pkg/descheduler/strategies/duplicates.go
```diff
@@ -196,6 +201,8 @@ func RemoveDuplicatePods(
 						klog.ErrorS(err, "Error evicting pod", "pod", klog.KObj(pod))
 						break
 					}
+					metrics.DuplicatePods.With(map[string]string{"workloadName": ownerKey.name,
+						"workloadKind": ownerKey.kind, "namespace": pod.Namespace, "node": nodeName}).Dec()
```
ingvagabund, line 205: I'd like to see a test which will validate the new metric does not increase indefinitely. E.g. in case some change gets merged which will cause the metrics to only get increased.

ingvagabund (changes requested):

pkg/descheduler/strategies/duplicates.go
ravisantoshgudimetla: Kind and Name are fine but we want to be generic enough to be used in workload context. The workload can be anything STS, deployment etc. I don't want to harp on this. I'll change it.

pkg/descheduler/evictions/evictions.go
```diff
@@ -130,6 +128,14 @@ func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod, node *v1.Node,
 		r := eventBroadcaster.NewRecorder(scheme.Scheme, v1.EventSource{Component: "sigs.k8s.io.descheduler"})
 		r.Event(pod, v1.EventTypeNormal, "Descheduled", fmt.Sprintf("pod evicted by sigs.k8s.io/descheduler%s", reason))
 		metrics.PodsEvicted.With(map[string]string{"result": "success", "strategy": strategy, "namespace": pod.Namespace}).Inc()
+		var ownerRefName, ownerRefKind string
```
ravisantoshgudimetla, line 131: I had to add these to get ownerRefs again. I am open to other suggestions.

pkg/descheduler/strategies/duplicates_test.go
```diff
@@ -420,3 +423,92 @@ func TestRemoveDuplicatesUniformly(t *testing.T) {
 		})
 	}
 }
+
+func TestDuplicatePodMetricValidation(t *testing.T) {
```
ravisantoshgudimetla, line 427: Increment and decrement are exercised in this test

ravisantoshgudimetla: I can add few more tests with pods on multiple nodes if needed

pkg/descheduler/evictions/evictions.go
```diff
@@ -134,6 +132,14 @@ func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod, node *v1.Node,
...
+		ownerRefs := pod.OwnerReferences
+		for _, ownerRef := range ownerRefs {
+			ownerRefName = ownerRef.Name
+			ownerRefKind = ownerRef.Kind
+		}
+		metrics.DuplicatePods.With(map[string]string{"name": ownerRefName,
+			"kind": ownerRefKind, "namespace": pod.Namespace, "node": node.Name}).Dec()
```
ingvagabund, line 142: This will get invoked by every strategy calling `EvictPod`, not just by `RemoveDuplicatePods`. Was there an issue decreasing the metrics right after `EvictPod` in `RemoveDuplicatePods`? Also, I'd rather expect `RemoveDuplicatePods` to keep count of `duplicated` and `evicted` pods and then set the metrics to `duplicated` - `evicted`. Is there any value in watching eviction of duplicated pods by one? Pods get evicted quite quickly since a strategy does not wait until a pod is terminated. It might be interesting to see it in case the strategy evicts hundreds of pods. Although, that might indicate a serious problem in the cluster (e.g. strategy constrains do not balance the pods properly).

PR #554, opened 2021-04-25
code cleanup: remove check on length

BinacsLee:
When `len(taints) == 0`, the `for-loop` will be skipped and return `true` directly.


damemi (commented):
/ok-to-test
/lgtm
/approve

Issue #486, opened 2021-01-22
Multiple descheduler strategies of the same type


irmiller22:
**Is your feature request related to a problem? Please describe.**

k8s version: 1.18
descheduler version: 1.18

Currently, we have `CronJob` resources that are getting stuck due to this k8s issue: https://github.com/kubernetes/kubernetes/issues/52172. We are using managed k8s (EKS), and 1.19 is not available yet. The issue is apparently fixed in 1.19, but we're unable to upgrade.

Long story short, a `CronJob` pod will spin up, and will immediately get stuck due to the bug described in the linked issue above, and the pod will ultimately end up in `Waiting` state with the reason being `CreateContainerError`.

```
State:          Waiting
  Reason:       CreateContainerError
 ```
 
 The only way to address this currently is to manually delete the problematic pods and have k8s re-schedule the CronJob objects during the next scheduled run. Currently, these problematic pods are blocking the CronJob resource from scheduling new pods.

**Describe the solution you'd like**
We'd like to leverage descheduler to handle this case by allowing for multiple `PodLifeTime` strategies. As I understand it from the documentation, we're only allowed to implement one definition per strategy. We'd like to have a policy that applies only to our `CronJob` objects (which have a priority class called `service-cron`), and another policy that applies to all other pods.

For example, the below configuration would be desired:

```yaml
apiVersion: "descheduler/v1alpha1"
kind: "DeschedulerPolicy"
strategies:
  "PodLifeTime":
     enabled: true
     params:
       podLifeTime:
         maxPodLifeTimeSeconds: 86400
         podStatusPhases:
         - "Pending"
  "PodLifeTime":
     enabled: true
     params:
       podLifeTime:
        maxPodLifeTimeSeconds: 86400
        podStatusPhases:
         - "Running"
        thresholdPriorityClassName: "service-cron" # this is a custom priority class
```

If there are other ways to accomplish this, please let me know! I'm open to suggestions.

**Describe alternatives you've considered**

We've considered upgrading our self-managed clusters to 1.19, but our managed clusters are unable to upgrade since 1.19 isn't available yet. We'd prefer not to run into version drift across our clusters.

**What version of descheduler are you using?**

descheduler version: 1.18

**Additional context**

damemi:
Hi @irmiller22, thanks for opening this issue. We have discussed similar requests before (I thought there was an open issue, but I couldn't find it), so there is definitely some validity to what you describe.

[...]

PR #556, opened 2021-04-26
Invert main strategy loop for performance and customizability

damemi:
I noticed this while thinking about the new API options we started discussing in https://github.com/kubernetes-sigs/descheduler/issues/486

Right now we are checking against every strategy function and then seeing if it's enabled. This does 2 things:
1. Wastes time for strategies that aren't enabled
2. Only allows strategies to be run in a fixed order

1. is probably negligible, but 2. is something we should consider providing. I don't know if the current order is hardcoded for a reason. There also hasn't been any demand for this specific functionality that I know of, but while we are considering improvements to the API I think a change like this could make it more flexible.

Either way, this does also add an error for invalid strategy names (where it currently just silently skips them)

### DIFF
--- a/pkg/descheduler/descheduler.go
+++ b/pkg/descheduler/descheduler.go
@@ -72,7 +72,7 @@ func RunDeschedulerStrategies(ctx context.Context, rs *options.DeschedulerServer
 	sharedInformerFactory.Start(stopChannel)
 	sharedInformerFactory.WaitForCacheSync(stopChannel)
 
-	strategyFuncs := map[string]strategyFunction{
+	strategyFuncs := map[api.StrategyName]strategyFunction{
 		"RemoveDuplicates":                            strategies.RemoveDuplicatePods,
 		"LowNodeUtilization":                          strategies.LowNodeUtilization,
 		"RemovePodsViolatingInterPodAntiAffinity":     strategies.RemovePodsViolatingInterPodAntiAffinity,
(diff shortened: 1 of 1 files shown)


pkg/descheduler/descheduler.go
```diff
@@ -136,9 +136,11 @@ func RunDeschedulerStrategies(ctx context.Context, rs *options.DeschedulerServer
 			ignorePvcPods,
 		)
 
-		for name, f := range strategyFuncs {
-			if strategy := deschedulerPolicy.Strategies[api.StrategyName(name)]; strategy.Enabled {
+		for name, strategy := range deschedulerPolicy.Strategies {
+			if f := strategyFuncs[name]; f != nil && strategy.Enabled {
 				f(ctx, rs.Client, strategy, nodes, podEvictor)
+			} else {
+				klog.Errorf("Unknown strategy name '%s', exiting", name)
```
ingvagabund, line 143: This will log even when the strategy exists but is not enabled.

damemi: Good catch, thanks

pkg/descheduler/descheduler.go
```diff
@@ -136,9 +136,13 @@ func RunDeschedulerStrategies(ctx context.Context, rs *options.DeschedulerServer
...
-		for name, f := range strategyFuncs {
-			if strategy := deschedulerPolicy.Strategies[api.StrategyName(name)]; strategy.Enabled {
-				f(ctx, rs.Client, strategy, nodes, podEvictor)
+		for name, strategy := range deschedulerPolicy.Strategies {
+			if f, ok := strategyFuncs[name]; ok {
+				if strategy.Enabled {
+					f(ctx, rs.Client, strategy, nodes, podEvictor)
+				}
+			} else {
+				klog.Errorf("Unknown strategy name '%s', exiting", name)
```
ingvagabund, line 145: Maybe skipping instead of existing? Exiting might give impression the descheduler is aborting actually.

Issue #529, opened 2021-03-17
Improve LowNodeUtilization's consideration of evictable pods


damemi:
Similar to how we check that a pod fits on other nodes before evicting in strategies like Affinity/TopologySpread, it would be an improvement if LowNodeUtilization checked that a pod from an overutilized node can fit on the intended underutilized node before evicting it (ie, check for taints/affinity). This could improve the performance of the strategy and prevent unnecessary infinite evictions each run.

RyanDevlin:
I can try this one.

RyanDevlin:
@damemi I've noticed that the Affinity and TopologySpread strategies check for affinity on other nodes, but don't necessarily check for taints. Is checking for taints on other nodes still something that should be included in this feature?

rustrial:
@RyanDevlin I guess eviction only makes sense if the Pod can be scheduled on another node, therefore it seems reasonable that all scheduling constraints (Affinity, TopologySpread and Taints) should be checked. Otherwise we would just evict a Pod to see it being scheduled to the same node again, which just increases the risk of service outages caused by Pod eviction (restart).

RyanDevlin:
@rustrial At the bottom of #551 @ingvagabund commented about how, for the purposes of this feature, it would be overkill to take into account every filtering plugin. I'm currently implementing this feature with checks for NodeAffinity, Taints, and NodeSelector. The optimization isn't perfect, but statistically it should improve performance.

ingvagabund:
Overkill on the code level view (duplicating the code since we can't import kubernetes/kubernetes code).

ingvagabund:
> Otherwise we would just evict a Pod to see it being scheduled to the same node again, which just increases the risk of service outages caused by Pod eviction (restart).

That's responsibility of PDB to take care of disruptions.

damemi:
> @RyanDevlin I guess eviction only makes sense if the Pod can be scheduled on another node

I mentioned this in the other thread (https://github.com/kubernetes-sigs/descheduler/issues/551#issuecomment-820542336), sorry for not bringing it up here... but there are use cases where a pod could be evicted regardless of if it fits on another node. For strategies like `PodLifetime`, you might actually want to evict a pod that will only be recreated on the same node

jecnua:
Still valid

damemi:
Bumped the NodeFit PR to get that moving again
/remove-lifecycle stale

ingvagabund:
[...]

Issue #551, opened 2021-04-15
RemoveDuplicates: take taints and node selector into account when computing the number of duplicates


ingvagabund:
**Is your feature request related to a problem? Please describe.**
In case there's a cluster with 3 master nodes and 3 workers and a workload (owned by an RC) is uniformly distributed among worker nodes (the expected distribution), the strategy still evicts pods of the workload since it takes master nodes into account as well. E.g. 12 pods, 4 pods bound to each worker node, 6 pods are evicted (6 nodes, 12 pods, 2 per node) which always end up on the same worker nodes again.

**Describe the solution you'd like**
Take at least node taints (e.g. taints repealing pods from master nodes) into account when selecting pods for eviction. If possible, take node selector into account as well.

**Describe alternatives you've considered**
Not considered

**What version of descheduler are you using?**
The latest master HEAD

descheduler version:
Assume 0.20

**Additional context**

damemi:
We could probably open an umbrella issue to get all strategies in line with this. I also have https://github.com/kubernetes-sigs/descheduler/issues/529 for LNU and @RyanDevlin pointed out the same with Affinity and PTS https://github.com/kubernetes-sigs/descheduler/issues/529#issuecomment-820393845

RyanDevlin:
It seems like this could be implemented as a global feature in `evictions.go`. Maybe there could be a modification to the Evictable function to take the context of cluster nodes into account. So instead of:
```
func (pe *PodEvictor) Evictable(opts ...func(opts *Options)) *evictable {}
```

We have: 
```
func (pe *PodEvictor) Evictable(nodes []*v1.Node, opts ...func(opts *Options)) *evictable {}
```

Inside this function (possibly with a call to a utility function), the node taints and pod tolerations, RequiredDuringSchedulingIgnoredDuringExecution affinity rules, and node selectors could be checked. All the strategies would most likely need to be modified after this addition to __not check__ affinity rules or taints if they already do.

This way, the optimization is globally implemented. Thoughts on this approach?

damemi:
[...]

PR #559, opened 2021-04-30
Working nodeFit feature

RyanDevlin:
Implements #529.

This feature adds a `nodeFit` boolean parameter to all descheduler strategies. When `nodeFit` is set to `true`, the descheduler strategy will take "node fit" into account when evicting pods. This increases the probability that pods can be re-scheduled to a new node successfully upon eviction.

For the current implementation, "node fit" is defined by the following criteria:
- A `nodeSelector` on the pod
- Any `Tolerations` on the pod and any `Taints` on the other nodes
- `nodeAffinity` on the pod
- Whether any of the other nodes are marked as `unschedulable`

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ Table of Contents
   - [Namespace filtering](#namespace-filtering)
   - [Priority filtering](#priority-filtering)
   - [Label filtering](#label-filtering)
+  - [Node Fit filtering](#node-fit-filtering)
 - [Pod Evictions](#pod-evictions)
   - [Pod Disruption Budget (PDB)](#pod-disruption-budget-pdb)
 - [Metrics](#metrics)
@@ -157,6 +158,7 @@ should include `ReplicaSet` to have pods created by Deployments excluded.
 |`namespaces`|(see [namespace filtering](#namespace-filtering))|
 |`thresholdPriority`|int (see [priority filtering](#priority-filtering))|
 |`thresholdPriorityClassName`|string (see [priority filtering](#priority-filtering))|
+|`nodeFit`|bool (see [node fit filtering](#node-fit-filtering))|
 
 **Example:**
 ```yaml
@@ -204,6 +206,7 @@ strategy evicts pods from `overutilized nodes` (those with usage above `targetTh
 |`numberOfNodes`|int|
 |`thresholdPriority`|int (see [priority filtering](#priority-filtering))|
 |`thresholdPriorityClassName`|string (see [priority filtering](#priority-filtering))|
+|`nodeFit`|bool (see [node fit filtering](#node-fit-filtering))|
 
 **Example:**
 
@@ -253,6 +256,7 @@ node.
 |`thresholdPriorityClassName`|string (see [priority filtering](#priority-filtering))|
 |`namespaces`|(see [namespace filtering](#namespace-filtering))|
 |`labelSelector`|(see [label filtering](#label-filtering))|
+|`nodeFit`|bool (see [node fit filtering](#node-fit-filtering))|
 
 **Example:**
 
@@ -291,6 +295,7 @@ podA gets evicted from nodeA.
 |`thresholdPriorityClassName`|string (see [priority filtering](#priority-filtering))|
 |`namespaces`|(see [namespace filtering](#namespace-filtering))|
 |`labelSelector`|(see [label filtering](#label-filtering))|
+|`nodeFit`|bool (see [node fit filtering](#node-fit-filtering))|
 
 **Example:**
 
@@ -320,6 +325,7 @@ and will be evicted.
 |`thresholdPriorityClassName`|string (see [priority filtering](#priority-filtering))|
 |`namespaces`|(see [namespace filtering](#namespace-filtering))|
 |`labelSelector`|(see [label filtering](#label-filtering))|
+|`nodeFit`|bool (see [node fit filtering](#node-fit-filtering))|
 
 **Example:**
 
@@ -348,6 +354,7 @@ include soft constraints.
 |`thresholdPriority`|int (see [priority filtering](#priority-filtering))|
 |`thresholdPriorityClassName`|string (see [priority filtering](#priority-filtering))|
 |`namespaces`|(see [namespace filtering](#namespace-filtering))|
+|`nodeFit`|bool (see [node fit filtering](#node-fit-filtering))|
 
 **Example:**
 
@@ -379,6 +386,7 @@ which determines whether init container restarts should be factored into that ca
 |`thresholdPriority`|int (see [priority filtering](#priority-filtering))|
 |`thresholdPriorityClassName`|string (see [priority filtering](#priority-filtering))|
 |`namespaces`|(see [namespace filtering](#namespace-filtering))|
+|`nodeFit`|bool (see [node fit filtering](#node-fit-filtering))|
 
 **Example:**
 
@@ -550,6 +558,49 @@ strategies:
           - {key: tier, operator: In, values: [cache]}
           - {key: environment, operator: NotIn, values: [dev]}
 ```
+
+
+### Node Fit filtering
+
+The following strategies accept a `nodeFit` boolean parameter which can optimize descheduling:
+* `RemoveDuplicates`
+* `LowNodeUtilization`
+* `RemovePodsViolatingInterPodAntiAffinity`
+* `RemovePodsViolatingNodeAffinity`
+* `RemovePodsViolatingNodeTaints`
+* `RemovePodsViolatingTopologySpreadConstraint`
+* `RemovePodsHavingTooManyRestarts`
+
+ If set to `true` the descheduler will consider whether or not the pods that meet eviction criteria will fit on other nodes before evicting them. If a pod cannot be rescheduled to another node, it will not be evicted. Currently the following criteria are considered when setting `nodeFit` to `true`:
+- A `nodeSelector` on the pod
+- Any `Tolerations` on the pod and any `Taints` on the other nodes
+- `nodeAffinity` on the pod
+- Whether any of the other nodes are marked as `unschedulable`
+
+E.g.
+
+```yaml
+apiVersion: "descheduler/v1alpha1"
+kind: "DeschedulerPolicy"
+strategies:
+  "LowNodeUtilization":
+     enabled: true
+     params:
+       nodeResourceUtilizationThresholds:
+         thresholds:
+           "cpu" : 20
+           "memory": 20
+           "pods": 20
+         targetThresholds:
+           "cpu" : 50
+           "memory": 50
+           "pods": 50
+        nodeFit: true
+```
+
+Note that node fit filtering references the current pod spec, and not that of it's owner. Thus, if the pod is owned by a ReplicationController (and that ReplicationController was modified recently), the pod may be running with an outdated spec, which the descheduler will reference when determining node fit. This is expected behavior as the descheduler is a "best-effort" mechanism. 
+
+Using Deployments instead of ReplicationControllers provides an automated rollout of pod spec changes, therefore ensuring that the descheduler has an up-to-date view of the cluster state.
 
 ## Pod Evictions
 
--- a/pkg/descheduler/evictions/evictions.go
+++ b/pkg/descheduler/evictions/evictions.go
@@ -32,6 +32,7 @@ import (
 	"k8s.io/client-go/tools/record"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/descheduler/metrics"
+	nodeutil "sigs.k8s.io/descheduler/pkg/descheduler/node"
 	podutil "sigs.k8s.io/descheduler/pkg/descheduler/pod"
 	"sigs.k8s.io/descheduler/pkg/utils"
 
@@ -47,6 +48,7 @@ type nodePodEvictedCount map[*v1.Node]int
 
 type PodEvictor struct {
 	client                  clientset.Interface
+	nodes                   []*v1.Node
 	policyGroupVersion      string
 	dryRun                  bool
 	maxPodsToEvictPerNode   int
@@ -74,6 +76,7 @@ func NewPodEvictor(
 
 	return &PodEvictor{
 		client:                  client,
+		nodes:                   nodes,
 		policyGroupVersion:      policyGroupVersion,
 		dryRun:                  dryRun,
 		maxPodsToEvictPerNode:   maxPodsToEvictPerNode,
@@ -164,6 +167,7 @@ func evictPod(ctx context.Context, client clientset.Interface, pod *v1.Pod, poli
 
 type Options struct {
 	priority *int32
+	nodeFit  bool
 }
 
 // WithPriorityThreshold sets a threshold for pod's priority class.
@@ -172,6 +176,16 @@ func WithPriorityThreshold(priority int32) func(opts *Options) {
 	return func(opts *Options) {
 		var p int32 = priority
 		opts.priority = &p
+	}
+}
+
+// WithNodeFit sets whether or not to consider taints, node selectors,
+// and pod affinity when evicting. A pod who's tolerations, node selectors,
+// and affinity match a node other than the one it is currently running on
+// is evictable.
+func WithNodeFit(nodeFit bool) func(opts *Options) {
+	return func(opts *Options) {
+		opts.nodeFit = nodeFit
 	}
 }
 
@@ -221,6 +235,14 @@ func (pe *PodEvictor) Evictable(opts ...func(opts *Options)) *evictable {
 		ev.constraints = append(ev.constraints, func(pod *v1.Pod) error {
 			if utils.IsPodWithPVC(pod) {
 				return fmt.Errorf("pod has a PVC and descheduler is configured to ignore PVC pods")
+			}
+			return nil
+		})
+	}
+	if options.nodeFit {
+		ev.constraints = append(ev.constraints, func(pod *v1.Pod) error {
+			if !nodeutil.PodFitsAnyOtherNode(pod, pe.nodes) {
+				return fmt.Errorf("pod does not fit on any other node because of nodeSelector(s), Taint(s), or nodes marked as unschedulable")
 			}
 			return nil
 		})
--- a/pkg/descheduler/evictions/evictions_test.go
+++ b/pkg/descheduler/evictions/evictions_test.go
@@ -73,19 +73,27 @@ func TestIsEvictable(t *testing.T) {
 	n1 := test.BuildTestNode("node1", 1000, 2000, 13, nil)
 	lowPriority := int32(800)
 	highPriority := int32(900)
+
+	nodeTaintKey := "hardware"
+	nodeTaintValue := "gpu"
+
+	nodeLabelKey := "datacenter"
+	nodeLabelValue := "east"
 	type testCase struct {
 		pod                     *v1.Pod
-		runBefore               func(*v1.Pod)
+		nodes                   []*v1.Node
+		runBefore               func(*v1.Pod, []*v1.Node)
 		evictLocalStoragePods   bool
 		evictSystemCriticalPods bool
 		priorityThreshold       *int32
+		nodeFit                 bool
 		result                  bool
 	}
 
 	testCases := []testCase{
 		{ // Normal pod eviction with normal ownerRefs
 			pod: test.BuildTestPod("p1", 400, 0, n1.Name, nil),
-			runBefore: func(pod *v1.Pod) {
+			runBefore: func(pod *v1.Pod, nodes []*v1.Node) {
 				pod.ObjectMeta.OwnerReferences = test.GetNormalPodOwnerRefList()
 			},
 			evictLocalStoragePods:   false,
@@ -93,7 +101,7 @@ func TestIsEvictable(t *testing.T) {
 			result:                  true,
 		}, { // Normal pod eviction with normal ownerRefs and descheduler.alpha.kubernetes.io/evict annotation
 			pod: test.BuildTestPod("p2", 400, 0, n1.Name, nil),
-			runBefore: func(pod *v1.Pod) {
+			runBefore: func(pod *v1.Pod, nodes []*v1.Node) {
 				pod.Annotations = map[string]string{"descheduler.alpha.kubernetes.io/evict": "true"}
 				pod.ObjectMeta.OwnerReferences = test.GetNormalPodOwnerRefList()
 			},
@@ -102,7 +110,7 @@ func TestIsEvictable(t *testing.T) {
 			result:                  true,
 		}, { // Normal pod eviction with replicaSet ownerRefs
 			pod: test.BuildTestPod("p3", 400, 0, n1.Name, nil),
-			runBefore: func(pod *v1.Pod) {
+			runBefore: func(pod *v1.Pod, nodes []*v1.Node) {
 				pod.ObjectMeta.OwnerReferences = test.GetReplicaSetOwnerRefList()
 			},
 			evictLocalStoragePods:   false,
@@ -110,7 +118,7 @@ func TestIsEvictable(t *testing.T) {
 			result:                  true,
 		}, { // Normal pod eviction with replicaSet ownerRefs and descheduler.alpha.kubernetes.io/evict annotation
 			pod: test.BuildTestPod("p4", 400, 0, n1.Name, nil),
-			runBefore: func(pod *v1.Pod) {
+			runBefore: func(pod *v1.Pod, nodes []*v1.Node) {
 				pod.Annotations = map[string]string{"descheduler.alpha.kubernetes.io/evict": "true"}
 				pod.ObjectMeta.OwnerReferences = test.GetReplicaSetOwnerRefList()
 			},
@@ -119,7 +127,7 @@ func TestIsEvictable(t *testing.T) {
 			result:                  true,
 		}, { // Normal pod eviction with statefulSet ownerRefs
 			pod: test.BuildTestPod("p18", 400, 0, n1.Name, nil),
-			runBefore: func(pod *v1.Pod) {
+			runBefore: func(pod *v1.Pod, nodes []*v1.Node) {
 				pod.ObjectMeta.OwnerReferences = test.GetStatefulSetOwnerRefList()
 			},
 			evictLocalStoragePods:   false,
@@ -127,7 +135,7 @@ func TestIsEvictable(t *testing.T) {
 			result:                  true,
 		}, { // Normal pod eviction with statefulSet ownerRefs and descheduler.alpha.kubernetes.io/evict annotation
 			pod: test.BuildTestPod("p19", 400, 0, n1.Name, nil),
-			runBefore: func(pod *v1.Pod) {
+			runBefore: func(pod *v1.Pod, nodes []*v1.Node) {
 				pod.Annotations = map[string]string{"descheduler.alpha.kubernetes.io/evict": "true"}
 				pod.ObjectMeta.OwnerReferences = test.GetStatefulSetOwnerRefList()
 			},
@@ -136,7 +144,7 @@ func TestIsEvictable(t *testing.T) {
 			result:                  true,
 		}, { // Pod not evicted because it is bound to a PV and evictLocalStoragePods = false
 			pod: test.BuildTestPod("p5", 400, 0, n1.Name, nil),
-			runBefore: func(pod *v1.Pod) {
+			runBefore: func(pod *v1.Pod, nodes []*v1.Node) {
 				pod.ObjectMeta.OwnerReferences = test.GetNormalPodOwnerRefList()
 				pod.Spec.Volumes = []v1.Volume{
 					{
@@ -154,7 +162,7 @@ func TestIsEvictable(t *testing.T) {
 			result:                  false,
 		}, { // Pod is evicted because it is bound to a PV and evictLocalStoragePods = true
 			pod: test.BuildTestPod("p6", 400, 0, n1.Name, nil),
-			runBefore: func(pod *v1.Pod) {
+			runBefore: func(pod *v1.Pod, nodes []*v1.Node) {
 				pod.ObjectMeta.OwnerReferences = test.GetNormalPodOwnerRefList()
 				pod.Spec.Volumes = []v1.Volume{
 					{
@@ -172,7 +180,7 @@ func TestIsEvictable(t *testing.T) {
 			result:                  true,
 		}, { // Pod is evicted because it is bound to a PV and evictLocalStoragePods = false, but it has scheduler.alpha.kubernetes.io/evict annotation
 			pod: test.BuildTestPod("p7", 400, 0, n1.Name, nil),
-			runBefore: func(pod *v1.Pod) {
+			runBefore: func(pod *v1.Pod, nodes []*v1.Node) {
 				pod.Annotations = map[string]string{"descheduler.alpha.kubernetes.io/evict": "true"}
 				pod.ObjectMeta.OwnerReferences = test.GetNormalPodOwnerRefList()
 				pod.Spec.Volumes = []v1.Volume{
@@ -191,7 +199,7 @@ func TestIsEvictable(t *testing.T) {
 			result:                  true,
 		}, { // Pod not evicted becasuse it is part of a daemonSet
 			pod: test.BuildTestPod("p8", 400, 0, n1.Name, nil),
-			runBefore: func(pod *v1.Pod) {
+			runBefore: func(pod *v1.Pod, nodes []*v1.Node) {
 				pod.ObjectMeta.OwnerReferences = test.GetDaemonSetOwnerRefList()
 			},
 			evictLocalStoragePods:   false,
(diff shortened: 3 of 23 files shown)


RyanDevlin:
/assign @damemi @ingvagabund 

Could you please review my PR? 

Specifically @damemi I made some changes to the TopologySpreadConstraint strategy to avoid checking for node affinity and taints twice (see line 290 of topologyspreadconstraint.go).

pkg/utils/predicates.go
```diff
@@ -107,3 +108,14 @@ func TolerationsTolerateTaintsWithFilter(tolerations []v1.Toleration, taints []v
 
 	return true
 }
+
+// GetNodeFitFromStrategyParams gets nodeFit configuration variable from the given StrategyParameters.
+// It will return false by default.
+func GetNodeFitFromStrategyParams(params *api.StrategyParameters) bool {
```
lixiang233, line 114: I'm not sure if this is the right place to put this helper function, maybe we can add a new `utils.go` file to `descheduler/pkg/descheduler/strategies/` to place it(and we can add more related helper functions to this file)?

lixiang233 (commented):
/ok-to-test

lixiang233:
> @RyanDevlin: The following test **failed**, say `/retest` to rerun all failed tests:
> 
> Test name	Commit	Details	Rerun command
> pull-descheduler-verify-master	[b7daf8e](https://github.com/kubernetes-sigs/descheduler/commit/b7daf8e87af4fea8e265a07d4d1d82ec8a4ce56f)	[link](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_descheduler/559/pull-descheduler-verify-master/1387965290997878784)	`/test pull-descheduler-verify-master`
> [Full PR test history](https://prow.k8s.io/pr-history?org=kubernetes-sigs&repo=descheduler&pr=559). [Your PR dashboard](https://prow.k8s.io/pr?query=is%3Apr%20state%3Aopen%20author%3ARyanDevlin). Please help us cut down on flakes by [linking to](https://git.k8s.io/community/contributors/devel/sig-testing/flaky-tests.md#filing-issues-for-flaky-tests) an [open issue](https://github.com/kubernetes-sigs/descheduler/issues?q=is:issue+is:open) when you hit one in your PR.

@RyanDevlin  It seems that you forgot to run `make verify` before pushing commits.

pkg/descheduler/evictions/evictions.go
```diff
@@ -164,6 +167,7 @@ func evictPod(ctx context.Context, client clientset.Interface, pod *v1.Pod, poli
 
 type Options struct {
 	priority *int32
+	nodeFit  *bool
```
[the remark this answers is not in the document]
RyanDevlin: Good point, that's more concise.

README.md
```diff
@@ -157,6 +157,7 @@ should include `ReplicaSet` to have pods created by Deployments excluded.
 |`namespaces`|(see [namespace filtering](#namespace-filtering))|
 |`thresholdPriority`|int (see [priority filtering](#priority-filtering))|
 |`thresholdPriorityClassName`|string (see [priority filtering](#priority-filtering))|
+|`nodeFit`|string (see [node fit filtering](#node-fit-filtering))|
```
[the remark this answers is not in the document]
RyanDevlin: Thanks, good catch!

pkg/utils/predicates.go
RyanDevlin: Good point, I wasn't really sure where to put that one. I've created `utils.go` and placed this function there.

damemi (commented):
```
./hack/verify-spelling.sh
./hack/verify-toc.sh
./hack/verify-conversions.sh
Generated output differs:
diff -Naupr pkg/api/v1alpha1/zz_generated.conversion.go ./hack/../_tmp/kube-verify.vLZmmb/descheduler/pkg/api/v1alpha1/zz_generated.conversion.go
--- pkg/api/v1alpha1/zz_generated.conversion.go	2021-04-30 11:32:04.717979648 +0000
+++ ./hack/../_tmp/kube-verify.vLZmmb/descheduler/pkg/api/v1alpha1/zz_generated.conversion.go	2021-04-30 11:34:02.331140501 +0000
@@ -294,6 +294,7 @@ func autoConvert_v1alpha1_StrategyParame
 	out.ThresholdPriority = (*int32)(unsafe.Pointer(in.ThresholdPriority))
 	out.ThresholdPriorityClassName = in.ThresholdPriorityClassName
 	out.LabelSelector = (*v1.LabelSelector)(unsafe.Pointer(in.LabelSelector))
+	out.NodeFit = in.NodeFit
 	return nil
 }
 
@@ -313,6 +314,7 @@ func autoConvert_api_StrategyParameters_
 	out.ThresholdPriority = (*int32)(unsafe.Pointer(in.ThresholdPriority))
 	out.ThresholdPriorityClassName = in.ThresholdPriorityClassName
 	out.LabelSelector = (*v1.LabelSelector)(unsafe.Pointer(in.LabelSelector))
+	out.NodeFit = in.NodeFit
 	return nil
 }
 
Generated conversions verify failed. Please run ./hack/update-conversions.sh
make: *** [Makefile:124: verify-gen] Error 1
```

`./hack/updat-conversions.sh` needs to be run

ingvagabund:
@RyanDevlin https://github.com/kubernetes-sigs/descheduler/issues/551 is more about deciding what's the average number of duplicates per node rather than checking if a pod can be evicted. The issue is about avoiding unnecessary evictions in the first place.

RyanDevlin:
@damemi I can't figure this one out. I only see `./hack/update-generated-conversions.sh` not `./hack/update-conversions.sh`. I ran `./hack/update-conversions.sh` just in case, and then `make verify` and it's still failing. I also tried running `make clean`, `make gen`, and then `make verify` and it's still reporting errors. Any thoughts?

@ingvagabund In the description of #551 it says "Take at least node taints (e.g. taints repealing pods from master nodes) into account when selecting pods for eviction. If possible, take node selector into account as well". I'm a bit confused how that differs from what my feature does? 

Since `RemoveDuplicas` filters pods by with the `Evicatable` function prior to calculating duplicates, would this feature not achieve the behavior desired in #551?

damemi:
> @damemi I can't figure this one out. I only see ./hack/update-generated-conversions.sh not ./hack/update-conversions.sh. I ran ./hack/update-conversions.sh just in case, and then make verify and it's still failing. I also tried running make clean, make gen, and then make verify and it's still reporting errors. Any thoughts?

Sorry, the script should be `./hack/update-generated-conversions.sh` (note that you need to `git commit` the changes before `make verify` will pass after running the update script)

RyanDevlin:
@damemi On my end, I think the problem is that the output of `./hack/update-generated-conversions.sh` produces files with errors. The `sigs.k8s.io/descheduler/pkg/` directory created by the script has `zz_generated.conversion.go` files with references to `*DeschedulerPolicy` that it says is undeclared.

Running `make clean` and then `make gen`, or `make clean` and `./hack/update-generated-conversions.sh` still produces files with errors in them. This later causes my `make verify` to fail. I'm not sure how to fix that.

go.sum
```diff
@@ -106,6 +106,7 @@ github.com/evanphx/json-patch v4.9.0+incompatible/go.mod h1:50XU6AFN0ol/bzJsmQLi
 github.com/fatih/color v1.7.0/go.mod h1:Zm6kSWBoL9eyXnKyktHP6abPY2pDugNf5KwzbycvMj4=
 github.com/form3tech-oss/jwt-go v3.2.2+incompatible h1:TcekIExNqud5crz4xD2pavyTgWiPvpYe4Xau31I0PRk=
 github.com/form3tech-oss/jwt-go v3.2.2+incompatible/go.mod h1:pbq4aXjuKjdthFRnoDwaVPLA+WlJuPGy+QneDUgJi2k=
+github.com/fsnotify/fsnotify v1.4.7 h1:IXs+QLmnXW2CcXuY+8Mzv/fWEsPGWxqefPtCP5CnV9I=
```
damemi, line 109: odd that there is a `go.sum` change in this PR

you are also committing the `kind` binary below this, it won't let me comment there but please remove that too

RyanDevlin: Oh that was dumb, I removed to go.sum line and kind binary.

ingvagabund:
> @ingvagabund In the description of #551 it says "Take at least node taints (e.g. taints repealing pods from master nodes) into account when selecting pods for eviction. If possible, take node selector into account as well". I'm a bit confused how that differs from what my feature does?
>
> Since RemoveDuplicas filters pods by with the Evicatable function prior to calculating duplicates, would this feature not achieve the behavior desired in #551?

#551 is focused mainly for adjusting https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/descheduler/strategies/duplicates.go#L107. As described in the description:
```
In case there's a cluster with 3 master nodes and 3 workers and a workload (owned by an RC) is uniformly
distributed among worker nodes (the expected distribution), the strategy still evicts pods of the workload
since it takes master nodes into account as well. E.g. 12 pods, 4 pods bound to each worker node, 6 pods
are evicted (6 nodes, 12 pods, 2 per node) which always end up on the same worker nodes again.
```

Currently, all 6 nodes are taken into account. So the average number of pods per node is 2. The idea is to adjust `nodeCount` to 3 (filter out masters which can not run evicted pods). With your PR the `nodeCount` is still 3 and there are two other worker nodes which can run the evicted pods. So instead of evicting 0 pods (since the pods are evenly distributed among the worker nodes), the strategy will still evict 6.

RyanDevlin:
@ingvagabund I see what you mean, thanks for that link! I'll remove #551 from the description of the PR.

RyanDevlin:
@damemi I'm still unable to resolve my generated conversions issue, but I have some more information if it helps. You can see the latest build logs here: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_descheduler/559/pull-descheduler-unit-test-master-master/1389200208998436864.

The log file shows all the places in my `zz_generated.conversion` files where there are errors, eg:
```
sigs.k8s.io/descheduler/pkg/api/v1alpha1/zz_generated.conversion.go:122:74: undefined: DeschedulerPolicy
```

You can see these errors yourself if you fetch my code and run `./hack/update-generated-conversions.sh`, then open the `sigs.k8s.io/descheduler/pkg/api/v1alpha1/zz_generated.conversion.go` file.

So far I haven't been able to figure out the cause of these errors. If there is something I'm doing wrong, or something I should try, please let me know.

damemi:
@RyanDevlin in that Prow output I see: `sigs.k8s.io/descheduler/sigs.k8s.io/descheduler/pkg/api/v1alpha1`, and from your other comment above (https://github.com/kubernetes-sigs/descheduler/pull/559#issuecomment-830402018) it sounds like the generator script is creating an entirely new directory called `sigs.k8s.io/descheduler` *within* the project?

If so, something is totally broken. The reason you're getting undefined errors is because those generated files don't have any type definitions to go along with them. In other words, that script should just be updating the *existing* `zz_**.go` files in the directories like https://github.com/kubernetes-sigs/descheduler/tree/master/pkg/api 

So, first off we need to delete the `sigs.k8s.io/descheduler` directory that's being created, and then figure out why the script is creating a new one in the project root on your environment.

RyanDevlin:
@damemi That makes so much more sense, I didn't initially understand that the conversion gen script should be editing the existing `zz_**.go` files. Also I've figured out why this is happening. Running `./conversion-gen --help` gives me:
```
  -o, --output-base string               Output base; defaults to $GOPATH/src/ or ./ if $GOPATH is not set. (default "./")
```

Since my $GOPATH is unset, this is defaulting to the root of the repo. The problem I face now is that even if I set my `$GOPATH`, my local copy of the repo is structured as a fork. Thus, my directory structure is:
```
~/go/src/github.com/RyanDevlin/descheduler
```
and not:
```
/go/src/github.com/sigs.k8s.io/descheduler
```
So the script will keep trying to look for the `sigs.k8s.io/descheduler/pkg/api` path, which I don't have. Is it a requirement to clone the repo into a `sigs.k8s.io folder`? Or should these scripts be agnostic of the directory they exist in?

damemi:
> Is it a requirement to clone the repo into a sigs.k8s.io folder?

When GOPATH was more of a requirement, yes, you should clone your forks to the location of the actual package name (ie, the upstream path) and add your fork as a Git remote.

Technically with go.mod now widely used, we should probably be able to function independently of that. In the meantime, I would try to change your project directory to match and see if that's the easy fix

RyanDevlin:
@damemi Okay I've fixed the issue by changing directories and using `GOPATH`. Sorry about all that confusion, I shouldn't run into this issue again.

Like you said, I'm thinking eventually we make it so the conversion-gen script doesn't rely on assumed paths. For now, this works though.

pkg/utils/utils.go
```diff
@@ -0,0 +1,16 @@
+package utils
+
+import (
+	"sigs.k8s.io/descheduler/pkg/api"
+)
+
+// GetNodeFitFromStrategyParams gets nodeFit configuration variable from the given StrategyParameters.
+// It will return false by default.
+func GetNodeFitFromStrategyParams(params *api.StrategyParameters) bool {
```
ingvagabund, line 9: Given the implementation of `GetNodeFitFromStrategyParams` is quite straightforward and the strategies already have `if strategy.Params != nil` clause, can you just inline the function? E.g.

```
	var nodeFit bool
	if strategy.Params != nil {
		...
		nodeFit = params.NodeFit
	}
```

ingvagabund (changes requested):

ingvagabund:
One thing to keep in mind. There's no guarantee an owner will keep the pod manifest as it is between descheduler runs. E.g. one might create 10 replicas of a pod and then edit a RS to change the taints/node selector. So when the Descheduler tries to decide if there's a different node that can run the evicted pod, a new pod can have completely different constraints to respect.

Worth mentioning that when enabling `nodeFit` param. The Descheduler still works in the "best-effort" approach.

damemi:
> There's no guarantee an owner will keep the pod manifest as it is between descheduler runs. E.g. one might create 10 replicas of a pod and then edit a RS to change the taints/node selector. So when the Descheduler tries to decide if there's a different node that can run the evicted pod, a new pod can have completely different constraints to respect.

@ingvagabund if the RS is modified, won't this trigger an immediate redeployment of all the pods with the new spec, and they'll be re-listed in the Descheduler cache? So we will have the new constraints anyway, or it will try acting on non-existent pods

ingvagabund:
It will not to my knowledge. Only when the number of replicas is decreased and then increased.

RyanDevlin:
@ingvagabund @damemi In that case, would some clarification in the README be sufficient to cover this topic in the context of the current PR?

pkg/utils/utils.go
RyanDevlin: This is a fair point. Initially I had moved this into it's own function because I wanted to minimized the overhead required for all strategies to utilize the feature. Since the `GetNodeFitFromStrategyParams()` function now resides in it's own file, this seems a bit convoluted. I'll make it inline.

README.md
```diff
@@ -551,6 +560,33 @@ strategies:
...
+strategies:
+  "PodLifeTime":
+     enabled: true
+     params:
+        podLifeTime:
+          maxPodLifeTimeSeconds: 86400
+        nodeFit: true
+```
+
+Note that node fit filtering references the current pod spec, and not that of it's owner. Thus, if the pod is owned by a ReplicationController (and that ReplicationController was modified recently), the pod may be running with an outdated spec, which the descheduler will reference when determining node fit. This is expected behavior as the descheduler is a "best-effort" mechanism.
```
ingvagabund, line 586: The pod can be also owned by a ReplicaSet or StatefulSet. What about `Thus, if the pod is managed by a controller which does not recreate all replicas everytime a pod template is updated, ...` ?

RyanDevlin: Added, thanks.

pkg/descheduler/evictions/evictions.go
```diff
@@ -225,6 +239,15 @@ func (pe *PodEvictor) Evictable(opts ...func(opts *Options)) *evictable {
 			return nil
 		})
 	}
+	//if options.nodeFit != nil && *options.nodeFit {
+	if options.nodeFit {
+		ev.constraints = append(ev.constraints, func(pod *v1.Pod) error {
+			if !nodeutil.PodFitsAnyOtherNode(pod, pe.nodes) {
+				return fmt.Errorf("pod does not fit on any other node because of nodeSelector(s), Taint(s), or nodes marked as unschedulable")
```
ingvagabund, line 246: nodeSelector(s), Taint(s), -> node selector, node affinity, taint(s) mismatch,

RyanDevlin: Got it, thank you.

pkg/descheduler/strategies/lownodeutilization.go
```diff
@@ -399,10 +403,10 @@ func evictPods(
 
 	if continueCond() {
 		for _, pod := range inputPods {
-			if !utils.PodToleratesTaints(pod, taintsOfLowNodes) {
+			/*if !utils.PodToleratesTaints(pod, taintsOfLowNodes) {
```
ingvagabund, line 406: `NodeFit` feature is not enabled by default. So this needs to stay here to keep the original functionality.

pkg/descheduler/strategies/lownodeutilization_test.go
```diff
@@ -514,6 +517,227 @@ func TestLowNodeUtilization(t *testing.T) {
...
+				v1.ResourceCPU:  30,
+				v1.ResourcePods: 30,
+			},
+			targetThresholds: api.ResourceThresholds{
+				v1.ResourceCPU:  50,
+				v1.ResourcePods: 50,
+			},
+			nodes: map[string]*v1.Node{
+				n1NodeName: test.BuildTestNode(n1NodeName, 4000, 3000, 9, nil),
+				n3NodeName: test.BuildTestNode(n3NodeName, 4000, 3000, 10, test.SetNodeUnschedulable),
```
ingvagabund, line 532: rather n2NodeName otherwise my first Q was where's the second node.

pkg/descheduler/strategies/lownodeutilization_test.go
```diff
@@ -514,6 +517,227 @@ func TestLowNodeUtilization(t *testing.T) {
...
+			},
+			targetThresholds: api.ResourceThresholds{
+				v1.ResourceCPU:  50,
+				v1.ResourcePods: 50,
+			},
+			nodes: map[string]*v1.Node{
+				n1NodeName: test.BuildTestNode(n1NodeName, 4000, 3000, 9, nil),
+				n3NodeName: test.BuildTestNode(n3NodeName, 4000, 3000, 10, func(node *v1.Node) {
+					node.ObjectMeta.Labels = map[string]string{
+						nodeSelectorKey: "east",
```
ingvagabund, line 587: I'd prefer `otherThanNodeSelectorValue` or similar  so it's clear `nodeSelectorValue` can't be equal to "east"

pkg/descheduler/strategies/lownodeutilization_test.go
```diff
@@ -514,6 +517,227 @@ func TestLowNodeUtilization(t *testing.T) {
...
+				v1.ResourceCPU:  30,
+				v1.ResourcePods: 30,
+			},
+			targetThresholds: api.ResourceThresholds{
+				v1.ResourceCPU:  50,
+				v1.ResourcePods: 50,
+			},
+			nodes: map[string]*v1.Node{
+				n1NodeName: test.BuildTestNode(n1NodeName, 4000, 3000, 9, nil),
+				n3NodeName: test.BuildTestNode(n3NodeName, 4000, 3000, 10, func(node *v1.Node) {
```
ingvagabund, line 585: rather n2NodeName otherwise my first Q was where's the second node.

pkg/descheduler/strategies/pod_antiaffinity_test.go
```diff
@@ -111,21 +159,28 @@ func TestPodAntiAffinity(t *testing.T) {
...
 
+		nodes := test.nodes
+
 		podEvictor := evictions.NewPodEvictor(
 			fakeClient,
 			policyv1.SchemeGroupVersion.String(),
 			false,
 			test.maxPodsToEvictPerNode,
-			[]*v1.Node{node},
+			nodes,
```
ingvagabund, line 172: Why not `test.nodes`? I don't see `nodes` used anywhere besides here and in `RemovePodsViolatingInterPodAntiAffinity`

pkg/descheduler/strategies/pod_lifetime_test.go
```diff
@@ -235,30 +257,59 @@ func TestPodLifeTime(t *testing.T) {
...
-			return true, node, nil
-		})
+
 		podEvictor := evictions.NewPodEvictor(
 			fakeClient,
 			policyv1.SchemeGroupVersion.String(),
 			false,
 			tc.maxPodsToEvictPerNode,
-			[]*v1.Node{node},
+			nodes,
```
ingvagabund, line 306: Why not tc.nodes directly?

pkg/descheduler/evictions/evictions.go
```diff
@@ -225,6 +239,14 @@ func (pe *PodEvictor) Evictable(opts ...func(opts *Options)) *evictable {
 			return nil
 		})
 	}
+	if options.nodeFit {
+		ev.constraints = append(ev.constraints, func(pod *v1.Pod) error {
+			if !nodeutil.PodFitsAnyOtherNode(pod, pe.nodes) {
```
ingvagabund, line 244: This will not fly for strategies like `PodLifeTime` for which it's valid to schedule the pod to the same node.

pkg/descheduler/strategies/pod_lifetime.go
```diff
@@ -74,7 +74,12 @@ func PodLifeTime(ctx context.Context, client clientset.Interface, strategy api.D
 		excludedNamespaces = strategy.Params.Namespaces.Exclude
 	}
 
-	evictable := podEvictor.Evictable(evictions.WithPriorityThreshold(thresholdPriority))
+	nodeFit := false
+	if strategy.Params != nil {
+		nodeFit = strategy.Params.NodeFit
+	}
+
+	evictable := podEvictor.Evictable(evictions.WithPriorityThreshold(thresholdPriority), evictions.WithNodeFit(nodeFit))
```
ingvagabund, line 82: It's valid for the strategies to schedule the pod to the same node. `evictions.WithNodeFit` does not belong here.

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -284,20 +291,6 @@ func balanceDomains(
 		// also (just for tracking), add them to the list of pods in the lower topology
 		aboveToEvict := sortedDomains[j].pods[len(sortedDomains[j].pods)-movePods:]
 		for k := range aboveToEvict {
-			// if the pod has a hard nodeAffinity or nodeSelector that only matches this node,
```
ingvagabund, line 287: > // don't bother evicting it as it will just end up back on the same node
> // however we still account for it "being evicted" so the algorithm can complete

The code still needs to stay here based on the comment. I will keep this for @damemi

ingvagabund (changes requested):

ingvagabund:
@RyanDevlin sorry for the late review. I finally found more time to check your changes more closely.

pkg/descheduler/evictions/evictions_test.go
```diff
@@ -311,20 +319,114 @@ func TestIsEvictable(t *testing.T) {
...
 		}
 
 		evictable := podEvictor.Evictable()
-		if test.priorityThreshold != nil {
+		if test.priorityThreshold != nil && test.nodeFit {
+			evictable = podEvictor.Evictable(WithPriorityThreshold(*test.priorityThreshold), WithNodeFit(test.nodeFit))
+		} else if test.priorityThreshold != nil && !test.nodeFit {
 			evictable = podEvictor.Evictable(WithPriorityThreshold(*test.priorityThreshold))
+		} else if test.nodeFit {
+			evictable = podEvictor.Evictable(WithNodeFit(test.nodeFit))
```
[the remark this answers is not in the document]
RyanDevlin: I like this a lot, much simpler.

pkg/descheduler/node/node_test.go
```diff
@@ -239,25 +245,33 @@ func TestPodFitsAnyNode(t *testing.T) {
...
+				{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: nodeNames[2],
+					},
+				},
 			},
 			success: true,
 		},
 		{
 			description: "Pod expected to fit one of the nodes (error node first)",
```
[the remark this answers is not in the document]
RyanDevlin: That specific test wasn't written by me, but I'm happy to still edit it. If I had to guess, I believe the intention of that test case was to test the short-circuit logic of the `PodFitsAnyOtherNode()` function. Would you still like me to remove it?

pkg/descheduler/strategies/lownodeutilization.go
RyanDevlin: Originally I had assumed this feature was to fix the fact that some strategies take taints, etc. into account while others do not. Following this assumption, I removed filtering for these conditions in the strategies (including `lownodeutilization` as well as `node_affinity`). My intention was to prevent this work from being repeated twice when `nodeFit` was enabled. If the user desired closer to optimal filtering, they would need to set `nodeFit` to `true`.

Is this assumption incorrect?

Also, I shouldn't have left the commented code in there. If we decide to remove it, I'll fully delete it.

pkg/descheduler/strategies/lownodeutilization_test.go
RyanDevlin: Fair point, fixed.

pkg/descheduler/strategies/pod_antiaffinity_test.go
RyanDevlin: I had that a few lines above, but it looks like I forgot to make it inline. It's there now.

pkg/descheduler/strategies/pod_lifetime.go
RyanDevlin: Good point, I've removed the feature from the `pod_lifetime` strategy. I'll also update the `README` accordingly.

pkg/descheduler/evictions/evictions.go
RyanDevlin: I've now removed this switch from the `pod_lifetime` strategy.

RyanDevlin:
@ingvagabund Thank you for taking the time to review my PR! I've resolved all of your comments except a few. These are left open with questions above. Please let me know if there is anything I missed.

RyanDevlin:
@ingvagabund have you had time to review my comments above? Specifically the second comment about strategies that filtered for node taints and affinity?

Just to reiterate, my line of thinking was that the goal of this feature was to get all strategies in line with a method for considering taints, affinity, etc. before descheduling. Some strategies already performed a subset of these checks before considering pods for descheduling.

In order to make this process more uniform, I removed some of the filtering for taints, etc. from `nodeAffinity`, `topologySpread`, and maybe one or two other strategies. I obviously never removed any filtering that was critical to the strategy (eg. the affinity filtering performed in `nodeAffinity`).

My assumption was that, if a user desired more advanced filtering, they would turn on the `nodeFit` feature in order to gain a closer to optimal eviction process.

pkg/descheduler/node/node_test.go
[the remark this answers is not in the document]
ingvagabund: Yeah, I don't see a point for such a test until we start measuring a performance.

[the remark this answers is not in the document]
RyanDevlin: Okay I'll remove it.

pkg/descheduler/strategies/lownodeutilization.go
ingvagabund: There's gonna be some duplication. Though, we need to keep the same behavior in case the feature is disabled. Unless @seanmalloy @damemi @lixiang233 thinks otherwise.

ingvagabund:
I like the point you are making. I am in favor of simplifying the code. On the other hand, we can't change the default behavior just because of that. I'd rather wait for a release or two and make it enabled by default. Then eventually remove the gate completely.

RyanDevlin:
@ingvagabund That's a fair point. In that case, I can revert any filtering changes I made to the strategies for now. Also I'll wait to hear from @damemi @lixiang233 @seanmalloy for extra thoughts on this before pushing anything.

For this type of follow up task, should I file a new issue so we can keep track of the future work?

ingvagabund:
> For this type of follow up task, should I file a new issue so we can keep track of the future work?

I'd like to hear from other as well what they think about it. Filling a tracking issue sounds good if agreed.

damemi:
> I like the point you are making. I am in favor of simplifying the code. On the other hand, we can't change the default behavior just because of that. I'd rather wait for a release or two and make it enabled by default. Then eventually remove the gate completely.

Yeah, let's not change the default behavior right now. Tracking it over the next 2 releases to be enabled by default makes sense, and falls in line with upstream's featuregating approach

pkg/descheduler/strategies/topologyspreadconstraint.go
damemi: Sorry, trying to remember exactly what I meant by this comment. I think removing the check here is okay. Pretty sure that `we still account for it "being evicted"` is actually referring to this line: https://github.com/kubernetes-sigs/descheduler/blob/24c0ca2/pkg/descheduler/strategies/topologyspreadconstraint.go#L294

Where all the pods in `aboveToEvict` are removed from the in-memory representation of `sortedDomains[j]`, whether or not they were actually added to `podsForEviction`.

tl;dr, we won't need this affinity/taints check if `evictable.IsEvictable` does it before actually evicting them. But it will result in more calls to `IsEvictable`, because some of the pods in `podsForEviction` aren't being skipped without it.

damemi: So, I think it's just that the comment was in a bad spot

pkg/descheduler/evictions/evictions.go
```diff
@@ -225,6 +239,15 @@ func (pe *PodEvictor) Evictable(opts ...func(opts *Options)) *evictable {
 			return nil
 		})
 	}
+	//if options.nodeFit != nil && *options.nodeFit {
```
ingvagabund, line 242: You can drop the comment

pkg/descheduler/strategies/node_affinity.go
```diff
@@ -80,8 +85,7 @@ func RemovePodsViolatingNodeAffinity(ctx context.Context, client clientset.Inter
 					node,
 					podutil.WithFilter(func(pod *v1.Pod) bool {
 						return evictable.IsEvictable(pod) &&
-							!nodeutil.PodFitsCurrentNode(pod, node) &&
-							nodeutil.PodFitsAnyNode(pod, nodes)
```
ingvagabund, line 84: `nodeutil.PodFitsAnyNode` needs to stay here as well

pkg/descheduler/strategies/pod_lifetime_test.go
```diff
@@ -34,15 +34,29 @@ import (
 
 func TestPodLifeTime(t *testing.T) {
 	ctx := context.Background()
-	node := test.BuildTestNode("n1", 2000, 3000, 10, nil)
+	node1 := test.BuildTestNode("n1", 2000, 3000, 10, nil)
+	/*node2 := test.BuildTestNode("n2", 2000, 3000, 10, func(node *v1.Node) {
```
ingvagabund, line 38: You can drop the comment

pkg/descheduler/strategies/pod_lifetime_test.go
```diff
@@ -235,30 +257,58 @@ func TestPodLifeTime(t *testing.T) {
 			},
 			maxPodsToEvictPerNode:   5,
 			pods:                    []v1.Pod{*p12, *p13},
+			nodes:                   []*v1.Node{node1},
 			expectedEvictedPodCount: 1,
 		},
+		/*{
```
ingvagabund, line 263: You can drop the comment

pkg/descheduler/strategies/lownodeutilization.go
```diff
@@ -330,9 +335,9 @@ func evictPodsFromTargetNodes(
 		v1.ResourceMemory: {},
 	}
 
-	var taintsOfLowNodes = make(map[string][]v1.Taint, len(lowNodes))
+	//var taintsOfLowNodes = make(map[string][]v1.Taint, len(lowNodes))
```
ingvagabund, line 338: Undo this change please

ingvagabund (changes requested):
The rest looks good (I checked the unit tests pretty fast this time, they will need some refactoring soon though, it's getting harder to review changes in those)

pkg/descheduler/strategies/node_affinity.go
RyanDevlin: This is the only place `PodFitsAnyNode()` was called prior to my feature. Originally I had rewrote this function as `PodFitsAnyOtherNode()` and added the ability to check for taints.

I've now added back in `PodFitsAnyNode()` alongside `PodFitsAnyOtherNode()` to restore the exact behavior of `node_affinity.go`. These functions are very similar so maybe eventually we should drop `PodFitsAnyNode()`.

RyanDevlin:
@ingvagabund @damemi Thanks for the input! I've restored the original functionality of the strategies. I also agree that the tests are becoming exceedingly complex, especially the e2e tests.

Lastly, I've opened issue #574 to track the removal of the optimization performed by this feature from the individual strategies. 

Hopefully that covers everything for this PR, but if not, please comment and I will make changes!

damemi (commented):
/approve
I'll leave any final points to @ingvagabund or @seanmalloy

ingvagabund:
/lgtm
Had a PTO the last week. Sorry for the delay :)

@RyanDevlin thank you for the patience.

RyanDevlin:
@damemi @ingvagabund Thank you both for the review!

PR #560, opened 2021-04-30
Update error messages in verify scripts to be more informative

damemi:
The error messages were incorrect, and leave out that the changes must be committed


damemi (commented):
noticed from https://github.com/kubernetes-sigs/descheduler/pull/559#issuecomment-830268646

PR #561, opened 2021-05-03
Add support for private registry creds

mekza:
This pull request adds support for private registry creds

### DIFF
--- a/charts/descheduler/README.md
+++ b/charts/descheduler/README.md
@@ -48,6 +48,7 @@ The following table lists the configurable parameters of the _descheduler_ chart
 | `image.repository`             | Docker repository to use                                                                                              | `k8s.gcr.io/descheduler/descheduler` |
 | `image.tag`                    | Docker tag to use                                                                                                     | `v[chart appVersion]`                |
 | `image.pullPolicy`             | Docker image pull policy                                                                                              | `IfNotPresent`                       |
+| `imagePullSecrets`             | Docker repository secrets                                                                                              | `[]`                       |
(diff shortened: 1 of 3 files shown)


charts/descheduler/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: descheduler
-version: 0.20.0
+version: 0.20.1
```
damemi, line 3: wondering if we should bump+release a new chart version now, or just wait for the v0.21 descheduler release (which is coming Soon) and release the updated chart then?

damemi (commented):
/ok-to-test

damemi:
Sorry, I don't think I was clear on what I meant above.

What we should have done was just merged the parameter change in (without bumping the chart version) *before* we released v0.21 (when we did bump the chart version).

Now that we have already released the 0.21 chart, merging this into the current release would require a bump to v0.21.1, as you have done here.

However, with k8s v0.22 code freeze less than a month away, I am inclined to just merge this into the current `master` (v0.22 development) branch, without any changes to the chart version, and let it get included into the v0.22 release with everything else for that version.

The alternative (releasing a v0.21.1 chart) would actually involve 2 PRs: one to merge to master, and another to cherry-pick it to the `release-1.21` branch. For a non-critical change I don't think it's worth the effort, though I am open to other opinions if this is an important feature. @mekza (and @seanmalloy @ingvagabund) wdyt?

mekza:
Merging it into master without any change on the chart version makes total sense to me!

damemi:
Thanks for understanding, @mekza, sorry we missed v0.21 with this!
/approve

mekza:
@damemi no problem! Thanks for your time

PR #563, opened 2021-05-05
RemoveDuplicates: take node taints, node affinity and node selector into account when computing a number of feasible nodes for the average occurence of pods per node

ingvagabund:
Nodes with taints which are not tolerated by evicted pods will never run the pods. The same holds for node affinity and node selector. So increase the number of pods per feasible nodes to decrease the number of evicted pods.

Fixes: https://github.com/kubernetes-sigs/descheduler/issues/551

**TODO**:
- [x] take taints into account
- [x] take node selector into account

### DIFF
--- a/pkg/descheduler/strategies/duplicates.go
+++ b/pkg/descheduler/strategies/duplicates.go
@@ -183,9 +183,17 @@ func RemoveDuplicatePods(
 	}
 
 	// 1. how many pods can be evicted to respect uniform placement of pods among viable nodes?
-	for ownerKey, nodes := range duplicatePods {
-		upperAvg := int(math.Ceil(float64(ownerKeyOccurence[ownerKey]) / float64(nodeCount)))
-		for nodeName, pods := range nodes {
+	for ownerKey, podNodes := range duplicatePods {
+		targetNodes := getTargetNodes(podNodes, nodes)
+
+		klog.V(2).InfoS("Adjusting feasible nodes", "owner", ownerKey, "from", nodeCount, "to", len(targetNodes))
+		if len(targetNodes) < 2 {
+			klog.V(1).InfoS("Less than two feasible nodes for duplicates to land, skipping eviction", "owner", ownerKey)
+			continue
+		}
+
+		upperAvg := int(math.Ceil(float64(ownerKeyOccurence[ownerKey]) / float64(len(targetNodes))))
+		for nodeName, pods := range podNodes {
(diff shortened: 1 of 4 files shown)


pkg/descheduler/strategies/duplicates.go
```diff
@@ -202,6 +210,67 @@ func RemoveDuplicatePods(
...
+	distinctPods := map[*v1.Pod]struct{}{}
+	for _, pods := range podNodes {
+		for _, pod := range pods {
+			duplicated := false
+			for dp := range distinctPods {
+				if utils.TolerationsEqual(pod.Spec.Tolerations, dp.Spec.Tolerations) &&
+					utils.NodeSelectorsEqual(getNodeAffinityNodeSelector(pod), getNodeAffinityNodeSelector(dp)) &&
+					reflect.DeepEqual(pod.Spec.NodeSelector, dp.Spec.NodeSelector) {
+					duplicated = true
+					continue
```
damemi, line 235: will this count pods as duplicate if they don't have tolerations/selectors/affinity?

damemi: also could you explain a bit more about what this check is for? I am a little confused about the check for duplicate node selector rules here (not that it's wrong, it's likely just me 🙂 )

ingvagabund: The idea is to assign each pod a (tolerations, nodeselectors, node affinity) triple (including nil instances). Also, there's some logic that reduces duplicated tolerations/nodeselectors/nodeaffinities terms and sorts the terms so it's easy to identify the duplicated terms. So when two triples are equal, evaluation of both corresponding pods wrt. feasible nodes will end up the same. Thus, reducing the number of (pod, node) evaluations with the same result. Thus, minimizing the set of pods where we check how many nodes are feasible for them once evicted. `getTargetNodes` is computing how many nodes are feasible for all the pods there were marked for eviction. So the average number of pods per node can be adjusted based what `getTargetNodes` returns.

The checks in `getTargetNodes` will be most probably all duplicates of checks in `IsEvictable`.

damemi: Ah, I see so this check is meant for efficiency in the later checks if I understand correctly?

ingvagabund: Right, efficiency purposes only.

damemi: I think that's a great idea. Would you mind adding that to the comment for this block, to explain that this is for efficiency? It wasn't clear to me why we were looking for distinct pods at this point

ingvagabund: Done

Issue #564, opened 2021-05-09
RemovePodsViolatingTopologySpreadConstraint includeSoftConstraints: true not evicting


jsalatiel:
**What version of descheduler are you using?**

descheduler version
* image: k8s.gcr.io/descheduler/descheduler:v0.20.0
* helm.sh/chart: descheduler-0.20.0

**Does this issue reproduce with the latest release?**
Yes.

**Please provide a copy of your descheduler policy config file**
```
  policy.yaml: |
    apiVersion: "descheduler/v1alpha1"
    kind: "DeschedulerPolicy"
    strategies:
      LowNodeUtilization:
        enabled: false
        params:
          nodeResourceUtilizationThresholds:
            targetThresholds:
              cpu: 50
              memory: 50
              pods: 50
            thresholds:
              cpu: 20
              memory: 20
              pods: 20
      RemoveDuplicates:
        enabled: true
      RemovePodsViolatingInterPodAntiAffinity:
        enabled: true
      RemovePodsViolatingNodeAffinity:
        enabled: true
        params:
          nodeAffinityType:
          - requiredDuringSchedulingIgnoredDuringExecution
      RemovePodsViolatingNodeTaints:
        enabled: true
      RemovePodsViolatingTopologySpreadConstraint:
        enabled: true
        params:
          includeSoftConstraints: true
```

**What k8s version are you using (`kubectl version`)?**

<details><summary><code>kubectl version</code> Output</summary><br><pre>
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.9", GitCommit:"9dd794e454ac32d97cde41ae10be801ae98f75df", GitTreeState:"clean", BuildDate:"2021-03-18T01:09:28Z", GoVersion:"go1.15.8", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.9", GitCommit:"9dd794e454ac32d97cde41ae10be801ae98f75df", GitTreeState:"clean", BuildDate:"2021-03-18T01:00:06Z", GoVersion:"go1.15.8", Compiler:"gc", Platform:"linux/amd64"}

</pre></details>

**What did you do?**
I have created a deployment with the following topologySpread
```
      topologySpreadConstraints:
      - labelSelector:
          matchLabels:
            app: whoami
        maxSkew: 1
        topologyKey: topology.kubernetes.io/zone
[...]

PR #565, opened 2021-05-10
Add test cases for soft constraints/multi constraints

damemi:
Fixes https://github.com/kubernetes-sigs/descheduler/issues/564
/kind bug

### DIFF
--- a/pkg/descheduler/strategies/topologyspreadconstraint.go
+++ b/pkg/descheduler/strategies/topologyspreadconstraint.go
@@ -121,12 +121,11 @@ func RemovePodsViolatingTopologySpreadConstraint(
 		}
 
 		// ...where there is a topology constraint
-		//namespaceTopologySpreadConstrainPods := make([]v1.Pod, 0, len(namespacePods.Items))
 		namespaceTopologySpreadConstraints := make(map[v1.TopologySpreadConstraint]struct{})
 		for _, pod := range namespacePods.Items {
 			for _, constraint := range pod.Spec.TopologySpreadConstraints {
 				// Ignore soft topology constraints if they are not included
-				if (strategy.Params == nil || !strategy.Params.IncludeSoftConstraints) && constraint.WhenUnsatisfiable != v1.DoNotSchedule {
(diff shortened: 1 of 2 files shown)


pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -126,7 +126,7 @@ func RemovePodsViolatingTopologySpreadConstraint(
 		for _, pod := range namespacePods.Items {
 			for _, constraint := range pod.Spec.TopologySpreadConstraints {
 				// Ignore soft topology constraints if they are not included
-				if (strategy.Params == nil || !strategy.Params.IncludeSoftConstraints) && constraint.WhenUnsatisfiable != v1.DoNotSchedule {
+				if constraint.WhenUnsatisfiable == v1.ScheduleAnyway && (strategy.Params == nil || !strategy.Params.IncludeSoftConstraints) {
```
damemi, line 129: changed this to be a bit more readable

damemi:
Turns out there wasn't actually a bug with this, but we should still have these test cases anyway

pkg/descheduler/strategies/topologyspreadconstraint_test.go
```diff
@@ -96,6 +96,41 @@ func TestTopologySpreadConstraint(t *testing.T) {
...
+					labels: map[string]string{"foo": "bar"},
+				},
+			}),
+			expectedEvictedCount: 1,
+			strategy:             api.DeschedulerStrategy{Params: &api.StrategyParameters{IncludeSoftConstraints: true}},
+			namespaces:           []string{"ns1"},
+		},
 		{
 			name: "NEW",
 			//name: "2 domains, sizes [3,1], maxSkew=1, no pods eligible, move 0 pods",
```
lixiang233, line 136: Can you update the name of this testcase by the way?

pkg/descheduler/strategies/topologyspreadconstraint.go
ingvagabund: +1 for inclusion, not exclusion

pkg/descheduler/strategies/topologyspreadconstraint_test.go
damemi: Ah whoops, thanks for catching that!

PR #545, opened 2021-04-08
Add verify script for deep-copies generator

pravarag:
This PR is related to: https://github.com/kubernetes-sigs/descheduler/issues/483

This PR contains changes for verifying deep-copies generator. Below are the logs,

1. When there are no changes,
```
-bash>descheduler(add-verify-script-deep-copies)$
 ->  make verify-gen
./hack/verify-conversions.sh
Generated conversions verified.
./hack/verify-deep-copies.sh
Generated deep-copies verified.
```
2. After making a change in the api similar to the change mentioned [here](https://github.com/kubernetes-sigs/descheduler/pull/507#issuecomment-792820077) for validating failed scenario,
```
-bash>descheduler(add-verify-script-deep-copies)$
 ->  ./hack/verify-deep-copies.sh
Generated deep-copies output differs:
diff -Naupr pkg/api/v1alpha1/zz_generated.deepcopy.go ./hack/../_tmp/kube-verify.whSrQp/descheduler/pkg/api/v1alpha1/zz_generated.deepcopy.go
--- pkg/api/v1alpha1/zz_generated.deepcopy.go	2021-04-08 11:16:08.000000000 +0530
+++ ./hack/../_tmp/kube-verify.whSrQp/descheduler/pkg/api/v1alpha1/zz_generated.deepcopy.go	2021-04-08 11:17:03.000000000 +0530
@@ -61,7 +61,6 @@ func (in *DeschedulerPolicy) DeepCopyInt
 		*out = new(int)
 		**out = **in
 	}
-	out.Foo = in.Foo
 	return
 }

Generated deep-copies verify failed.
```

PR #568, opened 2021-05-16
Add verify script for govet & fix pkg/descheduler/descheduler_test.go

BinacsLee:
We can only use `t.Fatal` in the main test goroutine.

The solution is similar to:
https://github.com/kubernetes/kubernetes/blob/master/cmd/kube-proxy/app/server_test.go#L501

Thanks for review!

### DIFF
--- a/pkg/descheduler/descheduler_test.go
+++ b/pkg/descheduler/descheduler_test.go
@@ -7,7 +7,7 @@ import (
 	"testing"
 	"time"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 	fakeclientset "k8s.io/client-go/kubernetes/fake"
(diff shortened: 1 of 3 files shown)


pkg/descheduler/descheduler_test.go
```diff
@@ -44,15 +44,20 @@ func TestTaintsUpdated(t *testing.T) {
...
+	select {
+	case err := <-errChan:
 		if err != nil {
 			t.Fatalf("Unable to run descheduler strategies: %v", err)
 		}
-	}()
+	case <-time.After(300 * time.Millisecond):
+	}
 
-	// Wait for few cycles and then verify the only pod still exists
```
lixiang233, line 54: I think this comment is useful, can you keep it?

BinacsLee: Done.
Thanks for review ^-^

damemi (commented):
Could you share what the failure is for this? I don't think we currently run staticcheck in our CI, if not this would be a good PR to add that check

BinacsLee:
> Could you share what the failure is for this? I don't think we currently run staticcheck in our CI, if not this would be a good PR to add that check

Hi @damemi , the failure is `call to (*T).Fatalf from a non-test goroutine`:

image

More about the failure at: https://golang.org/doc/go1.16#vet

And I will do some work about staticcheck as soon as possible.

damemi:
@BinacsLee ah, I see, so this is a failure from `go vet`. The title made me think this was coming from something like [`verify-staticcheck.sh`](https://github.com/kubernetes/kubernetes/blob/master/hack/verify-staticcheck.sh).

In that case, let's update this PR to add a script to verify go vet, similar to https://github.com/kubernetes/kubernetes/blob/master/hack/verify-govet.sh. We can have that script run as part of `make verify`, which will include it in our CI runs so we can catch things like this automatically in the future. As an example, you can see how we added verify scripts in some other PRs like https://github.com/kubernetes-sigs/descheduler/pull/545

BinacsLee:
I'm sorry to confuse you. The script for govet has been added. Let's see how CI turns out ^-^

damemi:
@BinacsLee no problem! Thanks for the update, this looks good to me now
/approve
/lgtm

PR #577, opened 2021-05-25
Use stable batch/v1 API Group for Kubernetes 1.21

a7i:
Kubernetes 1.21: CronJob Reaches GA (group version: `batch/v1`).
Ref: https://kubernetes.io/blog/2021/04/09/kubernetes-release-1.21-cronjob-ga/

WARNING: This PR should not be merged until the new Descheduler version is released.

fyi, I tried to compute the api version but it was using 1.20 for my local Kind cluster of version 1.21 so I decided against it.:
```yaml
{{/*
Return the appropriate apiVersion for CronJob.
*/}}
{{- define "cronjob.apiVersion" -}}
{{- if lt (int .Capabilities.KubeVersion.Minor) 21 -}}
{{- print "batch/v1beta1" -}}
{{- else -}}
{{- print "batch/v1" -}}
{{- end -}}
{{- end -}}
```

```yaml
apiVersion: {{ template "cronjob.apiVersion" . }}
kind: CronJob
```

### DIFF
--- a/charts/descheduler/values.yaml
+++ b/charts/descheduler/values.yaml
@@ -19,6 +19,7 @@ resources:
 nameOverride: ""
 fullnameOverride: ""
 
+cronJobApiVersion: "batch/v1"  # for k8s version < 0.21.0, use "batch/v1beta1"
 schedule: "*/2 * * * *"
 #startingDeadlineSeconds: 200
 #successfulJobsHistoryLimit: 1
--- a/charts/descheduler/values.yaml
+++ b/charts/descheduler/values.yaml
@@ -19,7 +19,7 @@ resources:
 nameOverride: ""
 fullnameOverride: ""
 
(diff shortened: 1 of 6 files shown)


charts/descheduler/values.yaml
```diff
@@ -19,6 +19,7 @@ resources:
 nameOverride: ""
 fullnameOverride: ""
 
+cronJobApiVersion: "batch/v1"  # for k8s version < 0.21.0, use "batch/v1beta1"
```
damemi, line 22: I don't think it's necessary to parameterize this, but if so we should make a note to remove this once `v1beta1` is removed as a TODO

damemi (commented):
/ok-to-test

charts/descheduler/values.yaml
a7i: Added a TODO. Happy to remove and hard-code but I was hoping to honor the three release rule:
"a particular descheduler release is only tested against the three latest k8s minor versions" from README

PR #581, opened 2021-06-04
Remove namespace from ClusterRoleBinding

jsravn:
It's not namespace scoped. This breaks some tools like kpt.


damemi (commented):
/ok-to-test
/approve
/lgtm
Thanks @jsravn!

PR #585, opened 2021-06-09
TestTaintsUpdated: Drop errChan, it's usefull only during the startup

ingvagabund:
Otherwise, one can get:
```
panic: send on closed channel

goroutine 76 [running]:
sigs.k8s.io/descheduler/pkg/descheduler.TestTaintsUpdated.func1(0x2d1b390, 0xc000126010, 0xc0000d8c60, 0xc00038fbd0, 0xc000111320, 0xc000239380)
	/home/travis/gopath/src/github.com/openshift/descheduler/_output/local/go/src/sigs.k8s.io/descheduler/pkg/descheduler/descheduler_test.go:51 +0xbe
created by sigs.k8s.io/descheduler/pkg/descheduler.TestTaintsUpdated
	/home/travis/gopath/src/github.com/openshift/descheduler/_output/local/go/src/sigs.k8s.io/descheduler/pkg/descheduler/descheduler_test.go:49 +0x6db
FAIL	sigs.k8s.io/descheduler/pkg/descheduler	0.845s
```

### DIFF
--- a/pkg/descheduler/descheduler_test.go
+++ b/pkg/descheduler/descheduler_test.go
@@ -44,20 +44,16 @@ func TestTaintsUpdated(t *testing.T) {
 	}
 	rs.Client = client
 	rs.DeschedulingInterval = 100 * time.Millisecond
-	errChan := make(chan error, 1)
-	defer close(errChan)
+
 	go func() {
 		err := RunDeschedulerStrategies(ctx, rs, dp, "v1beta1", stopChannel)
-		errChan <- err
-	}()
-	select {
-	case err := <-errChan:
 		if err != nil {
-			t.Fatalf("Unable to run descheduler strategies: %v", err)
+			t.Logf("ERROR: Unable to run descheduler strategies: %v", err)
 		}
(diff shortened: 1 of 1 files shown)


pkg/descheduler/descheduler_test.go
```diff
@@ -44,20 +44,16 @@ func TestTaintsUpdated(t *testing.T) {
 	}
 	rs.Client = client
 	rs.DeschedulingInterval = 100 * time.Millisecond
-	errChan := make(chan error, 1)
-	defer close(errChan)
+
 	go func() {
 		err := RunDeschedulerStrategies(ctx, rs, dp, "v1beta1", stopChannel)
-		errChan <- err
```
damemi, line 51: I don't think we should remove this pattern entirely (from when it was added in https://github.com/kubernetes-sigs/descheduler/pull/568)

Rather, I think the panic you're seeing could be fixed by moving the nil check to here from L55, so we're not always sending nil over this channel. wdyt?

Issue #586, opened 2021-06-10
Don't start when there are unknown keys in the config


der-eismann:
**Is your feature request related to a problem? Please describe.**
Today we decided to try out the descheduler and created a simple config to delete pods from two deployments after 24h. But we put the `labelSelector` in the wrong section so instead it was terminating **everything** that was older than 24h. Luckily we had enough PDBs defined and could stop it in time. Of course one should test everything in a staging environment first, but such a simple mistake shouldn't be enough to delete almost all of your pods.

**Describe the solution you'd like**
Let the program fail if there are unrecognized config keys in a part where they don't belong / use strict unmarshaling.

**Describe alternatives you've considered**
none

**What version of descheduler are you using?**
descheduler version: v0.21.0

**Additional context**
This is/was our config:
```
apiVersion: "descheduler/v1alpha1"
kind: "DeschedulerPolicy"
nodeSelector: node.kubernetes.io/role=worker
evictLocalStoragePods: true
evictSystemCriticalPods: false
ignorePvcPods: false
strategies:
  "PodLifeTime":
    enabled: true
    params:
      podLifeTime:
        maxPodLifeTimeSeconds: 86400
        labelSelector:
          matchExpressions:
            - {key: app, operator: In, values: [app1, app2]}
```
In this case I put the `labelSelector` under the `podLifeTime` key because it seemed logical to me. Instead it needs to be on the same level as `podLifeTime`. In this case the descheduler should fail to start because `podLifeTime` has no `labelSelector` field.

damemi:
+1, good suggestion, thank you @der-eismann 
Currently in many places we check for the expected parameters to be valid, but I don't think there are any checks for unexpected parameters.

I think a better refactoring of our parameter API could help strongly enforce this. For example, if the `params` field was an interface which each strategy cast to its own parameter type (`podLifeTime`), then that conversion would catch the unknown field and throw an error.

Perhaps we should add something like this to the proposal for the new `v1alpha2` API (https://docs.google.com/document/d/1S1JCh-0F-QCJvBBG-kbmXiHAJFF8doArhDIAKbOj93I/edit#), along with the changes from https://github.com/kubernetes-sigs/descheduler/issues/486

damemi:
We can probably close this, since it is a big requirement for the descheduler framework proposal to improve the config API. Ref https://github.com/kubernetes-sigs/descheduler/issues/837

Issue #557, opened 2021-04-28
Descheduler strategies as plugins


ingvagabund:
**Is your feature request related to a problem? Please describe.**

The number of strategies is increasing and we still have code of all the strategies under the same directory. Which increases risk of identifier name collisions. It’s also currently impractical to add custom strategies into the Descheduler (in case someone decides to fork the code and integrate e.g. proprietary strategies.).

Evicting pods outside the strategies will free us from passing pod evictor into each strategy. It will also allow us to run the Descheduler as a simulator more easily. E.g. by checking which strategies are evicting the same pods, how one strategy impacts other, etc.

**Describe the solution you'd like**

Create a factory like initialization of plugins and a new strategy interface with Deschedule method

**Describe alternatives you've considered**
NONE

**What version of descheduler are you using?**

Any version

**Additional context**

Google doc for open discussion: https://docs.google.com/document/d/1VuqbwakL69H4sbsGP8J6C10x1NfEpeAQY_cHibrZY0M/edit#

damemi:
+1, standardizing an interface for each strategy and moving them into their own packages is at the very least a big improvement for our code organization. I would go even further to say this is more than a feature, but a considerable redesign (and one well worth having). 

I'll leave some comments on the doc, thanks for starting this @ingvagabund !
/kind design

hanumanthan:
This is a great refractor. I would be interested to pick this up as it will give some good understanding of the code base. 
Is this in a shape to start? I can take a shot at it and raise doubts here/design doc or via PR as I encounter them

There is another recommendation of converting strategy params to an [interface](https://github.com/kubernetes-sigs/descheduler/issues/586#issuecomment-858675447). 
Can I attempt both in the same change ?

damemi:
> Can I attempt both in the same change ?

Let's keep different changes to their own PRs. Especially with refactorings like this, if something goes wrong that makes it easier to track down and revert

ingvagabund:
PoC in https://github.com/kubernetes-sigs/descheduler/pull/781

knelasevero:
@ingvagabund @damemi @a7i should we close this one? Considering the title, strategies are moved to plugins now

damemi:
I think it's worth keeping open until we've gotten off of the [migration wrapper](https://github.com/kubernetes-sigs/descheduler/blob/70df89601a943d18f96dad3fd911bea19f10a498/pkg/descheduler/strategy_migration.go#L43)

[...]

PR #587, opened 2021-06-12
Introducing descheduler profiles with v1alpha2 api version

### LINKED ISSUE #486: Multiple descheduler strategies of the same type

hanumanthan:
Closes #486 

1. Have added a new version `v1alpha2` while providing a backward compatibility with `v1alpha1` as defined in [design doc](https://docs.google.com/document/d/1S1JCh-0F-QCJvBBG-kbmXiHAJFF8doArhDIAKbOj93I/edit#)
2. Enabled all profiles and strategies by default and simplified strategy parameter struct as defined in the doc

The way I have converted two api policy versions may not be the best way as I have limited understanding of CRD's. Also, the changes in `readme` may not convey the details in the right way. Happy to fix both based on comments.


damemi (commented):
/ok-to-test

damemi:
Thank you for starting this work @hanumanthan. From a quick overview this looks to be the right direction but I will give it a more in depth review when I have a chance.

The one part I am still unsure about is if we will actually need the version conversions, or if we will be fine to just consume `v1alpha2` in our code. So, I will need to look more into how those work. 

I might recommend holding off on the embedded types for now, just so we can get the version bump with profiles merged smoothly. We could then merge the embedded types in the same release (or discuss if we would rather go with something like defining a params interface https://github.com/kubernetes-sigs/descheduler/issues/586#issuecomment-858675447).

hanumanthan:
Great @damemi I will wait for the review comments and remove the embedded struct changes along with fixing the review comments. 

Regarding the param interface, I am happy to pick that up. Have raised it [here](https://github.com/kubernetes-sigs/descheduler/issues/557)

hanumanthan:
hi @damemi Did you get a chance to review this ?

pkg/api/v1alpha2/types.go
```diff
@@ -0,0 +1,116 @@
...
+	// MaxNoOfPodsToEvictPerNode restricts maximum of pods to be evicted per node.
+	MaxNoOfPodsToEvictPerNode *int `json:"maxNoOfPodsToEvictPerNode,omitempty"`
+}
+
+type DeschedulerProfile struct {
+	// Name
+	Name string `json:"name,omitempty"`
+
+	// Enabled
+	Enabled *bool `json:"enabled,omitempty"`
```
damemi, line 53: I don't think this needs to be a pointer. I believe the reason we made the `strategy.Enabled` a pointer originally was part of converting most of our fields to pointers, so that they could be omitted from yaml/json when encoded. But this is a required field, so it shouldn't ever be empty anyway

damemi:
Hi @hanumanthan, very sorry but I haven't been able to take a deep look yet. I will prioritize that this or next week.

In the meantime, could you please remove the changes related to embedding the strategy params? We won't do that in this PR, and it will make it easier to review & test the changes around just the new api version

pkg/api/v1alpha2/types.go
hanumanthan: Its a pointer to allow the flag to be omitted from the config. Do we need `enabled` always present in the config, even if its `true`?
IMO this flag can be made optional if its only for debugging. Otherwise it feels a repetition of flag across all strategies and profiles. wdyt ?

pkg/api/v1alpha2/register.go
```diff
@@ -0,0 +1,61 @@
...
+
+func init() {
+	// We only register manually written functions here. The registration of the
+	// generated functions takes place in the generated files. The separation
+	// makes the code compile even when the generated files are missing.
+	localSchemeBuilder.Register(addKnownTypes, addDefaultingFuncs)
+}
+
+func addKnownTypes(scheme *runtime.Scheme) error {
+	// TODO this will get cleaned up with the scheme types are fixed
```
ingvagabund, line 55: Wondering what this comment actually mean (it's in v1alpha1 as well)

pkg/api/v1alpha2/types.go
```diff
@@ -0,0 +1,105 @@
...
+package v1alpha2
+
+import (
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+type DeschedulerPolicy struct {
```
ingvagabund, line 26: This does not correspond to v1alpha2 as specified in https://docs.google.com/document/d/1S1JCh-0F-QCJvBBG-kbmXiHAJFF8doArhDIAKbOj93I/edit#. I am expecting:
```
type DeschedulerPolicy struct {
  metav1.TypeMeta `json:”,inline”`

  Profiles []DeschedulerProfile `json:”profiles”`
  ...
}

type DeschedulerProfile struct {
  Name       string `json:”name”`
  Strategies StrategyList `json:”strategies”`
}
```
@hanumanthan any reason you did not incorporate the new type in the first commit?

pkg/api/v1alpha2/types.go
ingvagabund: +1 for not making it a pointer

ingvagabund (changes requested):

pkg/api/v1alpha2/types.go
hanumanthan: I seperated as two commits to 
1. Make review easier for this [file](pkg/descheduler/policyconfig.go)
2. To verify if the type conversion happens.

Would you like two commits to be merged as one ?
> This does not correspond to v1alpha2 as specified

Are you highlighting this in the context of first commit or this PR (confirming if the 2nd commit matches the design spec)?

pkg/api/v1alpha2/register.go
hanumanthan: TBH i didn't understand schemes, objects, runtime and CRD. I copied v1alpha1 making necessary changes. 

Any doc or code that I can refer to understand the above mentioned

Svisstack:
This is related to the error I'm facing installing it from helm chart?

`~# helm install descheduler --namespace kube-system descheduler/descheduler
Error: unable to build kubernetes objects from release manifest: unable to recognize "": no matches for kind "CronJob" in version "batch/v1"
`

pkg/api/v1alpha2/types.go
ingvagabund: Given the first commit is named "Adding v1alpha2 version with backward version compatibility" I'd assume you define the v1alpha2 version of the type properly there.

damemi:
@Svisstack no, this PR is to add a new version for the Descheduler's own config API

Your error is that you cluster is not recognizing the `v1` CronJob type. What version of kubernetes are you on? I believe CronJobs were just promoted to `v1` in 1.21. You can set the [`cronJobApiVersion` value in the helm chart](https://github.com/kubernetes-sigs/descheduler/blob/f07089d7b377a49f3d52362e581138d83c453715/charts/descheduler/values.yaml#L22) to `batch/v1beta1` if you are on an older version of kubernetes.

knelasevero:
@damemi @ingvagabund now that the strategies are migrated to the plugin approach, I imagine moving to v1alpha2 will be on the of next steps. I wanted to check what will be the plan regarding this new sub-effort. (EDIT: We still need to get rid of workarounds and other glues for the plugin migration https://github.com/kubernetes-sigs/descheduler/issues/557#issuecomment-1219737145)

1- 
https://github.com/kubernetes-sigs/descheduler/pull/587#issuecomment-861821017
> The one part I am still unsure about is if we will actually need the version conversions, or if we will be fine to just consume v1alpha2 in our code 

I think that, before the framework idea, conversion made some sense since the config would still be a bit similar. Now the config will look like something completely different, passing arguments to strategies in a different way, and so on. Making conversion now seems very very complicated and weird to maintain. Probably better to drop v1alpha1 altogether and directly move to v1alpha2 but with a big release note announcement... What are your thoughts on this? This will require a warning somewhere  making clear to users, that if they want to upgrade, and they forget to change the config file version and structure, things will break.

2-
@hanumantha

Do you still want to push this effort forward? Probably with getting familiar with https://github.com/kubernetes-sigs/descheduler/issues/837 and related issues and PRs. v1alpha2 would take another shape and would probably need to be another PR.

ingvagabund:
We need to support v1alpha1 for few releases to provide some deprecated time before users convert their configuration to v1alpha2. The conversion will be automatic without user intervention.

Issue #591, opened 2021-06-23
HighNodeUtilization README.md is duplication of LowNodeUtilization


chrisjohnson00:
Much of the HighNodeUtilization documentation in the README.md is a copy/paste from LowNodeUtilization with /s/Low/High replaced.  Other than the type change, the documentation is incorrect as it's a copy of Low Node's threshold functionality.

```
HighNodeUtilization

This strategy finds nodes that are under utilized and evicts pods in the hope that these pods will be scheduled compactly into fewer nodes. This strategy must be used with the scheduler strategy MostRequestedPriority. The parameters of this strategy are configured under nodeResourceUtilizationThresholds.
```

The above example explains what I mean, where it says it 

> ...finds nodes that are under utilized and evicts pods in the hope that these pods will be...

I don't know the inner workings of descheduler to propose this documentation update, but if other contributors are willing to review my PR with that in mind, I'd be willing to give it a crack at correcting it (LMK)

damemi:
Hi @chrisjohnson00,
The wording for the 2 strategies is very similar, because under the hood they are essentially the same functions with an invert. However I don't think the section you shared is incorrect:

> This strategy finds nodes that are under utilized and evicts pods in the hope that these pods will be scheduled compactly into fewer nodes.

For HighNodeUtilization, this means that it is looking for nodes that are *under* the utilization thresholds and evicts pods *from those nodes* to compact them onto *other* nodes

> This strategy finds nodes that are under utilized and evicts pods, if possible, from other nodes in the hope that recreation of evicted pods will be scheduled on these underutilized nodes

This is the matching section from LowNodeUtilization, which means that while it is still looking for nodes that are *under* the thresholds, it is instead evicting pods from *other* nodes onto those under utilized nodes.

This could probably be worded better in the docs though, so if you have the time to improve them then please feel free to submit a PR 🙂

seanmalloy:
Docs have been updated in #592.

/close

PR #592, opened 2021-06-24
docs: adding clarification to HighNodeUtilization's purpose

chrisjohnson00:
This PR resolves issue #591

### DIFF
--- a/README.md
+++ b/README.md
@@ -267,9 +267,11 @@ under utilized frequently or for a short period of time. By default, `numberOfNo
 
 ### HighNodeUtilization
 
-This strategy finds nodes that are under utilized and evicts pods in the hope that these pods will be scheduled compactly into fewer nodes.
-This strategy **must** be used with the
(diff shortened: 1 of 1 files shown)


chrisjohnson00:
I've signed the CLA

ingvagabund (changes requested):

damemi (commented):
/ok-to-test
/approve
thanks @chrisjohnson00, after @ingvagabund's comment I think this helps.

You'll also need to sign the CLA before this will merge. Thanks again!

chrisjohnson00:
> You'll also need to sign the CLA before this will merge. Thanks again!

I signed it yesterday, but `/check-cla` doesn't seem to be clearing the check... what can I do here?

damemi:
That's odd, usually it should pick it up pretty quick. Maybe the bot is just delayed right now for some reason, if it doesn't update we can look into it

damemi:
@chrisjohnson00 I'm not sure why the CLA check isn't passing for you, perhaps you could try filling it out again? You could also ask in #sig-contribex on k8s slack and they should be able to help

chrisjohnson00:
Woohoo!!

Issue #596, opened 2021-07-05
Allow deletion of failed pods


jaymebrd:
**Is your feature request related to a problem? Please describe.**

When pods are in status `Failed` the only mechanism to clean up these pods is the kubernetes[ garbage collector](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-garbage-collection) which is currently defaulted to 12500 which means that for most people this is useless.

There's been a few recent issues where the reason for the pod failures has been related to a kubernetes bug. One particular example is the recent nodeAffinity issue where the Kubelet rejects pod scheduled based on newly added node labels which have not been observed by the kubelet yet #93338

```
Status:               Failed
Reason:               NodeAffinity
Message:              Pod Predicate NodeAffinity failed
```

This can be particularly problematic to certain custom controllers which haven't implemented recreation of failed pod, or deletion of failed pods.

**Describe the solution you'd like**

Allow deletion of failed pods, possibly based on some reason exclusions (you may want to keep evicted pods around, for example)

**Describe alternatives you've considered**

The default garbage collection value in kubernetes could be made configurable, and/or given a better default than 12500, however, it seems more appropriate for a project like this to allow more configuration on which `reason` should be inclusive/exclusive of the automatic deletion.

**What version of descheduler are you using?**

descheduler version: N/A

**Additional context**

See more details here for a related the feature request made to kubernetes. https://github.com/kubernetes/kubernetes/issues/99986
This feels like a potential good fit for the descheduler.

damemi:
This is a hardcoded filter the descheduler uses when listing pods on a node: https://github.com/kubernetes-sigs/descheduler/blob/84b174e8417edbb13e51a3207191c3449a4ce309/pkg/descheduler/pod/pods.go#L83

So it's not just that it won't evict them, but it doesn't even see them at all. This makes sense because if it listed Failed pods, it would consider them in each strategy when calculating pods to evict, which could produce odd behavior (for example, evicting a pod for antiaffinity with a failed pod on the same node doesn't really make sense)

However, I see some of the use cases you have pointed out as valid. So we could add some (non-default) option to evict, or at least consider, Failed pods, but we should think of the best way to do that.

What about if we add a new strategy specifically for this kind of garbage collection? Something like `RemoveFailedPods`. Simple enough and doesn't interfere with core code that would affect the other strategies which don't really need failed pods in their logic

jaymebrd:
[...]

PR #610, opened 2021-07-29
Introduce RemoveFailedPods strategy

a7i:
Closes issue #596


README.md
```diff
@@ -122,7 +123,7 @@ See the [user guide](docs/user-guide.md) in the `/docs` directory.
 ## Policy and Strategies
 
 Descheduler's policy is configurable and includes strategies that can be enabled or disabled.
-Nine strategies
+Ten strategies
```
grosser, line 126:
```suggestion
Strategies:
```

README.md
```diff
@@ -132,6 +133,7 @@ Nine strategies
 7. `RemovePodsViolatingTopologySpreadConstraint`
 8. `RemovePodsHavingTooManyRestarts`
 9. `PodLifeTime`  
+10. `RemoveFailedPods`
```
grosser, line 136: why do we have 2 lists 🤷

README.md
```diff
@@ -520,6 +522,51 @@ strategies:
...
+|`include`|list(string)|
+|`exclude`|list(string)|
+
+**Example:**
+
+```yaml
+apiVersion: "descheduler/v1alpha1"
+kind: "DeschedulerPolicy"
+strategies:
+  "RemoveFailedPods":
```
grosser, line 557:
```suggestion
  RemoveFailedPods:
```

README.md
a7i: Yea, that's a good point. the ToC gets auto-generated from `./hack/update-toc.sh`. This list here doesn't bring any value.. just maintenance overhead

damemi: I agree, once we added the ToC this list became redundant especially as we add more and more strategies. I'm fine with removing it if you want to

grosser: best in it's own PR
also FYI github has a TOC builtin
Screen Shot 2021-07-30 at 10 19 49 AM

a7i: #611

README.md
a7i: Having quote is consistent with the rest of the examples.

damemi (commented):
/ok-to-test

README.md
```diff
@@ -515,6 +516,51 @@ strategies:
          - "Pending"
 ```
 
+### RemoveFailedPods
+
+This strategy evicts pods that are in failed status phase.
```
damemi, line 521: This could describe a lot more of what the strategy does, especially the parameters

README.md
```diff
@@ -515,6 +516,51 @@ strategies:
...
+|`nodeFit`|bool (see [node fit filtering](#node-fit-filtering))|
+
+
+**Reasons Parameters:**
+
+|Name|Type|
+|---|---|
+|`includingInitContainers`|bool|
+|`include`|list(string)|
+|`exclude`|list(string)|
```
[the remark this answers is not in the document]
damemi: thinking we could simplify this to not need an embedded type... if `includingInitContainers` was a higher-level option that applied to all reasons, then `reasons` could just be a list (if we also got rid of include/exclude)

pkg/descheduler/strategies/failedpods.go
```diff
@@ -0,0 +1,178 @@
...
+
+// RemoveFailedPods removes Pods that are in failed status phase.
+func RemoveFailedPods(
+	ctx context.Context,
+	client clientset.Interface,
+	strategy api.DeschedulerStrategy,
+	nodes []*v1.Node,
+	podEvictor *evictions.PodEvictor,
+) {
+	if err := validateRemoveFailedPodsParams(strategy.Params); err != nil {
```
damemi, line 27: Do you want to break this out into a `validateAndParse...` params function that returns a struct, similar to in https://github.com/kubernetes-sigs/descheduler/pull/576/files#diff-3ef994f616f4d3726bc08b98eb341fee83b26703bc6179277f5e91286d99b289R51-R60?

pkg/descheduler/strategies/failedpods.go
```diff
@@ -0,0 +1,178 @@
...
+		if podAgeSeconds < *failedPodsStrategy.MinPodLifetimeSeconds {
+			errs = append(errs, fmt.Errorf("ignoring pod for eviction as it does not exceed the min age seconds of %d", *failedPodsStrategy.MinPodLifetimeSeconds))
+		}
+	}
+
+	if len(excludeOwnerKinds) > 0 {
+		ownerRefList := podutil.OwnerRef(pod)
+		for _, owner := range ownerRefList {
+			if excludeOwnerKinds.Has(owner.Kind) {
+				errs = append(errs, fmt.Errorf("ignoring pod for eviction as owner kind of %s is excluded", owner.Kind))
```
damemi, line 120: another redundant `ignoring pod for eviction`

pkg/descheduler/strategies/failedpods.go
```diff
@@ -0,0 +1,178 @@
...
+		if failedPodsStrategy.Reasons != nil && failedPodsStrategy.Reasons.IncludingInitContainers {
+			reasons = append(reasons, getFailedContainerStatusReasons(pod.Status.InitContainerStatuses)...)
+		}
+
+		if len(includeReasons) > 0 && !includeReasons.HasAny(reasons...) {
+			errs = append(errs, fmt.Errorf("ignoring pod for eviction as it does not match any of the the include reasons"))
+		}
+
+		if len(excludeReasons) > 0 && excludeReasons.HasAny(reasons...) {
+			errs = append(errs, fmt.Errorf("ignoring pod for eviction as it matches a exclude reason"))
```
damemi, line 137: remove these 2 `ignoring pod for eviction` as well

damemi (commented):
looks pretty thorough, a couple notes but fundamentally solid imo

README.md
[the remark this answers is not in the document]
a7i: for my use-case, just having `reason` would suffice. I was just trying to make it more flexible but I agree that having an extra struct may not be ideal.

I'll pull the two fields up one level and just have `reasons` and `includingInitContainers`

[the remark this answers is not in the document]
damemi: Yeah, good thinking ahead but without any use case or demand for those extra fields we can keep it simple for now

damemi (commented):
This looks good to squash, then I can approve

One nit though, which I'd like your input on, is maybe the new StrategyParam validator should be in its own package? It's minor but it would be nice to keep the `strategies` package to strategy code.

I was also not really sure about the way it wraps the "common" params inside another struct for the new strategy, but I suppose that's the best way to accomplish that (with the common params embedded). Hopefully we can start moving other strategies to use this function too and clean up their duped code

a7i:
> I was also not really sure about the way it wraps the "common" params inside another struct for the new strategy, but I suppose that's the best way to accomplish that (with the common params embedded). Hopefully we can start moving other strategies to use this function too and clean up their duped code

Ok, I can do that. something like `pkg/descheduler/internal/validators/`? (hence `internal` to reduce public API surface).

As far as wrapping the common params, I don't have a clever idea. If it was just a validator, could use interface. But it's also a parser that returns the parsed params so the return type gets tricky.

damemi:
@a7i maybe just `pkg/descheduler/strategies/validation`?

damemi (commented):
/approve
I think the `validation` package will have some room to evolve as we implement it in other strategies but it looks good now. Thanks @a7i

pkg/descheduler/strategies/validation/failedpods_strategyparams.go
```diff
@@ -0,0 +1,52 @@
+package validation
```
ingvagabund, line 1: Let's move this under `pkg/descheduler/strategies/failedpods.go` unless @damemi has different plan for the validation code.

ingvagabund (changes requested):
Given all strategies use the same `StrategyParameters` type it makes sense to put all the validation under the same package as all the params are shared across multiple strategies. On the other hand each strategy accesses a subset of the parameters so some part of the validation needs to stay as close to each strategy as possible.

The rest looks good.

pkg/descheduler/strategies/validation/failedpods_strategyparams.go
damemi: Yeah, it is a tough one. I mentioned in another comment that I think this validation package will need some work over time but I do agree that perhaps it makes more sense to keep the strategy-specific validation colocated with the strategies themselves :\

Moving out the common code does help clean up a lot of the strategy code, so that much is at least still very helpful. @a7i really sorry to be back and forth on this, but what do you think?

a7i: I personally like it all in the validation package and move towards a pattern such as below:

```
validation.ValidateStrategyParams(...)
                 .withExcludeOwnerKinds(kinds []string)
                 .withReasons(reasons []string)
                 ...
```

My plan was to refactor the validation package in a subsequent PR. With that being said, in order to move the PR forward, I can go ahead and move it back to the strategy package and revisit in that PR.

ingvagabund: Please move the code back so we have more time to discuss how we are going to proceed in the next PR.

a7i: Done

PR #613, opened 2021-08-17
nodeutilization strategy: "numberOfNodes" should work as documented

derdanne:
The "numberOfNodes" paramter should take effect as documented, when underutilzed nodes count is above the configured value. Currently it takes also effect, when nodes count is equal the configured value.


damemi (commented):
/ok-to-test

damemi (commented):
/approve
/lgtm

PR #619, opened 2021-09-02
[helm values] Add note on evict-local-storage-pods

roim:
Setting it to true causes:

```
Error: unknown command "true" for "descheduler"
Run 'descheduler --help' for usage.
unknown command "true" for "descheduler"
```

This is confusing as other references to this parameter (outside of helm) do ask you to set it to true.

Would be better to just accept the value here, but for now this is what I can help with.

### DIFF
--- a/charts/descheduler/values.yaml
+++ b/charts/descheduler/values.yaml
@@ -35,7 +35,7 @@ deschedulingInterval: 5m
 
 cmdOptions:
   v: 3
-  # evict-local-storage-pods:
+  # evict-local-storage-pods: # To enable, just uncomment the key. Don't set its value to "true".
   # max-pods-to-evict-per-node: 10
   # node-selector: "key1=value1,key2=value2"
 


damemi (commented):
that's odd that it doesn't work, did you also get this error when setting it to `true` (without quotes)?

roim:
> that's odd that it doesn't work, did you also get this error when setting it to `true` (without quotes)?

That's actually what I set, and I don't think I tried it with quotes

ingvagabund:
```
ERROR: failed to create cluster: failed to pull image "kindest/node:v1.21.4": command "docker pull kindest/node:v1.21.4" failed with error: exit status 1
```
Opening https://github.com/kubernetes/test-infra/pull/23570

seanmalloy:
@damemi and @ingvagabund any idea why the CI bot thinks test `pull-descheduler-test-e2e-k8s-1-21-1-21` needs to be run? From what I can tell it should not be running this job for the master branch. See https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-sigs/descheduler/descheduler-presubmits-master.yaml.

damemi:
@seanmalloy that is weird, maybe closing and reopening this will help?
/close

PR #629, opened 2021-09-14
fix duplicate code in node_affinity_test.go

chenkaiyue:
Fix duplicate code in node_affinity_test.go.

### DIFF
--- a/pkg/descheduler/strategies/node_affinity_test.go
+++ b/pkg/descheduler/strategies/node_affinity_test.go
@@ -59,7 +59,7 @@ func TestRemovePodsViolatingNodeAffinity(t *testing.T) {
 	nodeWithoutLabels := test.BuildTestNode("nodeWithoutLabels", 2000, 3000, 10, nil)
 
 	unschedulableNodeWithLabels := test.BuildTestNode("unschedulableNodeWithLabels", 2000, 3000, 10, nil)
-	nodeWithLabels.Labels[nodeLabelKey] = nodeLabelValue
+	unschedulableNodeWithLabels.Labels[nodeLabelKey] = nodeLabelValue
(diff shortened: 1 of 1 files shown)


pkg/descheduler/strategies/node_affinity_test.go
```diff
@@ -59,7 +59,6 @@ func TestRemovePodsViolatingNodeAffinity(t *testing.T) {
 	nodeWithoutLabels := test.BuildTestNode("nodeWithoutLabels", 2000, 3000, 10, nil)
 
 	unschedulableNodeWithLabels := test.BuildTestNode("unschedulableNodeWithLabels", 2000, 3000, 10, nil)
-	nodeWithLabels.Labels[nodeLabelKey] = nodeLabelValue
```
damemi, line 62: this line is probably supposed to be `unschedulableNodeWithLabels.Labels[nodeLabelKey] = nodeLabelValue`

damemi (commented):
/ok-to-test

pkg/descheduler/strategies/node_affinity_test.go
ingvagabund: The line is from 2018-01-12. Anyway, the line has no effect since there's no node scheduled to `unschedulableNodeWithLabels` (two unit tests utilize the node). When the strategy is tested, there's either `nodeWithLabels` or `nodeWithoutLabels` nodes which play the major role.

+1 for just renaming `nodeLabelValue` into `unschedulableNodeWithLabels` rather than deleting the line. The test is old and needs to be refactored.

chenkaiyue: @damemi @ingvagabund  Thx, i have fixed it.

PR #635, opened 2021-09-30
Added request considerations to NodeFit Feature

### LINKED ISSUE #529: Improve LowNodeUtilization's consideration of evictable pods

### LINKED ISSUE #574: Remove redundant checks for taints and nodeSelectors from strategies.

Currently many features implement their own checks to determine if a pod will end up on the same node after eviction. Some strategies check for taints, others check for nodeSelectors, others check both, and some don't check for this edge case. With the addition of the `nodeFit` feature, checks for taints, nodeSelectors, and whether a node is unschedulable can be turned on and off for almost all strategies. The `nodeFit` feature allows for a consistent approach to this optimization so that all strategies are in line with one another.

Ideally, every strategy should explicitly *not* check for taints and/or nodeSelectors. If the consideration of these parameters is desired, one could enable the `nodeFit` strategy parameter and achieve the same optimization.

None.

Any version.

**Additional context**

See the discussion in #559 for more detailed information regarding exactly where edits should be made to the codebase.

### LINKED ISSUE #604: TopologySpreadConstraint should take target node's requests allocation into consideration before evicting

TopologySpreadConstraint strategy should take target node's requests allocation into consideration before evicting them.

For example, if I have 3 nodes in the cluster with 3 domains size `[7 5 3]`, `maxSkew=1`, `whenUnsatisfiable=ScheduleAnyways`, it will attempt to evict 2 pods from the node1 and place them on node3 without any consideration of whether node3 is over-allocated or not. If the cluster is not set to scale, we will get in a constant evict/schedule loop as the 2 pods get evicted (and since node3 is over-allocated), Scheduler will put it back on node1 again.

It would be ideal for Descheduler to take target node's requests allocation into consideration to see if the pods that are about to be evicted would fit on the target node.

No workarounds.

descheduler version: v0.21.0

**Additional context**

This is related to other non-utilization strategies as well.

RyanDevlin:
Fixes #604
Fixes #529
Fixes #574

This PR adds checks for resource requests when the [NodeFit](https://github.com/kubernetes-sigs/descheduler#node-fit-filtering) pod filtering flag is enabled. The `NodeFit` flag instructs the Descheduler to determine if there is a node in the cluster the pod might fit on before evicting the pod. In the past this check did not consider the CPU, memory, storage, and other `requests` on these nodes. It now considers any and all `requests` made by the pod and the nodes, including custom resources.

Originally I only intended to add checks for resource `requests` to the `TopologySpread` strategy. As I did this, I ended up modifying global functions in `pkg/descheduler/node/node.go` which control eviction filtering logic. This has the side effect of aligning all strategies that can enable the `NodeFit` flag with my original feature. So even though I didn't intend to fix #574 and #529, they were automatically implemented by fixing #604.

I would have broken these out into two PRs (as suggested by @damemi), but once I finished #604 I realized all that was needed for #529 and #574 were a few extra test cases.

I'd love some feedback on my implementation, especially regarding the changes I've made to the `TopologySpread` strategy.


RyanDevlin:
Also I've signed the CLA, I'm looking into why this check is failing.

PR #636, opened 2021-09-30
Added request considerations to NodeFit Feature

### LINKED ISSUE #529: Improve LowNodeUtilization's consideration of evictable pods

### LINKED ISSUE #574: Remove redundant checks for taints and nodeSelectors from strategies.

### LINKED ISSUE #579: Exception stops execution of descheduler

descheduler version: v0.20.0

yes

default helm chart from https://kubernetes-sigs.github.io/descheduler/

default helm chart from https://kubernetes-sigs.github.io/descheduler/

$ kubectl version
Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.1", GitCommit:"5e58841cce77d4bc13713ad2b91fa0d961e69192", GitTreeState:"clean", BuildDate:"2021-05-12T14:18:45Z", GoVersion:"go1.16.4", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.1", GitCommit:"c4d752765b3bbac2237bf87cf0b1c2e307844666", GitTreeState:"clean", BuildDate:"2020-12-18T12:00:47Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"linux/amd64"}
</pre></details>

**What did you do?**

1. Deployed Descheduler and a few other services.  
Descheduler does its job like its should.

2. Deployed Kata containers https://github.com/kata-containers/kata-containers/tree/main/tools/packaging/kata-deploy

Descheduler keeps working as before.

**What did you see instead?**

Descheduler fails to execute with this exception
```
E0528 08:41:48.775507       1 runtime.go:78] Observed a panic: &errors.errorString{s:"feature \"PodOverhead\" is not registered in FeatureGate \"k8s.io/apiserver/pkg/util/feature/feature_gate.go:28\""} (feature "PodOverhead" is not registered in FeatureGate "k8s.io/apiserver/pkg/util/feature/feature_gate.go:28")
goroutine 1 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic(0x153ddc0, 0xc00027a380)
	/go/src/sigs.k8s.io/descheduler/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:74 +0x95
k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)
	/go/src/sigs.k8s.io/descheduler/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:48 +0x89
panic(0x153ddc0, 0xc00027a380)
	/usr/local/go/src/runtime/panic.go:969 +0x1b9
k8s.io/component-base/featuregate.(*featureGate).Enabled(0xc0003a5140, 0x171d05b, 0xb, 0x0)
	/go/src/sigs.k8s.io/descheduler/vendor/k8s.io/component-base/featuregate/feature_gate.go:302 +0x237
sigs.k8s.io/descheduler/pkg/utils.PodRequestsAndLimits(0xc000210fa0, 0x157c, 0x6)
	/go/src/sigs.k8s.io/descheduler/pkg/utils/pod.go:142 +0x5dd
sigs.k8s.io/descheduler/pkg/descheduler/strategies.nodeUtilization(0xc000546900, 0xc0000cee00, 0x27, 0x40, 0xc000546900)
	/go/src/sigs.k8s.io/descheduler/pkg/descheduler/strategies/lownodeutilization.go:441 +0x347
sigs.k8s.io/descheduler/pkg/descheduler/strategies.getNodeUsage(0x190a9a0, 0xc000138020, 0x192a6c0, 0xc000420f20, 0xc0004664c0, 0x4, 0x4, 0xc000716a20, 0xc0007169f0, 0x0, ...)
	/go/src/sigs.k8s.io/descheduler/pkg/descheduler/strategies/lownodeutilization.go:220 +0x174
sigs.k8s.io/descheduler/pkg/descheduler/strategies.LowNodeUtilization(0x190a9a0, 0xc000138020, 0x192a6c0, 0xc000420f20, 0xc000496c01, 0x0, 0xc000408de0, 0xc0004664c0, 0x4, 0x4, ...)
	/go/src/sigs.k8s.io/descheduler/pkg/descheduler/strategies/lownodeutilization.go:102 +0x229
sigs.k8s.io/descheduler/pkg/descheduler.RunDeschedulerStrategies.func1()
	/go/src/sigs.k8s.io/descheduler/pkg/descheduler/descheduler.go:123 +0x30a
k8s.io/apimachinery/pkg/util/wait.BackoffUntil.func1(0xc0002a5b10)
	/go/src/sigs.k8s.io/descheduler/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:155 +0x5f
k8s.io/apimachinery/pkg/util/wait.BackoffUntil(0xc0000dbb10, 0x18cf8e0, 0xc00052d140, 0x1, 0xc000516060)
	/go/src/sigs.k8s.io/descheduler/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:156 +0xad
k8s.io/apimachinery/pkg/util/wait.JitterUntil(0xc0002a5b10, 0x0, 0x0, 0x1, 0xc000516060)
	/go/src/sigs.k8s.io/descheduler/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:133 +0x98
k8s.io/apimachinery/pkg/util/wait.Until(...)
	/go/src/sigs.k8s.io/descheduler/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:90
sigs.k8s.io/descheduler/pkg/descheduler.RunDeschedulerStrategies(0x190a9a0, 0xc000138020, 0xc00038de00, 0xc00047f600, 0xc0000468a0, 0xe, 0xc000516060, 0x0, 0x0)
	/go/src/sigs.k8s.io/descheduler/pkg/descheduler/descheduler.go:98 +0x536
sigs.k8s.io/descheduler/pkg/descheduler.Run(0xc00038de00, 0x18d11c0, 0xc0004662e0)
	/go/src/sigs.k8s.io/descheduler/pkg/descheduler/descheduler.go:60 +0x196
sigs.k8s.io/descheduler/cmd/descheduler/app.Run(...)
	/go/src/sigs.k8s.io/descheduler/cmd/descheduler/app/server.go:67
sigs.k8s.io/descheduler/cmd/descheduler/app.NewDeschedulerCommand.func1(0xc0003a7b80, 0xc0003a51c0, 0x0, 0x4)
	/go/src/sigs.k8s.io/descheduler/cmd/descheduler/app/server.go:52 +0x87
github.com/spf13/cobra.(*Command).execute(0xc0003a7b80, 0xc000140010, 0x4, 0x4, 0xc0003a7b80, 0xc000140010)
	/go/src/sigs.k8s.io/descheduler/vendor/github.com/spf13/cobra/command.go:830 +0x2c2
github.com/spf13/cobra.(*Command).ExecuteC(0xc0003a7b80, 0xc0002a5f58, 0x1, 0x1)
	/go/src/sigs.k8s.io/descheduler/vendor/github.com/spf13/cobra/command.go:914 +0x30b
github.com/spf13/cobra.(*Command).Execute(...)
	/go/src/sigs.k8s.io/descheduler/vendor/github.com/spf13/cobra/command.go:864
main.main()
	/go/src/sigs.k8s.io/descheduler/cmd/descheduler/descheduler.go:34 +0xbd
panic: feature "PodOverhead" is not registered in FeatureGate "k8s.io/apiserver/pkg/util/feature/feature_gate.go:28" [recovered]
	panic: feature "PodOverhead" is not registered in FeatureGate "k8s.io/apiserver/pkg/util/feature/feature_gate.go:28"
```

The cluster is running with default feature gates.  
PodOverhead is enabeld by default.

### LINKED ISSUE #604: TopologySpreadConstraint should take target node's requests allocation into consideration before evicting

RyanDevlin:
Fixes #604
Fixes #529
Fixes #574
Fixes #579

## Main Features 
- Requests are now calculated and used to determine if a pod can be evicted when `NodeFit` == `True`
- All strategies consider the exact same predicates when determining if a pod can be evicted

## Description
This PR adds checks for resource requests when the [NodeFit](https://github.com/kubernetes-sigs/descheduler#node-fit-filtering) pod filtering flag is enabled. The `NodeFit` flag instructs the Descheduler to determine if there is a node in the cluster the pod might fit on before evicting the pod. In the past this check did not consider the CPU, memory, storage, and other `requests` on these nodes. It now considers any and all `requests` made by the pod and the nodes, including custom resources.

Originally I only intended to add checks for resource `requests` to the `TopologySpread` strategy. As I did this, I ended up modifying global functions in `pkg/descheduler/node/node.go` which control eviction filtering logic. This has the side effect of aligning all strategies that can enable the `NodeFit` flag with my original feature. So even though I didn't intend to fix #574 and #529, they were automatically implemented by fixing #604.

I would have broken these out into two PRs (as suggested by @damemi), but once I finished #604 I realized all that was needed for #529 and #574 were a few extra test cases.

I'd love some feedback on my implementation, especially regarding the changes I've made to the `TopologySpread` strategy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -707,6 +707,7 @@ The following strategies accept a `nodeFit` boolean parameter which can optimize
 - A `nodeSelector` on the pod
 - Any `Tolerations` on the pod and any `Taints` on the other nodes
 - `nodeAffinity` on the pod
+- Resource `Requests` made by the pod and the resources available on other nodes 
 - Whether any of the other nodes are marked as `unschedulable`
 
 E.g.
--- a/pkg/descheduler/node/node.go
+++ b/pkg/descheduler/node/node.go
@@ -18,13 +18,16 @@ package node
 
 import (
 	"context"
+	"fmt"
 
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	coreinformers "k8s.io/client-go/informers/core/v1"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/klog/v2"
+	podutil "sigs.k8s.io/descheduler/pkg/descheduler/pod"
 	"sigs.k8s.io/descheduler/pkg/utils"
 )
 
@@ -96,32 +99,91 @@ func IsReady(node *v1.Node) bool {
 	return true
 }
 
-// PodFitsAnyOtherNode checks if the given pod fits any of the given nodes, besides the node
-// the pod is already running on. The node fit is based on multiple criteria, like, pod node selector
-// matching the node label (including affinity), the taints on the node, and the node being schedulable or not.
-func PodFitsAnyOtherNode(pod *v1.Pod, nodes []*v1.Node) bool {
+// NodeFit returns true if the provided pod can probably be scheduled onto the provided node.
+// This function is used when the NodeFit pod filtering feature of the Descheduler is enabled.
+// This function currently considers a subset of the Kubernetes Scheduler's predicates when
+// deciding if a pod would fit on a node, but more predicates may be added in the future.
+func NodeFit(nodeIndexer podutil.GetPodsAssignedToNodeFunc, pod *v1.Pod, node *v1.Node) []error {
+	// Check node selector and required affinity
+	var errors []error
+	ok, err := utils.PodMatchNodeSelector(pod, node)
+	if err != nil {
+		errors = append(errors, err)
+	} else if !ok {
+		errors = append(errors, fmt.Errorf("pod node selector does not match the node label"))
+	}
+	// Check taints (we only care about NoSchedule and NoExecute taints)
+	ok = utils.TolerationsTolerateTaintsWithFilter(pod.Spec.Tolerations, node.Spec.Taints, func(taint *v1.Taint) bool {
+		return taint.Effect == v1.TaintEffectNoSchedule || taint.Effect == v1.TaintEffectNoExecute
+	})
+	if !ok {
+		errors = append(errors, fmt.Errorf("pod does not tolerate taints on the node"))
+	}
+	// Check if the pod can fit on a node based off it's requests
+	ok, reqErrors := FitsRequest(nodeIndexer, pod, node)
+	if !ok {
+		errors = append(errors, reqErrors...)
+	}
+	// Check if node is schedulable
+	if IsNodeUnschedulable(node) {
+		errors = append(errors, fmt.Errorf("node is not schedulable"))
+	}
 
+	return errors
+}
+
+// PodFitsAnyOtherNode checks if the given pod will probably fit any of the given nodes, besides the node
+// the pod is already running on. The predicates used to determine if the pod will fit can be found in the NodeFit function.
+func PodFitsAnyOtherNode(nodeIndexer podutil.GetPodsAssignedToNodeFunc, pod *v1.Pod, nodes []*v1.Node) bool {
 	for _, node := range nodes {
 		// Skip node pod is already on
 		if node.Name == pod.Spec.NodeName {
 			continue
 		}
-		// Check node selector and required affinity
-		ok, err := utils.PodMatchNodeSelector(pod, node)
-		if err != nil || !ok {
-			continue
-		}
-		// Check taints (we only care about NoSchedule and NoExecute taints)
-		ok = utils.TolerationsTolerateTaintsWithFilter(pod.Spec.Tolerations, node.Spec.Taints, func(taint *v1.Taint) bool {
-			return taint.Effect == v1.TaintEffectNoSchedule || taint.Effect == v1.TaintEffectNoExecute
-		})
-		if !ok {
-			continue
-		}
-		// Check if node is schedulable
-		if !IsNodeUnschedulable(node) {
-			klog.V(2).InfoS("Pod can possibly be scheduled on a different node", "pod", klog.KObj(pod), "node", klog.KObj(node))
+
+		errors := NodeFit(nodeIndexer, pod, node)
+		if len(errors) == 0 {
+			klog.V(4).InfoS("Pod fits on node", "pod", klog.KObj(pod), "node", klog.KObj(node))
 			return true
+		} else {
+			klog.V(4).InfoS("Pod does not fit on node", "pod", klog.KObj(pod), "node", klog.KObj(node))
+			for _, err := range errors {
+				klog.V(4).InfoS(err.Error())
+			}
+		}
+	}
+	return false
+}
+
+// PodFitsAnyNode checks if the given pod will probably fit any of the given nodes. The predicates used
+// to determine if the pod will fit can be found in the NodeFit function.
+func PodFitsAnyNode(nodeIndexer podutil.GetPodsAssignedToNodeFunc, pod *v1.Pod, nodes []*v1.Node) bool {
+	for _, node := range nodes {
+		errors := NodeFit(nodeIndexer, pod, node)
+		if len(errors) == 0 {
+			klog.V(4).InfoS("Pod fits on node", "pod", klog.KObj(pod), "node", klog.KObj(node))
+			return true
+		} else {
+			klog.V(4).InfoS("Pod does not fit on node", "pod", klog.KObj(pod), "node", klog.KObj(node))
+			for _, err := range errors {
+				klog.V(4).InfoS(err.Error())
+			}
+		}
+	}
+	return false
+}
+
+// PodFitsCurrentNode checks if the given pod will probably fit onto the given node. The predicates used
+// to determine if the pod will fit can be found in the NodeFit function.
+func PodFitsCurrentNode(nodeIndexer podutil.GetPodsAssignedToNodeFunc, pod *v1.Pod, node *v1.Node) bool {
+	errors := NodeFit(nodeIndexer, pod, node)
+	if len(errors) == 0 {
+		klog.V(4).InfoS("Pod fits on node", "pod", klog.KObj(pod), "node", klog.KObj(node))
+		return true
+	} else {
+		klog.V(4).InfoS("Pod does not fit on node", "pod", klog.KObj(pod), "node", klog.KObj(node))
+		for _, err := range errors {
+			klog.V(4).InfoS(err.Error())
 		}
 	}
 	return false
@@ -133,39 +195,102 @@ func IsNodeUnschedulable(node *v1.Node) bool {
 	return node.Spec.Unschedulable
 }
 
-// PodFitsAnyNode checks if the given pod fits any of the given nodes, based on
-// multiple criteria, like, pod node selector matching the node label, node
-// being schedulable or not.
-func PodFitsAnyNode(pod *v1.Pod, nodes []*v1.Node) bool {
-	for _, node := range nodes {
+// FitsRequest determines if a pod can fit on a node based on that pod's requests. It returns true if
+// the pod will fit.
+func FitsRequest(nodeIndexer podutil.GetPodsAssignedToNodeFunc, pod *v1.Pod, node *v1.Node) (bool, []error) {
+	var insufficientResources []error
 
-		ok, err := utils.PodMatchNodeSelector(pod, node)
-		if err != nil || !ok {
-			continue
+	// Get pod requests
+	podRequests, _ := utils.PodRequestsAndLimits(pod)
+	resourceNames := make([]v1.ResourceName, 0, len(podRequests))
+	for name := range podRequests {
+		resourceNames = append(resourceNames, name)
+	}
+
+	availableResources, err := NodeAvailableResources(nodeIndexer, node, resourceNames)
+	if err != nil {
+		return false, []error{err}
+	}
+
+	podFitsOnNode := true
+	for _, resource := range resourceNames {
+		podResourceRequest := podRequests[resource]
+		var requestTooLarge bool
+		switch resource {
+		case v1.ResourceCPU:
+			requestTooLarge = podResourceRequest.MilliValue() > availableResources[resource].MilliValue()
+		default:
+			requestTooLarge = podResourceRequest.Value() > availableResources[resource].Value()
 		}
-		if !IsNodeUnschedulable(node) {
-			klog.V(2).InfoS("Pod can possibly be scheduled on a different node", "pod", klog.KObj(pod), "node", klog.KObj(node))
-			return true
+
+		if requestTooLarge {
+			insufficientResources = append(insufficientResources, fmt.Errorf("insufficient %v", resource))
+			podFitsOnNode = false
 		}
 	}
-	return false
+	return podFitsOnNode, insufficientResources
 }
 
-// PodFitsCurrentNode checks if the given pod fits on the given node if the pod
-// node selector matches the node label.
-func PodFitsCurrentNode(pod *v1.Pod, node *v1.Node) bool {
-	ok, err := utils.PodMatchNodeSelector(pod, node)
-
+// NodeAvailableResources returns resources mapped to the quanitity available on the node.
+func NodeAvailableResources(nodeIndexer podutil.GetPodsAssignedToNodeFunc, node *v1.Node, resourceNames []v1.ResourceName) (map[v1.ResourceName]*resource.Quantity, error) {
+	podsOnNode, err := podutil.ListPodsOnANode(node.Name, nodeIndexer, nil)
 	if err != nil {
-		klog.ErrorS(err, "Failed to match node selector")
-		return false
+		return nil, err
+	}
+	aggregatePodRequests := AggregatePodRequests(podsOnNode, resourceNames)
+	return nodeRemainingResources(node, aggregatePodRequests, resourceNames), nil
+}
+
+// AggregatePodRequests returns the resources requested by the given pods. Only resources supplied in the resourceNames parameter are calculated.
+func AggregatePodRequests(pods []*v1.Pod, resourceNames []v1.ResourceName) map[v1.ResourceName]*resource.Quantity {
+	totalReqs := map[v1.ResourceName]*resource.Quantity{
+		v1.ResourceCPU:    resource.NewMilliQuantity(0, resource.DecimalSI),
+		v1.ResourceMemory: resource.NewQuantity(0, resource.BinarySI),
+		v1.ResourcePods:   resource.NewQuantity(int64(len(pods)), resource.DecimalSI),
+	}
+	for _, name := range resourceNames {
+		if !IsBasicResource(name) {
+			totalReqs[name] = resource.NewQuantity(0, resource.DecimalSI)
+		}
 	}
 
-	if !ok {
-		klog.V(2).InfoS("Pod does not fit on node", "pod", klog.KObj(pod), "node", klog.KObj(node))
-		return false
+	for _, pod := range pods {
+		req, _ := utils.PodRequestsAndLimits(pod)
+		for _, name := range resourceNames {
+			quantity, ok := req[name]
+			if ok && name != v1.ResourcePods {
+				// As Quantity.Add says: Add adds the provided y quantity to the current value. If the current value is zero,
+				// the format of the quantity will be updated to the format of y.
+				totalReqs[name].Add(quantity)
+			}
+		}
 	}
 
-	klog.V(2).InfoS("Pod fits on node", "pod", klog.KObj(pod), "node", klog.KObj(node))
-	return true
+	return totalReqs
+}
+
+func nodeRemainingResources(node *v1.Node, aggregatePodRequests map[v1.ResourceName]*resource.Quantity, resourceNames []v1.ResourceName) map[v1.ResourceName]*resource.Quantity {
+	remainingResources := map[v1.ResourceName]*resource.Quantity{
+		v1.ResourceCPU:    resource.NewMilliQuantity(node.Status.Allocatable.Cpu().MilliValue()-aggregatePodRequests[v1.ResourceCPU].MilliValue(), resource.DecimalSI),
+		v1.ResourceMemory: resource.NewQuantity(node.Status.Allocatable.Memory().Value()-aggregatePodRequests[v1.ResourceMemory].Value(), resource.BinarySI),
+		v1.ResourcePods:   resource.NewQuantity(node.Status.Allocatable.Pods().Value()-aggregatePodRequests[v1.ResourcePods].Value(), resource.DecimalSI),
+	}
+	for _, name := range resourceNames {
+		if !IsBasicResource(name) {
+			allocatableResource := node.Status.Allocatable[name]
+			remainingResources[name] = resource.NewQuantity(allocatableResource.Value()-aggregatePodRequests[name].Value(), resource.DecimalSI)
+		}
+	}
+
+	return remainingResources
+}
+
+// IsBasicResource checks if resource is basic native.
+func IsBasicResource(name v1.ResourceName) bool {
+	switch name {
+	case v1.ResourceCPU, v1.ResourceMemory, v1.ResourcePods:
+		return true
+	default:
+		return false
+	}
 }
--- a/pkg/descheduler/node/node_test.go
+++ b/pkg/descheduler/node/node_test.go
@@ -21,9 +21,13 @@ import (
 	"testing"
 
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes/fake"
+	core "k8s.io/client-go/testing"
+	podutil "sigs.k8s.io/descheduler/pkg/descheduler/pod"
 	"sigs.k8s.io/descheduler/test"
 )
 
@@ -147,13 +151,13 @@ func TestPodFitsCurrentNode(t *testing.T) {
 					},
 				},
 			},
-			node: &v1.Node{
-				ObjectMeta: metav1.ObjectMeta{
-					Labels: map[string]string{
-						nodeLabelKey: nodeLabelValue,
-					},
-				},
-			},
+			node: test.BuildTestNode("node1", 64000, 128*1000*1000*1000, 200, func(node *v1.Node) {
+				node.ObjectMeta.Labels = map[string]string{
+					nodeLabelKey: nodeLabelValue,
+				}
+
+				node.Status.Allocatable[v1.ResourceEphemeralStorage] = *resource.NewQuantity(1000*1000*1000*1000, resource.DecimalSI)
+			}),
 			success: true,
 		},
 		{
@@ -181,27 +185,53 @@ func TestPodFitsCurrentNode(t *testing.T) {
 					},
 				},
 			},
-			node: &v1.Node{
-				ObjectMeta: metav1.ObjectMeta{
-					Labels: map[string]string{
-						nodeLabelKey: "no",
-					},
-				},
-			},
+			node: test.BuildTestNode("node1", 64000, 128*1000*1000*1000, 200, func(node *v1.Node) {
+				node.ObjectMeta.Labels = map[string]string{
+					nodeLabelKey: "no",
+				}
+
+				node.Status.Allocatable[v1.ResourceEphemeralStorage] = *resource.NewQuantity(1000*1000*1000*1000, resource.DecimalSI)
+			}),
 			success: false,
 		},
 	}
 
+	fakeClient := &fake.Clientset{}
+	fakeClient.Fake.AddReactor("list", "pods", func(action core.Action) (bool, runtime.Object, error) {
+		return true, &v1.PodList{Items: nil}, nil
+	})
+
 	for _, tc := range tests {
-		actual := PodFitsCurrentNode(tc.pod, tc.node)
-		if actual != tc.success {
-			t.Errorf("Test %#v failed", tc.description)
-		}
+		t.Run(tc.description, func(t *testing.T) {
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+
+			var objs []runtime.Object
+			objs = append(objs, tc.node)
+			objs = append(objs, tc.pod)
+
+			fakeClient := fake.NewSimpleClientset(objs...)
+
+			sharedInformerFactory := informers.NewSharedInformerFactory(fakeClient, 0)
+			podInformer := sharedInformerFactory.Core().V1().Pods()
+
+			getPodsAssignedToNode, err := podutil.BuildGetPodsAssignedToNodeFunc(podInformer)
+			if err != nil {
+				t.Errorf("Build get pods assigned to node function error: %v", err)
+			}
+
+			sharedInformerFactory.Start(ctx.Done())
+			sharedInformerFactory.WaitForCacheSync(ctx.Done())
+
+			actual := PodFitsCurrentNode(getPodsAssignedToNode, tc.pod, tc.node)
+			if actual != tc.success {
+				t.Errorf("Test %#v failed", tc.description)
+			}
+		})
 	}
 }
 
 func TestPodFitsAnyOtherNode(t *testing.T) {
-
 	nodeLabelKey := "kubernetes.io/desiredNode"
 	nodeLabelValue := "yes"
 	nodeTaintKey := "hardware"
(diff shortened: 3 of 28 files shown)


damemi (commented):
/ok-to-test
also, this would be a good PR to break into a couple of commits. At least the vendor changes could be split out, but if it's possible to put the refactors in one too that would help review (the ones that we mentioned could be a second PR)

RyanDevlin:
@damemi Is there a good way to undo git commit squashing?

RyanDevlin:
@damemi I tried pretty hard to undo my squashing and separate it into two commits, but unfortunately I ended up mangling it so badly I had to revert back.

damemi:
@RyanDevlin it's not so important in this PR, but for the future, have you tried these commands?
```
$ git reset HEAD~1
$ git add pkg/ test/ README.md
$ git commit -m "<code changes commit message>"
$ git add go.sum go.mod vendor/
$ git commit -m "Vendor updates"
```

Also, you have `kubectl` committed in here, so that needs to be removed and maybe .gitignore updated

RyanDevlin:
@damemi I actually did do that, but I have so many commit's and revisions in my reflog now that it's difficult to make sense of it all. I originally thought that all PRs must only have one commit, but now I realize I can group major revisions together into 2-3 commits. I'll be sure to do that in the future.

Also with regards to the `kubectl` commit, that's an amateur mistake sorry, removing it now....

RyanDevlin:
@damemi Finally got it! Thank you

pkg/utils/pod.go
```diff
@@ -53,11 +37,6 @@ func GetResourceRequestQuantity(pod *v1.Pod, resourceName v1.ResourceName) resou
 		requestQuantity = resource.Quantity{Format: resource.DecimalSI}
 	}
 
-	if resourceName == v1.ResourceEphemeralStorage && !utilfeature.DefaultFeatureGate.Enabled(LocalStorageCapacityIsolation) {
-		// if the local storage capacity isolation feature gate is disabled, pods request 0 disk
-		return requestQuantity
-	}
-
```
RyanDevlin, line 60: If you're wondering why I removed this, it's because it won't work so long as the Descheduler is separate from core Kubernetes. My rationale is below, but **TL;DR**: 
``The features referenced here (PodOverhead and LocalStorageCapacityIsolation) are only updated and tracked by the k8s.io/kubernetes/pkg/features package, which is not importable.``

**Longer explanation:**
The code I added in this PR originally used the same featuregate checking that's performed here, because I also needed to know the state of `PodOverhead` and `LocalStorageCapacityIsolation`. What I discovered through testing was that both of these features were not registered with the `DefaultFeatureGate`. When you run `utilfeature.DefaultFeatureGate.Enabled()` with a feature that hasn't been registered with the FeatureGate, it panics. What's also interesting is that line 56 and line 162 in `pod.go` are never executed in our test cases. This is what caused us to miss the bug in #579. Notice from that issue, we see:

```
Observed a panic: {s:"feature \"PodOverhead\" is not registered in FeatureGate \"k8s.io/apiserver/pkg/util/feature/feature_gate.go:28\""}
.
.
/go/src/sigs.k8s.io/descheduler/pkg/utils/pod.go:142
```
This specific message references line 142, but in reality this is now line 162 of `pod.go` on the master branch.

Regardless, in my tests I was seeing this same panic message. The problem is due to the fact that these features are never registered with the featuregate. After a lot of digging I found that the features are actually registered in the core [Kubernetes Features Package](https://pkg.go.dev/k8s.io/kubernetes/pkg/features). This package is not importable due to the fact that it lives in `k8s.io/kubernetes/*` (if you try to `go get` it you'll receive errors. Reading up on these errors shows that importing packages directly from core Kubernetes is not supported).


So this left me with two options. Firstly, we could just copy everything in the core [Kubernetes Features Package](https://pkg.go.dev/k8s.io/kubernetes/pkg/features) locally to the Descheduler. This way all features would be registered with the `DefaultFeatureGate`. The obvious downsides to this are:

1.) If someone uses anything but the defaults when executing the Kubelet, the Descheduler would still be unaware.

2.) These features are updated and removed regularly, and we would have to continuously track the state of that file.

So with all this in mind, I decided to remove all featuregate checks from both my new code, as well as the existing code. I've setup each section to operate as if the default value for that feature was set. For instance, if this is in the Descheduler now:

```
if utilfeature.DefaultFeatureGate.Enabled(PodOverhead) {
    // do stuff
}
```
and this is in `k8s.io/kubernetes/pkg/features/kube_features.go`:
```
PodOverhead:     {Default: true, PreRelease: featuregate.Beta},
```
Then the Deschduler code is now simply:
```
// do stuff
```

Assuming default settings is appropriate, in my opinion, because we won't ever be able to actually get these values, and the Descheduler is a best-effort mechanism. If some of our calculations for pod resources are slightly off because someone has disabled `podoverhead` checking in their cluster, the worst that will happen is we may evict a pod that will be scheduled to the same node.

damemi:
> I originally thought that all PRs must only have one commit, but now I realize I can group major revisions together into 2-3 commits

This is a bit more art than science, and there isn't a hard rule. But generally, vendor changes can always be their own commit, at least during reviews to make it easier.

RyanDevlin:
@damemi In the `TopologySpread` strategy, there is a check to consider if a pod fits on another node, [shown here](https://github.com/kubernetes-sigs/descheduler/blob/5b557941fac60cd85c1a3b709eb49a08d648fd8d/pkg/descheduler/strategies/topologyspreadconstraint.go#L276). I'd like to replace this with a call to `isEvictable()`. Doing this would then only check if a pod fits on another node if `NodeFit` is turned on.

Would that change allow the `TopologySpread` strategy to still function as desired when `NodeFit` is disabled?

pkg/descheduler/node/node.go
```diff
@@ -18,13 +18,18 @@ package node
...
 import (
 	"context"
+	"fmt"
+	"strings"
 
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/validation"
+	_ "k8s.io/apiserver/pkg/features"
```
eero-t, line 28: '_' as package alias?

RyanDevlin: Line 28 should have been removed, it's gone now.

SerenaTiede-Zen:
Hey I ran your branch of Descheduler and I don't think it handles #604. For example I spun up a kind cluster 
locally with size `[2 3 4]`, `maxSkew=1`, and `whenUnsatisfiable=ScheduleAnyways`. In this case node 1 is low on 
available resourcesDescheduler will evict a pod from node 3  and it will be scheduled onto node 2 (size is now 
`[2 4 3]`). The next time the cronjob runs Descheduler will evict a pod from node 2 and it will be scheduled onto node 3
(size is now `[2 3 4]`). 

Ideally in this resource constrained scenario Descheduler would not evict any pods for violating the 
TopologySpreadConstraint.

**Version of descheduler**

using @RyanDevlin 's fork at branch nodeFit-v2 at commit hash `91f37518dbe9fd6bf009f238f561c6f4222f05c1`

**Steps to reproduce**

I made up this very hacky shell script that will bring up a deployment named `nginx` with the above sizing. The 
deployment `resource-user` is to fill up the resources on node `descheduler-worker` to trigger this condition of evicted
pods moving between `descheduler-worker2` and `descheduler-worker3`

Let me know if you need more info

```shell
#!/usr/bin/env sh

set -o errexit

# create a cluster with the local registry enabled in containerd
cat <<EOF | kind create cluster --config=-
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
name: descheduler
nodes:
- role: control-plane
- role: worker
  kubeadmConfigPatches:
  - |
    kind: JoinConfiguration
    nodeRegistration:
      kubeletExtraArgs:
        node-labels: "topology.kubernetes.io/zone=local-a,ownedby=foo"
      taints:
      - effect: NoExecute
        key: ownedby
        value: foo
- role: worker 
  kubeadmConfigPatches:
  - |
    kind: JoinConfiguration
    nodeRegistration:
      kubeletExtraArgs:
        node-labels: "topology.kubernetes.io/zone=local-b,ownedby=foo"
      taints:
      - effect: NoExecute
        key: ownedby
        value: foo
- role: worker 
  kubeadmConfigPatches:
  - |
    kind: JoinConfiguration
    nodeRegistration:
      kubeletExtraArgs:
        node-labels: "topology.kubernetes.io/zone=local-c,ownedby=foo"
      taints:
      - effect: NoExecute
        key: ownedby
        value: foo
- role: worker 
  kubeadmConfigPatches:
  - |
    kind: JoinConfiguration
    nodeRegistration:
      kubeletExtraArgs:
        node-labels: "topology.kubernetes.io/zone=local-a"
EOF

# Create nginx-test namespace:
cat <<EOF | kubectl apply -f -
apiVersion: v1
kind: Namespace
metadata:
  name: nginx-test
EOF

# Create a Deployment with topology constraint for "topology.kubernetes.io/zone"
cat <<EOF | kubectl apply -f -
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx
  namespace: nginx-test
  labels:
    project: nginx
    role: nginx
spec:
  replicas: 0
  selector:
    matchLabels:
      project: nginx
      role: nginx
  template:
    metadata:
      labels:
        project: nginx
        role: nginx
    spec:
      containers:
      - name: nginx
        image: k8s.gcr.io/nginx-slim:0.8
        ports:
        - containerPort: 80
          name: http
        resources:
          requests:
            memory: 100Mi
            cpu: 350m
        volumeMounts:
        - name: config-volume
          mountPath: /var/config-fetcher
      topologySpreadConstraints:
      - maxSkew: 1
        topologyKey: "topology.kubernetes.io/zone"
        whenUnsatisfiable: "ScheduleAnyway"
        labelSelector:
          matchLabels:
            project: nginx
            role: nginx
      nodeSelector:
        ownedby: foo
      tolerations:
        - effect: NoExecute
          key: ownedby
          operator: Equal
          value: foo
      volumes:
        - name: config-volume
          emptyDir: {}
EOF

cat <<EOF | kubectl apply -f -
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app: busybox
  name: resource-user
  namespace: default
spec:
  progressDeadlineSeconds: 600
  replicas: 10
  revisionHistoryLimit: 10
  selector:
    matchLabels:
      app: busybox
  strategy:
    rollingUpdate:
      maxSurge: 25%
      maxUnavailable: 25%
    type: RollingUpdate
  template:
    metadata:
      creationTimestamp: null
      labels:
        app: busybox
    spec:
      containers:
      - command:
        - sleep
        - 1000s
        image: busybox
        imagePullPolicy: Always
        name: busybox
        resources:
          requests:
            memory: 256Mi
            cpu: 350m
        terminationMessagePath: /dev/termination-log
        terminationMessagePolicy: File
      dnsPolicy: ClusterFirst
      nodeSelector:
        ownedby: foo
        topology.kubernetes.io/zone: local-a
      restartPolicy: Always
      schedulerName: default-scheduler
      securityContext: {}
      terminationGracePeriodSeconds: 30
      tolerations:
      - effect: NoExecute
        key: ownedby
        operator: Equal
        value: foo
EOF

cat <<EOF | kubectl apply -f -
apiVersion: v1
kind: ConfigMap
metadata:
  name: descheduler-policy-configmap
  namespace: kube-system
data:
  policy.yaml: |
    kind: "DeschedulerPolicy"
    evictLocalStoragePods: true
    strategies:
      "RemovePodsViolatingTopologySpreadConstraint":
        enabled: true
        params:
          includeSoftConstraints: true
          nodeFit: true
EOF

kubectl scale deployment resource-user -n default --replicas=5 --timeout=30s

kubectl wait --for=condition=Ready pod -l app=busybox

sleep 5

kubectl cordon descheduler-worker
kubectl cordon descheduler-worker2

sleep 5

kubectl scale deployment nginx -n nginx-test --replicas=4 --timeout=30s

kubectl wait --for=condition=Ready pod -n nginx-test -l project=nginx

sleep 5

kubectl cordon descheduler-worker3
kubectl uncordon descheduler-worker2

sleep 5

kubectl scale deployment nginx -n nginx-test --replicas=7 --timeout=30s

kubectl wait --for=condition=Ready pod -n nginx-test -l project=nginx

sleep 5

kubectl cordon descheduler-worker2
kubectl uncordon descheduler-worker

sleep 5

kubectl scale deployment nginx -n nginx-test --replicas=9 --timeout=30s

kubectl wait --for=condition=Ready pod -n nginx-test -l project=nginx

kubectl uncordon descheduler-worker
kubectl uncordon descheduler-worker2
kubectl uncordon descheduler-worker3

sleep 5

if [ ! -d "./descheduler" ]; then
  git clone https://github.com/RyanDevlin/descheduler.git
fi
cd descheduler
git checkout nodeFit-v2
cd ..

local_image_name='local-descheduler:1.0.0'
docker build --build-arg=VERSION=1.0 --build-arg=ARCH=amd64 -t $local_image_name ./descheduler
kind load docker-image --name descheduler $local_image_name
kubectl create -f https://raw.githubusercontent.com/kubernetes-sigs/descheduler/v0.22.1/kubernetes/base/rbac.yaml
manifest=$(curl https://raw.githubusercontent.com/kubernetes-sigs/descheduler/v0.22.1/kubernetes/cronjob/cronjob.yaml)
echo "$manifest" | sed "s/k8s.gcr.io\\/descheduler\\/descheduler:v0.22.1/$local_image_name/g" |kubectl create -f - -o yaml
```

RyanDevlin:
@SerenaTiede-Zen  Thank you for the high level of detail in your comment! There's clearly something awry, though I'm not exactly sure what the cause is. Give me a few days to reproduce this myself and unravel the root cause. If I have any specific questions I'll definitely ping you in this thread. Thanks again for all the details!

a7i:
This unit test in [topologyspreadconstraint_test.go](https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/descheduler/strategies/topologyspreadconstraint_test.go) should help you reproduce as well:

```go
{
	name: "3 domains, sizes [2,3,4], maxSkew=1, not enough cpu on zoneA; nothing should be moved",
	nodes: []*v1.Node{
		test.BuildTestNode("n1", 100, 2000, 10, func(n *v1.Node) { n.Labels["zone"] = "zoneA" }),
		test.BuildTestNode("n2", 1000, 2000, 10, func(n *v1.Node) { n.Labels["zone"] = "zoneB" }),
		test.BuildTestNode("n3", 1000, 2000, 10, func(n *v1.Node) { n.Labels["zone"] = "zoneC" }),
	},
	pods: createTestPods([]testPodList{
		{
			count:       2,
			node:        "n1",
			labels:      map[string]string{"foo": "bar"},
			constraints: getDefaultTopologyConstraints(1),
		},
		{
			count:       3,
			node:        "n2",
			labels:      map[string]string{"foo": "bar"},
			constraints: getDefaultTopologyConstraints(1),
		},
		{
			count:       4,
			node:        "n3",
			labels:      map[string]string{"foo": "bar"},
			constraints: getDefaultTopologyConstraints(1),
		},
	}),
	expectedEvictedCount: 0,
	strategy: api.DeschedulerStrategy{Params: &api.StrategyParameters{NodeFit: true}},
	namespaces: []string{"ns1"},
},
```

I believe this issue happens due to [`validatePodFitsOnOtherNodes` ](https://github.com/kubernetes-sigs/descheduler/pull/636/files#diff-3ef994f616f4d3726bc08b98eb341fee83b26703bc6179277f5e91286d99b289L276-L280) being removed in the `topologyspreadconstraint` strategy when balancing the topology domains which is throwing off the topology size calculation. Perhaps it can call the new `PodFitsAnyOtherNode` to preserve the topology size calculation logic?

 It is unfortunate but we need to do this check twice: 1) when calculating the topology size 2) in Evictable

RyanDevlin:
@a7i I see, thanks for this! That last piece I couldn't figure out, but I see it now. I originally removed the `validatePodFitsOnOtherNodes` function to avoid running these calculations twice, but I didn't realize the edge case side effect that would have here.

There's definitely an optimization that could be made so we only do this once. Perhaps we can mark each pod as evictable or not as we perform the topology calculations. Regardless, it's likely this optimization would require too many changes for this PR.

So for now, I'll add back that function, have it call `PodFitsAnyOtherNode`, and open an issue to optimize this strategy in the future.

a7i:
> I didn't realize the edge case side effect that would have here.

We should definitely have unit tests to cover this edge case. This is not the first time we've seen this, see [here](https://github.com/kubernetes-sigs/descheduler/pull/639#discussion_r720249781).

On another note, I would like to request for `node.NodeFit` to return aggregated errors instead of a boolean. This allows the caller to log all failing "fit" conditions
```
klog.V(...).InfoS(fmt.Sprintf("ignoring pod for eviction due to: %s", err.Error()), "pod", klog.KObj(pod))
```
e.g. if a pod cannot fit due to affinity, taint, resource constraints. This can be very helpful in debugging issues.

pkg/descheduler/node/node.go
```diff
@@ -96,76 +100,400 @@ func IsReady(node *v1.Node) bool {
...
 		}
 	}
 	return false
 }
 
+// PodFitsCurrentNode checks if the given pod will probably fit onto the given node. The predicates used
+// to determine if the pod will fit can be found in the NodeFit function.
+func PodFitsCurrentNode(ctx context.Context, client clientset.Interface, pod *v1.Pod, node *v1.Node) bool {
+	if NodeFit(ctx, client, pod, node) {
+		klog.V(2).InfoS("Pod fits on node", "pod", klog.KObj(pod), "node", klog.KObj(node))
```
a7i, line 166: Can all "fits" logs be changed to v4? It's very noisy in our cluster of thousands of pods.

RyanDevlin: @a7i Great suggestions, I'm implementing these now. I'll report back here when everything is pushed.

RyanDevlin:
@damemi @a7i After digging into the behavior mentioned above, I think this problem needs to be moved into it's own issue. My reasoning is that it's out of scope for this PR to fix the topology spread behavior. Below are some details about my testing:

Test case that's still failing (NOTE: each pod requests `100 millicores` of CPU and `0 MB` of memory):
```
{
	name: "3 domains, sizes [2,3,4], maxSkew=1, NodeFit is enabled, and not enough cpu on zoneA; nothing should be moved",
	nodes: []*v1.Node{
		test.BuildTestNode("n1", 250, 2000, 10, func(n *v1.Node) { n.Labels["zone"] = "zoneA" }),
		test.BuildTestNode("n2", 1000, 2000, 10, func(n *v1.Node) { n.Labels["zone"] = "zoneB" }),
		test.BuildTestNode("n3", 1000, 2000, 10, func(n *v1.Node) { n.Labels["zone"] = "zoneC" }),
	},
	pods: createTestPods([]testPodList{
		{
			count:       2,
			node:        "n1",
			labels:      map[string]string{"foo": "bar"},
			constraints: getDefaultTopologyConstraints(1),
		},
		{
			count:       3,
			node:        "n2",
			labels:      map[string]string{"foo": "bar"},
			constraints: getDefaultTopologyConstraints(1),
		},
		{
			count:       4,
			node:        "n3",
			labels:      map[string]string{"foo": "bar"},
			constraints: getDefaultTopologyConstraints(1),
		},
	}),
	expectedEvictedCount: 0,
	strategy: api.DeschedulerStrategy{
		Params: &api.StrategyParameters{
			NodeFit: trdomains and fit.ue,
		},
	},
	namespaces: []string{"ns1"},
},
```

1.) I've modified the `balanceDomains()` function in the topology spread strategy so that line ~286 reads:
```
for k := range aboveToEvict {
                if !node.PodFitsAnyOtherNode(ctx, client, aboveToEvict[k], nodes) {
                        // unfinished Klog message printed here
                        continue
                }

                podsForEviction[aboveToEvict[k]] = struct{}{}
        }
```
I've tested the code with the old `validatePodFitsOnOtherNodes()` instead of `node.PodFitsAnyOtherNode()` and the behavior is exactly the same. It still fails the test case posted above.

2.) I've traced where the failure occurs during the test case, and it's in the logic of the topology strategy, not the new code in this PR. When the above test is run, the topology strategy will determine that `pod-8` on node `n3` should be evicted. When it calls `PodFitsAnyOtherNode()` in the `for` loop above, the new code correctly identifies that the pod will `not` fit on node `n1` because there is not enough CPU available. The new code then correctly reports back that the pod will fit on node `n2` because there is plenty of CPU available.

So it appears that the topology spread strategy is not taking into account the fact that scheduling the pod onto node `n2` will not improve/balance the topology of the pods. The point I'm trying to make here is that the code in this PR works as intended and correctly identifies whether a pod can be scheduled onto other nodes based off the pod's requests. I believe there needs  to be a separate effort to improve the topology spread strategy so that it correctly calls the `nodeFit` code and uses the information returned in a way that allows it to take these constraints into consideration when calculating where pods should move.

If it's okay with both of you, I'd like to remove that test case from my code (the test case is new and not already in the master branch). I'd like to then open a new issue, citing that test case, and asking for this functionality to be added to the Descheduler.

RyanDevlin:
On another note, I'll soon push code implementing the error handling and log level changes mentioned by @a7i. After that this PR should be ready for a final review.

RyanDevlin:
I've updated my code with the requested fixes. I also added a bugfix for the E2E tests because they were failing after I pulled down the current master branch. The bug was a race condition introduced recently. It was caused by the first test listing the nodes in a newly created `kind` cluster before the nodes were ready. This caused the first two tests to fail because these nodes were tainted with `node.kubernetes.io/not-ready`.

I can explain the fix in more detail if desired; it's a simple wait function added to `e2e_test.go`.

RyanDevlin:
@ingvagabund Do you or anyone else have time to review this PR? I will gladly walk someone through my changes if needed!

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -272,8 +271,7 @@ func balanceDomains(
 		// also (just for tracking), add them to the list of pods in the lower topology
 		aboveToEvict := sortedDomains[j].pods[len(sortedDomains[j].pods)-movePods:]
 		for k := range aboveToEvict {
-			if err := validatePodFitsOnOtherNodes(aboveToEvict[k], nodeMap); err != nil {
-				klog.V(2).InfoS(fmt.Sprintf("ignoring pod for eviction due to: %s", err.Error()), "pod", klog.KObj(aboveToEvict[k]))
+			if !node.PodFitsAnyOtherNode(ctx, client, aboveToEvict[k], nodes) {
 				continue
```
a7i, line 275:
```suggestion
				klog.V(2).InfoS("ignoring pod for eviction as it does not fit on any other node", "pod", klog.KObj(aboveToEvict[k]))
				continue
```

ingvagabund:
@RyanDevlin on my todo list. Slowly going through open PRs and issues.

pkg/descheduler/strategies/topologyspreadconstraint.go
RyanDevlin: @a7i Good point. I removed that because `PodFitsAnyOtherNode` logs these messages. I realize now that we should also log here at a lower level than in `PodFitsAnyOtherNode` in order to log what the strategy is doing.

RyanDevlin:
@ingvagabund Thank you. Let me know if you have any questions.

For some reason I have two checks failing, even though they don't fail on my local machine. The compiler in the pipeline thinks I have too many arguments to a function, but I've checked the commit that was uploaded and that's not the case. I'm not sure how to fix this since there's no error from my perspective?

damemi:
@RyanDevlin the `buildTestPod` signature was changed in https://github.com/kubernetes-sigs/descheduler/commit/2020642b6fe7b505732692783c1b366a001ff8c8, which it doesn't look like your branch has. Try rebasing your PR onto the current master branch and you should see the error

RyanDevlin:
@damemi I thought I had done that, but I guess I missed a recent update. It's fixed now, thanks!

ingvagabund (commented):
Observations:
- copy pasting code is not a good practice
- we already consider cpu, memory and pods as distinguished resources, everything else as extended (no need to assume existence of scalar resources or the ephemeral storage)
- insufficiency of a resource can be calculated generically:
  ```
  for _, resource := range resources {
    if podRequests[resource] > nodeAllocatable[resource] {
      insufficientResources = append(insufficientResources, fmt.Sprintf("Insufficient %v", resource))
    }
  }
  ```
- the descheduler already has a code operating over resources (https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/descheduler/strategies/nodeutilization/nodeutilization.go#L367) which can be reused/extended (e.g. when computing resources left on a node)

I wonder if it would be possible to reduce the complexity of the implementation by:
- replacing Resource and NodeInfo data type with a map of resources
- and reusing/extending code in pkg/descheduler/strategies/nodeutilization/nodeutilization.go

E.g.:
```
func FitsRequest(nodeIndexer podutil.GetPodsAssignedToNodeFunc, pod *v1.Pod, node *v1.Node) []error {
  var podRequests map[resource]Quantity
  var nodeAllocatable map[resource]Quantity
  var insufficientResources []error
  podRequests = getPodRequests(pod)
  nodeAllocatable = getNodeUtilization(nodeIndexer, node)
  resources = keys of podRequest and nodeAllocatable
  for _, resource := range resources {
    if podRequests[resource] > nodeAllocatable[resource] {
      insufficientResources = append(insufficientResources, fmt.Errorf("Insufficient %v", resource))
    }
  }
  return insufficientResources
}
```

RyanDevlin:
@ingvagabund Thank you for your review! A few points:

1.) I didn't intend to simply copy and paste, I tried to incorporate what was already used by the scheduler while cutting anything outside the scope of this PR. I originally thought it would be good to keep the Descheduler somewhat in line with how the scheduler calculates resources, but maybe I went about that wrong.

2.) You make some excellent points about the resource calculations in `nodeutilization.go` that I should have realized earlier. Let me refactor some of this today and I'll make the changes you mentioned.

test/e2e/e2e_test.go
```diff
@@ -128,9 +127,32 @@ func initializeClient(t *testing.T) (clientset.Interface, coreinformers.NodeInfo
 	sharedInformerFactory.Start(stopChannel)
 	sharedInformerFactory.WaitForCacheSync(stopChannel)
 
+	waitForNodesReady(context.Background(), t, clientSet, nodeInformer)
 	return clientSet, nodeInformer, getPodsAssignedToNode, stopChannel
 }
 
+func waitForNodesReady(ctx context.Context, t *testing.T, clientSet clientset.Interface, nodeInformer coreinformers.NodeInformer) {
```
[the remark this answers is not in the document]
RyanDevlin: I can absolutely put this into another PR. I originally put it here because my code would not pass the `e2e tests` without first eliminating this race condition. I'll try again without it, maybe it will work fine on the VMs in the pipeline. Either way, I'll file a new PR with this change.

ingvagabund:
> 1.) I didn't intend to simply copy and paste, I tried to incorporate what was already used by the scheduler while cutting anything outside the scope of this PR. I originally thought it would be good to keep the Descheduler somewhat in line with how the scheduler calculates resources, but maybe I went about that wrong.

I have noticed and appreciate your effort to be aligned with the scheduler and use only the minimum of the code from the scheduler code base. We have been discussing multiple times where is the right line between syncing and bringing the functionality of the scheduler into the descheduler and re-implementing the bits on our own. We are still in a process of keeping the right balance. If there's a chance, simpler solution is preferable. So both kube-scheduler and descheduler code bases have minimum assumptions about each other.

RyanDevlin:
@ingvagabund That makes sense, I'll definitely keep it in mind moving forward!

test/e2e/e2e_test.go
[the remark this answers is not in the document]
ingvagabund: Wonderful. I will review the PR so we can merge it and have this PR to pass all the tests.

damemi:
I apologize but I haven't been able to follow some of the recent updates to this PR lately.

In terms of the v0.23 descheduler release, @RyanDevlin is this something we should hold off on or could it fall into the next release?

RyanDevlin:
@damemi I just put in all my final updates/changes requested. If possible I'd love to get this merged sooner rather than later. Ultimately I'll defer to you on whether or not this should wait until next release.

RyanDevlin:
@ingvagabund Just a heads up, I've made the changes you requested and greatly simplified the code in `node.go`. I'd love to hear your feedback!

pkg/descheduler/strategies/nodeutilization/highnodeutilization_test.go
```diff
@@ -385,6 +385,50 @@ func TestHighNodeUtilization(t *testing.T) {
...
+				test.BuildTestPod("p4", 400, 50, n1NodeName, test.SetDSOwnerRef),
+				test.BuildTestPod("p5", 400, 100, n2NodeName, func(pod *v1.Pod) {
+					// A pod requesting more memory than is available on node1
+					test.SetRSOwnerRef(pod)
+				}),
+			},
+			expectedPodsEvicted: 0,
+		},
+		{
+			name: "Other node does not have enough Memory",
```
ingvagabund, line 411: Looks like this test is duplicated?

ingvagabund (changes requested):
Thank you for refactoring and simplifying the code. I have reviewed only the code. I will review the changes in the tests later.

pkg/descheduler/node/node_test.go
```diff
@@ -181,27 +185,53 @@ func TestPodFitsCurrentNode(t *testing.T) {
...
+					nodeLabelKey: "no",
+				}
+
+				node.Status.Allocatable[v1.ResourceEphemeralStorage] = *resource.NewQuantity(1000*1000*1000*1000, resource.DecimalSI)
+			}),
 			success: false,
 		},
 	}
 
+	fakeClient := &fake.Clientset{}
```
ingvagabund, line 199: Does not look to be used anywhere. The `t.Run` redefines the `fakeClient` variable.

pkg/descheduler/node/node_test.go
```diff
@@ -215,238 +245,527 @@ func TestPodFitsAnyOtherNode(t *testing.T) {
 		pod         *v1.Pod
 		nodes       []*v1.Node
 		success     bool
+		podsOnNodes []*v1.Pod
 	}{
 		{
 			description: "Pod fits another node matching node affinity",
-			pod:         createPodManifest(nodeNames[2], nodeLabelKey, nodeLabelValue),
```
ingvagabund, line 221: The `createPodManifest` is/was setting `Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution` whereas here the test is changed to set `spec.NodeSelector`. Which is not the same as the node affinity. Would you elaborate on why this change was performed?

pkg/descheduler/node/node_test.go
```diff
@@ -215,238 +245,527 @@ func TestPodFitsAnyOtherNode(t *testing.T) {
...
+
+					node.Status.Allocatable[v1.ResourceEphemeralStorage] = *resource.NewQuantity(1000*1000*1000*1000, resource.DecimalSI)
+				}),
+				test.BuildTestNode(nodeNames[2], 0, 0, 0, nil),
+			},
+			podsOnNodes: []*v1.Pod{},
+			success:     true,
+		},
+		{
+			description: "Pod expected to fit one of the nodes",
```
ingvagabund, line 278: How does this test differs from `Pod fits another node matching node affinity`? I only see node labels of `node1` and `node2` switched.

pkg/descheduler/strategies/nodeutilization/lownodeutilization_test.go
```diff
@@ -643,6 +643,112 @@ func TestLowNodeUtilization(t *testing.T) {
...
+					pod.Namespace = "kube-system"
+					priority := utils.SystemCriticalPriority
+					pod.Spec.Priority = &priority
+				}),
+				test.BuildTestPod("p9", 0, 0, n2NodeName, test.SetRSOwnerRef),
+			},
+			expectedPodsEvicted: 3,
+		},
+		{
+			name: "without priorities, but only other node doesn't match pod node affinity for p4 and p5",
```
ingvagabund, line 700: This test looks to be duplicate of the previous one.

pkg/descheduler/strategies/nodeutilization/lownodeutilization_test.go
```diff
@@ -643,6 +643,112 @@ func TestLowNodeUtilization(t *testing.T) {
 			},
 			expectedPodsEvicted: 3,
 		},
+		{
+			name: "without priorities, but only other node doesn't match pod node affinity for p4 and p5",
```
ingvagabund, line 647: I don't see any pod node affinities set in this test. Would you elaborate on what this test is expected to test?

ingvagabund (changes requested):

README.md
```diff
@@ -707,6 +707,7 @@ The following strategies accept a `nodeFit` boolean parameter which can optimize
 - A `nodeSelector` on the pod
 - Any `Tolerations` on the pod and any `Taints` on the other nodes
 - `nodeAffinity` on the pod
+- Resource `Requests` made by the pod and the resources available on other nodes
```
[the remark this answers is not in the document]
RyanDevlin: Is it okay with you if I fix the `tolerations` and `taints` capitalization in this PR? The change is in the `NodeFit` section only.

pkg/descheduler/strategies/nodeutilization/nodeutilization.go
```diff
@@ -354,44 +356,7 @@ func getResourceNames(thresholds api.ResourceThresholds) []v1.ResourceName {
...
-				totalReqs[name].Add(quantity)
-			}
-		}
-	}
-
-	return totalReqs
-}
-
-func classifyPods(pods []*v1.Pod, filter func(pod *v1.Pod) bool) ([]*v1.Pod, []*v1.Pod) {
+func classifyPods(ctx context.Context, pods []*v1.Pod, filter func(pod *v1.Pod) bool) ([]*v1.Pod, []*v1.Pod) {
```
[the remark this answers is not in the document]
RyanDevlin: Good catch, that's a mistake caused by a rebase. In the original PR I needed `context.Context` to list the pods which would have been used in that filter function. With a recent PR this changed to a `nodeIndexer` so the context isn't needed here.

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -285,64 +286,14 @@ func balanceDomains(
...
-
 // sortDomains sorts and splits the list of topology domains based on their size
 // it also sorts the list of pods within the domains based on their node affinity/selector and priority in the following order:
 // 1. non-evictable pods
 // 2. pods with selectors or affinity
 // 3. pods in descending priority
 // 4. all other pods
 // We then pop pods off the back of the list for eviction
-func sortDomains(constraintTopologyPairs map[topologyPair][]*v1.Pod, isEvictable func(*v1.Pod) bool) []topology {
+func sortDomains(ctx context.Context, constraintTopologyPairs map[topologyPair][]*v1.Pod, isEvictable func(pod *v1.Pod) bool) []topology {
```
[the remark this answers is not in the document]
RyanDevlin: Same reason as in `nodeutilization.go`, removed.

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -285,64 +286,14 @@ func balanceDomains(
 	}
 }
 
-// validatePodFitsOnOtherNodes performs validation based on scheduling predicates for affinity and toleration.
-// It excludes the current node because, for the sake of domain balancing only, we care about if there is any other
-// place it could theoretically fit.
-// If the pod doesn't fit on its current node, that is a job for RemovePodsViolatingNodeAffinity, and irrelevant to Topology Spreading
-func validatePodFitsOnOtherNodes(pod *v1.Pod, nodeMap map[string]*v1.Node) error {
```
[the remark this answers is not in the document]
RyanDevlin: @damemi and I discussed removing this function on a call. But I also would like him to review this change one more time.

pkg/descheduler/strategies/nodeutilization/highnodeutilization_test.go
RyanDevlin: This test should have been the reciprocal of the above test. Fixed.

README.md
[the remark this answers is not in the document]
ingvagabund: By all means :)

pkg/descheduler/node/node.go
```diff
@@ -96,32 +99,91 @@ func IsReady(node *v1.Node) bool {
...
+	}
+	// Check taints (we only care about NoSchedule and NoExecute taints)
+	ok = utils.TolerationsTolerateTaintsWithFilter(pod.Spec.Tolerations, node.Spec.Taints, func(taint *v1.Taint) bool {
+		return taint.Effect == v1.TaintEffectNoSchedule || taint.Effect == v1.TaintEffectNoExecute
+	})
+	if !ok {
+		errors = append(errors, fmt.Errorf("pod does not tolerate taints on the node"))
+	}
+	// Check if the pod can fit on a node based off it's requests
+	ok, reqErrors := FitsRequest(nodeIndexer, pod, node)
```
damemi, line 123: is `FitsRequest` called anywhere else? Wondering the reasoning behind making it its own exported function

pkg/descheduler/strategies/topologyspreadconstraint.go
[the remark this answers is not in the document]
damemi: The functionality looks the same, but I would like to preserve the comments from here explaining the effect on topology spread constraint and the TODO

damemi:
This has gone through a lot of review and now has sat for a few months, @ingvagabund is there anything blocking from your previous review or can we move forward with this? I'm open to some refactoring todos if necessary

ingvagabund:
Checking the last review I found some minor nits in the tests. Though, it looks good in overall.

damemi:
> Checking the last review I found some minor nits in the tests. Though, it looks good in overall.

would you be open to rebasing this and addressing the nits if @RyanDevlin isn't able to?

ingvagabund:
Sure thing

ingvagabund:
Merged through https://github.com/kubernetes-sigs/descheduler/pull/790. Thank You @RyanDevlin for putting the code together.
/close

Issue #638, opened 2021-10-01
Ignore Pods With Deletion Timestamp


seanmalloy:
As a follow-up, some of the other strategies could benefit from this as well. Possibly a good place would be [ListPodsOnANode](https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/descheduler/pod/pods.go#L159)), for those that use that.

_Originally posted by @a7i in https://github.com/kubernetes-sigs/descheduler/pull/621#pullrequestreview-754599387_

seanmalloy:
The `RemovePodsViolatingTopologySpreadConstraint` strategy now ignores pods with `.metadata.DeletionTimestamp` set. Need to figure out which other strategies would benefit from this.

/kind feature

JaneLiuL:
/assign 
I would like to have a try :)

a7i:
Fixed in #639 

/close

Issue #618, opened 2021-08-31
TopologySpreadConstraint should not take terminating pods into account when balancing domains


a7i:
**What version of descheduler are you using?**

descheduler version: 0.21.0

**Does this issue reproduce with the latest release?**
Yes

**Which descheduler CLI options are you using?**
`--logging-format text --policy-config-file /policy-dir/policy.yaml --v 4`

**Please provide a copy of your descheduler policy config file**
```
apiVersion: "descheduler/v1alpha1"
kind: "DeschedulerPolicy"
evictLocalStoragePods: true
strategies:
  RemovePodsViolatingTopologySpreadConstraint:
    enabled: true
    params:
      includeSoftConstraints: true
      labelSelector:
        matchExpressions:
        - {key: descheduler.k8s.io/opt-in, operator: In, values: ["true"]}
```

**What k8s version are you using (`kubectl version`)?**

<details><summary><code>kubectl version</code> Output</summary><br><pre>
$ kubectl version
Server Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.9", GitCommit:"7a576bc3935a6b555e33346fd73ad77c925e9e4a", GitTreeState:"clean", BuildDate:"2021-07-15T20:56:38Z", GoVersion:"go1.15.14", Compiler:"gc", Platform:"linux/amd64"}
</pre></details>

**What did you do?**
We have a deployment that has a high termination grace period.  During rollouts, the old pods stick around for some time and depending on when Descheduler runs, this can lead to Descheduler not having an accurate view of topology spread across domains.

This is mainly due to the fact that terminating is not a pod phase and the actual pod phase is running. The only distintion is the `metadata.deletionTimestamp`

**What did you expect to see?**
Descheduler to filter out terminating pods when running the strategy.

**What did you see instead?**
Descheduler takes terminating pods into account (i.e. filter by pods missing `metadata.deletionTimestamp`) when running the strategy.

PR #639, opened 2021-10-01
Ignore Pods With Deletion Timestamp

JaneLiuL:
Enhance ignore Pods With Deletion Timestamp for all strategies  
Issue: https://github.com/kubernetes-sigs/descheduler/issues/638

### DIFF
--- a/pkg/descheduler/evictions/evictions.go
+++ b/pkg/descheduler/evictions/evictions.go
@@ -290,6 +290,10 @@ func (ev *evictable) IsEvictable(pod *v1.Pod) bool {
 		checkErrs = append(checkErrs, fmt.Errorf("pod is a static pod"))
 	}
 
+	if utils.IsPodTerminating(pod) {
+		checkErrs = append(checkErrs, fmt.Errorf("pod is terminating"))
+	}
+
 	for _, c := range ev.constraints {
 		if err := c(pod); err != nil {
 			checkErrs = append(checkErrs, err)
--- a/pkg/descheduler/strategies/pod_antiaffinity_test.go
+++ b/pkg/descheduler/strategies/pod_antiaffinity_test.go
@@ -55,6 +55,10 @@ func TestPodAntiAffinity(t *testing.T) {
 	p6 := test.BuildTestPod("p6", 100, 0, node1.Name, nil)
 	p7 := test.BuildTestPod("p7", 100, 0, node1.Name, nil)
 	p8 := test.BuildTestPod("p8", 100, 0, node1.Name, nil)
+	p9 := test.BuildTestPod("p9", 100, 0, node1.Name, nil)
+	p10 := test.BuildTestPod("p10", 100, 0, node1.Name, nil)
+	p9.DeletionTimestamp = &metav1.Time{}
+	p10.DeletionTimestamp = &metav1.Time{}
 
 	criticalPriority := utils.SystemCriticalPriority
 	nonEvictablePod := test.BuildTestPod("non-evict", 100, 0, node1.Name, func(pod *v1.Pod) {
@@ -72,6 +76,8 @@ func TestPodAntiAffinity(t *testing.T) {
 	test.SetNormalOwnerRef(p5)
 	test.SetNormalOwnerRef(p6)
 	test.SetNormalOwnerRef(p7)
+	test.SetNormalOwnerRef(p9)
+	test.SetNormalOwnerRef(p10)
 
 	// set pod anti affinity
 	setPodAntiAffinity(p1, "foo", "bar")
@@ -80,6 +86,8 @@ func TestPodAntiAffinity(t *testing.T) {
 	setPodAntiAffinity(p5, "foo1", "bar1")
 	setPodAntiAffinity(p6, "foo1", "bar1")
 	setPodAntiAffinity(p7, "foo", "bar")
+	setPodAntiAffinity(p9, "foo", "bar")
+	setPodAntiAffinity(p10, "foo", "bar")
 
 	// set pod priority
 	test.SetPodPriority(p5, 100)
@@ -149,6 +157,13 @@ func TestPodAntiAffinity(t *testing.T) {
 			nodes:                   []*v1.Node{node1, node3},
 			expectedEvictedPodCount: 0,
 			nodeFit:                 true,
+		},
+		{
+			description:             "No pod to evicted since all pod terminating",
+			maxPodsToEvictPerNode:   0,
+			pods:                    []v1.Pod{*p9, *p10},
+			nodes:                   []*v1.Node{node1},
+			expectedEvictedPodCount: 0,
 		},
 	}
 
--- a/pkg/descheduler/strategies/pod_lifetime_test.go
+++ b/pkg/descheduler/strategies/pod_lifetime_test.go
@@ -125,6 +125,17 @@ func TestPodLifeTime(t *testing.T) {
 	p12.ObjectMeta.OwnerReferences = ownerRef1
 	p13.ObjectMeta.OwnerReferences = ownerRef1
 
+	p14 := test.BuildTestPod("p14", 100, 0, node1.Name, nil)
+	p15 := test.BuildTestPod("p15", 100, 0, node1.Name, nil)
+	p14.Namespace = "dev"
+	p15.Namespace = "dev"
+	p14.ObjectMeta.CreationTimestamp = olderPodCreationTime
+	p15.ObjectMeta.CreationTimestamp = olderPodCreationTime
+	p14.ObjectMeta.OwnerReferences = ownerRef1
+	p15.ObjectMeta.OwnerReferences = ownerRef1
+	p14.DeletionTimestamp = &metav1.Time{}
+	p15.DeletionTimestamp = &metav1.Time{}
+
 	var maxLifeTime uint = 600
 	testCases := []struct {
 		description             string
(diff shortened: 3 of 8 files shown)


damemi (commented):
/ok-to-test

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -140,11 +140,6 @@ func RemovePodsViolatingTopologySpreadConstraint(
 			// (this loop is where we count the number of pods per topologyValue that match this constraint's selector)
 			var sumPods float64
 			for i := range namespacePods.Items {
-				// skip pods that are being deleted.
-				if namespacePods.Items[i].DeletionTimestamp != nil {
```
a7i, line 144: Removing this throws off the calculation of the topology size.

/cc @damemi to confirm.

pkg/utils/pod.go
```diff
@@ -89,6 +89,11 @@ func IsMirrorPod(pod *v1.Pod) bool {
 	return ok
 }
 
+// IsPodActive return true if the pod deletiontimestamp is nil.
+func IsPodActive(pod *v1.Pod) bool {
+	return pod.DeletionTimestamp != nil
```
a7i, line 94: This is backward? The comment is correct but the implementation is not.

pkg/descheduler/strategies/topologyspreadconstraint.go
damemi: Correct this check needs to be here specifically so that pods with a DeletionTimestamp aren't included in the calculation of the topology size, see https://github.com/kubernetes-sigs/descheduler/issues/618

It's unfortunately a bit redundant to the eviction check but necessary

pkg/descheduler/evictions/evictions.go
```diff
@@ -290,6 +290,10 @@ func (ev *evictable) IsEvictable(pod *v1.Pod) bool {
 		checkErrs = append(checkErrs, fmt.Errorf("pod is a static pod"))
 	}
 
+	if !utils.IsPodActive(pod) {
```
damemi, line 293: maybe to be more consistent with the other checks, change this function to `IsPodTerminating` and have it return `pod.DeletionTimestamp != nil`

JaneLiuL: Good idea~ Fix done :)

damemi:
> Thoughts on updating README to indicate that going forward Pods that are being deleted are ignored. This is important because Pods that have a long termination grace period will no longer be considered for eviction so it's a change in behavior.

This should definitely be updated in the readme, yeah. We might even want to make this a non-default behavior

JaneLiuL:
> > Thoughts on updating README to indicate that going forward Pods that are being deleted are ignored. This is important because Pods that have a long termination grace period will no longer be considered for eviction so it's a change in behavior.
> 
> This should definitely be updated in the readme, yeah. We might even want to make this a non-default behavior

Update done, please check~

pkg/descheduler/strategies/pod_lifetime_test.go
```diff
@@ -245,12 +253,34 @@ func TestPodLifeTime(t *testing.T) {
 			pods:                    []v1.Pod{*p12, *p13},
 			nodes:                   []*v1.Node{node1},
 			expectedEvictedPodCount: 1,
+			testFunc:                func(podList []v1.Pod) {},
```
ingvagabund, line 256: There's no need for `testFunc` which sets the `DeletionTimestamp` only for pods `p12` and `p13`. Can you create pods `p14` and `p15` instead as copies of `p12` and `p13` and set the `DeletionTimestamp` explicitly?

pkg/descheduler/strategies/pod_antiaffinity_test.go
```diff
@@ -149,12 +156,26 @@ func TestPodAntiAffinity(t *testing.T) {
...
 			nodeFit:                 true,
+			testFunc:                func(podList []v1.Pod) {},
+		},
+		{
+			description:             "No pod to evicted since all pod terminating",
+			maxPodsToEvictPerNode:   0,
+			pods:                    []v1.Pod{*p1, *p2, *p3, *p4},
+			nodes:                   []*v1.Node{node1},
+			expectedEvictedPodCount: 3,
+			testFunc: func(podList []v1.Pod) {
```
ingvagabund, line 167: Why 4 pods? Are not two pods sufficient enough?

pkg/descheduler/strategies/pod_antiaffinity_test.go
```diff
@@ -149,12 +156,26 @@ func TestPodAntiAffinity(t *testing.T) {
...
+				for _, pod := range podList {
+					pod.DeletionTimestamp = &metav1.Time{}
```
ingvagabund, line 169: This actually modifies the pods permanently. Each test needs a freshly created pod. Instead, can you create new pods (e.g. p5 - p8) which have the `DeletionTimestamp` set right away?

ingvagabund (changes requested):

pkg/descheduler/strategies/pod_lifetime_test.go
JaneLiuL: Hi @ingvagabund  thanks for the reivew~~
I think `testFunc` is more easy to extend on the test case. This time we only enhance for DeletionTimestamp, maybe next time we need to do other actions on the Pod. So for me, testFunc is more like a `BuildTestNode` or `BuildTestPod`.

pkg/descheduler/strategies/pod_antiaffinity_test.go
JaneLiuL: Hi . This is for comare with above test case.
```
			description:             "Maximum pods to evict - 3",
			maxPodsToEvictPerNode:   3,
			pods:                    []v1.Pod{*p1, *p2, *p3, *p4},
			nodes:                   []*v1.Node{node1},
			expectedEvictedPodCount: 3,
```

pkg/descheduler/strategies/pod_antiaffinity_test.go
JaneLiuL: Thanks~ I learn it. I can not check that testcase need a freashly created pod. So I would create a new pod for this test case, is it ok for you?

pkg/descheduler/strategies/pod_antiaffinity_test.go
JaneLiuL: Fix done~ Please check :)

pkg/descheduler/strategies/pod_antiaffinity_test.go
JaneLiuL: Fix done~ Please check :)

damemi: @JaneLiuL if you are creating p9 and p10 now specifically for this test, do you still need `testFunc` at all? you can just specify those pods with an empty timestamp

pkg/descheduler/strategies/pod_antiaffinity_test.go
```diff
@@ -149,12 +162,26 @@ func TestPodAntiAffinity(t *testing.T) {
...
+				}
+			},
 		},
 	}
 
 	for _, test := range tests {
 		// create fake client
 		fakeClient := &fake.Clientset{}
+		test.testFunc([]v1.Pod{*p9, *p10})
```
damemi, line 184: this is calling `testFunc` on p9 and p10 for every test, even though they are only used in one test. in this case you should just define those pods with the settings you need

pkg/descheduler/strategies/pod_lifetime_test.go
damemi: @JaneLiuL that's not bad thinking to think ahead, but we don't have any use cases or evidence that we'll need more extensibility at this time. If we do come across that, the test can be refactored then, but for now adding a testFunc is unnecessary

pkg/descheduler/strategies/pod_lifetime_test.go
```diff
@@ -245,12 +253,34 @@ func TestPodLifeTime(t *testing.T) {
...
+					pod.DeletionTimestamp = &metav1.Time{}
+				}
+			},
 		},
 	}
 
 	for _, tc := range testCases {
 		fakeClient := &fake.Clientset{}
-
+		tc.testFunc([]v1.Pod{*p12, *p13})
```
damemi, line 283: this is again calling testFunc for just p12 and p13 on every test, even though the pods are only used in 1 test. this shouldn't be necessary

pkg/descheduler/strategies/pod_lifetime_test.go
JaneLiuL: @ingvagabund  @damemi  Thanks for your comment~
Then I will try to remove the testFunc, and create the new testPod with deleteTimestamp, is it better?

JaneLiuL: Already remove `testFunc`, may I click `Resolve conversation`?

pkg/descheduler/strategies/pod_antiaffinity_test.go
JaneLiuL: Already remove `testFunc`, may I click `Resolve conversation`?

pkg/descheduler/strategies/pod_antiaffinity_test.go
JaneLiuL: Already remove `testFunc`, may I click `Resolve conversation`?

pkg/descheduler/strategies/pod_lifetime_test.go
JaneLiuL: Already remove `testFunc`, may I click `Resolve conversation`?

pkg/descheduler/strategies/pod_lifetime_test.go
damemi: Yeah this is looking a lot cleaner now, thanks!

PR #510, opened 2021-03-01
Filter pods by labelSelector

lixiang233:
Fixes: #195 

TODO:
- [x] API change
- [x] new labelSelector option
- [x] Add option to strategies
- [x] README and e2e-tests

/kind feature

PR #642, opened 2021-10-07
README: Clarify podRestartThreshold applying to the sum over containers

wking:
`calcContainerRestarts` [sums over containers][1].  The new language makes that clear, avoiding potential confusion vs. an altenative that looked for pods where a single container had passed the configured threshold. For example, with three containers with 50 restarts and a threshold of 100, the actual "sum over containers" logic makes that pod a candidate for descheduling, but the "largest single container restart count" hypothetical would not have made it a candidate.

Also shifts `labelSelector` into the parameter table, because when it was added in 29ade13ce7 (#510), it landed a few lines too high.

[1]: https://github.com/kubernetes-sigs/descheduler/blob/bfd5feaf60a202cecafcf2d30cf171a138c04885/pkg/descheduler/strategies/toomanyrestarts.go#L110-L122

### DIFF
--- a/README.md
+++ b/README.md
@@ -456,9 +456,9 @@ strategies:
 
 This strategy makes sure that pods having too many restarts are removed from nodes. For example a pod with EBS/PD that
 can't get the volume/disk attached to the instance, then the pod should be re-scheduled to other nodes. Its parameters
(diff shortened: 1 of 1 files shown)


damemi:
Good suggestion @wking! I have opened a PR here to do just that: https://github.com/kubernetes/test-infra/pull/23956

Issue #644, opened 2021-10-08
Allow to remove Failed pods with no OwnerRef


weisdd:
**Is your feature request related to a problem? Please describe.**
Currently, Failed pods with no OwnerRef cannot be evicted by RemoveFailedPods policy.
Manually created pods are taken out of consideration as they will not be recreated by controllers. Though, I think it should not be a concern for Failed pods.

**Describe the solution you'd like**
Make Failed pods evictable despite lack of OwnerRef.

**Describe alternatives you've considered**
Manually deleting pods

**What version of descheduler are you using?**
descheduler version: 0.22.1

damemi:
I [commented in your PR](https://github.com/kubernetes-sigs/descheduler/pull/645#pullrequestreview-775000570) before I saw this issue, but just for reference here applying the [`descheduler.alpha.kubernetes.io/evict` annotation](https://github.com/kubernetes-sigs/descheduler#pod-evictions) to any pod will make it evictable regardless of the various eviction checks. So I don't think this is necessary for just one strategy

weisdd:
@damemi well, when there are many people experimenting in the same dev-cluster, it's hard to force everyone to always add the annotation. E.g. when people create Jobs and then never clean them up. So, I thought it would be much easier to simply deschedule Failed pods in that case. Setting up mutation webhook would be way more difficult.

damemi:
So we don't have to duplicate the comments, I propose we keep the discussion for this in this issue rather than the PR 🙂  From there (so others can see):

>@weisdd I understand your point, but couldn't the same be said for other strategies like PodLifetime? Maybe we should delineate a group of strategies that apply to "expired" pods to consider this kind of behavior.
>
>If we do add this, it should be as a non-default option since it is a change in behavior from the current default.

weisdd:
> So we don't have to duplicate the comments, I propose we keep the discussion for this in this issue rather than the PR

@damemi agreed, was about to suggest the same :)

I've actually used descheduler for the first time today (it's been on my list of things to try for ages), so my perception of how things can potentially be done might be too far from the project's maintainers vision :)

Making Failed pods evictable felt in line with what I'd expect from descheduler. The change in the PR seemed to be the easiest solution and, which is more important, to be based on a safe assumption that nobody would want to have those pods around for longer than `minPodLifeTimeSeconds` (+ safe enough to make it default). Though, I'm open to whatever feels better to you.

[...]

PR #645, opened 2021-10-08
Allow to evict Failed pods with no OwnerRef

weisdd:
The suggested change allows descheduler to evict Failed pods with no OwnerRef. The PR was opened in support of the feature request: https://github.com/kubernetes-sigs/descheduler/issues/644

### DIFF
--- a/pkg/descheduler/evictions/evictions.go
+++ b/pkg/descheduler/evictions/evictions.go
@@ -278,7 +278,7 @@ func (ev *evictable) IsEvictable(pod *v1.Pod) bool {
 		checkErrs = append(checkErrs, fmt.Errorf("pod is a DaemonSet pod"))
 	}
 
-	if len(ownerRefList) == 0 {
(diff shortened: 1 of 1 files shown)


weisdd:
@damemi well, when there are many people experimenting in the same dev-cluster, it's hard to force everyone to always add the annotation. E.g. when people create Jobs and then never clean them up. So, I thought it would be much easier to simply deschedule Failed pods in that case. Setting up mutation webhook would be way more difficult.

damemi:
@weisdd @babygoat are we going to close this PR in favor of https://github.com/kubernetes-sigs/descheduler/pull/688? They are trying to solve the same issue in different ways

weisdd:
@damemi please, feel free to close this PR. As discussed in the related issue, my implementation was not optimal.

damemi:
@weisdd thanks for the clarification, just wanted to check to make sure we didn't have duplicate fixes open!
/close

Issue #503, opened 2021-02-19
Improve podEvictor statistics


damemi:
As suggested in https://github.com/kubernetes-sigs/descheduler/issues/501#issuecomment-781967812, it would be nice to improve the pod evictor type to report eviction statistics for individual strategies. Some suggestions were:

> ```
> Number of evicted pods (in this strategy): XX
> Number of evicted pods in this run: XX
> Total number of evicted pods in all strategies: XX
> ```

x-ref this could also be reported as Prometheus metrics (https://github.com/kubernetes-sigs/descheduler/issues/348)

ingvagabund:
I prefer to report the statistics in metrics. So we don't have to cumulative much in the pod evictor itself.

damemi:
I am just suggesting that, since those metrics will have to be calculated somewhere, doing it in podEvictor makes sense because it already has access to the information. Metrics can then use the podEvictor instance to report them when requested.

a7i:
@damemi I would be happy to contribute to this. Any docs highlighting the decision made to-date?

damemi:
@a7i nothing concrete, though if you would like to put some ideas together and share a doc that would be a great place to start the discussion. Right now we have 1 metric [`pods_evicted`](https://github.com/kubernetes-sigs/descheduler/blob/1c5b327/metrics/metrics.go) that's reported by the PodEvictor after a run.

As suggested above, it would be good to have some similar reports on a per-strategy basis. From there we could probably even come up with some additional meta metrics that are specific to the different strategies themselves.

pravarag:
I'd like to work on this issue if someone is working on it already 🙂  @damemi @ingvagabund

ingvagabund:
Not aware of anyone working on this atm. Although, this requires some design and probably starting some discussion (e.g. in a google doc). @damemi wdyt?

damemi:
Yeah I think we have some good patterns started in the code for metrics reporting already that could be fleshed out more. @pravarag feel free to take this on if you'd like

a7i:
It would be great to have the following:

- pods evicted success
- pods evicted failed
- pods skipped
- total pods under consideration

Overall and per strategy.

pravarag:
@damemi @ingvagabund I'm trying to replicate the eviction of pods in a local cluster for better understanding of the way statistics are currently being represented. I have a 3 node cluster with the resources not that much heavily utilized for the three nodes cluster, the stats stand here:

```
[...]

PR #648, opened 2021-10-17
Add new podEvictor statistics

pravarag:
Fixes https://github.com/kubernetes-sigs/descheduler/issues/503

As part of this change, we are adding newer metrics for improving podEvictor statistics. Currently there is only one metric which is being calculated under: [evictions.go](https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/descheduler/evictions/evictions.go)

### DIFF
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -38,6 +38,38 @@ var (
 			StabilityLevel: metrics.ALPHA,
 		}, []string{"result", "strategy", "namespace", "node"})
 
+	PodsEvictedSuccess = metrics.NewCounterVec(
+		&metrics.CounterOpts{
+			Subsystem:      DeschedulerSubsystem,
+			Name:           "pods_evicted_success",
+			Help:           "Number of pods successfully evicted per strategy",
+			StabilityLevel: metrics.ALPHA,
+		}, []string{"strategy", "namespace"})
+
+	PodsEvictedFailed = metrics.NewCounterVec(
+		&metrics.CounterOpts{
+			Subsystem:      DeschedulerSubsystem,
+			Name:           "pods_evicted_failed",
+			Help:           "Number of pods failed to evict per strategy",
+			StabilityLevel: metrics.ALPHA,
+		}, []string{"strategy", "namespace"})
+
+	TotalPodsEvicted = metrics.NewCounterVec(
+		&metrics.CounterOpts{
+			Subsystem:      DeschedulerSubsystem,
+			Name:           "total_pods_evicted",
+			Help:           "Total pods evicted so far",
+			StabilityLevel: metrics.ALPHA,
+		}, []string{"result", "namespace"})
+
+	TotalPodsSkipped = metrics.NewCounterVec(
+		&metrics.CounterOpts{
+			Subsystem:      DeschedulerSubsystem,
+			Name:           "total_pods_skipped",
+			Help:           "Total pods skipped for a single run",
+			StabilityLevel: metrics.ALPHA,
+		}, []string{"result", "namespace"})
+
 	buildInfo = metrics.NewGauge(
 		&metrics.GaugeOpts{
 			Subsystem:      DeschedulerSubsystem,
@@ -51,6 +83,10 @@ var (
 	metricsList = []metrics.Registerable{
 		PodsEvicted,
 		buildInfo,
+		PodsEvictedSuccess,
+		PodsEvictedFailed,
+		TotalPodsEvicted,
+		TotalPodsSkipped,
 	}
 )
 
--- a/pkg/descheduler/evictions/evictions.go
+++ b/pkg/descheduler/evictions/evictions.go
@@ -124,29 +124,29 @@ func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod, node *v1.Node, 
 		reason += " (" + strings.Join(reasons, ", ") + ")"
 	}
 	if pe.maxPodsToEvictPerNode != nil && pe.nodepodCount[node]+1 > *pe.maxPodsToEvictPerNode {
-		if pe.metricsEnabled {
-			metrics.PodsEvicted.With(map[string]string{"result": "maximum number of pods per node reached", "strategy": strategy, "namespace": pod.Namespace, "node": node.Name}).Inc()
-		}
+		metrics.PodsEvicted.With(map[string]string{"result": "maximum number of pods per node reached", "strategy": strategy, "namespace": pod.Namespace, "node": node.Name}).Inc()
+		metrics.TotalPodsSkipped.With(map[string]string{"result": "total pods skipped so far", "namespace": pod.Namespace}).Inc()
 		return false, fmt.Errorf("Maximum number %v of evicted pods per %q node reached", *pe.maxPodsToEvictPerNode, node.Name)
 	}
 
 	if pe.maxPodsToEvictPerNamespace != nil && pe.namespacePodCount[pod.Namespace]+1 > *pe.maxPodsToEvictPerNamespace {
-		if pe.metricsEnabled {
-			metrics.PodsEvicted.With(map[string]string{"result": "maximum number of pods per namespace reached", "strategy": strategy, "namespace": pod.Namespace, "node": node.Name}).Inc()
-		}
+		metrics.PodsEvicted.With(map[string]string{"result": "maximum number of pods per namespace reached", "strategy": strategy, "namespace": pod.Namespace, "node": node.Name}).Inc()
+		metrics.TotalPodsSkipped.With(map[string]string{"result": "total pods skipped so far", "namespace": pod.Namespace}).Inc()
 		return false, fmt.Errorf("Maximum number %v of evicted pods per %q namespace reached", *pe.maxPodsToEvictPerNamespace, pod.Namespace)
 	}
 
 	err := evictPod(ctx, pe.client, pod, pe.policyGroupVersion)
+	// increment TotalPodsEvicted
+	metrics.TotalPodsEvicted.With(map[string]string{"result": "total pods evicted so far", "namespace": pod.Namespace}).Inc()
 	if err != nil {
 		// err is used only for logging purposes
 		klog.ErrorS(err, "Error evicting pod", "pod", klog.KObj(pod), "reason", reason)
-		if pe.metricsEnabled {
-			metrics.PodsEvicted.With(map[string]string{"result": "error", "strategy": strategy, "namespace": pod.Namespace, "node": node.Name}).Inc()
-		}
+		metrics.PodsEvicted.With(map[string]string{"result": "error", "strategy": strategy, "namespace": pod.Namespace, "node": node.Name}).Inc()
+		metrics.PodsEvictedFailed.With(map[string]string{"strategy": strategy, "namespace": pod.Namespace}).Inc()
 		return false, nil
 	}
 
+	metrics.PodsEvictedSuccess.With(map[string]string{"strategy": strategy, "namespace": pod.Namespace}).Inc()
 	pe.nodepodCount[node]++
 	pe.namespacePodCount[pod.Namespace]++
 
@@ -184,6 +184,7 @@ func evictPod(ctx context.Context, client clientset.Interface, pod *v1.Pod, poli
 	err := client.PolicyV1beta1().Evictions(eviction.Namespace).Evict(ctx, eviction)
 
 	if apierrors.IsTooManyRequests(err) {
+		metrics.TotalPodsSkipped.With(map[string]string{"result": "total pods skipped so far", "namespace": pod.Namespace}).Inc()
 		return fmt.Errorf("error when evicting pod (ignoring) %q: %v", pod.Name, err)
 	}
 	if apierrors.IsNotFound(err) {


a7i:
🥇 

It would be nice to have more insights around these metrics. What are your (and others) thoughts around including the following info?

1. Include `namespace`
> I think this should be easy to do and I would think this is something that most users would care about. `PodsEvicted` metric already does this.

2. Include all `ownerReferences[].name`
> I don't have a use-case around this, just an idea. 

3. Inject custom labels into the metrics.
> This one is tricky but my company has standards around labels (e.g. `project=` `team=`), it would be nice to add those in the DeschedulerPolicy somehow and inject them as "custom" labels in the metrics. I realize this is outside the scope of this PR so I can take this on after this PR is merged.

4. this conflicts with #3 but another option would be to include [well-know labels](https://kubernetes.io/docs/reference/labels-annotations-taints/)
> I personally prefer no. 3 but just an idea

pravarag:
@a7i thanks for the information shared above. Please feel free to suggest any changes I can implement as part of this PR as well.
@damemi @ingvagabund kindly review and let me know if I can update or make any further changes :)

a7i:
> @a7i thanks for the information shared above. Please feel free to suggest any changes I can implement as part of this PR as well. @damemi @ingvagabund kindly review and let me know if I can update or make any further changes :)

I would really like to see `namespace` included in the metrics. `PodsEvicted` metric already does this.

damemi (commented):
/ok-to-test
@pravarag could you please move the PR out of draft if it's ready for review and remove the `WIP` from the title?

metrics/metrics.go
```diff
@@ -38,6 +38,38 @@ var (
...
+
+	TotalPodsEvicted = metrics.NewCounterVec(
+		&metrics.CounterOpts{
+			Subsystem:      DeschedulerSubsystem,
+			Name:           "total_pods_evicted",
+			Help:           "Total pods evicted so far",
+			StabilityLevel: metrics.ALPHA,
+		}, []string{"result"})
+
+	TotalPodsSkipped = metrics.NewCounterVec(
```
damemi, line 65: is there anywhere you're updating `TotalPodsSkipped`?

pkg/descheduler/evictions/evictions.go
```diff
@@ -112,14 +113,16 @@ func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod, node *v1.Node,
 	}
 	if pe.maxPodsToEvictPerNode > 0 && pe.nodepodCount[node]+1 > pe.maxPodsToEvictPerNode {
 		metrics.PodsEvicted.With(map[string]string{"result": "maximum number reached", "strategy": strategy, "namespace": pod.Namespace}).Inc()
-		return false, fmt.Errorf("Maximum number %v of evicted pods per %q node reached", pe.maxPodsToEvictPerNode, node.Name)
+		metrics.PodsEvictedSuccess.With(map[string]string{"strategy": strategy}).Inc()
```
a7i, line 116: Why is `PodsEvictedSuccess` being incremented when maxPodsToEvictPerNode has been reached? Should this be `TotalPodsSkipped`?

pkg/descheduler/evictions/evictions.go
```diff
@@ -99,6 +99,7 @@ func (pe *PodEvictor) TotalEvicted() int {
 	for _, count := range pe.nodepodCount {
 		total += count
 	}
+	metrics.TotalPodsEvicted.With(map[string]string{"result": "total pods evicted so far"}).Inc()
```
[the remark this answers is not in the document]
pravarag: The reason for incrementing the metrics here is to capture the total pods evicted as per the count goes up in line 100. I thought of keeping it align with the evicted pod count going up in `evictions.go` so that it might be easier to record the total count. Let me know your thoughts otherwise I'll make some changes.

pkg/descheduler/evictions/evictions.go
pravarag: Thanks for identifying this, I'll move `PodsEvictedSuccess` to a different and more appropriate place. I might have misinterpreted it's usage here.

pkg/descheduler/evictions/evictions.go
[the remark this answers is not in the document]
a7i: I see. It seems odd that it's being incremented in a method that is only supposed to get total evicted count.
```
// TotalEvicted gives a number of pods evicted through all nodes
```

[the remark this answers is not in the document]
pravarag: Updated this as well to [this](https://github.com/kubernetes-sigs/descheduler/pull/648/files#diff-695c3df113e19fd5eaebf2b4bd6f6d4ba498c1bd0ae97875816360b66f6f8044R133) line.

metrics/metrics.go
pravarag: I was not able to figure out as where I can update `TotalPodsSkipped` if it could be below this [line](https://github.com/kubernetes-sigs/descheduler/blob/8b5bb6a12d9d13b7a3906368137fc37970599cb3/pkg/descheduler/evictions/evictions.go#L122) and this [line](https://github.com/kubernetes-sigs/descheduler/blob/8b5bb6a12d9d13b7a3906368137fc37970599cb3/pkg/descheduler/evictions/evictions.go#L127). If you  could suggest me anything here, will be helpful.

pravarag:
@a7i @damemi I've one doubt regarding the utilization of metric `TotalPodsSkipped`. So for it's use-cases, I could think of two possible situations when we might want to skip a pod and then increment the respective metrics for it:
1) I think in these two cases when we have reached maximum number of pods per Node, Namespace: [L120-125](https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/descheduler/evictions/evictions.go#L120-L125)
2) After this [line](https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/descheduler/evictions/evictions.go#L174) if there is any error faced while trying to evict a pod.

Any suggestions on where else I can utilize TotalPodsSkipped metrics would be helpful.

damemi:
@pravarag I think those are the two ways I would think to increment it as well

pravarag:
I've updated the `evictions.go` to include PodsSkipped metrics, kindly review and let me know for any changes @damemi @a7i

ingvagabund:
@pravarag Apologies for waiting for all the handy improvements in your PR to merge for so long. The review has been on for some time and I believe it's near its end. @damemi @a7i please prove me wrong if there's anything else to review.

Would you mind waiting for https://github.com/kubernetes-sigs/descheduler/pull/677 to merge before the final approval? The PR makes sure the metrics are incrementing only when enabled.

a7i:
> @pravarag Apologies for waiting for all the handy improvements in your PR to merge for so long. The review has been on for some time and I believe it's near its end. @damemi @a7i please prove me wrong if there's anything else to review.
> 
> As part of the improvements, would you mind waiting for #677 to merge before the final approval? The PR makes sure the metrics are incrementing only when enabled.

I do not believe the following are being incremented:
```
PodsEvictedSuccess
PodsEvictedFailed
```

Either the metric definition/registration should be removed or they should be incremented.

pravarag:
@a7i @ingvagabund thanks for reviewing this PR and letting me know about missing parts.

> I do not believe the following are being incremented:
> 
> PodsEvictedSuccess
> PodsEvictedFailed
> 
> Either the metric definition/registration should be removed or they should be incremented.

I missed on incrementing the PodsEvictedFailed which I've added now here: [link](https://github.com/kubernetes-sigs/descheduler/pull/648/files#diff-695c3df113e19fd5eaebf2b4bd6f6d4ba498c1bd0ae97875816360b66f6f8044R139) though I had a thought of completely removing this line [here](https://github.com/kubernetes-sigs/descheduler/pull/648/files#diff-695c3df113e19fd5eaebf2b4bd6f6d4ba498c1bd0ae97875816360b66f6f8044R138) and only increment `PodsEvictedFailed` at this point since we are adding a particular failed metric. Kindly let me know your thoughts around it.

metrics/metrics.go
pravarag: Updated.

damemi:
@pravarag I think this looks good, just needs a rebase
/approve

pravarag:
Thanks @damemi , I've rebased the branch as suggested.

metrics/metrics.go
```diff
@@ -38,6 +38,38 @@ var (
...
+			Name:           "total_pods_evicted",
+			Help:           "Total pods evicted so far",
+			StabilityLevel: metrics.ALPHA,
+		}, []string{"result", "namespace"})
+
+	TotalPodsSkipped = metrics.NewCounterVec(
+		&metrics.CounterOpts{
+			Subsystem:      DeschedulerSubsystem,
+			Name:           "total_pods_skipped",
+			Help:           "Total pods skipped for a single run",
```
ingvagabund, line 69: I wonder what a single run means in this context? Is it meant as a single [descheduling cycle](https://github.com/kubernetes-sigs/descheduler/blob/901a16ecbcaaf41c11318920f8a047c918b6c30d/pkg/descheduler/descheduler.go#L218)? `TotalPodsSkipped` metric is used only inside `PodEvictor` where there's currently no way to distinguish when a single run finished and a new one started.

pkg/descheduler/evictions/evictions.go
```diff
@@ -184,6 +184,7 @@ func evictPod(ctx context.Context, client clientset.Interface, pod *v1.Pod, poli
 	err := client.PolicyV1beta1().Evictions(eviction.Namespace).Evict(ctx, eviction)
 
 	if apierrors.IsTooManyRequests(err) {
+		metrics.TotalPodsSkipped.With(map[string]string{"result": "total pods skipped so far", "namespace": pod.Namespace}).Inc()
```
ingvagabund, line 187: Setting the `result` to `total pods skipped so far` is redundant here given the metric's name is `TotalPodsSkipped`. I wonder if it would make more sense to set the `result` to `pod skipped due to TooManyRequests error`?

pkg/descheduler/evictions/evictions.go
```diff
@@ -124,29 +124,29 @@ func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod, node *v1.Node,
...
 		klog.ErrorS(err, "Error evicting pod", "pod", klog.KObj(pod), "reason", reason)
-		if pe.metricsEnabled {
-			metrics.PodsEvicted.With(map[string]string{"result": "error", "strategy": strategy, "namespace": pod.Namespace, "node": node.Name}).Inc()
-		}
+		metrics.PodsEvicted.With(map[string]string{"result": "error", "strategy": strategy, "namespace": pod.Namespace, "node": node.Name}).Inc()
+		metrics.PodsEvictedFailed.With(map[string]string{"strategy": strategy, "namespace": pod.Namespace}).Inc()
 		return false, nil
 	}
 
+	metrics.PodsEvictedSuccess.With(map[string]string{"strategy": strategy, "namespace": pod.Namespace}).Inc()
```
ingvagabund, line 149: `PodsEvictedSuccess` is a special case of `PodsEvicted` (with `result` label set to `success` and `node` label ignored).

pkg/descheduler/evictions/evictions.go
```diff
@@ -124,29 +124,29 @@ func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod, node *v1.Node,
...
+	// increment TotalPodsEvicted
+	metrics.TotalPodsEvicted.With(map[string]string{"result": "total pods evicted so far", "namespace": pod.Namespace}).Inc()
 	if err != nil {
 		// err is used only for logging purposes
 		klog.ErrorS(err, "Error evicting pod", "pod", klog.KObj(pod), "reason", reason)
-		if pe.metricsEnabled {
-			metrics.PodsEvicted.With(map[string]string{"result": "error", "strategy": strategy, "namespace": pod.Namespace, "node": node.Name}).Inc()
-		}
+		metrics.PodsEvicted.With(map[string]string{"result": "error", "strategy": strategy, "namespace": pod.Namespace, "node": node.Name}).Inc()
+		metrics.PodsEvictedFailed.With(map[string]string{"strategy": strategy, "namespace": pod.Namespace}).Inc()
```
ingvagabund, line 145: `PodsEvictedFailed` is a special case of `PodsEvicted` (with `result` label set to `error` and `node` label ignored). I wonder what is benefit of creating this metric compared to `PodsEvictedFailed`?

pkg/descheduler/evictions/evictions.go
```diff
@@ -124,29 +124,29 @@ func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod, node *v1.Node,
...
+		metrics.TotalPodsSkipped.With(map[string]string{"result": "total pods skipped so far", "namespace": pod.Namespace}).Inc()
 		return false, fmt.Errorf("Maximum number %v of evicted pods per %q node reached", *pe.maxPodsToEvictPerNode, node.Name)
 	}
 
 	if pe.maxPodsToEvictPerNamespace != nil && pe.namespacePodCount[pod.Namespace]+1 > *pe.maxPodsToEvictPerNamespace {
-		if pe.metricsEnabled {
-			metrics.PodsEvicted.With(map[string]string{"result": "maximum number of pods per namespace reached", "strategy": strategy, "namespace": pod.Namespace, "node": node.Name}).Inc()
-		}
+		metrics.PodsEvicted.With(map[string]string{"result": "maximum number of pods per namespace reached", "strategy": strategy, "namespace": pod.Namespace, "node": node.Name}).Inc()
+		metrics.TotalPodsSkipped.With(map[string]string{"result": "total pods skipped so far", "namespace": pod.Namespace}).Inc()
```
ingvagabund, line 134: `TotalPodsSkipped` called here is a special case of `PodsEvicted` (with `result` label set to `maximum number of pods per namespace reached` and `node` label ignored).

pkg/descheduler/evictions/evictions.go
```diff
@@ -124,29 +124,29 @@ func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod, node *v1.Node,
...
 		return false, fmt.Errorf("Maximum number %v of evicted pods per %q namespace reached", *pe.maxPodsToEvictPerNamespace, pod.Namespace)
 	}
 
 	err := evictPod(ctx, pe.client, pod, pe.policyGroupVersion)
+	// increment TotalPodsEvicted
+	metrics.TotalPodsEvicted.With(map[string]string{"result": "total pods evicted so far", "namespace": pod.Namespace}).Inc()
```
ingvagabund, line 140: At this point it is unknown if a pod was evicted or not. Thus, incrementing the `TotalPodsEvicted` does not reflect the actual pod eviction. The metric currently captures how many times the `evictPod` function was invoked.

ingvagabund (commented):
I wonder what is benefit of the new `TotalPodsSkipped`, `PodsEvictedSuccess` and `PodsEvictedFailed` metrics besides simpler expressions when querying the metrics? Currently they are subset of what is already provided by `PodsEvicted`.

When running the descheduler with metrics enabled the `PodsEvicted` metric gets continuously incremented over time with pods getting regularly evicted. In order to provide a number of pods evicted/skipped/... in a single run the pod evictor needs to be signaled a new run/descheduling cycle started.

ingvagabund:
Within each [descheduling cycle](https://github.com/kubernetes-sigs/descheduler/blob/901a16ecbcaaf41c11318920f8a047c918b6c30d/pkg/descheduler/descheduler.go#L263) a new pod evictor instance is created. Which eliminates any way of telling the pod evictor "a run has ended". The `PodsEvicted` metric is shared across cycles so there's no distinction of which increments belongs to which descheduling cycle. Assuming the descheduling cycle is set to e.g. 1h there's no need for such distinction as one can either see the steps in a graph or use `delta{...[1h]}` expression to see the increments per cycle.

metrics/metrics.go
damemi: Maybe we could introduce a copy of these metrics named something like `PodsSkippedLastRun`, for example. It would require some refactoring to the PodEvictor, maybe pass it a timestamp at the start of a new run, and if that timestamp differs from the last one it remembers then reset the counters for those metrics. But I think knowing the totals vs individual runs would be helpful.

I suppose you could get the total by just summing the individual runs, in which case we don't need the `Total..` metrics anymore. Either way, I think this is a good start and the refactors could come as a follow up. Wdyt?

jklaw90:
@pravarag hey, i was looking into similar metrics around evictions. are you still planning on finishing this pr up or do you need any help?

pravarag:
> 

Hey @jklaw90 , I'm working on addressing the latest review comments for this PR. If you have any suggestions/comments, please feel free to share on this PR and I'll definitely check those 🙂 .

metrics/metrics.go
pravarag: With `single run` I meant a single descheduling cycle (somewhat mentioned here in this comment: https://github.com/kubernetes-sigs/descheduler/issues/503#issuecomment-861813502). So just wanted to check will it still be a good idea to introduce this metric or not @damemi @ingvagabund

pkg/descheduler/evictions/evictions.go
pravarag: So this was again inclined towards capturing Failed metrics for a single run. Are you suggesting `PodsEvictedFailed` as compared to `PodsEvicted` metric alone?

pkg/descheduler/evictions/evictions.go
pravarag: I see, I'll update this part. Thanks!

pravarag:
> Within each [descheduling cycle](https://github.com/kubernetes-sigs/descheduler/blob/901a16ecbcaaf41c11318920f8a047c918b6c30d/pkg/descheduler/descheduler.go#L263) a new pod evictor instance is created. Which eliminates any way of telling the pod evictor "a run has ended". The `PodsEvicted` metric is shared across cycles so there's no distinction of which increments belongs to which descheduling cycle. On the other hand when the descheduling cycle is set to e.g. 1h there's no need for such distinction as one can either see the steps in a graph or use `delta{...[1h]}` expression to see the increments per cycle.

@ingvagabund @damemi thanks for the detailed review. So few questions which I have now:
1. Do we still want to capture metrics for a single run?
2. Are there any other metrics that will need to be added or removed as part of this PR?

pravarag:
@damemi @ingvagabund are we planning to merge these changes as part of release 1.24. If so, I would like you suggestions in moving this forward with all the changes required :)

damemi:
@pravarag I would like to merge this, and sorry it's taken so long.

I think the only thing still being discussed was the fact that this is reporting cumulative pod evictions rather than single-run as implied. @ingvagabund is that all that was left to sort out? I wonder if maybe a different metric type, like a histogram, could solve this and provide easier access to single-run metrics

pravarag:
> @pravarag I would like to merge this, and sorry it's taken so long.
> 
> I think the only thing still being discussed was the fact that this is reporting cumulative pod evictions rather than single-run as implied. @ingvagabund is that all that was left to sort out? I wonder if maybe a different metric type, like a histogram, could solve this and provide easier access to single-run metrics

I can give a try by implementing histogram type metric for single-run of pod.

pravarag:
Thanks @ingvagabund @damemi for your reviews on this. And like mentioned above, that we may not need to capture metrics for a single run I guess, I can close this PR then? and since I've already put some effort around it, I still don't want to leave it unfinished as this has come out to be a good learning for me as well. Kindly let me know in ways we can rethink on implementing newer metrics as part of this issue? I'm definitely open for a discussion around it be it on this PR itself or maybe going back to the original issue :)

damemi:
@pravarag sorry about this. it sounds like we've come back around to not needing single-run metrics. I think per-strategy metrics could be a good option though. what do you think?

pravarag:
> @pravarag sorry about this. it sounds like we've come back around to not needing single-run metrics. I think per-strategy metrics could be a good option though. what do you think?

I think that's a good idea @damemi @ingvagabund  and I can work on implementing those. But wanted to check how to approach the same? Do I need to open a `Proposal` or new discussion around it, mayb a new issue? Or we are good to continue in #503 issue itself?

pravarag:
On the other hand, I was thinking if I could just close this PR and open a fresh PR with new (per strategy) changes 🤔 and we can continue in that same issue.

ingvagabund:
Opening a new PR sounds reasonable so we can start fresh. The issue can stay open though.

pravarag:
Closing this PR based on above discussions, will open a fresh one based on newer changes.

PR #649, opened 2021-10-24
Rewrite the LowNodeUtilization documentation section

Jille:
I had serious difficulty figuring out whether I needed HighNodeUtilization or LowNodeUtilization. The names were exactly opposite from what I expected of them. On top of that, I found the documentation about them pretty hard to read.

I rewrote the LowNodeUtilization section with more focus on the descheduling off overutilized nodes. I hope that'll make it clearer for others in the future.

I also bumped the underutilization threshold in the example. I think we want to make clearer to people that they should probably set their underutilized percentages pretty high, because otherwise descheduler won't deschedule pods if they don't have underutilized nodes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,27 +194,35 @@ strategies:
 
 ### LowNodeUtilization
 
-This strategy finds nodes that are under utilized and evicts pods, if possible, from other nodes
-in the hope that recreation of evicted pods will be scheduled on these underutilized nodes. The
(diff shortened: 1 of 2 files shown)


ingvagabund:
+1 for making the description more clear. Also agree with @damemi about the thresholds in the example file. Every cluster wrt. its workload type might have different settings so bumping them does not make much sense here.

PR #653, opened 2021-11-10
Add support for descheduler.alpha.kubernetes.io/noevict annotation

ssanders1449:
Add support for a descheduler.alpha.kubernetes.io/noevict annotation which will make a pod ineligible for eviction even if passes all other tests.

The use case is that I want to, in general, allow eviction of pods with local storage. However, for a few specific deployments (e.g. caches) I want to not evict them.

I know that I can configure the descheduler to not evict pods with local storage and then annotate all of the non-cache deployments with the descheduler.alpha.kubernetes.io/evict annotation. However, there are many more of those and this PR looks simple enough

### DIFF
--- a/README.md
+++ b/README.md
@@ -749,6 +749,8 @@ never evicted because these pods won't be recreated.
 best effort pods are evicted before burstable and guaranteed pods.
 * All types of pods with the annotation `descheduler.alpha.kubernetes.io/evict` are eligible for eviction. This
   annotation is used to override checks which prevent eviction and users can select which pod is evicted.
+* All types of pods with the annotation `descheduler.alpha.kubernetes.io/noevict` are NOT eligible for eviction. This
+  annotation is used to override pods that would otherwise be evicted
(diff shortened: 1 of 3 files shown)


damemi (commented):
Thanks for the contribution @ssanders1449 

This idea has been discussed a couple of times before, mainly in https://github.com/kubernetes-sigs/descheduler/issues/329 (for some background).

The general consensus on this is that [PriorityClass filtering](https://github.com/kubernetes-sigs/descheduler#priority-filtering) is a more functional and descriptive way to achieve the same desired result. It also allows easier updates to be made, for example if you eventually decide to evict some of the pods then you only need to update the threshold in Descheduler rather than remove the annotation from every pod.

For that reason we have chosen not to implement this annotation. If you have more details of your use case or perhaps reasons why Priority filtering won't work, please feel free to share and we can discuss them. Otherwise, our recommendation is to filter the pods you wish to exclude using the already provided methods.

Hope that helps!

PR #661, opened 2021-11-24
Add pod.Status.Reason to the list of reasons

kirecek:
Hello,

I noticed that `RemoveFailedPods` strategy is actually not checking reason in pod status.

This causes that pods in the state like in the example below are ignored by the policy.

```yaml
status:
  message: Node is shutting, evicting pods
  phase: Failed
  reason: Shutdown
```

This MR updates the reasons so descheduler can evict Shutdown pods.

Example of the policy:

```yaml
    strategies:
      "RemoveFailedPods":
        enabled: true
        params:
          failedPods:
            excludeOwnerKinds:
            - Job
            includingInitContainers: true
            reasons:
            - Shutdown
```

### DIFF
--- a/pkg/descheduler/strategies/failedpods.go
+++ b/pkg/descheduler/strategies/failedpods.go
@@ -147,6 +147,10 @@ func validateFailedPodShouldEvict(pod *v1.Pod, strategyParams validatedFailedPod
 	if len(strategyParams.reasons) > 0 {
 		reasons := getFailedContainerStatusReasons(pod.Status.ContainerStatuses)
 
+		if pod.Status.Phase == v1.PodFailed && pod.Status.Reason != "" {
+			reasons = append(reasons, pod.Status.Reason)
+		}
+
 		if strategyParams.includingInitContainers {
 			reasons = append(reasons, getFailedContainerStatusReasons(pod.Status.InitContainerStatuses)...)
 		}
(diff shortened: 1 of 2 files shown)


damemi (commented):
@kirecek thanks for the suggestion, could you please add some tests so we could see this change more in action?
/ok-to-test

kirecek:
Thank you. TBH I am not sure what is the best way to set reason in `buildTestPod`. Will think about it, but meanwhile I've added 2 simple unit tests to see it "in the action".

pkg/descheduler/strategies/failedpods.go
```diff
@@ -143,6 +144,10 @@ func validateFailedPodShouldEvict(pod *v1.Pod, strategyParams validatedFailedPod
 	if len(strategyParams.reasons) > 0 {
 		reasons := getFailedContainerStatusReasons(pod.Status.ContainerStatuses)
 
+		if pod.Status.Reason != "" {
```
ingvagabund, line 147: With `pod.Status.Phase == v1.PodFailed` as well just in case `validateFailedPodShouldEvict` gives more than just failed pods. Even though `ListPodsOnANodeWithFieldSelector` with the `status.phase=Failed` field selector is currently present, the future refactoring may break the assumption. The check is very simple.

ingvagabund (changes requested):
Can you also squash the commits? The changes are not extensive so it's still doable to review the changes in a single PR.

pkg/descheduler/strategies/failedpods.go
kirecek: Applied, thank you.

kirecek:
Of course. Commits squashed.

PR #662, opened 2021-11-29
Drop deprecated flags

ingvagabund:
To drop https://github.com/kubernetes-sigs/descheduler/blob/master/cmd/descheduler/app/options/options.go#L89-L93:
```
	fs.StringVar(&rs.NodeSelector, "node-selector", rs.NodeSelector, "DEPRECATED: selector (label query) to filter on, supports '=', '==', and '!='.(e.g. -l key1=value1,key2=value2)")
	// max-no-pods-to-evict limits the maximum number of pods to be evicted per node by descheduler.
	fs.IntVar(&rs.MaxNoOfPodsToEvictPerNode, "max-pods-to-evict-per-node", rs.MaxNoOfPodsToEvictPerNode, "DEPRECATED: limits the maximum number of pods to be evicted per node by descheduler")
	// evict-local-storage-pods allows eviction of pods that are using local storage. This is false by default.
	fs.BoolVar(&rs.EvictLocalStoragePods, "evict-local-storage-pods", rs.EvictLocalStoragePods, "DEPRECATED: enables evicting pods using local storage by descheduler")
```


pkg/apis/componentconfig/types.go
```diff
@@ -41,15 +41,6 @@ type DeschedulerConfiguration struct {
...
 
-	// Node selectors
-	NodeSelector string
-
-	// MaxNoOfPodsToEvictPerNode restricts maximum of pods to be evicted per node.
-	MaxNoOfPodsToEvictPerNode int
-
-	// EvictLocalStoragePods allows pods using local storage to be evicted.
-	EvictLocalStoragePods bool
-
```
damemi, line 52: I'd like to figure out our eventual plan for the componentconfig types. I started an issue for it a while back here https://github.com/kubernetes-sigs/descheduler/issues/431, but I don't know if it's just redundant with `Policy` or what

pkg/descheduler/descheduler.go
```diff
@@ -109,11 +109,6 @@ func RunDeschedulerStrategies(ctx context.Context, rs *options.DeschedulerServer
 		ignorePvcPods = *deschedulerPolicy.IgnorePVCPods
 	}
 
-	maxNoOfPodsToEvictPerNode := rs.MaxNoOfPodsToEvictPerNode
-	if deschedulerPolicy.MaxNoOfPodsToEvictPerNode != nil {
-		maxNoOfPodsToEvictPerNode = *deschedulerPolicy.MaxNoOfPodsToEvictPerNode
-	}
-
```
damemi, line 116: shouldn't we keep this nil check instead of passing the variable directly?

ingvagabund: I check for the nil inside the pod evictor

pkg/descheduler/evictions/evictions.go
```diff
@@ -110,9 +110,9 @@ func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod, node *v1.Node,
 	if len(reasons) > 0 {
 		reason += " (" + strings.Join(reasons, ", ") + ")"
 	}
-	if pe.maxPodsToEvictPerNode > 0 && pe.nodepodCount[node]+1 > pe.maxPodsToEvictPerNode {
```
damemi, line 113: I believe this was checking `>0` because the default of `0 = unlimited`. So, removing this and making nil the default changes the existing behavior. I think we should still respect 0 as no limit

pkg/descheduler/descheduler.go
damemi: I noticed that after I commented here, sorry. Still, is there any reason not to do that check earlier with the rest of the flags that are checked similarly here?

pkg/descheduler/evictions/evictions.go
ingvagabund: Based on https://github.com/kubernetes-sigs/descheduler#policy-and-strategies:

The policy includes a common configuration that applies to all the strategies:
| Name | Default Value | Description |
|------|---------------|-------------|
| `nodeSelector` | `nil` | limiting the nodes which are processed |
| `evictLocalStoragePods` | `false` | allows eviction of pods with local storage |
| `evictSystemCriticalPods` | `false` | [Warning: Will evict Kubernetes system pods] allows eviction of pods with any priority, including system pods like kube-dns |
| `ignorePvcPods` | `false` | set whether PVC pods should be evicted or ignored |
| `maxNoOfPodsToEvictPerNode` | `nil` | maximum number of pods evicted from each node (summed through all strategies) |

`maxNoOfPodsToEvictPerNode` is nil by default and there's no mention of 0 being the unlimited case. Also, using 0 for unlimited is not intuitive since without any other information setting it to 0 is like saying "do not evict any pod from any node".

pkg/descheduler/descheduler.go
ingvagabund: Nil is a valid value

pkg/descheduler/evictions/evictions.go
damemi: Ah okay, I was looking at the componentconfig types where it wasn't a pointer. So, 0 was just unlimited for the internal representation. Sounds good

pkg/descheduler/descheduler.go
damemi: Got it, the other thread cleared that up for me too thanks

pkg/apis/componentconfig/types.go
ingvagabund: Good point. I will put those back.

damemi (commented):
/lgtm

Issue #641, opened 2021-10-07
Kubernetes 1.23 Release Cycle


damemi:
The 1.23 release cycle is proposed as follows:

- **Monday, August 23, 2021**: Week 1 - Release cycle begins
- **Thursday, September 2, 2021**: Week 2 - [Production Readiness Soft Freeze](https://groups.google.com/g/kubernetes-sig-architecture/c/a6_y81N49aQ)
- **Thursday, September 9, 2021**: Week 3 - Enhancements Freeze
- **Tuesday, November 16, 2021**: Week 13 - Code Freeze
- **Tuesday, November 30, 2021**: Week 15 - Docs must be completed and reviewed
- **Tuesday, December 7, 2021**: Week 16 - **Kubernetes v1.23.0 released**

The Descheduler v0.23 release will be targeted soon after Kubernetes 1.23.0 is released in December

damemi:
Draft release notes:
```
Container image for this release is available at: `k8s.gcr.io/descheduler/descheduler:v0.23.0`

The Helm chart for this release is published [here](https://github.com/kubernetes-sigs/descheduler/releases/tag/descheduler-helm-chart-0.23.0).

## New Features :rainbow:
* #658 Add maxNoOfPodsToEvictPerNamespace policy @JaneLiuL
* #663 Bump go version in go.mod to go1.17 @ingvagabund
* #639 Ignore Pods With Deletion Timestamp @JaneLiuL
* #621 Ignore pods being deleted @uthark
* #662 Added support for cronjob suspend @yutachaos

## Bug Fixes :bug:
* #632 RemoveFailedPods: guard against nil descheduler strategy (e.g. in case of default that loads all strategies) @a7i
* #629 fix duplicate code in node_affinity_test.go @chenkaiyue

## Others :running:
* #665 images: use k8s-staging-test-infra/gcb-docker-gcloud @spiffxp
* #664 fix typo minPodLifeTimeSeconds @stpabhi
* #642 README: Clarify podRestartThreshold applying to the sum over containers @wking
* #627 Add E2E test case cover duplicatepods strategy @JaneLiuL
* #637 Document suspend helm chart configuration option @seanmalloy
```

there are some other PRs in progress that we'll try to merge before the release too

damemi:
The 1.23 bump has now merged https://github.com/kubernetes-sigs/descheduler/pull/667

damemi:
[...]

PR #667, opened 2021-12-06
bump: k8s to 1.23

damemi:
In preparation for the 1.23 release, bumping to the 1.23 release candidate for soak time

ref https://github.com/kubernetes-sigs/descheduler/issues/641


damemi (commented):
/cc @ingvagabund 
helm test is failing, but it's not clear to me why. Maybe @stevehipwell could look too?

stevehipwell:
@damemi haven't PSPs been removed in K8s v1.23?

seanmalloy:
> @damemi haven't PSPs been removed in K8s v1.23?

If I recall correctly they are still deprecated. They will be removed in v1.25.

seanmalloy:
> > @damemi haven't PSPs been removed in K8s v1.23?
> 
> If I recall correctly they are still deprecated. They will be removed in v1.25.

https://kubernetes.io/blog/2021/04/06/podsecuritypolicy-deprecation-past-present-and-future/

stevehipwell:
Thanks @seanmalloy, that would have been too easy.

@damemi are there any API version overrides in the values being used? Maybe for the cronjob? I'm not in front of my computer, so just guessing but might be able to take a proper look tomorrow.

damemi:
@stevehipwell oh yeah! probably https://github.com/kubernetes-sigs/descheduler/pull/577

damemi:
Hm, actually looks like we should still be good with v1beta1 until 1.25 https://kubernetes.io/docs/reference/using-api/deprecation-guide/#cronjob-v125
/retest
maybe it's just a flake?

damemi:
Now that 1.23 has released, I'm just updating this to use the GA tag. Still need to sort out why Helm test is failing, we have some ideas above.

In the meantime, we also need to make the regular updates to test-infra to run against k8s 1.23 and add a test for the release-1.23 branch. If there are any volunteers to make that change please feel free!

damemi:
Ran this locally and saw this error in the pod logs:
```
$ kubectl logs pod/descheduler-ci-27319252-5lpdg -n kube-system
Error: unknown flag: --v
```

which looks like we need the change to initialize klog, similar to [this](https://github.com/openshift/cluster-kube-controller-manager-operator/pull/581/commits/c339a947622142be4e6e0df7799cb2a1445e1eb2). I'll add that and report back

damemi:
Green now, so this should be ready to merge

ingvagabund:
Running locally with `-v=4` I don't see the descheduler logs

damemi:
> Running locally with `-v=4` I don't see the descheduler logs

well that's not good...
/hold

ingvagabund:
Klog seems to work, it's just the `--v` option that's not properly wired to it probably

ingvagabund:
Because https://github.com/kubernetes/kubernetes/pull/106090

damemi:
Thanks again @ingvagabund for solving that problem!

PR #673, opened 2021-12-11
list pods assigned to a node by pod informer cache

### LINKED ISSUE #671: Too slow to list pods when cluster is huge

It always appears becase the way we list pods is not reasonable. See [here](https://github.com/kubernetes-sigs/descheduler/blob/49ad197dfcc96f5c06c86c192899dd339814a3b4/pkg/descheduler/pod/pods.go#L81-L166). We list pods on a node by field selector while it means the descheduler counts on `kube-apiserver` to get the pods on each node. 

When the cluster is large scale, like 3k+ nodes and 90k+ pods. The descheduler is not able to work because every interval it will request `kube-apiserver` for almost 30k+ times(enable 10 plugins and every plugins request for 3k+ times) with a `FieldSelector`. Meanwhile, it leads to a lot of pressure for `kube-apiserver`.

Now I have a very simple idea to fix this problem. We could new a `PodInformer` to establish a cache for all pods. And then we use an informer [indexer](https://github.com/kubernetes-sigs/descheduler/blob/master/vendor/k8s.io/client-go/tools/cache/index.go#L35-L57) to establish a relationship between the pods and their assigned nodes. So we could easily get the pods that assigned to a node and the action is very cheap and quick.

Well, I think this solution is the best way to fix this problem. It makes descheduler work normally in a large scale cluster.

descheduler version: v0.22.1

**Additional context**

Garrybest:
**What this PR does / why we need it**:
Now there is a perf problem about the way we list pods that assigned to a node. When the cluster is large scale, like 3k+ nodes and 90k+ pods. The descheduler is not able to work because every interval it will request kube-apiserver for almost 30k+ times(enable 10 plugins and every plugins request for 3k+ times) with a FieldSelector. Meanwhile, it leads to a lot of pressure for kube-apiserver. So I try to make a revision by using pod informer cache.

**Which issue(s) this PR fixes**:
Fixes #671

**Special notes for your reviewer**:
I'm afraid the PR is too large to review (beacause listing pods is a really basic action and the function is called everywhere in strategy files and test files). So I separate this PR into 3 commits in order to make code review easier.😄

1. **Reform ListPodsOnANode by using pod informer and indexer**
This commit is the most important one. The code is very concise. I establish an indexer to map the pods and their assigned nodes by an informer [indexer](https://github.com/kubernetes-sigs/descheduler/blob/master/vendor/k8s.io/client-go/tools/cache/index.go#L35-L57). So every time we list pods we could use informer cache instead of requesting `kube-apiserver` by filed selectors.

2. Reform all strategies by using getPodsAssignedToNode
This commit let all other files of descheduler adapt to listing pods by function `getPodsAssignedToNode`. Not so important, just some adaptions.

3. Reform all test files
This commit is not important. Just a lot of adaptions of test files and it took me a lot of time to do the adaptions (cause the test cases and strategies are so many). 🤣

### DIFF
--- a/pkg/descheduler/descheduler.go
+++ b/pkg/descheduler/descheduler.go
@@ -33,6 +33,7 @@ import (
 	"sigs.k8s.io/descheduler/pkg/descheduler/evictions"
 	eutils "sigs.k8s.io/descheduler/pkg/descheduler/evictions/utils"
 	nodeutil "sigs.k8s.io/descheduler/pkg/descheduler/node"
+	podutil "sigs.k8s.io/descheduler/pkg/descheduler/pod"
 	"sigs.k8s.io/descheduler/pkg/descheduler/strategies"
 	"sigs.k8s.io/descheduler/pkg/descheduler/strategies/nodeutilization"
 )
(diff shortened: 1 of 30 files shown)


ingvagabund:
@Garrybest you were faster sir :) https://github.com/kubernetes-sigs/descheduler/pull/674

ingvagabund:
@Garrybest I need this change merged pretty soon. I have another PR on the way that relies on the pod lister and removing the field selector. What is your availability?

Garrybest:
> What is your availability?

Hi @ingvagabund, I'm totally available recently and I'm willing to make cooperations.😄

By the way, according to what @damemi said in https://github.com/kubernetes-sigs/descheduler/issues/671#issuecomment-991051305, I have some experimental results and facts to show.

1. I have used an informer indexer, which makes the code simple and concise.

According to the [first commit](https://github.com/kubernetes-sigs/descheduler/pull/673/commits/08ed129a078fa40c41f168ae8911696f61f72d13) of my PR, I build a `GetPodsAssignedToNodeFunc` to provide a method to get pods that assigned to a node. The code is very simple, just establish a inform indexer between pods and assgined nodes.

2. Using filter function to replace the `field selector`.

Actually, the `field selector` is a special kind of pod filter. So I aggregate the origin incoming filter function and our filed selector filter function as a final pod filter in `ListPodsOnANode`. So actually the `field selector` still works but as another form.

3. Some test result in my own clusters of production environment.

Test method: Run `ListPodsOnANode` without any extra filter functions looped by all ready nodes and then measure the elapsed time.
Test machine: A cloud VM with 6 cpu cores and 9.4GB memory.

- A cluster with 233 nodes and 1933 assigned pods.
  - Before: 96s
  - After: 0.2s
  - Accelerate rate: 480 times
- A cluster with 5235 nodes and 42477 assigned pods.
  - Before: More than 30min
  - After: 3.36s
  - Accelerate rate: more than 535 times

I think the benefit of this PR is very obvious. **Note that this test is only listing once.** If we enable all plugins, the accelerate rate will be a very large number since the listing of all nodes will be called more than once.

ingvagabund:
This PR is well done!!!

@Garrybest thank you. I like the changes you made to the code.

/approve
/lgtm

PR #674, opened 2021-12-13
Flip to informers

ingvagabund:
Use pod lister and client side field filtering instead of directly listing pods through a clientset and using field selector.


ingvagabund:
Closing in favor of https://github.com/kubernetes-sigs/descheduler/pull/673

PR #675, opened 2021-12-13
RemoveFailedPods: match casing of field LifeTime to other strategies

a7i:
Strategies such as `PodLifeTime` contain field [`MaxPodLifeTimeSeconds`](https://github.com/kubernetes-sigs/descheduler/blob/49ad197dfcc96f5c06c86c192899dd339814a3b4/pkg/api/v1alpha1/types.go#L107) (notice `LifeTime`) while `RemoveFailedPods` contains field [`MinPodLifetimeSeconds `](https://github.com/kubernetes-sigs/descheduler/blob/49ad197dfcc96f5c06c86c192899dd339814a3b4/pkg/api/v1alpha1/types.go#L113) which does not match the exact casing.

Examples and documents are accurate but this PR tries to address the inconsistency between these two strategies. I realize that this is a breaking API change so I am ok with this PR being closed due to low-value / breaking-change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -524,7 +524,7 @@ strategies:
 This strategy evicts pods that are in failed status phase.
 You can provide an optional parameter to filter by failed `reasons`.
 `reasons` can be expanded to include reasons of InitContainers as well by setting the optional parameter `includingInitContainers` to `true`.
-You can specify an optional parameter `minPodLifetimeSeconds` to evict pods that are older than specified seconds.
+You can specify an optional parameter `minPodLifeTimeSeconds` to evict pods that are older than specified seconds.
 Lastly, you can specify the optional parameter `excludeOwnerKinds` and if a pod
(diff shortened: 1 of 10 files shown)


damemi (commented):
/ok-to-test
@a7i thanks for catching this. As you pointed out, it is a breaking change so I'm not sure we'll go forward with it. I don't have a strong opinion either way (because it's not a terribly big change), but open to hear what others think

ingvagabund:
For the exact reasons mentioned, we can not merge this. Although, we can change the name as part of implementing [v1alpha2 design proposal](https://docs.google.com/document/d/1S1JCh-0F-QCJvBBG-kbmXiHAJFF8doArhDIAKbOj93I/edit#heading=h.6lniphgnfcmi)

damemi:
Like @ingvagabund mentioned, I think this could rather be a change for the next API version
@a7i is it okay to close this PR then?

PR #677, opened 2021-12-14
Use a fake client when evicting pods by individual strategies to accumulate the evictions

ingvagabund:
Currently, when the descheduler is running with the --dry-run on, no strategy actually evicts a pod so every strategy always starts with a complete list of pods. E.g. when the PodLifeTime strategy evicts few pods, the RemoveDuplicatePods strategy still takes into account even the pods eliminated by the PodLifeTime strategy. Which does not correspond to the real case scenarios as the same pod can be evicted multiple times. Instead, use a fake client and evict/delete the pods from its cache so the strategies evict each pod at most once as it would be normally done in a real cluster.

### DIFF
--- a/pkg/descheduler/descheduler.go
+++ b/pkg/descheduler/descheduler.go
@@ -21,10 +21,19 @@ import (
 	"fmt"
 
 	v1 "k8s.io/api/core/v1"
+	policy "k8s.io/api/policy/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/informers"
 	clientset "k8s.io/client-go/kubernetes"
+	fakeclientset "k8s.io/client-go/kubernetes/fake"
+	core "k8s.io/client-go/testing"
 	"k8s.io/klog/v2"
+
+	corev1informers "k8s.io/client-go/informers/core/v1"
+	schedulingv1informers "k8s.io/client-go/informers/scheduling/v1"
 
 	"sigs.k8s.io/descheduler/cmd/descheduler/app/options"
 	"sigs.k8s.io/descheduler/metrics"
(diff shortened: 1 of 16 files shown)


pkg/descheduler/evictions/evictions.go
```diff
@@ -74,6 +75,7 @@ func NewPodEvictor(
 	evictSystemCriticalPods bool,
 	ignorePvcPods bool,
 	evictFailedBarePods bool,
+	metricsEnabled bool,
```
ravisantoshgudimetla, line 78: why this?

pkg/descheduler/descheduler.go
```diff
@@ -67,10 +76,92 @@ func Run(rs *options.DeschedulerServer) error {
...
 type strategyFunction func(ctx context.Context, client clientset.Interface, strategy api.DeschedulerStrategy, nodes []*v1.Node, podEvictor *evictions.PodEvictor, getPodsAssignedToNode podutil.GetPodsAssignedToNodeFunc)
 
+func cachedClient(
+	realClient clientset.Interface,
+	podInformer corev1informers.PodInformer,
+	nodeInformer corev1informers.NodeInformer,
+	namespaceInformer corev1informers.NamespaceInformer,
+	priorityClassInformer schedulingv1informers.PriorityClassInformer,
+) (clientset.Interface, error) {
+	fakeClient := fakeclientset.NewSimpleClientset()
```
ravisantoshgudimetla, line 86: Do we need fakeClient, can't we get away with a local cache of pods just like we do in case of scheduler?

ravisantoshgudimetla (commented):
The intention LGTM. I've few questions about implementation. Can you answer them?

pkg/descheduler/descheduler.go
ingvagabund: I used the fake client so the strategies can stay unchanged. Replacing one client for another. Creating a local cache of pods would require more changes. Perhaps I don't see the full picture. Would you mind describing the approach in more detail? How a strategy is expected to consume the local cache?

pkg/descheduler/evictions/evictions.go
ingvagabund: To avoid accumulating the metrics in case the metrics server is not enabled. To save some instructions.

ingvagabund:
@lixiang233 @seanmalloy PTAL

ravisantoshgudimetla (approved):
/approve

Thank you for working on this @ingvagabund

Issue #652, opened 2021-10-28
livenessProbe


gabegorelick:
**Is your feature request related to a problem? Please describe.**

Descheduler doesn't specify a livenessProbe on its containers. This means that it could theoretically hang in such a way that the container is still running but not actually working. I have no idea if this is actually possible, but I thought I would raise it for discussion. 

**Describe the solution you'd like**

Add a livenessProbe to container manifests.

**Describe alternatives you've considered**

Closing this issue with an explanation as to why a livenessProbe is overkill is totally fine.

**What version of descheduler are you using?**

descheduler version: v0.22.1

**Additional context**

A lot of Kubernetes linters will flag missing livenessProbes, which means extra work when deploying Descheduler in a cluster with (arguably overzealous) admission requirements.

damemi:
@gabegorelick that's a good question. We don't have any health endpoints implemented in the descheduler, so unless there is a generic endpoint we could use then we'll have to add one of those too

ingvagabund:
+1 for the liveness probe. From https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/:
```
The kubelet uses liveness probes to know when to restart a container. For example, liveness probes could catch a deadlock, where an application is running, but unable to make progress. Restarting a container in such a state can help to make the application more available despite bugs.
```

The descheduler has a descheduling [loop](https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/descheduler/descheduler.go#L117) running every DeschedulingInterval seconds. Would be great to take that into account verifying the last time from the last loop was at most DeschedulingInterval ago + time it takes to run all the strategies. Nevertheless, to at least make sure the binary is not hanging, we may provide a simple ping liveness probe. E.g. https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/server/healthz/healthz.go#L42-L55.

a7i:
@JaneLiuL are you still working on this? I have an immediate need to get this into the next release and can submit a PR within a day.

JaneLiuL:
So sorry I forget it, already submit a PR.

seanmalloy:
This was completed in #685.

/close

PR #685, opened 2021-12-24
add  liveness probe

JaneLiuL:
Fix issue: https://github.com/kubernetes-sigs/descheduler/issues/652
Add liveness probe


cmd/descheduler/app/server.go
```diff
@@ -79,6 +80,16 @@ func NewDeschedulerCommand(out io.Writer) *cobra.Command {
 				}
 			}
 
+			// healthz returns true when health check pass
+			var checker []healthz.HealthChecker
+			checker = append(checker,  healthz.NamedCheck("Descheduler", healthz.PingHealthz.Check))
+			newMux := mux.NewPathRecorderMux("healthz")
+			healthz.InstallHandler(newMux, checker...)
+
+			if _, err := SecureServing.Serve(newMux, 0, ctx.Done()); err != nil {
```
ingvagabund, line 89: Unfortunately, this does not work for me. Though this does:
```
diff --git a/cmd/descheduler/app/server.go b/cmd/descheduler/app/server.go
index 555beaf41..129756c3f 100644
--- a/cmd/descheduler/app/server.go
+++ b/cmd/descheduler/app/server.go
@@ -20,6 +20,7 @@ package app
 import (
        "context"
        "io"
+
        "k8s.io/apiserver/pkg/server/healthz"
 
@@ -70,24 +71,19 @@ func NewDeschedulerCommand(out io.Writer) *cobra.Command {
                        }
 
                        ctx := context.TODO()
+                       pathRecorderMux := mux.NewPathRecorderMux("descheduler")
                        if !s.DisableMetrics {
-                               pathRecorderMux := mux.NewPathRecorderMux("descheduler")
                                pathRecorderMux.Handle("/metrics", legacyregistry.HandlerWithReset())
-
-                               if _, err := SecureServing.Serve(pathRecorderMux, 0, ctx.Done()); err != nil {
-                                       klog.Fatalf("failed to start secure server: %v", err)
-                                       return
-                               }
                        }
 
                        // healthz returns true when health check pass
                        var checker []healthz.HealthChecker
-                       checker = append(checker,  healthz.NamedCheck("Descheduler", healthz.PingHealthz.Check))
-                       newMux := mux.NewPathRecorderMux("healthz")
-                       healthz.InstallHandler(newMux, checker...)
+                       checker = append(checker, healthz.NamedCheck("Descheduler", healthz.PingHealthz.Check))
+                       healthz.InstallHandler(pathRecorderMux, checker...)
 
-                       if _, err := SecureServing.Serve(newMux, 0, ctx.Done()); err != nil {
+                       if _, err := SecureServing.Serve(pathRecorderMux, 0, ctx.Done()); err != nil {
                                klog.Fatalf("failed to start secure server: %v", err)
+                               return
                        }
 
                        err := Run(s)
```

Maybe creating two muxes is not the right way to go.

kubernetes/cronjob/cronjob.yaml
```diff
@@ -31,6 +31,14 @@ spec:
               requests:
                 cpu: "500m"
                 memory: "256Mi"
+            livenessProbe:
+              failureThreshold: 3
+              httpGet:
+                path: /healthz
+                port: 10258
+                scheme: HTTP
```
ingvagabund, line 39: HTTPS

kubernetes/deployment/deployment.yaml
```diff
@@ -33,6 +33,14 @@ spec:
           ports:
           - containerPort: 10258
             protocol: TCP
+          livenessProbe:
+            failureThreshold: 3
+            httpGet:
+              path: /healthz
+              port: 10258
+              scheme: HTTP
```
ingvagabund, line 41: HTTPS

kubernetes/job/job.yaml
```diff
@@ -29,6 +29,14 @@ spec:
             requests:
               cpu: "500m"
               memory: "256Mi"
+          livenessProbe:
+            failureThreshold: 3
+            httpGet:
+              path: /healthz
+              port: 10258
+              scheme: HTTP
```
ingvagabund, line 37: HTTPS

cmd/descheduler/app/server.go
JaneLiuL: Yes, you are right, so sorry for I don't found it, I will fix it soon.

kubernetes/job/job.yaml
JaneLiuL: Fix done, please check :)

kubernetes/deployment/deployment.yaml
JaneLiuL: Fix done, please check :)

kubernetes/cronjob/cronjob.yaml
JaneLiuL: Fix done, please check :)

jklaw90:
@JaneLiuL is it possible to update the helm chart while you're at it. if not i can when this merges.

JaneLiuL:
> if not i can when this merges.

Sure. so sorry that I forget the helm chart, I will update soon.

cmd/descheduler/app/server.go
```diff
@@ -68,15 +69,21 @@ func NewDeschedulerCommand(out io.Writer) *cobra.Command {
...
+
+			// healthz returns true when health check pass
+			var checker []healthz.HealthChecker
 
-				if _, err := SecureServing.Serve(pathRecorderMux, 0, ctx.Done()); err != nil {
-					klog.Fatalf("failed to start secure server: %v", err)
-					return
-				}
+			checker = append(checker, healthz.NamedCheck("Descheduler", healthz.PingHealthz.Check))
+			healthz.InstallHandler(pathRecorderMux, checker...)
```
a7i, line 82: You can just pass a single arg to variadic functions (no ned to create a slice)
```suggestion
			healthz.InstallHandler(pathRecorderMux, healthz.NamedCheck("Descheduler", healthz.PingHealthz.Check))
```

kubernetes/cronjob/cronjob.yaml
```diff
@@ -31,6 +31,14 @@ spec:
...
                 cpu: "500m"
                 memory: "256Mi"
+            livenessProbe:
+              failureThreshold: 3
+              httpGet:
+                path: /healthz
+                port: 10258
+                scheme: HTTPS
+              initialDelaySeconds: 3
+              periodSeconds: 3
```
a7i, line 41: Isn't `periodSeconds` of 3 too aggressive given that min interval is 1 minute?

kubernetes/job/job.yaml
```diff
@@ -29,6 +29,14 @@ spec:
             requests:
               cpu: "500m"
               memory: "256Mi"
+          livenessProbe:
```
a7i, line 32: Restart policy of the job is defined as:
```
      restartPolicy: "Never"
```

Then why even have a liveness for types Job / CronJob?

kubernetes/cronjob/cronjob.yaml
JaneLiuL: Which `periodSeconds` you prefer? 30 seconds is better or?

kubernetes/job/job.yaml
JaneLiuL: I check from the https://pkg.go.dev/k8s.io/api/batch/v1 that, job or cronjob both have podspec, so it can have `livenessProbe`. And I think for the cronjob that even the restartPolicy is Never, it need to tell that is it running success or not. What do you think so?

cmd/descheduler/app/server.go
ingvagabund: +1

kubernetes/cronjob/cronjob.yaml
ingvagabund: @a7i would you point me to the place where you see the 1 minute interval?

kubernetes/job/job.yaml
ingvagabund: @JaneLiuL has a good point. In a cluster with high number of nodes and pods a single descheduler pod can run for minutes. Either a strategy needs more time to decide which pods are candidates for eviction or sending request to the `/pods/evict` subresource is throttled.

cmd/descheduler/app/server.go
JaneLiuL: Fix done, please check

kubernetes/cronjob/cronjob.yaml
a7i: The lowest cronjob interval you can have is 1 minute.

a7i: IMO, we should just remove it to use the default value of 10 seconds

kubernetes/job/job.yaml
a7i: If liveness probe fails, kubelet will kill the container. However, the job's restart policy is "Never" so it will never re-start.

ingvagabund: The question here is whether we want to keep an unhealthy pod running or kill it sooner when the liveness probe fails. Even though it's a job. @a7i do you think it is worth taking the case into account?

kubernetes/cronjob/cronjob.yaml
ingvagabund: In the descheduler case there's no significant different between 3s and 10s. Looks fine to me use the default value.

kubernetes/job/job.yaml
a7i: yeah, that's a good point. Let's leave it as-is.

kubernetes/cronjob/cronjob.yaml
JaneLiuL: It seems we align this comment and not need to update, so may I click resolve conversation?

kubernetes/job/job.yaml
JaneLiuL: It seems we align this comment and not need to update, so may I click resolve conversation?

a7i: Yes

kubernetes/cronjob/cronjob.yaml
a7i: > Looks fine to me to use the default value in all the cases.

Basically removing `periodSeconds` should default to 10.

JaneLiuL: Already change  `periodSeconds` to 10, I will click "Resolve conversation"

JaneLiuL:
@a7i  @ingvagabund Already fix all comments, would you help to approve and lgtm label?
Thanks a lot.

ingvagabund:
/approve
/lgtm

@JaneLiuL Thank you for taking the time to add the healthz check!!!
@a7i Many thanks for the review

PR #691, opened 2022-01-06
Eliminated race condition in E2E tests

RyanDevlin:
## Fix
This PR fixes a bug in the E2E tests involving a race condition that can occur in some `Kind` clusters.

## Description
This bug occurs because the first E2E test (currently `TestRemoveDuplicates`) lists the nodes. Since the E2E automation builds a `Kind` cluster and then immediately executes the tests, the nodes may or may not be ready at the time the first test lists them. This can cause unpredictable behavior, because if one or more nodes are not ready when the test lists them, the nodes the test case sees will be tainted with `node.kubernetes.io/not-ready`. This ultimately results in the Deschedulder avoiding evictions because it only has knowledge of tainted nodes, which causes the test case to fail.

## Solution
To fix this, I've added a `waitForNodesReady()` function which is called by the `initializeClient()` setup routine. This allows the test cases to wait until all nodes are `Ready` before running. The waiting period is typically only a second or two.

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -128,7 +128,30 @@ func initializeClient(t *testing.T) (clientset.Interface, coreinformers.NodeInfo
 	sharedInformerFactory.Start(stopChannel)
 	sharedInformerFactory.WaitForCacheSync(stopChannel)
 
+	waitForNodesReady(context.Background(), t, clientSet, nodeInformer)
 	return clientSet, nodeInformer, getPodsAssignedToNode, stopChannel
+}
+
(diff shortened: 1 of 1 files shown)


RyanDevlin:
@ingvagabund Moved from #636 to here.

RyanDevlin:
@ingvagabund It looks like the `descheduler-helm-test` is failing to build the control plane? Any idea why this might occur?

damemi:
/retest
I saw a similar helm test failure in another PR... don't know what's up with that but trying another retest to see if it's a flake. Don't think we merged anything recently that would break it

PR #692, opened 2022-01-06
NonSlidingUntil for deployment

jklaw90:
Use [NonSlidingUntil](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#NonSlidingUntil) over Until in order to account for function duration.  
Trying to make the run times on more of an interval in order to make executions times more predictable. 

Using NonSlidingWindow will make the deployment behave more like the cronjob timing wise.

### DIFF
--- a/pkg/descheduler/descheduler.go
+++ b/pkg/descheduler/descheduler.go
@@ -116,7 +116,7 @@ func RunDeschedulerStrategies(ctx context.Context, rs *options.DeschedulerServer
 		ignorePvcPods = *deschedulerPolicy.IgnorePVCPods
 	}
 
-	wait.Until(func() {
+	wait.NonSlidingUntil(func() {
 		nodes, err := nodeutil.ReadyNodes(ctx, rs.Client, nodeInformer, nodeSelector)
 		if err != nil {
 			klog.V(1).InfoS("Unable to get ready nodes", "err", err)


damemi (commented):
/ok-to-test
Sounds good to me, thanks for the suggestion @jklaw90 !

pkg/descheduler/descheduler.go
```diff
@@ -116,7 +116,7 @@ func RunDeschedulerStrategies(ctx context.Context, rs *options.DeschedulerServer
 		ignorePvcPods = *deschedulerPolicy.IgnorePVCPods
 	}
 
-	wait.Until(func() {
+	wait.NonSlidingUntil(func() {
```
ingvagabund, line 119: The intent of the jitter factor is to make sure all clients sending a request to the apiserver do not sent the request at the same time or too close to each other in time. Given the running time of all the strategies is unknown in advance, using the sliding version helps to increase the randomness.

@jklaw90 I wonder if there are measurements which might tell us if using the non-sliding version has any benefits for the overall resource utilization of the cluster.

> Trying to make the run times on more of an interval in order to make executions times more predictable.

What's the motivation for making the execution time more predictable?

ingvagabund:
@jklaw90 thank you for opening the topic of the descheduler improvements. Is your motivation to improve the debugability, performance or other aspect of the descheduler?

jklaw90:
@ingvagabund debugging and testability are why i think we should use the non sliding window. 
we have smoke tests that were testing the old cronjob which always ran on an exact interval. The `Until` method causes the "interval" to not be respected. From our findings when we set the interval to 10m and we were seeing runs ~12 - 15 minutes apart. 

Here is an example since the documentation wasn't the clearest.  
```
func main() {
	wait.NonSlidingUntil(func() {
		log.Println("running")
		time.Sleep(8 * time.Second)
	}, 10*time.Second, nil)
}
// 2022/01/10 08:47:54 running
// 2022/01/10 08:48:04 running
// 2022/01/10 08:48:14 running
// 2022/01/10 08:48:24 running

func main() {
	wait.Until(func() {
		log.Println("running")
		time.Sleep(8 * time.Second)
	}, 10*time.Second, nil)
}
// 2022/01/10 08:48:50 running
// 2022/01/10 08:49:08 running
// 2022/01/10 08:49:26 running
// 2022/01/10 08:49:44 running
```

ingvagabund:
@jklaw90 @damemi the use cases you mentioned make clear this change improve the user experience. Which is also very valuable.

/approve

PR #695, opened 2022-01-12
make livenessprobe consistent across manifests

a7i:
Addresses 2 inconsistencies from #685 :
+ `livenessProbe.periodSeconds` was not consistent across manifests (`10` in one, `3` in others)
+ `charts/descheduler/templates/cronjob.yaml` was missing livenessProbe but `kubernetes/cronjob/cronjob.yaml` had it

Also:
+ Makes it possible to override via `values.yaml` for Helm


damemi (commented):
/ok-to-test

damemi (commented):
/approve

PR #698, opened 2022-01-20
add v1alpha2 design proposal

JaneLiuL:
Draft for add descheduler v1alpha2 design proposal, not finish yet.
Most for reference to k8s scheduler framework.
welcome for your comments.

I check that one of the PR is for v1alpha2, but it seems long time not to update, so I just raise a PR for discuss. 
If some of you already have new idea or already have new PR for it, feel for fo comments.


JaneLiuL:
kindly ping @damemi

PR #700, opened 2022-01-25
Use the root context cancellation

jklaw90:
using root context cancellation from sigint sigterm to control our http server and our run loop.
noticed the todo context while going through the code, figured we should use the sigterm signal to cancel it.

### DIFF
--- a/cmd/descheduler/app/server.go
+++ b/cmd/descheduler/app/server.go
@@ -20,6 +20,9 @@ package app
 import (
 	"context"
 	"io"
+	"os/signal"
+	"syscall"
+
 	"k8s.io/apiserver/pkg/server/healthz"
 
 	"sigs.k8s.io/descheduler/cmd/descheduler/app/options"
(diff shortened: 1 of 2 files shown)


damemi (commented):
/approve

ingvagabund:
Testing the functionality with the following change applied:
```
diff --git a/pkg/descheduler/descheduler.go b/pkg/descheduler/descheduler.go
index 05239d343..0aeb66fd9 100644
--- a/pkg/descheduler/descheduler.go
+++ b/pkg/descheduler/descheduler.go
@@ -73,6 +73,7 @@ func Run(ctx context.Context, rs *options.DeschedulerServer) error {
        stopChannel := make(chan struct{})
        go func() {
                <-ctx.Done()
+               fmt.Printf("closing the stopChannel")
                close(stopChannel)
        }()
        return RunDeschedulerStrategies(ctx, rs, deschedulerPolicy, evictionPolicyGroupVersion, stopChannel)
```

When running the descheduler from this PR with the change applied locally and sending the both sigterm and sigint signals (`kill -n 15 PID`, `kill -n 2 PID`) causes the descheduler to terminate and print the `closing the stopChannel` string. Whereas, when running the descheduler from the master branch with the following change applied
```
diff --git a/pkg/descheduler/descheduler.go b/pkg/descheduler/descheduler.go
index eaa403f24..157f14453 100644
--- a/pkg/descheduler/descheduler.go
+++ b/pkg/descheduler/descheduler.go
@@ -71,6 +71,11 @@ func Run(rs *options.DeschedulerServer) error {
        }
 
        stopChannel := make(chan struct{})
+       go func() {
+               <-ctx.Done()
+               fmt.Printf("closing the stopChannel")
+               close(stopChannel)
+       }()
        return RunDeschedulerStrategies(ctx, rs, deschedulerPolicy, evictionPolicyGroupVersion, stopChannel)
 }
```
and sending the both signals the string does not get printed when the binary terminates.

/lgtm

Issue #701, opened 2022-01-25
Proposal: Fail Fast Validation


jklaw90:
**Is your feature request related to a problem? Please describe.**
Currently a misconfiguration per strategy would allow descheduler to run with just an error log line every iteration. It may be preferable to fail early, rather than throwing throwing error logs. We won't notice there is a misconfiguration without looking at logs, it'll start and just throw error logs every iteration.

**Describe the solution you'd like**
I think we should change our strategy to an interface like
```go
// what we currently have
type strategyFunction func(ctx context.Context, client clientset.Interface, strategy api.DeschedulerStrategy, nodes []*v1.Node, podEvictor *evictions.PodEvictor, getPodsAssignedToNode podutil.GetPodsAssignedToNodeFunc)

// proposed interface
type Strategy interface {
    Name() string
    Enabled() bool
    Validate(ctx context.Context) error
    Run func(ctx context.Context, nodes []*v1.Node, podEvictor *evictions.PodEvictor, getPodsAssignedToNode podutil.GetPodsAssignedToNodeFunc)
}

// Example
type StrategyX struct {
    client clientset.Interface,
    strategy api.DeschedulerStrategy
}
func NewStrategyX(client clientset.Interface, strategy api.DeschedulerStrategy) Strategy {}
func (s *StrategyX) Name() string
func (s *StrategyX) Enabled() bool
func (s *StrategyX) Validate() error {}
func (s *StrategyX) Run(ctx context.Context, nodes []*v1.Node, podEvictor *evictions.PodEvictor, getPodsAssignedToNode podutil.GetPodsAssignedToNodeFunc) {}


// Example of how our loop can look
...
strategyList := []Strategy{
    strategies.NewStrategyX(rs.Client, strategyConfig),
}
...
// VALIDATE
var configErrors []errors
for _, strategy := range strategyList {
    if err := strategy.Validate(); err != nil {
        klog.Errors(err, "strategy config error")
        configErrors = append(configErrors, err)
    }
}
// FAIL early before we start the loop.
if len(configErrors > 0 {
   return errors.New("invalid descheduler configuration")
}
...
// our usual loop
wait.NonSlidingUntil(func() {
    for _, strategy := range strategyList {
        if strategy.Enabled() {
	    strategy.Run(ctx, nodes, podEvictor, getPodsAssignedToNode)
        } else {
	    klog.Debug("skipping strategy", "strategy", strategy.Name())
        }
    }
[...]

PR #706, opened 2022-01-28
Validation fail fast

jklaw90:
Still a wip, but the goal i'm trying to solve is to fail validation of our config before we begin running.
Alternatively we could add metrics to show the strategies failing due to misconfiguration. 

Relates to https://github.com/kubernetes-sigs/descheduler/issues/701

### DIFF
--- a/pkg/descheduler/descheduler.go
+++ b/pkg/descheduler/descheduler.go
@@ -80,6 +80,13 @@ func Run(ctx context.Context, rs *options.DeschedulerServer) error {
 
 type strategyFunction func(ctx context.Context, client clientset.Interface, strategy api.DeschedulerStrategy, nodes []*v1.Node, podEvictor *evictions.PodEvictor, getPodsAssignedToNode podutil.GetPodsAssignedToNodeFunc)
 
+type Strategy interface {
+	Name() api.StrategyName
+	Enabled() bool
+	Validate() error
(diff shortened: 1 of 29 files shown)


pkg/descheduler/descheduler.go
```diff
@@ -80,6 +80,13 @@ func Run(ctx context.Context, rs *options.DeschedulerServer) error {
 
 type strategyFunction func(ctx context.Context, client clientset.Interface, strategy api.DeschedulerStrategy, nodes []*v1.Node, podEvictor *evictions.PodEvictor, getPodsAssignedToNode podutil.GetPodsAssignedToNodeFunc)
 
+type Strategy interface {
```
ingvagabund, line 83: This actually align with the effort of moving to plugins. During the "descheduling framework" we would first initialize and validate all strategies.

jklaw90: i can close this pr and hold off till a proposal is defined if you'd like.

damemi: I think let's keep it open for now so it's easier to find. Thanks for doing this @jklaw90, I think this has some big picture potential to improve the descheduler

I'd like to start gathering some feedback on what a descheduling framework would look like, so I'll open an issue to get some discussion going on that and include this as an example.

PR #708, opened 2022-01-30
Clarify resource calculations in NodeUtilization strategy Readmes

damemi:
Adding these as it is a frequent source of confusion for users as to why Descheduler is showing different consumption values than kubelet


damemi:
test lol

damemi:
bump
@seanmalloy could you please tag this? thanks!

PR #716, opened 2022-02-07
fix: add imagePullSecrets for deployment resource

eminaktas:
Signed-off-by: emin.aktas <eminaktas34@gmail.com>
Co-authored-by: yasintahaerol <yasintahaerol@gmail.com>
Co-authored-by: Dentrax <furkan.turkal@trendyol.com>

### DIFF
--- a/charts/descheduler/templates/deployment.yaml
+++ b/charts/descheduler/templates/deployment.yaml
@@ -27,6 +27,10 @@ spec:
       priorityClassName: {{ .Values.priorityClassName }}
       {{- end }}
       serviceAccountName: {{ template "descheduler.serviceAccountName" . }}
+      {{- with .Values.imagePullSecrets }}
+      imagePullSecrets:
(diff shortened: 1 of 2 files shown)


charts/descheduler/templates/deployment.yaml
```diff
@@ -27,6 +27,10 @@ spec:
       priorityClassName: {{ .Values.priorityClassName }}
       {{- end }}
       serviceAccountName: {{ template "descheduler.serviceAccountName" . }}
+      {{- with .Values.imagePullSecrets }}
```
damemi, line 30: I'm not an expert on the helm chart syntax, how does `with` handle empty values? Everywhere else we have been using `if`, but if the behavior is the same then it's probably fine

damemi (commented):
/ok-to-test

liangyuanpeng (approved):

Issue #633, opened 2021-09-22
Stop eviction of Pods with Local Storage PVC


zuzzas:
**Is your feature request related to a problem? Please describe.**

Here's the part where the descheduler decides whether Pod has a Local Storage. It does not take into account PVCs that are provisioned via a variation of local storage provisioned. Such as the default `kubernetes.io/no-provisioner`.

https://github.com/kubernetes-sigs/descheduler/blob/d7dc0abd7b1ce2600666101dd9947b380341d632/pkg/utils/pod.go#L113-L122

**Describe the solution you'd like**

I'd like to have a way to specify a list of provisioners for which Pods that are bound to a PVC with such provisioner should not be evicted.

**Describe alternatives you've considered**

There is the `IgnorePVCPods` option, but it applies to all Pods with any PVCs, and I'd like to discriminate between them.

**What version of descheduler are you using?**

descheduler version: v0.21.0

damemi:
Hi @zuzzas, could you explain more how this check would look? Is this another field on the pod spec that would be checked against? I'm assuming this would be a config field for the descheduler that takes a list of label keys?

zuzzas:
>Is this another field on the pod spec that would be checked against?

Resolve StorageClass's [`provisioner`](https://github.com/kubernetes/api/blob/master/storage/v1/types.go#L40-L41) field and

>I'm assuming this would be a config field for the descheduler that takes a list of label keys?

compare it against keys from a config field.

It would also be possible to have a list of StorageClass names, but it would be impossible to default.

EvgenySamoylov:
Hi. We still need this issue to be fixed. Please don't close.

erikschul:
Still an issue. The Kubernetes docs actively recommend against using hostPath, so using local storage PVCs is the most correct approach:

https://kubernetes.io/docs/concepts/storage/volumes/#hostpath

erikschul:
It would also be nice to update the docs that `evictLocalStoragePods` only checks hostPath. The name is misleading. It should probably be `evictHostPathPods`.

ricardomaraschini:
I am taking this over.

PR #721, opened 2022-02-11
Add IgnoreLocalPVCPods parameter

briend:
Motivation:

Currently, pods that have PV `local` or PV `hostPath` storage via a PVC will be evicted even if `evictLocalStoragePods` is `false`. There is no way to prevent this except by ignoring _all_ PVC pods with `ignorePvcPods`.  This PR adds a new option to ignore Pods that have  PVC with a Persistent Volume of type `local` or `hostPath`: https://kubernetes.io/docs/concepts/storage/persistent-volumes/#types-of-persistent-volumes.  Crucially, if we do not `ignoreLocalPvcPods`, this PR adds functionality to delete the PVC before the eviction of the Pod, so that it can be rescheduled on another node if necessary. Unfortunately this is function (apparently) outside the scope of the Eviction Api, so there is a chance to mess things up and delete a PVC for a pod that is not actually evicted. We do a dryrun first to be sure the pod would be evicted, but that doesn't seem 100% guaranteed to be the case when the true eviction happens (even if it is a moment later).

`ignoreLocalPvcPods`: `true` is probably a better safe default as it will only ignore additional pods, but this is a behavior change.  Previously, those pods would be restarted on the same node they were on before (due to the PVC binding), which for many eviction strategies is not productive.

related: https://github.com/kubernetes-sigs/descheduler/issues/633

I have a passing unit-test for `PodLifeTime` strategy, and tested in a real k8s cluster to evict an Elasticsearch StatefulSet using local PVCs (with `local` PVs). The statefulset was evicted successfully (using PodLifeTime strategy) one pod at a time without violating the Pod Disruption Budget, and rescheduled on different nodes with new PVCs.

Current Issues

E2E test fails as the PVCs that I create won't bind to the PVs made in the Kind cluster, so the pods get stuck and timeout.

```
go test -run ^TestRemoveLocalPVCPods$ ${PRJ_PREFIX}/test/e2e/ -v
=== RUN   TestRemoveLocalPVCPods
    e2e_localpvcpods_test.go:54: Creating testing namespace
    e2e_localpvcpods_test.go:61: Creating StorageClass
    e2e_localpvcpods_test.go:78: &StorageClass{ObjectMeta:{local-ssd  e2e-testremovelocalpvcpods    0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [] []  []},Provisioner:kubernetes.io/no-provisioner,Parameters:map[string]string{},ReclaimPolicy:*Delete,MountOptions:[],AllowVolumeExpansion:nil,VolumeBindingMode:*WaitForFirstConsumer,AllowedTopologies:[]TopologySelectorTerm{},}
    e2e_localpvcpods_test.go:90: Creating local Persistent Volumes
    e2e_localpvcpods_test.go:186: Creating localPVC pods
=== RUN   TestRemoveLocalPVCPods/Should_not_evict_pods_when_ignoreLocalPvcPods_is_true
    e2e_localpvcpods_test.go:298: Creating statefulset localpvcpods in e2e-testremovelocalpvcpods namespace
    e2e_duplicatepods_test.go:191: Waiting for 2 pods to be running, got 0 instead
    e2e_duplicatepods_test.go:191: Waiting for 2 pods to be running, got 0 instead
    e2e_duplicatepods_test.go:191: Waiting for 2 pods to be running, got 0 instead
    e2e_duplicatepods_test.go:191: Waiting for 2 pods to be running, got 0 instead
    e2e_duplicatepods_test.go:191: Waiting for 2 pods to be running, got 0 instead
    e2e_duplicatepods_test.go:191: Waiting for 2 pods to be running, got 0 instead
    e2e_duplicatepods_test.go:191: Waiting for 2 pods to be running, got 0 instead
    e2e_duplicatepods_test.go:191: Waiting for 2 pods to be running, got 0 instead
    e2e_duplicatepods_test.go:202: Error waiting for pods running: timed out waiting for the condition
--- FAIL: TestRemoveLocalPVCPods (60.28s)
    --- FAIL: TestRemoveLocalPVCPods/Should_not_evict_pods_when_ignoreLocalPvcPods_is_true (60.04s)
FAIL
FAIL	sigs.k8s.io/descheduler/test/e2e	60.792s
FAIL
```

We can see the PVC is pending, causing the pods to time out as well

```
get pv,pvc,pods -A
NAME                   CAPACITY   ACCESS MODES   RECLAIM POLICY   STATUS      CLAIM   STORAGECLASS   REASON   AGE
persistentvolume/pv1   2G         RWO            Delete           Available           local-ssd               18s
persistentvolume/pv2   2G         RWO            Delete           Available           local-ssd               18s

NAMESPACE                    NAME                                       STATUS    VOLUME   CAPACITY   ACCESS MODES   STORAGECLASS   AGE
e2e-testremovelocalpvcpods   persistentvolumeclaim/www-localpvcpods-0   Pending   www      0                         local-ssd      18s

NAMESPACE                    NAME                                             READY   STATUS    RESTARTS   AGE
e2e-testremovelocalpvcpods   pod/localpvcpods-0                               0/1     Pending   0          18s
kube-system                  pod/coredns-558bd4d5db-c4qnn                     1/1     Running   0          152m
kube-system                  pod/coredns-558bd4d5db-krkgw                     1/1     Running   0          152m
kube-system                  pod/etcd-kind-control-plane                      1/1     Running   0          152m
kube-system                  pod/kindnet-hhcdk                                1/1     Running   0          152m
kube-system                  pod/kindnet-qtvtp                                1/1     Running   0          152m
kube-system                  pod/kindnet-vwdrr                                1/1     Running   0          152m
kube-system                  pod/kube-apiserver-kind-control-plane            1/1     Running   0          152m
kube-system                  pod/kube-controller-manager-kind-control-plane   1/1     Running   0          152m
kube-system                  pod/kube-proxy-gffph                             1/1     Running   0          152m
kube-system                  pod/kube-proxy-ng66v                             1/1     Running   0          152m
kube-system                  pod/kube-proxy-xfgg4                             1/1     Running   0          152m
kube-system                  pod/kube-scheduler-kind-control-plane            1/1     Running   0          152m
local-path-storage           pod/local-path-provisioner-85494db59d-xnv8p      1/1     Running   0          152m
```

Since Kind uses the rancher local-path-provisioner, I even tested that by creating a PVC as described here:

https://github.com/rancher/local-path-provisioner#usage

And that does indeed work, however if I do the same thing with the StatefulSet test (using the default storageClass), the PV is never created and the same situation unfolds (Pending PVCs and Pods).

-----

Another issue appears randomly when the eviction run is finished, seems like a race condition closing the channel twice. It seems to be more of a nuisance than an error, but obviously shouldn't happen.
```
I0209 22:46:11.691166       1 descheduler.go:325] "Number of evicted pods" totalEvicted=0
I0209 22:46:11.691346       1 reflector.go:225] Stopping reflector *v1.Namespace (0s) from k8s.io/client-go/informers/factory.go:134
panic: close of closed channel
```


briend:
> Hi @briend, thank you for the suggestion. I think I can see the use case for this, but it would be nice to discuss some of the implementation options. Our settings for evicting pods with certain storage is growing to be complex so this could even be part of a bigger refactor.
> 
> I see that this is still a WIP, so whenever you are ready to discuss it please remove the tag and ping me or any of the reviewers. You'll also need to sign the CLA as described in the bot comment.

@damemi I had a bit of issue fixing my commits to match my correct email address, and the PR was closed.  Could it be reopened?  Happy to discuss options here, it would be great if the eviction api could handle these local/hostPath PV issues for us!

damemi:
@briend sure, once you get your commits fixed up you should be able to reopen this PR with /reopen (or you can feel free to open another one).

briend:
Opened a new PR for this here: https://github.com/kubernetes-sigs/descheduler/pull/723

PR #722, opened 2022-02-11
feat(leaderelection): impl leader election for HA Deployment

### LINKED ISSUE #720: HA Mode for Deployment using Leader Election
I'm not so sure whether it has been discussed before or not but dropping the idea here so we can discuss further it.

Assume that one of the nodes that contain your _descheduler pod_ has crashed. If we keep auto-scaling the nodes by joining/removing to cluster, so descheduler couldn't do its job properly for a while until the broken node gets fixed. This is where I came up with this idea.

* Implement [leaderelection](https://pkg.go.dev/k8s.io/client-go/tools/leaderelection) client
  * One of the pods should be leader
  * It should set `leases.coordination.k8s.io`
* Ability to set more than `1` pods for deployment

Are we really want to run in HA mode? Would it be over-engineered?

descheduler version: `0.23.1`

**Additional context**
`-`

Dentrax:
Fixes #720

**TODO:**
- [x] update README
- [X] pass namespace dynamically (used `BindLeaderElectionFlags`)
- [X] pass durations dynamically from cmdOptions (used LeaderElectionFlag set)
- [X] better id handle: POD-NAME_UUID (used `os.Hostname` + UUID)
- [X] add e2e test
- [X] update Helm Chart (Pod Static 1)
  - [X] add leaderElection config
  - [X] add podAntiAffinity example
  - [X] enforce replica should not greater than 1 if leaderElection is not set

Signed-off-by: Furkan <furkan.turkal@trendyol.com>
Co-authored-by: Emin <emin.aktas@trendyol.com>
Co-authored-by: Yasin <yasintaha.erol@trendyol.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,8 @@ Table of Contents
   - [Node Fit filtering](#node-fit-filtering)
 - [Pod Evictions](#pod-evictions)
   - [Pod Disruption Budget (PDB)](#pod-disruption-budget-pdb)
+- [High Availability](#high-availability)
+  - [Configure HA Mode](#configure-ha-mode)
 - [Metrics](#metrics)
 - [Compatibility Matrix](#compatibility-matrix)
 - [Getting Involved and Contributing](#getting-involved-and-contributing)
@@ -773,6 +775,23 @@ Setting `--v=4` or greater on the Descheduler will log all reasons why any pod i
 Pods subject to a Pod Disruption Budget(PDB) are not evicted if descheduling violates its PDB. The pods
 are evicted by using the eviction subresource to handle PDB.
 
+## High Availability
+
+In High Availability mode, Descheduler starts [leader election](https://github.com/kubernetes/client-go/tree/master/tools/leaderelection) process in Kubernetes. You can activate HA mode
+if you choose to deploy your application as Deployment.
+
+Deployment starts with 1 replica by default. If you want to use more than 1 replica, you must consider
+enable High Availability mode since we don't want to run descheduler pods simultaneously.
+
+### Configure HA Mode
+
+The leader election process can be enabled by setting `--leader-elect` in the CLI. You can also set
+`--set=leaderElection.enabled=true` flag if you are using Helm.
+
+To get best results from HA mode some additional configurations might require:
+* Configure a [podAntiAffinity](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node) rule if you want to schedule onto a node only if that node is in the same zone as at least one already-running descheduler
+* Set the replica count greater than 1
+
 ## Metrics
 
 | name	| type	| description |
@@ -792,17 +811,16 @@ v0.18 should work with k8s v1.18, v1.17, and v1.16.
 Starting with descheduler release v0.18 the minor version of descheduler matches the minor version of the k8s client
 packages that it is compiled with.
 
-Descheduler  | Supported Kubernetes Version
--------------|-----------------------------
-v0.22        | v1.22
-v0.21        | v1.21
-v0.20        | v1.20
-v0.19        | v1.19
-v0.18        | v1.18
-v0.10        | v1.17
-v0.4-v0.9    | v1.9+
-v0.1-v0.3    | v1.7-v1.8
-
+| Descheduler | Supported Kubernetes Version |
+|-------------|------------------------------|
+| v0.22       | v1.22                        |
+| v0.21       | v1.21                        |
+| v0.20       | v1.20                        |
+| v0.19       | v1.19                        |
+| v0.18       | v1.18                        |
+| v0.10       | v1.17                        |
+| v0.4-v0.9   | v1.9+                        |
+| v0.1-v0.3   | v1.7-v1.8                    |
 
 ## Getting Involved and Contributing
 
--- a/charts/descheduler/templates/deployment.yaml
+++ b/charts/descheduler/templates/deployment.yaml
@@ -6,7 +6,14 @@ metadata:
   labels:
     {{- include "descheduler.labels" . | nindent 4 }}
 spec:
+  {{- if gt .Values.replicas 1.0}}
+  {{- if not .Values.leaderElection.enabled }}
+  {{- fail "You must set leaderElection to use more than 1 replica"}}
+  {{- end}}
+  replicas: {{ required "leaderElection required for running more than one replica" .Values.replicas }}
+  {{- else }}
   replicas: 1
+  {{- end }}
   selector:
     matchLabels:
       {{- include "descheduler.selectorLabels" . | nindent 6 }}
@@ -48,6 +55,7 @@ spec:
             - {{ $value | quote }}
             {{- end }}
             {{- end }}
+            {{- include "descheduler.leaderElection" . | nindent 12 }}
           ports:
             - containerPort: 10258
               protocol: TCP
--- a/cmd/descheduler/app/options/options.go
+++ b/cmd/descheduler/app/options/options.go
@@ -19,13 +19,15 @@ package options
 
 import (
 	"github.com/spf13/pflag"
-
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	apiserveroptions "k8s.io/apiserver/pkg/server/options"
 	clientset "k8s.io/client-go/kubernetes"
-
+	componentbaseconfig "k8s.io/component-base/config"
+	componentbaseoptions "k8s.io/component-base/config/options"
 	"sigs.k8s.io/descheduler/pkg/apis/componentconfig"
 	"sigs.k8s.io/descheduler/pkg/apis/componentconfig/v1alpha1"
 	deschedulerscheme "sigs.k8s.io/descheduler/pkg/descheduler/scheme"
+	"time"
 )
 
 const (
@@ -58,7 +60,17 @@ func NewDeschedulerServer() (*DeschedulerServer, error) {
 }
 
 func newDefaultComponentConfig() (*componentconfig.DeschedulerConfiguration, error) {
-	versionedCfg := v1alpha1.DeschedulerConfiguration{}
+	versionedCfg := v1alpha1.DeschedulerConfiguration{
+		LeaderElection: componentbaseconfig.LeaderElectionConfiguration{
+			LeaderElect:       false,
+			LeaseDuration:     metav1.Duration{Duration: 137 * time.Second},
+			RenewDeadline:     metav1.Duration{Duration: 107 * time.Second},
+			RetryPeriod:       metav1.Duration{Duration: 26 * time.Second},
+			ResourceLock:      "leases",
+			ResourceName:      "descheduler",
+			ResourceNamespace: "kube-system",
+		},
+	}
 	deschedulerscheme.Scheme.Default(&versionedCfg)
 	cfg := componentconfig.DeschedulerConfiguration{}
 	if err := deschedulerscheme.Scheme.Convert(&versionedCfg, &cfg, nil); err != nil {
@@ -75,6 +87,8 @@ func (rs *DeschedulerServer) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&rs.PolicyConfigFile, "policy-config-file", rs.PolicyConfigFile, "File with descheduler policy configuration.")
 	fs.BoolVar(&rs.DryRun, "dry-run", rs.DryRun, "execute descheduler in dry run mode.")
 	fs.BoolVar(&rs.DisableMetrics, "disable-metrics", rs.DisableMetrics, "Disables metrics. The metrics are by default served through https://localhost:10258/metrics. Secure address, resp. port can be changed through --bind-address, resp. --secure-port flags.")
+
+	componentbaseoptions.BindLeaderElectionFlags(&rs.LeaderElection, fs)
 
 	rs.SecureServing.AddFlags(fs)
 }
--- a/pkg/descheduler/descheduler.go
+++ b/pkg/descheduler/descheduler.go
@@ -69,7 +69,22 @@ func Run(ctx context.Context, rs *options.DeschedulerServer) error {
 		return err
 	}
 
-	return RunDeschedulerStrategies(ctx, rs, deschedulerPolicy, evictionPolicyGroupVersion)
+	runFn := func() error {
+		return RunDeschedulerStrategies(ctx, rs, deschedulerPolicy, evictionPolicyGroupVersion)
+	}
+
+	if rs.LeaderElection.LeaderElect && rs.DeschedulingInterval.Seconds() == 0 {
+		return fmt.Errorf("leaderElection must be used with deschedulingInterval")
+	}
+
+	if rs.LeaderElection.LeaderElect && !rs.DryRun {
+		if err := NewLeaderElection(runFn, rsclient, &rs.LeaderElection, ctx); err != nil {
+			return fmt.Errorf("leaderElection: %w", err)
+		}
+		return nil
+	}
+
+	return runFn()
 }
 
 type strategyFunction func(ctx context.Context, client clientset.Interface, strategy api.DeschedulerStrategy, nodes []*v1.Node, podEvictor *evictions.PodEvictor, getPodsAssignedToNode podutil.GetPodsAssignedToNodeFunc)
--- a/pkg/descheduler/leaderelection.go
+++ b/pkg/descheduler/leaderelection.go
@@ -1,0 +1,99 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package descheduler
+
+import (
+	"context"
+	"fmt"
+	"k8s.io/apimachinery/pkg/util/uuid"
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/leaderelection"
+	"k8s.io/client-go/tools/leaderelection/resourcelock"
+	componentbaseconfig "k8s.io/component-base/config"
+	"k8s.io/klog/v2"
+	"os"
+)
+
+// NewLeaderElection starts the leader election code loop
+func NewLeaderElection(
+	run func() error,
+	client clientset.Interface,
+	LeaderElectionConfig *componentbaseconfig.LeaderElectionConfiguration,
+	ctx context.Context,
+) error {
+	var id string
+
+	if hostname, err := os.Hostname(); err != nil {
+		// on errors, make sure we're unique
+		id = string(uuid.NewUUID())
+	} else {
+		// add a uniquifier so that two processes on the same host don't accidentally both become active
+		id = hostname + "_" + string(uuid.NewUUID())
+	}
+
+	klog.V(3).Infof("Assigned unique lease holder id: %s", id)
+
+	if len(LeaderElectionConfig.ResourceNamespace) == 0 {
+		return fmt.Errorf("namespace may not be empty")
+	}
+
+	if len(LeaderElectionConfig.ResourceName) == 0 {
+		return fmt.Errorf("name may not be empty")
+	}
+
+	lock, err := resourcelock.New(
+		LeaderElectionConfig.ResourceLock,
+		LeaderElectionConfig.ResourceNamespace,
+		LeaderElectionConfig.ResourceName,
+		client.CoreV1(),
+		client.CoordinationV1(),
+		resourcelock.ResourceLockConfig{
+			Identity: id,
+		},
+	)
+	if err != nil {
+		return fmt.Errorf("unable to create leader election lock: %v", err)
+	}
+
+	leaderelection.RunOrDie(ctx, leaderelection.LeaderElectionConfig{
+		Lock:            lock,
+		ReleaseOnCancel: true,
+		LeaseDuration:   LeaderElectionConfig.LeaseDuration.Duration,
+		RenewDeadline:   LeaderElectionConfig.RenewDeadline.Duration,
+		RetryPeriod:     LeaderElectionConfig.RetryPeriod.Duration,
+		Callbacks: leaderelection.LeaderCallbacks{
+			OnStartedLeading: func(ctx context.Context) {
+				klog.V(1).InfoS("Started leading")
+				err := run()
+				if err != nil {
+					klog.Error(err)
+				}
+			},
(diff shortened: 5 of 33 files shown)


charts/descheduler/templates/deployment.yaml
```diff
@@ -6,7 +6,14 @@ metadata:
   labels:
     {{- include "descheduler.labels" . | nindent 4 }}
 spec:
+  {{- if gt .Values.replicas 1.0}}
+  {{- if and .Values.leaderElection .Values.leaderElection.enabled }}
```
damemi, line 10: it could also be helpful to check that users aren't doing the inverse, ie setting >1 replicas without leaderelection

pkg/descheduler/descheduler.go
```diff
@@ -75,7 +76,21 @@ func Run(ctx context.Context, rs *options.DeschedulerServer) error {
 		<-ctx.Done()
 		close(stopChannel)
 	}()
-	return RunDeschedulerStrategies(ctx, rs, deschedulerPolicy, evictionPolicyGroupVersion, stopChannel)
+
+	runFn := func() error {
+		return RunDeschedulerStrategies(ctx, rs, deschedulerPolicy, evictionPolicyGroupVersion, stopChannel)
+	}
```
damemi, line 82: It feels a bit odd to be wrapping the function like this, is there a different way we could pass it to `leader.NewLeaderElection` with all of the parameters intact?

pkg/descheduler/descheduler.go
```diff
@@ -75,7 +76,21 @@ func Run(ctx context.Context, rs *options.DeschedulerServer) error {
...
+
+	// if deployed as a Deployment,
+	if rs.LeaderElection.LeaderElect {
+		if rs.DeschedulingInterval.Seconds() > 0 {
+			leader.NewLeaderElection(runFn, rsclient, rs.LeaderElection, ctx)
+			return nil
+		}
+		klog.V(1).InfoS("Warning: You set LeaderElection to True but server does not use Deployment kind.")
```
damemi, line 90: technically `DeschedulingInterval` can be set on cronjobs too, so it's not purely indicative of the kind. Maybe reword this error to not imply that (for example a user would be confused if they are using Deployment kind but forgot to set DeschedulingInterval)

damemi (commented):
/ok-to-test
I had mostly nits on docs and logging, but the overall idea looks pretty straightforward. Thanks for the contribution @Dentrax !

README.md
```diff
@@ -89,6 +91,9 @@ kubectl create -f kubernetes/base/configmap.yaml
 kubectl create -f kubernetes/deployment/deployment.yaml
 ```
 
+Deployment starts with 1 replica by default. If you want use more than 1 replica, please refer to 
+[High Availability](#high-availability) mode since we don't want to run descheduler pods simultaneously.
```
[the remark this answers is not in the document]
Dentrax: Makes sense! Rephrased and moved to `HA` section.

README.md
```diff
@@ -761,6 +766,22 @@ Setting `--v=4` or greater on the Descheduler will log all reasons why any pod i
...
+In High Availability mode, Descheduler starts [leader election](https://github.com/kubernetes/client-go/tree/master/tools/leaderelection) process in Kubernetes. You can activate HA mode
+if you choose to deploy your application as Deployment.
+
+### Configure HA Mode
+
+You can install descheduler in HA using Helm Chart.
+
+To get best results from HA mode some additional configurations might require:
+* Configure [podAntiAffinity](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node) in the Helm Chart
+* Set replica count to at least 3
```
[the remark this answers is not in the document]
Dentrax: Clarified!

charts/descheduler/templates/deployment.yaml
Dentrax: Actually, the following condition already checks it, isn't it?

```
{{- if gt .Values.replicas 1.0}}
{{- if and .Values.leaderElection .Values.leaderElection.enabled }}
```

it means: _setting >1 replicas without leaderelection_

pkg/descheduler/descheduler.go
Dentrax: Had same feeling. But couldn't find a better way to pass `RunDeschedulerStrategies()` func to `leader.NewLeaderElection`, any ideas and advices here?

eminaktas:
There is an error happening [here](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_descheduler/722/pull-descheduler-verify-master/1495740448386322432#1:build-log.txt%3A163). @Dentrax and I tried to add it manually but it didn't work. Can you help us with this?

Thanks

cc @damemi @ingvagabund @Dentrax

pkg/descheduler/descheduler.go
ingvagabund: Does `NewLeaderElection` need to be under a new directory? The function might go under `pkg/descheduler/descheduler.go` or a new `pkg/descheduler/leaderelection.go`. I do not see the `NewLeaderElection` function to be used anywhere but here so it could be just inlined:
```
if rs.LeaderElection.LeaderElect {
		if rs.DeschedulingInterval.Seconds() > 0 {
			...
			leaderelection.RunOrDie(ctx, leaderelection.LeaderElectionConfig{
					Lock:            lock,
					ReleaseOnCancel: true,
					LeaseDuration:   LeaderElectionConfig.LeaseDuration.Duration,
					RenewDeadline:   LeaderElectionConfig.RenewDeadline.Duration,
					RetryPeriod:     LeaderElectionConfig.RetryPeriod.Duration,
					Callbacks: leaderelection.LeaderCallbacks{
						OnStartedLeading: func(ctx context.Context) {
							klog.V(1).InfoS("Started leading")
							err := RunDeschedulerStrategies(ctx, rs, deschedulerPolicy, evictionPolicyGroupVersion, stopChannel)
							if err != nil {
								klog.Error(err)
							}
						},
			...
			return nil
		}
		...
}
```

pkg/descheduler/descheduler.go
```diff
@@ -75,7 +76,21 @@ func Run(ctx context.Context, rs *options.DeschedulerServer) error {
...
+		return RunDeschedulerStrategies(ctx, rs, deschedulerPolicy, evictionPolicyGroupVersion, stopChannel)
+	}
+
+	// if deployed as a Deployment,
+	if rs.LeaderElection.LeaderElect {
+		if rs.DeschedulingInterval.Seconds() > 0 {
+			leader.NewLeaderElection(runFn, rsclient, rs.LeaderElection, ctx)
+			return nil
+		}
+		klog.V(1).InfoS("Warning: DeschedulingInterval must be greater than 0 to use LeaderElection")
```
ingvagabund, line 90: Would you mind returning an error here? If the leader election is enabled but can not proceed, we need to error and tell the user to set the descheduling interval.

ingvagabund:
> T[here](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_descheduler/722/pull-descheduler-verify-master/1495740448386322432#1:build-log.txt%3A163) is an error happening here. @Dentrax and I tried to add it manually but it didn't work. Can you help us with this?

@eminaktas @Dentrax did you try running `./hack/update-generated-conversions.sh`?

eminaktas:
> > There is an error happening [here](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_descheduler/722/pull-descheduler-verify-master/1495740448386322432#1:build-log.txt%3A163). @Dentrax and I tried to add it manually but it didn't work. Can you help us with this?
> 
> @eminaktas @Dentrax did you try running `./hack/update-generated-conversions.sh`?

Yes, we did but it returns an irrational output in the cloud.

```go
+	// WARNING: in.LeaderElection requires manual conversion: does not exist in peer-type
 	out.Logging = in.Logging
 	return nil
 }
-
-// Convert_componentconfig_DeschedulerConfiguration_To_v1alpha1_DeschedulerConfiguration is an autogenerated conversion function.
-func Convert_componentconfig_DeschedulerConfiguration_To_v1alpha1_DeschedulerConfiguration(in *componentconfig.DeschedulerConfiguration, out *DeschedulerConfiguration, s conversion.Scope) error {
-	return autoConvert_componentconfig_DeschedulerConfiguration_To_v1alpha1_DeschedulerConfiguration(in, out, s)
-}
```

When I tried it in my local, It verifies but gives me this output with the current state of this PR:

```bash
➜  descheduler git:(feature/leaderelection) ./hack/update-generated-conversions.sh 
E0222 16:22:22.731871   49961 conversion.go:755] Warning: could not find nor generate a final Conversion function for sigs.k8s.io/descheduler/pkg/apis/componentconfig.DeschedulerConfiguration -> sigs.k8s.io/descheduler/pkg/apis/componentconfig/v1alpha1.DeschedulerConfiguration
E0222 16:22:22.732092   49961 conversion.go:756]   the following fields need manual conversion:
E0222 16:22:22.732100   49961 conversion.go:758]       - LeaderElection
➜  descheduler git:(feature/leaderelection) ./hack/verify-conversions.sh 
E0222 16:23:14.442056   50171 conversion.go:755] Warning: could not find nor generate a final Conversion function for sigs.k8s.io/descheduler/pkg/apis/componentconfig.DeschedulerConfiguration -> ./pkg/apis/componentconfig/v1alpha1.DeschedulerConfiguration
E0222 16:23:14.442272   50171 conversion.go:756]   the following fields need manual conversion:
E0222 16:23:14.442280   50171 conversion.go:758]       - LeaderElection
Generated conversions verified.
```

damemi:
@eminaktas you might need to add the new [`LeaderElection` field](https://github.com/kubernetes-sigs/descheduler/pull/722/files#diff-8cb3a832fd89345b8caf52788b477463aaf7cd78c7a947d55eef89fdc1d5e12fR57) to `pkg/apis/componentconfig/v1alpha1/types.go` as well, so that it can generate the auto-conversion between versions

charts/descheduler/templates/deployment.yaml
damemi: Yup, at least if I'm reading it right, currently this code will fail if `replicas > 1` and `leaderElection != nil`, but we want what you just posted

pkg/descheduler/descheduler.go
Dentrax: We removed the `leader` package and move it into `leaderelection.go` file. I think `NewLeaderElection` function makes things cleaner in a separated file, and it would a bit verbose to put entire logic in the `descheduler.go` file. But I agree with you, we don't use _that_ function again anywhere. 🤔

damemi: Instead of nesting the `interval` check inside the `leaderElect` check, you could break both of those into an error check above. That might open up an option to make this flow better. Something like:
```go
if rs.LeaderElection.LeaderElect && rs.DeschedulingInterval.Seconds() == 0 {
  return fmt.Errorf("LeaderElection must be used with deschedulingInterval...")
}
if rs.LeaderElection.LeaderElect {
  // start leader election call
} else {
  // start normal run call
}
```

something like that. it's a little redundant but it makes the code clearer and removes the deep nesting

eminaktas:
> @eminaktas you might need to add the new [`LeaderElection` field](https://github.com/kubernetes-sigs/descheduler/pull/722/files#diff-8cb3a832fd89345b8caf52788b477463aaf7cd78c7a947d55eef89fdc1d5e12fR57) to `pkg/apis/componentconfig/v1alpha1/types.go` as well, so that it can generate the auto-conversion between versions

I had to change a few things in the code, now it is fine. Thanks for the advice.

charts/descheduler/templates/deployment.yaml
Dentrax: We tested the following scenarios:

1. Works as is
```yaml
replicas: 1
leaderElection: {}
```

2. Got: `You must set leaderElection to use more than 1 replica`
```yaml
replicas: 2
leaderElection: {}
```

3. Got: `You must set leaderElection to use more than 1 replica`
```yaml
replicas: 2
leaderElection:
  enabled: false
```

4. Works smoothly

```yaml
replicas: 2
leaderElection:
  enabled: true
``

5. Works. We allowed this case since user might want to scale their deployment later
```
replicas: 1
leaderElection:
  enabled: true
```

Comment (eminaktas) [reply charts/descheduler/templates/deployment.yaml]: We've also added a note that we think may be useful to users.

```
{{- if eq .Values.kind "Deployment" }}
{{- if eq .Values.replicas 1.0}}
WARNING: You set replica count as 1 and workload kind as Deployment however leaderElection is not enabled. Consider enabling Leader Election for HA mode.
{{- end}}
{{- end}}
```

File: pkg/descheduler/descheduler.go [hunk 7]
```diff
@@ -69,7 +69,22 @@ func Run(ctx context.Context, rs *options.DeschedulerServer) error {
...
-	return RunDeschedulerStrategies(ctx, rs, deschedulerPolicy, evictionPolicyGroupVersion)
+	runFn := func() error {
+		return RunDeschedulerStrategies(ctx, rs, deschedulerPolicy, evictionPolicyGroupVersion)
+	}
+
+	if rs.LeaderElection.LeaderElect && rs.DeschedulingInterval.Seconds() == 0 {
+		return fmt.Errorf("leaderElection must be used with deschedulingInterval")
+	}
+
+	if rs.LeaderElection.LeaderElect {
```
ingvagabund, line 80: `if rs.LeaderElection.LeaderElect && !rs.DryRun {`

No need to have the leader election on when the descheduler is run in the dry run mode. There's a use case for combination of the dry-run mode and the metrics enabled when two or more deschedulers (each with a different configuration) are run in parallel with metrics reported with goal to find a configuration which leads to an optional eviction.

pkg/descheduler/leaderelection.go
```diff
@@ -0,0 +1,86 @@
...
+	LeaderElectionConfig *componentbaseconfig.LeaderElectionConfiguration,
+	ctx context.Context,
+) error {
+	hostname, err := os.Hostname()
+	if err != nil {
+		return fmt.Errorf("unable to get hostname: %v", err)
+	}
+
+	id := fmt.Sprintf("%s-%s", hostname, uuid.New().String())
+
```
ingvagabund, line 44:
```go
if len(LeaderElectionConfig.Namespace) == 0 {
	return fmt.Errorf("namespace may not be empty")
}
if len(LeaderElectionConfig.Name) == 0 {
	return fmt.Errorf("name may not be empty")
}
```

pkg/descheduler/leaderelection.go
```diff
@@ -0,0 +1,86 @@
...
+	client clientset.Interface,
+	LeaderElectionConfig *componentbaseconfig.LeaderElectionConfiguration,
+	ctx context.Context,
+) error {
+	hostname, err := os.Hostname()
+	if err != nil {
+		return fmt.Errorf("unable to get hostname: %v", err)
+	}
+
+	id := fmt.Sprintf("%s-%s", hostname, uuid.New().String())
```
ingvagabund, line 43:
```go
import "k8s.io/apimachinery/pkg/util/uuid"
...
if hostname, err := os.Hostname(); err != nil {
	// on errors, make sure we're unique
	id = string(uuid.NewUUID())
} else {
	// add a uniquifier so that two processes on the same host don't accidentally both become active
	id = hostname + "_" + string(uuid.NewUUID())
}
```

ingvagabund (changes requested):

ingvagabund:
@Dentrax @eminaktas would you mind extending the [e2e test suite](https://github.com/kubernetes-sigs/descheduler/tree/master/test/e2e) with a test validating only a single instance is active at a time? E.g.
- create 2 namespaces (e.g. A, B)
- create e.g. 5 pods in each
- wait for 5 or more seconds, configure both deschedulers to evict pods running longer than 5 seconds, each descheduler targeting a specific namespace (includeNamespaces)
- validate only pods from a single namespace (A xor B) are evicted.

This will be a bit tricky as you will need to invoke `Run` from the `pkg/descheduler/descheduler.go` or some higher function so you are running both deschedulers from within the same binary.

Dentrax:
> @Dentrax @eminaktas would you mind extending the [e2e test suite](https://github.com/kubernetes-sigs/descheduler/tree/master/test/e2e?rgh-link-date=2022-03-14T09%3A38%3A02Z) with a test validating only a single instance is active at a time?

We've tried to write a simple test by following your instructions, here is test output:

```
=== RUN   TestLeaderElection
    e2e_leaderelection_test.go:45: Creating testing namespace e2e-testleaderelection-a
    e2e_leaderelection_test.go:52: Creating testing namespace e2e-testleaderelection-b
    e2e_leaderelection_test.go:175: Creating deployment leaderelection for namespace e2e-testleaderelection-a
    e2e_leaderelection_test.go:175: Creating deployment leaderelection for namespace e2e-testleaderelection-b
    e2e_leaderelection_test.go:203: Pod leaderelection-76dbf88fff-kpv7d not running yet, is Pending instead
    e2e_leaderelection_test.go:203: Pod leaderelection-76dbf88fff-v98g4 not running yet, is Pending instead
    e2e_leaderelection_test.go:94: starting deschedulers
I0317 12:40:08.240823   75207 leaderelection.go:248] attempting to acquire leader lease kube-system/descheduler...
I0317 12:40:08.270859   75207 leaderelection.go:258] successfully acquired lease kube-system/descheduler
I0317 12:40:11.232282   75207 leaderelection.go:248] attempting to acquire leader lease kube-system/descheduler...
    e2e_leaderelection_test.go:135: Total evicted Pods count for e2e-testleaderelection-a: is 5 and e2e-testleaderelection-b: is 14
--- PASS: TestLeaderElection (43.24s)
PASS
ok      sigs.k8s.io/descheduler/test/e2e        44.618s
```

We also had to put two new policy files since we couldn't pass a `PolicyConfig` into `DeschedulerServer` struct, so we bring the following files:

```
test/e2e/policy_leaderelection_a.yaml
test/e2e/policy_leaderelection_b.yaml
```

Eventually, we're asserting that only pods from a single namespace are evicted:
```go
singleNamespaceEvicted := (podsA == 5 && podsB != 5) || (podsA != 5 && podsB == 5)
```


Thanks, @ingvagabund! I think we've resolved the following above reviews.

a7i:
Re: RBAC, I would have expected the `Role` definition to include the following:
```
  - apiGroups: ["coordination.k8s.io"]
    resources: ["leases"]
    verbs: ["create"]
  - apiGroups: ["coordination.k8s.io"]
    resources: ["leases"]
    resourceNames: ["descheduler"]
    verbs: ["get", "patch", "delete"]
```

test/e2e/e2e_leaderelection_test.go
```diff
@@ -0,0 +1,195 @@
...
+	s, err := options.NewDeschedulerServer()
+	if err != nil {
+		t.Fatalf("unable to initialize server: %v", err)
+	}
+	s.Client = clientSet
+	s.DeschedulingInterval = 5 * time.Second
+	s.LeaderElection.LeaderElect = true
+	s.KubeconfigFile = os.Getenv("KUBECONFIG")
+
+	s1 := s
```
ingvagabund, line 90:
```go
s1, err := options.NewDeschedulerServer()
s2, err := options.NewDeschedulerServer()
```
rather than copying `s` to avoid sharing data.

ingvagabund (changes requested):

test/e2e/e2e_leaderelection_test.go
Dentrax: Actually, I did it intentionally to avoid duplication for the following lines:

```go
if err != nil {
		t.Fatalf("unable to initialize server: %v", err)
	}
	s.Client = clientSet
	s.DeschedulingInterval = 5 * time.Second
	s.LeaderElection.LeaderElect = true
	s.KubeconfigFile = os.Getenv("KUBECONFIG")
```

ingvagabund: It is ok to duplicate those when it comes to the initialization. Other option is to create a function which will produce the same. Invoke twice then.

eminaktas: Hi @ingvagabund, I updated the changes as you mentioned.

pkg/descheduler/leaderelection.go
```diff
@@ -0,0 +1,99 @@
...
+			},
+			OnStoppedLeading: func() {
+				klog.V(1).InfoS("Leader lost")
+			},
+			OnNewLeader: func(identity string) {
+				// Just got the lock
+				if identity == id {
+					return
+				}
+				klog.V(1).Infof("New leader elected: ", identity)
```
ingvagabund, line 94: `"New leader elected: "` -> `"New leader elected: %v"` (missing %v)

cmd/descheduler/app/options/options.go
```diff
@@ -58,7 +60,17 @@ func NewDeschedulerServer() (*DeschedulerServer, error) {
 }
 
 func newDefaultComponentConfig() (*componentconfig.DeschedulerConfiguration, error) {
-	versionedCfg := v1alpha1.DeschedulerConfiguration{}
+	versionedCfg := v1alpha1.DeschedulerConfiguration{
+		LeaderElection: componentbaseconfig.LeaderElectionConfiguration{
+			LeaderElect:       false,
+			LeaseDuration:     metav1.Duration{Duration: 15 * time.Second},
+			RenewDeadline:     metav1.Duration{Duration: 10 * time.Second},
+			RetryPeriod:       metav1.Duration{Duration: 2 * time.Second},
```
ingvagabund, line 68: This will generate a lot of requests to the apiserver. What about to go with LeaseDuration=137s, RenewDealine=107s, RetryPeriod=26s by default? Those are the defaults we use in OpenShift by experience.

test/e2e/e2e_leaderelection_test.go
```diff
@@ -0,0 +1,201 @@
...
+			t.Errorf("unable to start descheduler: %v", err)
+			return
+		}
+	}()
+
+	defer clientSet.CoordinationV1().Leases(s1.LeaderElection.ResourceNamespace).Delete(ctx, s1.LeaderElection.ResourceName, metav1.DeleteOptions{})
+	defer clientSet.CoordinationV1().Leases(s2.LeaderElection.ResourceNamespace).Delete(ctx, s2.LeaderElection.ResourceName, metav1.DeleteOptions{})
+
+	// wait for a while so all the pods are 5 seconds older
+	time.Sleep(5 * time.Second)
```
ingvagabund, line 125: 7 is better since 5s is exactly the boundary given the second descheduler instance is created after `time.Sleep(1 * time.Second)`.

ingvagabund:
Few more nits I noticed. Durations is the most important here. We need to set those wisely to reduce the traffic between the descheduler instances and the apiserver.

cmd/descheduler/app/options/options.go
Dentrax: Thanks, updated. We forgot to recheck these values whether these appropriate or not. 

Did you use any calculation formula to get these values? For example, why `137` instead of `120` (which is 2m)?

ingvagabund: https://github.com/openshift/enhancements/blob/master/CONVENTIONS.md#handling-kube-apiserver-disruption will help with the context though it is not a must. We can go lower though I would stick with at least 60s for the renew deadline and lease duration.

PR #723, opened 2022-02-11
Add IgnoreLocalPVCPods parameter

briend:
New PR because I messed up my email address on the previous commits here https://github.com/kubernetes-sigs/descheduler/pull/721

Motivation:

Currently, pods that have PV `local` or PV `hostPath` storage via a PVC will be evicted even if `evictLocalStoragePods` is `false`. There is no way to prevent this except by ignoring _all_ PVC pods with `ignorePvcPods`.  This PR adds a new option to ignore Pods that have  PVC with a Persistent Volume of type `local` or `hostPath`: https://kubernetes.io/docs/concepts/storage/persistent-volumes/#types-of-persistent-volumes.  

`ignoreLocalPvcPods`: `true` is probably a better safe default as it will only ignore additional pods, but this is a behavior change.  Previously, those pods would be restarted on the same node they were on before (due to the PVC binding), which for many eviction strategies is not productive.

related: https://github.com/kubernetes-sigs/descheduler/issues/633

I have a passing unit-test for `PodLifeTime` strategy, and tested in a real k8s cluster to ignore an Elasticsearch StatefulSet using local PVCs (with `local` PVs). 

-----

An issue appears randomly when the eviction run is finished, seems like a race condition closing the channel twice. It seems to be more of a nuisance than an error, but obviously shouldn't happen.
```
I0209 22:46:11.691166       1 descheduler.go:325] "Number of evicted pods" totalEvicted=0
I0209 22:46:11.691346       1 reflector.go:225] Stopping reflector *v1.Namespace (0s) from k8s.io/client-go/informers/factory.go:134
panic: close of closed channel
```


damemi:
/ok-to-test
@briend could you please squash your commits down to the fewest logical steps? It also looks like you inadvertently pulled in an extra commit (https://github.com/kubernetes-sigs/descheduler/pull/723/commits/2910f63e5cb692937f5c5ec95af0a4d324808bf4)

charts/descheduler/templates/clusterrole.yaml
```diff
@@ -24,6 +24,12 @@ rules:
 - apiGroups: ["scheduling.k8s.io"]
   resources: ["priorityclasses"]
   verbs: ["get", "watch", "list"]
+- apiGroups: [""]
+  resources: ["persistentvolumeclaims"]
+  verbs: ["get", "watch", "list", "delete"]
+- apiGroups: [""]
+  resources: ["persistentvolumes"]
+  verbs: ["get", "watch", "list"]
```
a7i, line 32: Should these be guarded by the policy's condition?

```suggestion
{{- if not (default true | deschedulerPolicy.IgnoreLocalPVCPods) }}
- apiGroups: [""]
  resources: ["persistentvolumeclaims"]
  verbs: ["get", "watch", "list", "delete"]
- apiGroups: [""]
  resources: ["persistentvolumes"]
  verbs: ["get", "watch", "list"]
{{ end }}
```

pkg/descheduler/descheduler.go
```diff
@@ -138,7 +140,29 @@ func cachedClient(
 
 	for _, item := range namespaces {
 		if _, err := fakeClient.CoreV1().Namespaces().Create(context.TODO(), item, metav1.CreateOptions{}); err != nil {
-			return nil, fmt.Errorf("unable to copy node: %v", err)
+			return nil, fmt.Errorf("unable to copy namespace: %v", err)
+		}
+	}
+
+	persistentVolumes, err := pvInformer.Lister().List(labels.Everything())
```
a7i, line 147: Can these also be guarded by conditions for when it's actually needed?

pkg/utils/pvc.go
```diff
@@ -0,0 +1,54 @@
...
+		if err != nil {
+			return nil, err
+		}
+		return pvc, nil
+	}
+	return nil, nil
+}
+
+// GetPV returns a PersistentVolume matching provided name string
+func GetPV(ctx context.Context, client clientset.Interface, name string) (pv *v1.PersistentVolume, err error) {
```
a7i, line 45: maybe name the file volumes to allow for PV and PVC utils?

briend: Agree, it's a bit inaccurate to name it just pvc.

charts/descheduler/templates/clusterrole.yaml
briend: Good idea, if `ignoreLocalPvcPods` is `true` then there is no reason to have `delete` ability for PVCs

pkg/descheduler/descheduler.go
briend: I've only added one parameter, `ignoreLocalPvcPods`, and in either case (true for false) it needs to be able check the PVC and the PV to determine if it is indeed a "local" (or hostPath) type PV.  We could break it up into two config options, one for checking and one for deleting, but that might get (more) confusing?

JaneLiuL (commented):
there have duplicate `if isPodWithLocalPVC`

pkg/descheduler/evictions/evictions.go
```diff
@@ -137,7 +140,48 @@ func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod, node *v1.Node,
...
+		if err != nil {
+			klog.InfoS("Error checking for Local PVC for pod", "pod", klog.KObj(pod), "error", err)
+			return false, err
+		}
+		if isPodWithLocalPVC {
+			klog.V(1).InfoS("Pod HAS Local PVC Storage")
+		} else {
+			klog.V(1).InfoS("Pod does NOT have Local PVC Storage")
+		}
+		if isPodWithLocalPVC {
```
JaneLiuL, line 166: there are two `if isPodWithLocalPVC` and duplicate with line 161.

pkg/descheduler/evictions/evictions.go
```diff
@@ -137,7 +140,48 @@ func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod, node *v1.Node,
...
+			// dry run first to see if we should delete the PVC
+			klog.V(1).InfoS("Pod has Local PVC, performing dryrun eviction")
+			dryRunErr := evictPod(ctx, pe.client, pod, pe.policyGroupVersion, true)
```
JaneLiuL, line 169: don't need a new variable, just use ` evictPod(ctx, pe.client, pod, pe.policyGroupVersion, true)` is enough

pkg/descheduler/evictions/evictions.go
briend: I removed the redundant conditional in there

pkg/descheduler/evictions/evictions.go
briend: Thanks, I simplified it a bit

pkg/descheduler/evictions/evictions.go
```diff
@@ -76,6 +77,7 @@ func NewPodEvictor(
 	ignorePvcPods bool,
 	evictFailedBarePods bool,
 	metricsEnabled bool,
+	ignoreLocalPvcPods bool,
```
ingvagabund, line 80: Would you mind moving the `ignoreLocalPvcPods` next to `ignorePvcPods`?

pkg/descheduler/evictions/evictions.go
```diff
@@ -137,7 +140,48 @@ func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod, node *v1.Node,
 		return false, fmt.Errorf("Maximum number %v of evicted pods per %q namespace reached", *pe.maxPodsToEvictPerNamespace, pod.Namespace)
 	}
 
-	err := evictPod(ctx, pe.client, pod, pe.policyGroupVersion)
+	if pe.ignoreLocalPvcPods {
+		klog.V(1).InfoS("checking if pod has Local PVC Storage")
+		isPodWithLocalPVC, _ := utils.IsPodWithLocalPVC(ctx, pe.client, pod)
+		if isPodWithLocalPVC {
+			return false, fmt.Errorf("This pod %q/%q has Local PVC Storage and ignoreLocalPvcPods is set to true", pod.Namespace, pod.Name)
```
ingvagabund, line 147: s/This pod/Pod

pkg/descheduler/evictions/evictions.go
ingvagabund: There can be other factors in play such as PDB violation. Running `evictPod` in the dry mode will only check the client side constraints.

ingvagabund (commented):
Has "removing the PVCs before evicting pods" approach been discussed with the SIG-storage community? I do not have the right knowledge here to conclude if this approach is valid. The PVC controller might try to recreate the resource. Or, there might be other resources that need to be deleted alongside. Also, there's no guarantee a pod gets evicted right after its underlying PVC(s) get deleted. Unfortunately, deleting a pod with PVC(s) is not an atomic operation.

pkg/descheduler/evictions/evictions.go
briend: Pod Disruption Budget was the primary thing I was testing against (and relying upon), and seemed to work fine with the dryrun eviction. I will double check.  I know the dryrun option doesn't work at all with the Fake client, which is unfortunate, but it would be really surprising if dryrun wasn't accurate against a real k8s api.

briend:
> Has "removing the PVCs before evicting pods" approach been discussed with the SIG-storage community? I do not have the right knowledge here to conclude if this approach is valid. The PVC controller might try to recreate the resource. Or, there might be other resources that need to be deleted alongside. Also, there's no guarantee a pod gets evicted right after its underlying PVC(s) get deleted. Unfortunately, deleting a pod with PVC(s) is not an atomic operation.

I don't know if this is the right approach either, but at the very least this technique could be used to avoid needlessly evicting these types of pods (since they would simply get rescheduled back on the same node).  I'll try to find any discussions relevant to this over here https://github.com/kubernetes/community/tree/master/sig-storage and link to them here

pkg/descheduler/evictions/evictions.go
briend: Sure, wasn't sure what the best practice was there

pkg/descheduler/evictions/evictions.go
briend: Deleting the PVC definitely seems like a sketchy idea, even if it appears to work. I've updated this PR to only _ignore_ these pods instead of evicting them.

briend:
I mentioned this in a thread, but this PR only ignores pods with local PVCs, instead of trying to delete anything. 
/retest

JaneLiuL:
hello,  Thank you for the improvements. Would you mind squashing the commits into a single commit?

pkg/utils/pod.go
```diff
@@ -136,6 +138,50 @@ func IsPodWithPVC(pod *v1.Pod) bool {
 	return false
 }
 
+// DeleteLocalPVCSForPod will loop through a pods Volumes and delete any PVCs that are of Local type
+// These PVCs will likely be in Terminating state until the pod is evicted.
+func DeleteLocalPVCsForPod(ctx context.Context, client kubernetes.Interface, pod *v1.Pod) (err error) {
```
ingvagabund, line 143: Is this function still needed?

pkg/utils/volume.go
```diff
@@ -0,0 +1,54 @@
...
+		klog.ErrorS(err, "Failed to get PV for PVC")
+		return false, err
+	}
+	return false, nil
+}
+
+// GetPVC returns a PersistentVolumeClaim matching provided name string in namespace
+func GetPVC(ctx context.Context, client clientset.Interface, name string, namespace string) (pv *v1.PersistentVolumeClaim, err error) {
+	if name != "" {
+		pvc, err := client.CoreV1().PersistentVolumeClaims(namespace).Get(ctx, name, metav1.GetOptions{})
```
ingvagabund, line 35: This performs a direct request to the server. Getting it invoked for every pod to be evicted will create a lot of traffic between the descheduler and the apiserver. Better to use informers here.

pkg/utils/volume.go
```diff
@@ -0,0 +1,54 @@
...
+		}
+		return pvc, nil
+	}
+	return nil, nil
+}
+
+// GetPV returns a PersistentVolume matching provided name string
+func GetPV(ctx context.Context, client clientset.Interface, name string) (pv *v1.PersistentVolume, err error) {
+	if name != "" {
+		pv, err := client.CoreV1().PersistentVolumes().Get(ctx, name, metav1.GetOptions{})
```
ingvagabund, line 47: This performs a direct request to the server. Getting it invoked for every pod to be evicted will create a lot of traffic between the descheduler and the apiserver. Better to use informers here.

briend:
Closing as our original use case for this is too dangerous to implement (deleting PVCs automatically) 
https://github.com/kubernetes-sigs/descheduler/issues/633

Issue #690, opened 2022-01-05
Evicting pods is too deterministic


HitkoDev:
**Is your feature request related to a problem? Please describe.**

Consider having the following pods (all pods have the same priority and QoS class):
```
bar-cchwg
bar-fkpbw
bar-ictdt
bar-oanma
bar-ypagq
foo-goiou
foo-pclma
foo-vzxpc
foo-wozdq
foo-yxgfx
```

Since they share the same priority and QoS class, descheduler will always evict them in the same (alphabetical) order (`bar` before `foo`). This behaviour implicitly gives pods a secondary "priority" based on name, and causes `bar` pods to be evicted over and over while `foo` pods never get evicted despite having the same priority. It also causes all pods belonging to the same group to be evicted together instead of evicting pods across all groups with the same priority.

**Describe the solution you'd like**

Pods with the same priority and QoS class should have the same probability of being evicted.

**Describe alternatives you've considered**

Changing priority or QoS class doesn't help, since pods within the same group still get evicted in deterministic order.

**What version of descheduler are you using?**

descheduler version: 0.22.1

lianghao208:
Pods with the same priority and QoS class should be evicted in:
1. creation timestamp order (the older pods are more likely to be evicted)
2. random order
I wonder which order is better.

maartengo:
Edit: I meant option 1 😄

HitkoDev:
Option 1 doesn't really solve the problem imo, since most of the time all `foo` pods will have one timestamps, and all `bar` pods will have another timestamp. Descheduler will therefore still evict all pods from one group before starting to evict pods from the other group.

The only improvement option 1 has over the current implementation is that it will *eventually* get to all pods, whereas the current implementation keeps evicting the same pods over and over.

lianghao208:
> Descheduler will therefore still evict all pods from one group before starting to evict pods from the other group.

Option 1 can prevent the same pods being evicted over and over. Because pods `bar` will be newer than pods `foo` after the eviction.

damemi:
I think adding timestamp as another sorting criteria will help this. We did something similar for kube-controller-manager to add pseudo-random selection for scaling down replica sets as well, so there is some precedent for it

HitkoDev:
> > Descheduler will therefore still evict all pods from one group before starting to evict pods from the other group.
> 
> Option 1 can prevent the same pods from being evicted over and over. Because pods `bar` will be newer than pods `foo` after the eviction.

[...]

PR #724, opened 2022-02-14
feat: Add EvictionStrategy option for DeschedulerPolicy

lianghao208:
fix: #690 

According to the discussion in https://github.com/kubernetes-sigs/descheduler/issues/690

Pods eviction may cause pods sharing the same priority and QoS class to be evicted over and over. Because pods are in alphabetically order in podIndexer.

If pods share the same priority and QoS class, the older pods should be more likely to be evicted.


knelasevero:
@a7i @ingvagabund @damemi with the new framework/plugin approach, this PR could be a new custom plugin, am I right? Is this PR now obsolete with the latest changes in the descheduler?

damemi:
> @a7i @ingvagabund @damemi with the new framework/plugin approach, this PR could be a new custom plugin, am I right? Is this PR now obsolete with the latest changes in the descheduler?

Yeah, this change would be better as a Sorting plugin in the new framework code

knelasevero:
We let the bot close the issue, but we left the PR open? What's the plan here?

PR #726, opened 2022-02-15
Eviction Logs

jklaw90:
Currently it's hard to tell where the pod is being evicted from in the `RemovePodsViolatingTopologySpreadConstraint` strategy. This should make it easier to debug for any strategy.


ingvagabund:
@jklaw90 appreciated, this is an useful improvement

damemi (commented):
/lgtm

PR #729, opened 2022-02-16
Skip draining nodes that are already cordoned

adriananeci:
We ran into a situation were we had to disable `Descheduler` because it was attempting to drain nodes which were cordoned, causing conflicts with cluster upgrade.

We found that `Descheduler` was attempting to evict _every outdated Node in the cluster_ during the worker re-image because it reacts to a Node taint which is added by the node controller when a Node is marked as unschedulable. This caused the cluster to be disrupted at a high rate which made it very unstable.

This PR ignores nodes that are marked as `unschedulable` so that `Descheduler` won't conflict with node controller taints.
Not sure why the code was in place, but commented out.

### DIFF
--- a/pkg/descheduler/node/node.go
+++ b/pkg/descheduler/node/node.go
@@ -89,10 +89,10 @@ func IsReady(node *v1.Node) bool {
 		}*/
 	}
 	// Ignore nodes that are marked unschedulable
-	/*if node.Spec.Unschedulable {
-		klog.V(4).InfoS("Ignoring node since it is unschedulable", "node", klog.KObj(node.Name))
+	if node.Spec.Unschedulable {
+		klog.V(4).InfoS("Ignoring node since it is unschedulable", "node", klog.KObj(node))
 		return false
-	}*/
+	}
 	return true
 }
 
--- a/pkg/descheduler/node/node_test.go
+++ b/pkg/descheduler/node/node_test.go
@@ -39,16 +39,16 @@ func TestReadyNodes(t *testing.T) {
 	node5.Status.Conditions = []v1.NodeCondition{{Type: v1.NodeReady, Status: v1.ConditionFalse}}
 
 	if !IsReady(node1) {
-		t.Errorf("Expected %v to be ready", node2.Name)
+		t.Errorf("Expected %v to be ready", node1.Name)
 	}
 	if !IsReady(node2) {
-		t.Errorf("Expected %v to be ready", node3.Name)
+		t.Errorf("Expected %v to be ready", node2.Name)
 	}
 	if !IsReady(node3) {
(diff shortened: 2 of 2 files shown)


damemi (commented):
/ok-to-test
I'm not sure why this code is in place/commented-out either, but it looks like it's been here for a while. We should look into this as a bigger task

This change should also probably be gated by a feature setting. I'll have to look more into the details of what sets `spec.Unschedulable` on a node but I believe there can be valid reasons for wanting to evict from unschedulable nodes.

And as a side note, technically the descheduler doesn't immediately react to any taints on a node. If the Node controller is tainting a node and instantly the node is being drained, that is the effect of [taint based evictions](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/#taint-based-evictions), which is handled by the node lifecycle controller. (Though in your case, you might have seen that the descheduler is eventually evicting the nodes from its logs, I don't know, just pointing out TBE in case you weren't aware of it)

adriananeci:
@damemi Thanks for your thoughts.
Please be aware that in case of `Taint based Evictions`:
```
Note:
Kubernetes automatically adds a toleration for node.kubernetes.io/not-ready and node.kubernetes.io/unreachable with tolerationSeconds=300, unless you, or a controller, set those tolerations explicitly.

These automatically-added tolerations mean that Pods remain bound to Nodes for 5 minutes after one of these problems is detected.
```

In this case `Descheduler` might trigger at any time in this 5 minutes (default) timeframe.

Unfortunately I don't have the logs anymore since the issue happed a while ago and they were rotated. 

If agreed, I'm happy to gate this behaviour with a feature setting.

ingvagabund:
Hello @adriananeci.

> Not sure why the code was in place, but commented out.

Some strategies like `LowNodeUtilization`, `HighNodeUtilization` or `RemoveDuplicates` need to take into account all nodes when selecting nodes for eviction. Given the `ReadyNodes` (invoker of `IsReady`) is invoked in the main descheduling loop, we need to take into account even nodes which are getting drained/unschedulable.

Instead of ignoring nodes which are unschedulable I suggest we perform a check in the [pod evictor](https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/descheduler/evictions/evictions.go#L51) and refuse to evict a pod when a node has certain [list of known taints](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/#taint-based-evictions) added (e.g. `node.kubernetes.io/not-ready`, `node.kubernetes.io/unreachable`, `node.kubernetes.io/unschedulable`).

JaneLiuL:
agree with @ingvagabund , some pods have tolerations , we still need to check some strategies on all nodes.

Issue #728, opened 2022-02-16
Panic with K8s v1.22: Close of closed channel


stephan2012:
**What version of descheduler are you using?** 

descheduler version: 0.23.0

**Does this issue reproduce with the latest release?**

yes

**Which descheduler CLI options are you using?**

```yaml
          containers:
          - args:
            - --policy-config-file
            - /policy-dir/policy.yaml
            - --v
            - "3"
            command:
            - /bin/descheduler
```

**Please provide a copy of your descheduler policy config file**

```yaml
apiVersion: v1
data:
  policy.yaml: |
    apiVersion: "descheduler/v1alpha1"
    kind: "DeschedulerPolicy"
    strategies:
      LowNodeUtilization:
        enabled: true
        params:
          nodeResourceUtilizationThresholds:
            targetThresholds:
              cpu: 50
              memory: 50
              pods: 50
            thresholds:
              cpu: 20
              memory: 20
              pods: 20
      RemoveDuplicates:
        enabled: true
      RemovePodsViolatingInterPodAntiAffinity:
        enabled: true
      RemovePodsViolatingNodeAffinity:
        enabled: true
        params:
          nodeAffinityType:
          - requiredDuringSchedulingIgnoredDuringExecution
      RemovePodsViolatingNodeTaints:
        enabled: true
      RemovePodsViolatingTopologySpreadConstraint:
        enabled: true
kind: ConfigMap
metadata:
  annotations:
    meta.helm.sh/release-name: descheduler
    meta.helm.sh/release-namespace: kube-system
  creationTimestamp: "2021-09-14T08:04:24Z"
  labels:
    app.kubernetes.io/instance: descheduler
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: descheduler
    app.kubernetes.io/version: 0.23.0
    helm.sh/chart: descheduler-0.23.1
  name: descheduler
  namespace: kube-system
  resourceVersion: "281483144"
  uid: 21c26712-c359-4c6e-ad47-9f315c66fa1f
```

**What k8s version are you using (`kubectl version`)?**

<details><summary><code>kubectl version</code> Output</summary><br><pre>
$ kubectl version
[...]

PR #731, opened 2022-02-18
Bugfix: Cronjob ctx cancel

jklaw90:
Removed the stop channel and just use ctx.
Fixes https://github.com/kubernetes-sigs/descheduler/issues/728

### DIFF
--- a/cmd/descheduler/app/server.go
+++ b/cmd/descheduler/app/server.go
@@ -73,7 +73,7 @@ func NewDeschedulerCommand(out io.Writer) *cobra.Command {
 			}
 
 			ctx, done := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
-			defer done()
+
 			pathRecorderMux := mux.NewPathRecorderMux("descheduler")
 			if !s.DisableMetrics {
 				pathRecorderMux.Handle("/metrics", legacyregistry.HandlerWithReset())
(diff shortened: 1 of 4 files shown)


cmd/descheduler/app/server.go
```diff
@@ -73,23 +73,28 @@ func NewDeschedulerCommand(out io.Writer) *cobra.Command {
...
 				return
 			}
 
-			err := Run(ctx, s)
+			err = Run(ctx, s)
 			if err != nil {
 				klog.ErrorS(err, "descheduler server")
 			}
+
+			done()
```
ingvagabund, line 95: Would moving `done` back under the `defer` produce the same behavior?

pkg/descheduler/descheduler.go
```diff
@@ -69,13 +69,7 @@ func Run(ctx context.Context, rs *options.DeschedulerServer) error {
 		return err
 	}
 
-	// tie in root ctx with our wait stopChannel
-	stopChannel := make(chan struct{})
```
ingvagabund, line 73: I like removing this part!!!

ingvagabund (commented):
Looks good!!! Left one nit

cmd/descheduler/app/server.go
jklaw90: I changed it this way to notify the metrics server to close which blocks on the next line until all requests are done. If I kept it as a defer it's going to hang forever on the blocking channel for the cronjob case.

ingvagabund:
There's a valid building error in the e2e:
```
vet: test/e2e/e2e_test.go:989:105: too many arguments in call to descheduler.RunDeschedulerStrategies
```

cmd/descheduler/app/server.go
ingvagabund: Understood.

ingvagabund:
@jklaw90 would you mind squashing the commits? Looks good.

Issue #718, opened 2022-02-08
How to scrape metrics


mateuszdrab:
Hi

I can see that recently, a PR was merged to enable metrics support - however no documentation has been created to provide instructions how to use it.
I tried to create a podmonitor like this:
```
apiVersion: monitoring.coreos.com/v1
kind: PodMonitor
metadata:
  creationTimestamp: "2022-02-07T17:45:56Z"
  generation: 2
  labels:
    app: descheduler
  name: descheduler
  namespace: kube-system
spec:
  jobLabel: descheduler
  namespaceSelector:
    matchNames:
    - kube-system
  podMetricsEndpoints:
  - honorLabels: true
    interval: 5s
    path: /metrics
    scheme: http
    targetPort: 10258
  selector:
    matchLabels:
      app: descheduler
```

Unfortunately, no metrics are collected. I'm running descheduler as a CronJob. Could it be that I need to run it as a deployment and create a service monitor instead?

Perhaps a sample could be provided for the monitor or better, included in the helm chart?

Thanks

damemi:
/kind support
cc @ingvagabund, do you have any examples for this?

ingvagabund:
Hi @mateuszdrab, in OpenShift we deploy a service monitor [1] as you mentioned. I agree it would be practical to create a similar deployment for Kubernetes as well.

> I'm running descheduler as a CronJob. Could it be that I need to run it as a deployment and create a service monitor instead?

That depends on how long the descheduler runs. Though, the primary use of the metrics is for long running descheduler.

[1] https://github.com/openshift/cluster-kube-descheduler-operator/tree/master/bindata/assets/kube-descheduler

mateuszdrab:
Thanks for the link, I'll redeploy as a deployment and try to modify the servicemonitor to work. Too bad this isn't part of the helm charts.

PR #732, opened 2022-02-21
feat: Add metric scrape configs in Helm Chart

eminaktas:
This adds service and servicemonitor resources along with necessary helm chart values

Signed-off-by: eminaktas <emin.aktas@trendyol.com>


ingvagabund:
@mateuszdrab FYI

mateuszdrab:
@ingvagabund 

Thanks for the ping, will update to the new chart when it's merged 🙂

ingvagabund:
@mateuszdrab would you mind helping review this PR and incorporating your changes into the new charts here? I see you set `honorLabels`, `metricRelabelings` and other fields in your [manifests](https://github.com/kubernetes-sigs/descheduler/issues/718#issuecomment-1032909197).

@eminaktas would you mind taking a look at the manifests @mateuszdrab provided? Would be great to join the efforts.

@eminaktas @mateuszdrab thank You for putting the bits together.

mateuszdrab:
> @mateuszdrab would you mind helping review this PR and incorporating your changes into the new charts here? I see you set `honorLabels`, `metricRelabelings` and other fields in your [manifests](https://github.com/kubernetes-sigs/descheduler/issues/718#issuecomment-1032909197).
> 
> @eminaktas would you mind taking a look at the manifests @mateuszdrab provided? Would be great to join the efforts.
> 
> @eminaktas @mateuszdrab thank You for putting the bits together.

I ended up removing the metricRelabelings, it came from a sample config but I wasn't able to verify whether it worked or not because the descheduler didn't actually deschedule any pods.

Currently, the spec I'm using currently which does scrape is:

```
spec:
  endpoints:
  - honorLabels: true
    interval: 30s
    path: /metrics
    port: metrics
    scheme: https
    tlsConfig:
      insecureSkipVerify: true
  jobLabel: app.kubernetes.io/name
  namespaceSelector:
    matchNames:
    - kube-system
  selector:
    matchLabels:
      app.kubernetes.io/name: descheduler
```

The thing is, we probably need to get rid of insecureSkipVerify: true. I just did it because I wanted it to work.

Are you aware of any Grafana dashboard to visualize this?

eminaktas:
`metricRelabelings` and `relabelings` can be added through [values.yaml](https://github.com/kubernetes-sigs/descheduler/pull/732/files#diff-9e280a2849938ec31773c6cd202a9b6c7731e4642758237121b8e21d552866c3R131).

I deployed without this:

```yaml
    metricRelabelings:
    - action: replace
      sourceLabels:
      - exported_namespace
      targetLabel: pod_namespace
```

As you can see it has namespace label already. I don't know why this config is added there but I don't think it is needed.

Screen Shot 2022-02-22 at 12 38 01


`honorLabels` is not needed either but we can let the users choose. It simply keeps(or chooses) the scraped data's labels when labels are on collisions with target labels.

@mateuszdrab you might want to use Prometheus Remote Write for the cronjob. Since I haven't done it myself, I don't know how you can do it.

By default, the metric server is HTTPS, without `insecureSkipVerify: true`, it won't work. We need to create certs and distribute them to descheduler and prometheus.

mateuszdrab:
> `metricRelabelings` and `relabelings` can be added through [values.yaml](https://github.com/kubernetes-sigs/descheduler/pull/732/files#diff-9e280a2849938ec31773c6cd202a9b6c7731e4642758237121b8e21d552866c3R131).
> 
> I deployed without this:
> 
> ```yaml
>     metricRelabelings:
>     - action: replace
>       sourceLabels:
>       - exported_namespace
>       targetLabel: pod_namespace
> ```
> 
> As you can see it has namespace label already. I don't know why this config is added there but I don't think it is needed.
> 
> Screen Shot 2022-02-22 at 12 38 01
> 
> `honorLabels` is not needed either but we can let the users choose. It simply keeps(or chooses) the scraped data's labels when labels are on collisions with target labels.
> 
> @mateuszdrab you might want to use Prometheus Remote Write for the cronjob. Since I haven't done it myself, I don't know how you can do it.
> 
> By default, the metric server is HTTPS, without `insecureSkipVerify: true`, it won't work. We need to create certs and distribute them to descheduler and prometheus.

Indeed, looks like your template has all it needs and has space for the necessary customization. Much better than what I could produce :)

charts/descheduler/values.yaml
```diff
@@ -22,6 +22,7 @@ resources:
   #   cpu: 100m
   #   memory: 128Mi
 
+namespace: kube-system
```
a7i, line 25: Curious why as Helm has its own `--namespace` flag.

```
  -n, --namespace string            namespace scope for this request
```

charts/descheduler/templates/service.yaml
```diff
@@ -0,0 +1,21 @@
+{{- if eq .Values.kind "Deployment" }}
+{{- if eq .Values.service.enabled true }}
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    {{- include "descheduler.labels" . | nindent 4 }}
+  name: {{ template "descheduler.fullname" . }}-service
```
a7i, line 8:
```suggestion
  name: {{ template "descheduler.fullname" . }}
```

charts/descheduler/values.yaml
eminaktas: Thanks for mentioning it. I changed the namespace values with `Release.Namespace` in the templates. This should make sense now for the `--namespace` flag.

>Release.Namespace: The namespace to be released into (if the manifest doesn’t override)
Ref: https://helm.sh/docs/chart_template_guide/builtin_objects/

eminaktas:
This commit is waiting for your approval @damemi @a7i  @ingvagabund

ingvagabund:
1. deploying locally with `helm template descheduler --release-name descheduler` | kubectl apply -f -
2. kubectl port-forward service/descheduler 10258:10258
3. curl -k https://localhost:10258/metrics

Don't have a Prometheus instance to validate the service monitor though.

Will keep the lgtm part of others.

damemi:
@eminaktas we usually try to keep merge commits out of the PRs themselves, would you mind squashing or doing a rebase instead?

eminaktas:
> @eminaktas we usually try to keep merge commits out of the PRs themselves, would you mind squashing or doing a rebase instead?

@damemi thanks for the reminder. It is fixed now.

rouke-broersma:
Can the OCI helm chart with these changes be released? We tried to implement this and were wondering why the servicemonitor was not being deployed 😅

rouke-broersma:
We can roll our own 👌

PR #734, opened 2022-02-23
fix: change versioning from v20220127-v0.22.0-79-gafb1d75ce to v0.22.0

eminaktas:
Currently, Descheduler version is being recorded as v20220127-v0.22.0-79-gafb1d75ce. This commit changes it to v0.22.0

Signed-off-by: Emin Aktas <emin.aktas@trendyol.com>
Co-authored-by: Fatih Sarhan <fatih.sarhan@trendyol.com>
Co-authored-by: Necatican Yildirim <necatican.yildirim@trendyol.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,14 +15,14 @@
 .PHONY: test
 
 # VERSION is based on a date stamp plus the last commit
-VERSION?=v$(shell date +%Y%m%d)-$(shell git describe --tags --match "v*")
+VERSION?=$(shell git describe --tags --match "v*" --abbrev=0)
 BRANCH?=$(shell git branch --show-current)
 SHA1?=$(shell git rev-parse HEAD)
-BUILD=$(shell date +%FT%T%z)
+BUILD_DATE=$(shell date +%FT%T%z)
 LDFLAG_LOCATION=sigs.k8s.io/descheduler/pkg/version
 ARCHS = amd64 arm arm64
 
-LDFLAGS=-ldflags "-X ${LDFLAG_LOCATION}.version=${VERSION} -X ${LDFLAG_LOCATION}.buildDate=${BUILD} -X ${LDFLAG_LOCATION}.gitbranch=${BRANCH} -X ${LDFLAG_LOCATION}.gitsha1=${SHA1}"
+LDFLAGS=-ldflags "-X ${LDFLAG_LOCATION}.version=${VERSION} -X ${LDFLAG_LOCATION}.buildDate=${BUILD_DATE} -X ${LDFLAG_LOCATION}.gitbranch=${BRANCH} -X ${LDFLAG_LOCATION}.gitsha1=${SHA1}"
 
 GOLANGCI_VERSION := v1.43.0
 HAS_GOLANGCI := $(shell ls _output/bin/golangci-lint 2> /dev/null)


damemi (commented):
@eminaktas curious if you could explain more what is the use case for this? Afaik, this format is used to uniquely identify the automated builds that happen after each PR merge (https://console.cloud.google.com/gcr/images/k8s-staging-descheduler/GLOBAL/descheduler)

/ok-to-test

eminaktas:
> @eminaktas curious if you could explain more what is the use case for this? Afaik, this format is used to uniquely identify the automated builds that happen after each PR merge (https://console.cloud.google.com/gcr/images/k8s-staging-descheduler/GLOBAL/descheduler)
> 
> /ok-to-test

Thanks for sharing that part. I think we should separate the versioning since when the actual releases are taken the version is being shown as this `v20220127-v0.22.0-79-gafb1d75ce`.

Also, I was about to create another bug. With the latest image, the version is also wrong. This is what I see when I run the app.

```
v20220127-v0.22.0-79-gafb1d75ce
```

```bash
crane digest k8s.gcr.io/descheduler/descheduler:v0.23.0
sha256:1a40adaa9b92624d09fee9e8611f0177949949c6f2fa633e43457eec96860c65
```

damemi:
@eminaktas yeah the `VERSION` is just the git tag, I create that tag from the master branch when I published the release through Github's UI (which should do the same thing, I'm assuming, see https://github.com/kubernetes-sigs/descheduler/tree/v0.23.0)

It would be nice to figure out why our git tag seems 1 version behind

damemi:
@eminaktas the process for publishing the images involves manually taking the most recent SHA and updating it in the `images.yaml` file you linked. I made that change in this PR: https://github.com/kubernetes/k8s.io/pull/3344

That PR merged on Feb 3, which is when (and how) the tag you're referring to was published (`v20220203-v0.23.0`). So, that is actually the image tag *created* by the promotion process (see https://console.cloud.google.com/gcr/images/k8s-artifacts-prod/us/descheduler/descheduler, the SHA of the most recent `prod` matches what you're pointing to from `staging`.

Maybe this could be fixed in the future by tagging the release before starting the image promotion, but I think this will still require a change to merge to the new tag in order for the automated build to pick it up and provide us with a SHA from the new tag to promote.

There's more details on how this promotion works in https://github.com/kubernetes/k8s.io/tree/main/k8s.gcr.io#staging-repos, if you are interested

eminaktas:
@damemi 
We have to change this before the new release. I will try to help you with tihs. 

I still believe the images are somehow messed up. I mean that if you check the shas and the dates along with tags the images shas are the same. Somehow the prod has the wrong image. You can see it in the screenshots below.

first
Screenshot is taken from [here](https://console.cloud.google.com/gcr/images/k8s-staging-descheduler/GLOBAL/descheduler).

second
Screenshot is taken from [here](https://console.cloud.google.com/gcr/images/k8s-artifacts-prod/us/descheduler/descheduler).

damemi:
@eminaktas my point above was that the prod `v0.23` image is pointing to a `v0.22` SHA in staging because we need to provide an existing SHA in order to create the prod `v0.23` image. So updating images.yaml doesn't fix the problem.

I think for the next release we can try tagging the Git branch before doing the image promotion so that we can provide a staging SHA that is git-tagged with `v0.24`. Previously, I've been waiting until the prod image was promoted so that we can tag the branch when we publish the [release on Github](https://github.com/kubernetes-sigs/descheduler/releases/tag/v0.23.0) (https://docs.github.com/en/repositories/releasing-projects-on-github/managing-releases-in-a-repository#creating-a-release is what I usually follow)

I don't know if there is a good way to fix this for the existing release without just publishing a new patch version. We could see if it's possible to update the shas that are pointed to in images.yaml to promote a new image that points to the existing promoted image

damemi:
@eminaktas based on the discussion above, I think the fix for this is more based on the order of steps in our release process (ie, ensuring that the images get tagged at the right time). Based on that, are you ok closing this PR? I don't think we should drop the timestamp versioning since it is used to differentiate individual automatic builds

eminaktas:
@damemi
Thanks for reminding me this PR. I couldn't work on this PR and the release process. But I aggree with you. If we remove the timestamp, we will probably need a way to name the images in a different way. Consequently, I am closing it now. If I find a better idea, I'll bring it up 🤓

PR #740, opened 2022-02-25
fix doc about NPD description

### LINKED ISSUE #606: Docs around autohealing are misleading
The [docs around autohealing](https://github.com/kubernetes-sigs/descheduler/blob/master/docs/user-guide.md#autoheal-node-problems) are a bit misleading in my opinion.

They link off to Node Problem Detector, claiming that `Node Problem Detector can detect specific Node problems and taint any Nodes which have those problems.`. In fact, NPD doesn't do any tainting. It's the `TaintNodeByCondition` feature of the node controller that takes _some_ conditions and turns them in to taints. However this only works for the default node conditions: `PIDPressure`, `MemoryPressure`, `DiskPressure`, `Ready`, and some cloud provider specific conditions.

There is an [open PR](https://github.com/kubernetes/node-problem-detector/pull/565) on NPD that wants to add this tainting behaviour, but the maintainers seem to think it shouldn't be NPD that does the tainting.

The effect is that the autoheal cycle describe doesn't actually work, at least not for custom conditions. It would be wonderful if it did, because it's quite a compelling outcome, and it would be amazing if it were offered exclusively in terms of Kubernetes first party tooling.

At the very least, we should change the wording in the docs to make it clear that NPD doesn't really participate in the autohealing. At most, I'm hoping that by raising this issue, the fact that this cycle doesn't currently work as intended can get a little more visibility. My guess at possible solutions:
* Merge the PR linked above, so that NPD creates taints
* Extend the node controller to add condition based taints for _all_ conditions, including custom ones
* Create a new project to convert conditions in to taints
* Add a new strategy in this repo that allows for descheduling based on Conditions

JaneLiuL:
fix #606

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -137,7 +137,14 @@ strategies:
 Descheduler's `RemovePodsViolatingNodeTaints` strategy can be combined with
 [Node Problem Detector](https://github.com/kubernetes/node-problem-detector/) and
 [Cluster Autoscaler](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler) to automatically remove
-Nodes which have problems. Node Problem Detector can detect specific Node problems and taint any Nodes which have those
-problems. The Descheduler will then deschedule workloads from those Nodes. Finally, if the descheduled Node's resource
+Nodes which have problems. Node Problem Detector can detect specific Node problems and report them to the API server.
+There is a feature called TaintNodeByCondition of the node controller that takes some conditions and turns them into taints. Currently, this only works for the default node conditions: PIDPressure, MemoryPressure, DiskPressure, Ready, and some cloud provider specific conditions.
+The Descheduler will then deschedule workloads from those Nodes. Finally, if the descheduled Node's resource
 allocation falls below the Cluster Autoscaler's scale down threshold, the Node will become a scale down candidate
 and can be removed by Cluster Autoscaler. These three components form an autohealing cycle for Node problems.
+---
+**NOTE**
+
+Once [kubernetes/node-problem-detector#565](https://github.com/kubernetes/node-problem-detector/pull/565) is available in NPD, we need to update this section.
+
+---


docs/user-guide.md
```diff
@@ -137,7 +137,8 @@ strategies:
 Descheduler's `RemovePodsViolatingNodeTaints` strategy can be combined with
 [Node Problem Detector](https://github.com/kubernetes/node-problem-detector/) and
```
damemi, line 138: It seems like NPD doesn't really do anything to help with autohealing, so we could probably just remove it entirely from this section. Or, add a note in here that it could eventually be combined here when https://github.com/kubernetes/node-problem-detector/pull/565 is available in NPD

docs/user-guide.md
```diff
@@ -137,7 +137,8 @@ strategies:
...
 [Cluster Autoscaler](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler) to automatically remove
-Nodes which have problems. Node Problem Detector can detect specific Node problems and taint any Nodes which have those
+Nodes which have problems. Node Problem Detector can detect specific Node problems and reports them to apiserver.
```
damemi, line 140: nit: `and report them to the API server`

docs/user-guide.md
```diff
@@ -137,7 +137,8 @@ strategies:
...
+There is a feature call TaintNodeByCondition of the node controller that takes some conditions and turns them in to taints and this only works for the default node conditions: PIDPressure, MemoryPressure, DiskPressure, Ready, and some cloud provider specific conditions.
```
damemi, line 141:
```suggestion
There is a feature called TaintNodeByCondition of the node controller that takes some conditions and turns them into taints. Currently, this only works for the default node conditions: PIDPressure, MemoryPressure, DiskPressure, Ready, and some cloud provider specific conditions.
```

docs/user-guide.md
```diff
@@ -137,7 +137,8 @@ strategies:
...
 problems. The Descheduler will then deschedule workloads from those Nodes. Finally, if the descheduled Node's resource
```
damemi, line 142:
```suggestion
The Descheduler will then deschedule workloads from those Nodes. Finally, if the descheduled Node's resource
```

damemi (commented):
/ok-to-test
/approve
just some minor suggestions

docs/user-guide.md
JaneLiuL: thanks, i add a note now :)

docs/user-guide.md
JaneLiuL: fix done~ please check :)

docs/user-guide.md
JaneLiuL: fix done~ please check :)

docs/user-guide.md
JaneLiuL: fix done~ please check :)

JaneLiuL: may i click resolve conversation?

PR #316, opened 2020-06-03
Fix 'make gen' target outside of GOPATH

pmundt:
When running 'make gen' outside of a GOPATH, it fails to update the
generated files in-place, and further gets confused, as it's unable
to locate the working descheduler directory.

The update tools further cannot handle an absolute path for the
package imports, so this must be relative to the OS_ROOT. Output
files must further be written in place, which doesn't work if
one has both an active GOPATH and is working on descheduler outside
of it. This is addressed by specifying the output base, which maps
properly for both the in-GOPATH and out-of-GOPATH cases.

This resolves #313.

PR #777, opened 2022-03-23
Add RemovePodsViolatingNodeTaints taint exclusion

### LINKED ISSUE #772: Add support for ignorelisting taints considered by RemovePodsViolatingNodeTaints

When a node pressure taint is applied to a node (e.g. `node.kubernetes.io/disk-pressure`), the `RemovePodsViolatingNodeTaints` strategy will deschedule pods that don't tolerate the node pressure taint. We would rather wait for such situations to be resolved by a [node pressure eviction](https://kubernetes.io/docs/concepts/scheduling-eviction/node-pressure-eviction/) to avoid unnecessary disruptions.

To address the problem described above (but with a more general solution) we propose the following: 

The `RemovePodsViolatingNodeTaints` strategy should support an ignorelist of taints. Pods that don't tolerate ignore-listed taints would not be considered for RemovePodsViolatingNodeTaints descheduling (unless other non-ignorelisted taints were also not tolerated).

We've considered disabling the RemovePodsViolatingNodeTaints strategy, but we would prefer to keep it in place to handle non-node-pressure taints.

descheduler version: [v0.23.1](https://github.com/kubernetes-sigs/descheduler/releases/tag/v0.23.1)

JacobHenner:
Add taint exclusion to RemovePodsViolatingNodeTaints. This permits node
taints to be ignored by allowing users to specify ignored taint keys or
ignored taint key=value pairs.

Fixes #772

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -89,6 +89,7 @@ type StrategyParameters struct {
 	LabelSelector                     *metav1.LabelSelector
 	NodeFit                           bool
 	IncludePreferNoSchedule           bool
+	ExcludedTaints                    []string
 }
 
 type Percentage float64
--- a/pkg/descheduler/strategies/node_taint.go
+++ b/pkg/descheduler/strategies/node_taint.go
@@ -55,12 +55,15 @@ func RemovePodsViolatingNodeTaints(ctx context.Context, client clientset.Interfa
 		return
 	}
 
-	var includedNamespaces, excludedNamespaces sets.String
+	var includedNamespaces, excludedNamespaces, excludedTaints sets.String
 	var labelSelector *metav1.LabelSelector
 	if strategy.Params != nil {
 		if strategy.Params.Namespaces != nil {
 			includedNamespaces = sets.NewString(strategy.Params.Namespaces.Include...)
 			excludedNamespaces = sets.NewString(strategy.Params.Namespaces.Exclude...)
+		}
+		if strategy.Params.ExcludedTaints != nil {
+			excludedTaints = sets.NewString(strategy.Params.ExcludedTaints...)
 		}
 		labelSelector = strategy.Params.LabelSelector
 	}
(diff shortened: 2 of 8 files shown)


pkg/api/types.go
```diff
@@ -89,6 +89,7 @@ type StrategyParameters struct {
 	LabelSelector                     *metav1.LabelSelector
 	NodeFit                           bool
 	IncludePreferNoSchedule           bool
+	ExcludedTaints                    []string
```
ingvagabund, line 92: `pkg/apis/v1alpha1/types.go` needs to be updated as well since we [convert v1alpha1 into the internal](https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/descheduler/policyconfig.go#L50) + running the deepcopy/conversion generators

ingvagabund (changes requested):
One nit, the rest looks great!!! Thank You for noticing this and fixing it right away.

pkg/api/types.go
JacobHenner: I've updated v1alpha1/types.go, but `make gen` doesn't seem to be modifying the auto-generated files. I'll look into this.

JacobHenner: Pushed the auto-generated changes - the issue was https://github.com/kubernetes-sigs/descheduler/pull/316.

pkg/descheduler/strategies/node_taint.go
```diff
@@ -89,10 +99,14 @@ func RemovePodsViolatingNodeTaints(ctx context.Context, client clientset.Interfa
 		return
 	}
 
-	taintFilterFnc := func(taint *v1.Taint) bool { return taint.Effect == v1.TaintEffectNoSchedule }
+	excludeTaint := func(taint *v1.Taint) bool {
```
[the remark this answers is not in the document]
JacobHenner: Sure, done.

ingvagabund:
@JacobHenner much appreciated
/approve
/lgtm

PR #793, opened 2022-05-06
add strategy defaults and fix nodeFit for node affinity strategy

### LINKED ISSUE #640: nodeFit = false doesn't work as expected with RemovePodsViolatingNodeAffinity

descheduler version: 0.22

Yes

            - "--policy-config-file"
            - "/policy-dir/policy.yaml"
            - "--descheduling-interval"
            - "30s"
            - "--v"
            - "4"

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: descheduler-policy-configmap
  namespace: kube-system
data:
  policy.yaml: |
    kind: "DeschedulerPolicy"
    strategies:
      "RemovePodsViolatingNodeAffinity":
        enabled: true
        params:
          nodeAffinityType:
            - "requiredDuringSchedulingIgnoredDuringExecution"
          labelSelector:
            matchExpressions:
              - {key: "foo", operator: In, values: [bar]}
          nodeFit: false
```

$ kubectl version
Client Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.9", GitCommit:"7a576bc3935a6b555e33346fd73ad77c925e9e4a", GitTreeState:"clean", BuildDate:"2021-07-15T21:01:38Z", GoVersion:"go1.15.14", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.9", GitCommit:"7a576bc3935a6b555e33346fd73ad77c925e9e4a", GitTreeState:"clean", BuildDate:"2021-07-15T20:56:38Z", GoVersion:"go1.15.14", Compiler:"gc", Platform:"linux/amd64"}
</pre></details>

**What did you do?**

I created a deployment with a node selector forcing the pods to land on a specific single node. The deployment pod template had a node affinity rule, which was not triggered during scheduling. After scheduling, I changed the node labels so that the affinity should trigger descheduling, and labeled the pods with the required foo=bar label.

I expected the descheduler to evict my pod, and the pod to end up in PENDING state as there is no node where it could fit.

**What did you see instead?**
Descheduler prints that the pod will not fit the node, but it won't evict it.

The culprit is at the strategy, node_affinity.go line 89. https://github.com/kubernetes-sigs/descheduler/blob/5b557941fac60cd85c1a3b709eb49a08d648fd8d/pkg/descheduler/strategies/node_affinity.go#L89

The node affinity strategy will always check for finding a node that fits, regardless of how nodeFit is set. The documentation of the nodeFit filtering would suggest that the default is not to check for node fitting: https://github.com/kubernetes-sigs/descheduler/blob/5b557941fac60cd85c1a3b709eb49a08d648fd8d/README.md#L703-L707

It would be logical, that when nodeFit=false, the pod would be evicted regardless of whether it fits somewhere, or not. A one-line change would fix this. Line 89 could be e.g. changed to something like:
`(!nodeFit || nodeutil.PodFitsAnyNode(pod, nodes))`

uniemimu:
This adds defaults for strategies. Now each strategy may have
different default parameters, if some are omitted from the configmap.

Since the NodeFit parameter was a bool instead of a pointer, it
couldn't be detected whether it was being set in the configmap, or
not. Hence it is changed to a pointer. If it is not set in the
configmap, the default is taken.

Except for node affinity, all other  strategies are now having a
default of false for an omitted NodeFit parameter, which matches
exactly how things have worked.

For the node affinity strategy, the legacy mode of operation has
always been what a "true" NodeFit would be. Thus an omitted NodeFit
parameter for that strategy is set to true, which matches how
node affinity has been working, effectively not evicting anything
if the pod doesn't fit into another node.

The change is in how things work when the NodeFit is explicitly set
to "false". The hard-coded check for pod fitting any node is removed,
and the NodeFit feature has control. NodeFit being false, node
affinity strategy will evict the pod regardless of whether there is
another node where it would fit.

Fixes #640 

Signed-off-by: Ukri Niemimuukko <ukri.niemimuukko@intel.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -735,6 +735,9 @@ The following strategies accept a `nodeFit` boolean parameter which can optimize
 * `RemovePodsHavingTooManyRestarts`
 * `RemoveFailedPods`
 
+If omitted, the `nodeFit` boolean will default to `false` except for `RemovePodsViolatingNodeAffinity`
+for which it defaults to `true` due to backwards compatibility reasons.
+
  If set to `true` the descheduler will consider whether or not the pods that meet eviction criteria will fit on other nodes before evicting them. If a pod cannot be rescheduled to another node, it will not be evicted. Currently the following criteria are considered when setting `nodeFit` to `true`:
 - A `nodeSelector` on the pod
 - Any `tolerations` on the pod and any `taints` on the other nodes
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -87,7 +87,7 @@ type StrategyParameters struct {
 	ThresholdPriority                 *int32
 	ThresholdPriorityClassName        string
 	LabelSelector                     *metav1.LabelSelector
-	NodeFit                           bool
+	NodeFit                           *bool
 	IncludePreferNoSchedule           bool
 	ExcludedTaints                    []string
 }
--- a/pkg/api/v1alpha1/defaults.go
+++ b/pkg/api/v1alpha1/defaults.go
@@ -16,7 +16,51 @@ limitations under the License.
 
 package v1alpha1
 
-import "k8s.io/apimachinery/pkg/runtime"
+import (
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/utils/pointer"
+)
+
+var (
+	nodeAffinityDefaults = DeschedulerStrategy{
+		Params: &StrategyParameters{
+			NodeFit: pointer.Bool(true),
+		},
+	}
+	genericStrategyDefaults = DeschedulerStrategy{
+		Params: &StrategyParameters{
+			NodeFit: pointer.Bool(false),
+		},
+	}
+
+	strategyDefaults = StrategyList{
+		StrategyName("RemoveDuplicates"):                            genericStrategyDefaults,
+		StrategyName("LowNodeUtilization"):                          genericStrategyDefaults,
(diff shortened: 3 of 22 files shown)


uniemimu:
Nit for self, and for conversation: PodFitsAnyNode has no users anymore if this gets merged. Should it be removed at the same time?

pkg/utils/types.go
```diff
@@ -0,0 +1,6 @@
+package utils
+
+// NewBool returns a pointer to a new initialized bool value.
+func NewBool(value bool) *bool {
```
damemi, line 4: You can use https://pkg.go.dev/github.com/bitrise-io/go-utils/pointers#NewBoolPtr instead of adding a new function

uniemimu: Okayh, I will back off to that one.

uniemimu:
Squashed and cleaned. Sorry for the slipped in hacks in the intermediates.

ingvagabund:
@uniemimu thanks for keeping the defaulting in mind. As part of moving to the descheduling framework, this becomes part of the core changes in the code base after all plugins are migrated to plugins.

uniemimu:
> @uniemimu thanks for keeping the defaulting in mind. As part of moving to the descheduling framework, this becomes part of the core changes in the code base after all plugins are migrated to plugins.

Ok please let me know if this is worth rebasing on top of some WIP PR which is considered better off going in first.

PR #792, opened 2022-05-04
Update Docs and Manifests for v0.24.0

pravarag:
Changes covered as part of this PR are:

- [ ] Added v0.24 references to README
- [ ]  Update k8s manifests with v0.24.0 references
- [ ] Added table with list of supported architectures by release

PR #795, opened 2022-05-12
Update e2e test versions

damemi:
Possibly fixing https://github.com/kubernetes-sigs/descheduler/pull/792#issuecomment-1124569018


damemi (commented):
cc @pravarag 
wondering if this is causing the error we saw in CI for 1.24

Issue #736, opened 2022-02-23
Currently, v0.23.0 image has wrong app version


eminaktas:
**What version of descheduler are you using?**

descheduler version: v0.23.0

```bash
crane digest k8s.gcr.io/descheduler/descheduler:v0.23.0
sha256:1a40adaa9b92624d09fee9e8611f0177949949c6f2fa633e43457eec96860c65
```

**Does this issue reproduce with the latest release?**

Yes.

**Which descheduler CLI options are you using?**

Default options.

**Please provide a copy of your descheduler policy config file**

Default policy yaml

**What k8s version are you using (`kubectl version`)?**

<details><summary><code>kubectl version</code> Output</summary><br><pre>
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.4", GitCommit:"b695d79d4f967c403a96986f1750a35eb75e75f1", GitTreeState:"clean", BuildDate:"2021-11-17T15:48:33Z", GoVersion:"go1.16.10", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.7", GitCommit:"132a687512d7fb058d0f5890f07d4121b3f0a2e2", GitTreeState:"clean", BuildDate:"2021-05-12T12:32:49Z", GoVersion:"go1.15.12", Compiler:"gc", Platform:"linux/amd64"}
</pre></details>

**What did you do?**

Nothing special.

**What did you expect to see?**

`v0.23.0`

**What did you see instead?**

`v20220127-v0.22.0-79-gafb1d75ce`

eminaktas:
#734

eminaktas:
Kindly reminder:

Unfortunately, we still have the wrong Descheduler version with the latest image. We should consider publishing an image to fix the tag.

```
descheduler_build_info{DeschedulerVersion="v20220228-descheduler-helm-chart-0.23.2-2-gfd6c9a68a", GoVersion="go1.17.7",..
```

cc @damemi

JaneLiuL:
@eminaktas  do you use helm chart install or use which for install? i need to have a test, also have to reinstall recently?

eminaktas:
> @eminaktas do you use helm chart install or use which for install? i need to have a test, also have to reinstall recently? since pr #747 just merge about 9 days ago, so you need to reinstall again for it

[...]

PR #796, opened 2022-05-13
Update helm chart version to v0.24

JaneLiuL:


damemi:
Thanks @JaneLiuL!
I'm going to hold this for now just so it doesn't merge accidentally, I want to double check the order that we merge helm chart stuff so we can try to fix the tagging issue from https://github.com/kubernetes-sigs/descheduler/issues/736

damemi (commented):
/approve
/lgtm
/hold cancel
`v0.24.0` tag is created, so the image built from this change should end up with that as its version git tag

PR #801, opened 2022-05-18
Bump To Go 1.18.2

seanmalloy:
The main k/k repo was updated to Go 1.18.2 for the
k8s v1.24.0 release. See below PR for reference.

https://github.com/kubernetes/kubernetes/pull/110044


seanmalloy:
Need to update CI to use Go 1.18 before merging.

/hold

damemi:
Thanks for catching this @seanmalloy. I'm holding the image promotion PR (https://github.com/kubernetes/k8s.io/pull/3739) so we should be able to include this change in the v0.24.0 release

seanmalloy:
Looks like CI was already updated to Go 1.18 in https://github.com/kubernetes/test-infra/pull/25833

/hold cancel

damemi (commented):
/approve
this looks good (we actually should have gotten the 1.18 updates earlier). my only concern is how we want to handle this with our own `v0.24` release

The upstream change is for 1.24.1, which hasn't been released yet. So do we think it's ok to preemptively make the go1.18.2 change for `v0.24.0`, or should we include that in a `v0.24.1` that we publish after k8s publishes their own patch and is actually on go1.18.2? (At the very least, we should update `v0.24.0` to go1.18.1 in all these places, which 1.24 is on already.)

I also just want to point out that this won't be in our `v0.24.0` git tag, which I think is okay since we already effectively define the "release" code as the `release-*` branch.

ingvagabund:
I wonder if the `Dockerfile` is used anywhere in the building process rather than for users building the descheduler image manually. Anyway, as @damemi pointed out, the building pipeline was bumped to 1.18 anyway so all the changes in this PR are necessary but not blocking.

/lgtm

ingvagabund:
Also, https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-sigs/descheduler/descheduler-presubmits-master.yaml#L16 is set to 1.18, not 1.18.z. I still prefer to pin to 1.18.z so the building pipeline does not pull the latest 1.18 which can be buggy/contain CVEs.

damemi:
@ingvagabund @seanmalloy any input on my question from above?

> The upstream change is for 1.24.1, which hasn't been released yet. So do we think it's ok to preemptively make the go1.18.2 change for v0.24.0, or should we include that in a v0.24.1 that we publish after k8s publishes their own patch and is actually on go1.18.2? (At the very least, we should update v0.24.0 to go1.18.1 in all these places, which 1.24 is on already.)

seanmalloy:
> @ingvagabund @seanmalloy any input on my question from above?
> 
> > The upstream change is for 1.24.1, which hasn't been released yet. So do we think it's ok to preemptively make the go1.18.2 change for v0.24.0, or should we include that in a v0.24.1 that we publish after k8s publishes their own patch and is actually on go1.18.2? (At the very least, we should update v0.24.0 to go1.18.1 in all these places, which 1.24 is on already.)

My perspective is that the [v0.24.0](https://github.com/kubernetes-sigs/descheduler/releases/tag/v0.24.0) tag has already been created, so we can finish the release process by promoting the container images. I confirmed that the Go version in the `Dockerfile` is what the container images are built with, so v0.24.0 would be built with Go 1.17.7.

I can back port this change to to the `release-1.24` branch once it is created from the `v0.24.0` tag and bump the k8s deps to v1.24.1 once it is released. I can take care of updating the CI jobs to use Go 1.18.2 instead of Go 1.18.

@damemi and @ingvagabund let me know if you have any concerns with this approach.

seanmalloy:
> > @ingvagabund @seanmalloy any input on my question from above?
> > > The upstream change is for 1.24.1, which hasn't been released yet. So do we think it's ok to preemptively make the go1.18.2 change for v0.24.0, or should we include that in a v0.24.1 that we publish after k8s publishes their own patch and is actually on go1.18.2? (At the very least, we should update v0.24.0 to go1.18.1 in all these places, which 1.24 is on already.)
> 
> My perspective is that the [v0.24.0](https://github.com/kubernetes-sigs/descheduler/releases/tag/v0.24.0) tag has already been created, so we can finish the release process by promoting the container images. I confirmed that the Go version in the `Dockerfile` is what the container images are built with, so v0.24.0 would be built with Go 1.17.7.
> 
> I can back port this change to to the `release-1.24` branch once it is created from the `v0.24.0` tag and bump the k8s deps to v1.24.1 once it is released. I can take care of updating the CI jobs to use Go 1.18.2 instead of Go 1.18.
> 
> @damemi and @ingvagabund let me know if you have any concerns with this approach.

Forgot to mention we would then cut a descheduler v0.24.1 release once I complete all of that work. This release would use Go 1.18.2.

damemi:
@seanmalloy I think that sounds good, and gets us a 1.24 release out the door that's compatible with k8s 1.24. But the patch release raises a good point that we need to work on addressing https://github.com/kubernetes-sigs/descheduler/issues/736 better. I think I came up with a solution in https://github.com/kubernetes-sigs/descheduler/issues/736#issuecomment-1133191939. Please double check my logic when you have a chance

PR #804, opened 2022-05-22
Chart tests.enabled flag

jklaw90:
If using kustomize to render the chart and add extra transformations you can't disable the tests with the normal `--skip-tests` see [issue](https://github.com/kubernetes-sigs/kustomize/issues/3831).
Just adding this this flag to get around this case.


charts/descheduler/templates/tests/test-descheduler-pod.yaml
```diff
@@ -28,3 +29,4 @@ spec:
             chmod +x ./kubectl &&
             mv ./kubectl /usr/local/bin/kubectl &&
             /usr/local/bin/kubectl get pods --namespace {{ .Release.Namespace }} --token "$(cat /var/run/secrets/kubernetes.io/serviceaccount/token)" | grep "descheduler" | grep "Completed"
+{{- end }}
```
damemi, line 32: nit: newline?

charts/descheduler/values.yaml
```diff
@@ -167,3 +167,6 @@ serviceMonitor:
     #   targetLabel: nodename
     #   replacement: $1
     #   action: replace
+
+tests:
+  enabled: true
```
damemi, line 172: newline

damemi (commented):
/ok-to-test

jklaw90:
yeah, i tested it locally works fine and i agree it is odd.
I wonder if kustomize doesn't care to support this case since helm tests are pretty rare.
Our test seems kind of unnecessary too, would it even pass for a cronjob that is deployed for the first time to a new cluster?

damemi:
@jklaw90 sorry, I haven't forgotten about this (just had other stuff come up). I'm still not sure how I feel about this particular approach, and you bring up a good point about the helm tests too. So I'd like to look into those a bit more before going ahead with this. Is it a critical issue that's blocking kustomize users?

PR #805, opened 2022-05-23
cleanup lownodeutilization code

xiaoanyunfei:
The code is duplicated and useless

### DIFF
--- a/pkg/descheduler/strategies/nodeutilization/lownodeutilization.go
+++ b/pkg/descheduler/strategies/nodeutilization/lownodeutilization.go
@@ -118,20 +118,6 @@ func LowNodeUtilization(ctx context.Context, client clientset.Interface, strateg
 	klog.V(1).InfoS("Criteria for a node under utilization", keysAndValues...)
 	klog.V(1).InfoS("Number of underutilized nodes", "totalNumber", len(lowNodes))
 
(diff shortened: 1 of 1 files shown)


a7i:
@xiaoanyunfei thank you for your contribution!

/lgtm
/assign @damemi

damemi:
@a7i or @xiaoanyunfei would you mind opening a PR to revert this? Since the removed log is for targetThresholds, that should still be printed. Maybe update the comment to more clearly indicate that

damemi:
@a7i this is printing both of the thresholds for lownodeutilization. ie, the threshold for nodes that are overutilized as well as the threshold for matching candidate underutilized nodes. so, both should be printed for lownodeutilization.

highnodeutilization only really relies on one threshold since it is focused on bin-packing toward 100% utilization

a7i:
> @a7i this is printing both of the thresholds for lownodeutilization. ie, the threshold for nodes that are overutilized as well as the threshold for matching candidate underutilized nodes. so, both should be printed for lownodeutilization.
> 
> highnodeutilization only really relies on one threshold since it is focused on bin-packing toward 100% utilization

@damemi took a stab at it here:
https://github.com/kubernetes-sigs/descheduler/pull/830

PR #818, opened 2022-05-25
Update release guide docs

damemi:
Updating the release guide with the new flowchart and steps, including pre-release code changes


docs/release-guide.md
```diff
@@ -1,36 +1,79 @@
 # Release Guide
 
-## Container Image
-
-### Semi-automatic
```
damemi, line 5: the semi-automatic and manual image pushing steps were all the same except for the command to manually push, so I removed these sections and added a note that it can be manually pushed instead

damemi (commented):
/cc @seanmalloy @ingvagabund @JaneLiuL ptal

JaneLiuL (commented):
/lgtm

PR #820, opened 2022-05-26
[release-1.24] Update helm chart for v0.24.1

damemi:
This prepares the v0.24.1 release

(this branch already tagged with `v0.24.1`, so after this merges we should be able to do the image promotion)


JaneLiuL (commented):
/lgtm

PR #823, opened 2022-05-31
Fix version command to parse helm chart tags

damemi:
Using `--match "v*"` actually matches on the `descheduler-helm-chart-` tags as well since we prefix all automated build tags with `v2022..` date timestamps. This updates the `version` command to just add another use cases for scenarios where helm chart tags are the most recent ones. 

Related: https://github.com/kubernetes-sigs/descheduler/issues/736#issuecomment-1062608866

I tried putting together a release process to address this when we're doing a release, but that just means we have a race between the chart-releaser action and the image auto builder. This is more of an actual fix

PR #824, opened 2022-05-31
[release-1.24] Fix version command to parse helm chart tags

damemi:
Porting https://github.com/kubernetes-sigs/descheduler/pull/823 to 1.24

### DIFF
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -80,9 +80,19 @@ func splitVersion(version string) (string, string) {
 	// Version from an automated container build environment(not a tag) or a local build. For example v20201009-v0.18.0-46-g939c1c0.
 	m2, _ := regexp.MatchString(`^v\d{8}-v\d+\.\d+\.\d+-\w+-\w+$`, version)
 
+	// Version tagged by helm chart releaser action
+	helm, _ := regexp.MatchString(`^v\d{8}-descheduler-helm-chart-\d+\.\d+\.\d+$`, version)
+	// Dirty version where helm chart is the last known tag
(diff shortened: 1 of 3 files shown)


damemi (commented):
/hold
to make sure https://github.com/kubernetes-sigs/descheduler/pull/823 works once it merges

pkg/version/version.go
```diff
@@ -80,9 +80,19 @@ func splitVersion(version string) (string, string) {
 	// Version from an automated container build environment(not a tag) or a local build. For example v20201009-v0.18.0-46-g939c1c0.
 	m2, _ := regexp.MatchString(`^v\d{8}-v\d+\.\d+\.\d+-\w+-\w+$`, version)
 
+	// Version tagged by helm chart releaser action
+	helm, _ := regexp.MatchString(`^v\d{8}-descheduler-helm-chart-\d+\.\d+\.\d+$`, version)
+	// Dirty version where helm chart is the last known tag
+	helm2, _ := regexp.MatchString(`^v\d{8}-descheduler-helm-chart-\d+\.\d+\.\d+-\w+-\w+$`, version)
```
a7i, line 86: can probably be moved closer to where the variables used.

damemi: true, but I also like having all the regex's near each other

a7i: fair enough

damemi:
/hold cancel
/approve
(also re-adding lgtm, just picked https://github.com/kubernetes-sigs/descheduler/pull/825 was the only change)

PR #826, opened 2022-05-31
TopologySpreadConstraint strategy to support a fallback for Pods that do not have constraints defined

### LINKED ISSUE #609: TopologySpreadConstraint to fall back to KubeSchedulerConfiguration if Pod Topology Spread is not defined

Currently, RemovePodsViolatingTopologySpreadConstraint requires PodTopologySpread to be defined on every Pod. This can be challenging for organizations that already have thousands of workloads, especially given that PodTopologySpread was recently introduced in k8s 1.18.

We have defined our default `KubeSchedulerConfiguration` to be spread across a topology with maxSkew of 1 as a default cluster-level constraint without having to define it for every workload.

A param in the `RemovePodsViolatingTopologySpreadConstraint` policy that instructs the strategy to fall back to the default `KubeSchedulerConfiguration`.

Admission webhook to mutate the Pod object and include the PodTopologySpread, if not defined.

descheduler version: v0.21.0

**Additional context**

https://kubernetes.io/docs/concepts/workloads/pods/pod-topology-spread-constraints/#cluster-level-default-constraints
```yaml
apiVersion: kubescheduler.config.k8s.io/v1beta1
kind: KubeSchedulerConfiguration
profiles:
  - pluginConfig:
      - name: PodTopologySpread
        args:
          defaultConstraints:
            - maxSkew: 1
              topologyKey: topology.kubernetes.io/zone
              whenUnsatisfiable: ScheduleAnyway
          defaultingType: List
```

a7i:
Closes #609 

`topologySpreadConstraint` provides a `defaultConstraints` field in order to provide a list of
`TopologySpreadConstraints` as a fallback when the Pod spec does not have any defined.

_p.s. in the future when we're ready to bump the descheduler config API version, we should move `includeSoftConstraints` from top-level to `topologySpreadConstraint`._


pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -359,3 +360,14 @@ func comparePodsByPriority(iPod, jPod *v1.Pod) bool {
...
+
+// getTopologySpreadConstraints is a helper to get topology spread constraints in the following order:
+// - from Pod Spec (if defined)
+// - from the strategy params (if defined)
+func getTopologySpreadConstraints(pod v1.Pod, strategy api.DeschedulerStrategy) []v1.TopologySpreadConstraint {
+	topologySpreadConstraints := pod.Spec.TopologySpreadConstraints
+	if len(topologySpreadConstraints) == 0 && strategy.Params != nil && strategy.Params.TopologySpreadConstraint != nil && len(strategy.Params.TopologySpreadConstraint.DefaultConstraints) > 0 {
+		topologySpreadConstraints = strategy.Params.TopologySpreadConstraint.DefaultConstraints
+	}
+	return topologySpreadConstraints
```
damemi, line 372: maybe rewrite to show the hierarchy a bit more:

```go
if len(pod.Spec.TopologySpreadConstraints) > 0 {
  return pod.Spec.TopologySpreadConstraints
}

if strategy.Params != nil && strategy.Params.TopologySpreadConstraint != nil && len(strategy.Params.TopologySpreadConstraint.DefaultConstraints) > 0 {
  return strategy.Params.TopologySpreadConstraint.DefaultConstraints
}

return nil
```

a7i:
While the idea sounds great, I am leaning towards closing the PR for 2 reasons:

1. There's a possibility to get eviction loops since Scheduler will not have the PodTopologySpread definition. Moreover, if the `KubeSchedulerConfiguration` and descheduler default policy ever get out of sync, we will see unexpected eviction loops.
2. Combination of components such as Scheduler, cluster-autoscaler, and Descheduler not being in sync can lead to unexpected behaviors, mostly leading to unnecessary evictions.

damemi:
@a7i I share those concerns too. I figured it would be reasonable to offer this with a bold warning that it must be kept in sync with the scheduler config, but there is still a risk (and burden on the user) with that as you pointed out. Adding in other components like cluster autoscaler complicates it too

PR #833, opened 2022-06-05
PodLifeTime: sort pods by creation timestamp

### LINKED ISSUE #816: Descheduler should sort pods by age for PodLifeTime

descheduler version: 0.23.1

Unknown

```
deschedulerPolicy:
  evictLocalStoragePods: true
  ignorePvcPods: true
  nodeSelector: ""
  strategies:
    HighNodeUtilization:
      enabled: false
      params:
        nodeFit: true
          thresholds:
            cpu: 40
            memory: 70
            pods: 80
    LowNodeUtilization:
      enabled: false
    PodLifeTime:
      enabled: true
      params:
        labelSelector:
          matchLabels:
            app.umbra.space/deschedule-pod-lifetime: "true"
        podLifeTime:
          maxPodLifeTimeSeconds: 120
    RemoveDuplicates:
      enabled: false
    RemoveFailedPods:
      enabled: false
    RemovePodsHavingTooManyRestarts:
      enabled: false
      enabled: false
      enabled: false
      enabled: false
    RemovePodsViolatingTopologySpreadConstraint:
      enabled: false
kind: Deployment
nodeSelector:
  node.umbra.space/basenode: "true"
```

1.22

**What did you do?**

Set `maxPodLifeTimeSeconds` to 120 and `deschedulingInterval` to 5m.  Using an example of a two replica deployment with a pod disruption budget  `minAvailable` set to 1.  This will result in descheduler trying to evict both pods because both will always be older than 120s.  One eviction will fail due to the pdb.

The descheduler should sort by the pods by age and evict the oldest one first.  The descheduler seems to always evict the "first" pod as listed in `kubectl get pods` which sorts alphanumerically.

The descheduler should evict the older pod first.

**What did you see instead?**

The descheduler always evicts the first pod listed and therefore the second continues to grow in age.

a7i:
Closes #816 

In the event that PDB or settings such maxNoOfPodsToEvictPer* prevent too much eviction, we need to make sure that oldest pod is evicted first.

### DIFF
--- a/pkg/descheduler/pod/pods_test.go
+++ b/pkg/descheduler/pod/pods_test.go
@@ -18,8 +18,11 @@ package pod
 
 import (
 	"context"
+	"fmt"
+	"math/rand"
 	"reflect"
 	"testing"
+	"time"
 
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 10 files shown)


a7i:
I think it's because of the pointer lib: `"k8s.io/utils/pointer"`

pkg/descheduler/pod/pods_test.go
```diff
@@ -154,3 +157,27 @@ func TestSortPodsBasedOnPriorityLowToHigh(t *testing.T) {
...
+
+	for i := 0; i < len(podList); i++ {
+		podList[i] = test.BuildTestPod(fmt.Sprintf("p%d", i), 1, 32, n1.Name, func(pod *v1.Pod) {
+			creationTimestamp := metav1.Now().Add(time.Minute * time.Duration(i))
+			pod.ObjectMeta.SetCreationTimestamp(metav1.NewTime(creationTimestamp))
+		})
+	}
+
+	rand.Seed(time.Now().UnixNano())
+	rand.Shuffle(len(podList), func(i, j int) { podList[i], podList[j] = podList[j], podList[i] })
```
ingvagabund, line 173: Nit: I wonder what's the percentage of cases where the shuffled list is identical to its original ordering.

ingvagabund (approved):

pkg/descheduler/pod/pods_test.go
a7i: I think it should be factorial so pretty low. If my calculation is correct `1 / 40320`

But I can refactor it in another PR

a7i: Addressed in: https://github.com/kubernetes-sigs/descheduler/pull/856

PR #834, opened 2022-06-05
PodLifeTime: support container states PodInitializing and ContainerCreating

### LINKED ISSUE #580: podStatusPhases support containercreating status

some network issue or cni issue may cause pods always containercreating 
sth like https://github.com/kubernetes-sigs/descheduler/issues/62

podStatusPhases now only support  `Running` and `Pending` should add some other status

a7i:
Closes #580

This will allow users to define strategies to:
- If a Pod is stuck in container state PodInitializing for longer than ..
- If a Pod is stuck in container state ContainerCreating for longer than ..

### DIFF
--- a/README.md
+++ b/README.md
@@ -524,15 +524,17 @@ strategies:
 
 This strategy evicts pods that are older than `maxPodLifeTimeSeconds`.
 
-You can also specify `podStatusPhases` to `only` evict pods with specific `StatusPhases`, currently this parameter is limited
-to `Running` and `Pending`.
+You can also specify `states` parameter to **only** evict pods matching the following conditions:
+  - [Pod Phase](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase) status of: `Running`, `Pending`
+  - [Container State Waiting](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#container-state-waiting) condition of: `PodInitializing`, `ContainerCreating`
 
 **Parameters:**
 
 |Name|Type|
 |---|---|
 |`maxPodLifeTimeSeconds`|int|
-|`podStatusPhases`|list(string)|
+|`podStatusPhases` (**Deprecated**. Use `states` instead)|list(string)|
+|`states`|list(string)|
 |`thresholdPriority`|int (see [priority filtering](#priority-filtering))|
 |`thresholdPriorityClassName`|string (see [priority filtering](#priority-filtering))|
 |`namespaces`|(see [namespace filtering](#namespace-filtering))|
@@ -549,8 +551,9 @@ strategies:
      params:
        podLifeTime:
          maxPodLifeTimeSeconds: 86400
-         podStatusPhases:
+         states:
          - "Pending"
+         - "PodInitializing"
 ```
 
 ### RemoveFailedPods
--- a/pkg/descheduler/strategies/pod_lifetime.go
+++ b/pkg/descheduler/strategies/pod_lifetime.go
@@ -32,7 +32,7 @@ import (
 )
 
 var (
-	podLifetimeAllowedPhases = sets.NewString(
+	podLifeTimeAllowedStates = sets.NewString(
 		string(v1.PodRunning),
 		string(v1.PodPending),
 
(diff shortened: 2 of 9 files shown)


JaneLiuL (commented):
/lgtm

pkg/descheduler/strategies/pod_lifetime.go
```diff
@@ -84,7 +93,14 @@ func PodLifeTime(ctx context.Context, client clientset.Interface, strategy api.D
 				if string(pod.Status.Phase) == phase {
```
ingvagabund, line 93: From https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase:
```
The number and meanings of Pod phase values are tightly guarded.
Other than what is documented here, nothing should be assumed
about Pods that have a given phase value.
```

We use `pod status phase` term which might be easy to confuse with `pod phase`. Also, the upstream documentation mentions `container status` with three possible values: `Waiting`, `Running`, and `Terminated`.

With `PodInitializing` and `ContainerCreating` as the pod status phase values it will get even more confusing what is what. Although, I agree it's practical and preferable to have a single field for all the values.

If anything, we need to be explicit about what the field is used for in https://github.com/kubernetes-sigs/descheduler#podlifetime. Though, I prefer to use more generic word "state" rather than "status" if the list of values gets extended with `PodInitializing` and `ContainerCreating`. Worth adding a new field and deprecating `podStatusPhases`.

CC @knelasevero

a7i: That makes sense to me.

If we have a consensus, then I can add the new field.

knelasevero: Even if it is a bit confusing, wouldn't it be better to take advantage of the version being v1alpha1 and deprecate "before it's too late"? Alpha versions are for that, right?

a7i: There is a second [commit](https://github.com/kubernetes-sigs/descheduler/pull/834/commits/446dd5a11fb7c3927ec9fb4090d7511c8e2b2e6b) in this PR that adds the new field `States` and deprecates `PodStatusPhases`

README.md
```diff
@@ -524,15 +524,17 @@ strategies:
 
 This strategy evicts pods that are older than `maxPodLifeTimeSeconds`.
 
-You can also specify `podStatusPhases` to `only` evict pods with specific `StatusPhases`, currently this parameter is limited
-to `Running` and `Pending`.
+You can also specify `states` parameter to **only** evict pods matching the following conditions:
+  - [Pod Phase](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase) status of: `Running`, `Pending`
+  - [`ContainerStateWaiting`](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#container-states) condition of: `PodInitializing`, `ContainerCreating`
```
damemi, line 529: Is this supposed to say `ContainerStateWaiting` or `ContainerState`? The link talks about Waiting, Running, and Terminating states. Maybe this section could be clarified what the states are, feel free to add more description outside of the bullet points

pkg/descheduler/strategies/pod_lifetime.go
```diff
@@ -32,16 +32,26 @@ import (
...
 		return fmt.Errorf("MaxPodLifeTimeSeconds not set")
 	}
 
-	if params.PodLifeTime.PodStatusPhases != nil {
-		for _, phase := range params.PodLifeTime.PodStatusPhases {
-			if phase != string(v1.PodPending) && phase != string(v1.PodRunning) {
-				return fmt.Errorf("only Pending and Running phases are supported in PodLifeTime")
-			}
+	states := getPodLifeTimeStates(params.PodLifeTime)
+	if states != nil {
```
damemi, line 52: do you need a `nil` check for the empty list?

pkg/descheduler/strategies/pod_lifetime.go
```diff
@@ -78,15 +88,24 @@ func PodLifeTime(ctx context.Context, client clientset.Interface, strategy api.D
 	evictable := podEvictor.Evictable(evictions.WithPriorityThreshold(thresholdPriority))
 
 	filter := evictable.IsEvictable
-	if strategy.Params.PodLifeTime.PodStatusPhases != nil {
-		filter = func(pod *v1.Pod) bool {
-			for _, phase := range strategy.Params.PodLifeTime.PodStatusPhases {
-				if string(pod.Status.Phase) == phase {
-					return evictable.IsEvictable(pod)
+
+	states := getPodLifeTimeStates(strategy.Params.PodLifeTime)
```
damemi, line 92: `validatePodLifeTimeParams` could return the states so that this doesn't need to be its own function called twice

README.md
a7i: Currently PodLifeTime will only filter on `Reasons` of `State.Waiting`

Code [here](https://github.com/kubernetes-sigs/descheduler/pull/834/files#diff-4b7a41236384b65d820a644397e833df40452f002d63e2ade2b145904fe04170R101)

pkg/descheduler/strategies/pod_lifetime.go
a7i: True. It would be somewhat odd for the validate params to return only a subset of the params.

a7i: Applied

README.md
a7i: good catch on the link, updated

ingvagabund (changes requested):

README.md
```diff
@@ -524,15 +524,17 @@ strategies:
 
 This strategy evicts pods that are older than `maxPodLifeTimeSeconds`.
 
-You can also specify `podStatusPhases` to `only` evict pods with specific `StatusPhases`, currently this parameter is limited
-to `Running` and `Pending`.
+You can also specify `states` parameter to **only** evict pods matching the following conditions:
+  - [Pod Phase](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase) status of: `Running`, `Pending`
+  - [`ContainerStateWaiting`](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#container-state-waiting) condition of: `PodInitializing`, `ContainerCreating`
```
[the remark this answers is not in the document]
a7i: Is that in reference to `ContainerStateWaiting`? If so, probably not.

pkg/descheduler/strategies/pod_lifetime.go
a7i: ping

pkg/descheduler/strategies/pod_lifetime.go
```diff
@@ -31,33 +31,48 @@ import (
...
-				return fmt.Errorf("only Pending and Running phases are supported in PodLifeTime")
-			}
-		}
+		states = append(states, params.PodLifeTime.PodStatusPhases...)
+	}
+	if params.PodLifeTime.States != nil {
+		states = append(states, params.PodLifeTime.States...)
+	}
+	if !podLifeTimeAllowedStates.HasAll(states...) {
+		return nil, fmt.Errorf("states must be one of %v", podLifeTimeAllowedStates.List())
```
damemi, line 58: nit: this would be nice if it said which state was invalid, but that can be a todo for later

a7i: that's a good idea, PR coming up soon

README.md
```diff
@@ -524,15 +524,17 @@ strategies:
 
 This strategy evicts pods that are older than `maxPodLifeTimeSeconds`.
 
-You can also specify `podStatusPhases` to `only` evict pods with specific `StatusPhases`, currently this parameter is limited
-to `Running` and `Pending`.
+You can also specify `states` parameter to **only** evict pods matching the following conditions:
```
blampe, line 527: @a7i the latest release (0.24.1) doesn't actually support the `states` parameter, and silently ignores it if it's specified. Someone reading these instructions who installs 0.24.1 and enables the `PodLifeTime` policy with `states` other than `Running` will be unpleasantly surprised when their `Running` pods are terminated... 

I'd recommend clarifying the instructions to specify which versions accept the `state` parameter. I'd also suggest reconsidering whether `Running` pods should be targeted by default; it would be more reasonable (IMO) to require `states` so there's no room for surprises like this.

a7i: Hi @blampe thanks for the feedback.

Looking at the [0.24.1 README](https://github.com/kubernetes-sigs/descheduler/tree/v0.24.1), I do not see `states` as a configuration so I assume you're suggesting that regardless of which README you're looking at (main branch or a tag), the policy description should specify the minimum version a field is supported in. As an alternative, what is your feedback on [fail fast validation](https://github.com/kubernetes-sigs/descheduler/pull/706) where "erroneous" fields cause Descheduler to fail?

As far as excluding `Running` from `states`, this would be a breaking change. Would it be sufficient to update the documentation to indicate that ALL pods regardless of state/status are candidates for eviction if `states` / `podStatusPhases` is not specified?

a7i: https://github.com/kubernetes-sigs/descheduler/pull/928

PR #836, opened 2022-06-05
TopologySpreadConstraint: only evaluate nodes below ideal avg when balancing domains

### LINKED ISSUE #835: TopologySpreadConstraint strategy considers already balanced nodes when checking for nodeFit

descheduler version: 0.24.0

**Does this issue reproduce with the latest release?** Yes

**Which descheduler CLI options are you using?** N/A

```yaml
kind: "DeschedulerPolicy"
strategies:
  RemovePodsViolatingTopologySpreadConstraint:
    enabled: true
    params:
      nodeFit: true
      includeSoftConstraints: true
```

$ kubectl version
Client Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.3", GitCommit:"c92036820499fedefec0f847e2054d824aea6cd1", GitTreeState:"clean", BuildDate:"2021-10-27T18:34:20Z", GoVersion:"go1.16.10", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.1", GitCommit:"5e58841cce77d4bc13713ad2b91fa0d961e69192", GitTreeState:"clean", BuildDate:"2021-05-21T23:01:33Z", GoVersion:"go1.16.4", Compiler:"gc", Platform:"linux/amd64"}
</pre></details>

**What did you do?** 
See [thread](https://github.com/kubernetes-sigs/descheduler/pull/636#issuecomment-983751152) for additional background/info.

If I have `3` nodes in the cluster with 3 domains size `[2 3 4]`, `maxSkew=1` (where is skew is violated)
- `node1` (2 pods scheduled) is at max capacity (requests cpu/mem 100%)
- `node2` (3 pods scheduled) has adequate capacity (requests cpu/mem very low)
- `node3` (4 pods scheduled) has adequate capacity (requests cpu/mem very low)

I would expect no Pods to be evicted.
Given that `node1` has no capacity to give and `node2` is already balanced, nothing should be evicted from `node3`.

**What did you see instead?**
One Pod was evicted from `node3` because `node2` has capacity.

**Additional Notes**
`balanceDomains` function will identify that aboveAvg is 4 and needs to evict one since `PodFitsAnyOtherNode` function returns true given that the already balanced node (i.e. `node2`) still has adequate capacity. This function should ignore nodes/domains that are already balanced, i.e. only nodes `belowAvg`

cc: @RyanDevlin

a7i:
Closes #835 #839

TopologySpreadConstraint should only evaluate nodes below ideal avg when balancing domains to prevent eviction loop when nodes aboveAvg are taken into account.

### DIFF
--- a/pkg/descheduler/strategies/topologyspreadconstraint.go
+++ b/pkg/descheduler/strategies/topologyspreadconstraint.go
@@ -159,7 +159,7 @@ func RemovePodsViolatingTopologySpreadConstraint(
 					continue
 				}
 
-				// 5. If the pod's node matches this constraint'selector topologyKey, create a topoPair and add the pod
+				// 5. If the pod's node matches this constraint's topologyKey, create a topoPair and add the pod
 				node, ok := nodeMap[namespacePods.Items[i].Spec.NodeName]
 				if !ok {
 					// If ok is false, node is nil in which case node.Labels will panic. In which case a pod is yet to be scheduled. So it's safe to just continue here.
@@ -179,7 +179,7 @@ func RemovePodsViolatingTopologySpreadConstraint(
 				klog.V(2).InfoS("Skipping topology constraint because it is already balanced", "constraint", constraint)
 				continue
 			}
-			balanceDomains(client, getPodsAssignedToNode, podsForEviction, constraint, constraintTopologies, sumPods, evictorFilter.Filter, nodes)
+			balanceDomains(getPodsAssignedToNode, podsForEviction, constraint, constraintTopologies, sumPods, evictorFilter.Filter, nodes)
 		}
 	}
 
@@ -244,7 +244,6 @@ func topologyIsBalanced(topology map[topologyPair][]*v1.Pod, constraint v1.Topol
 // [5, 5, 5, 5, 5, 5]
 // (assuming even distribution by the scheduler of the evicted pods)
 func balanceDomains(
-	client clientset.Interface,
 	getPodsAssignedToNode podutil.GetPodsAssignedToNodeFunc,
 	podsForEviction map[*v1.Pod]struct{},
 	constraint v1.TopologySpreadConstraint,
@@ -255,6 +254,8 @@ func balanceDomains(
 
 	idealAvg := sumPods / float64(len(constraintTopologies))
 	sortedDomains := sortDomains(constraintTopologies, isEvictable)
+
+	nodesBelowIdealAvg := filterNodesBelowIdealAvg(nodes, sortedDomains, constraint.TopologyKey, idealAvg)
 
 	// i is the index for belowOrEqualAvg
 	// j is the index for aboveAvg
@@ -306,7 +307,8 @@ func balanceDomains(
 			// In other words, PTS can perform suboptimally if some of its chosen pods don't fit on other nodes.
 			// This is because the chosen pods aren't sorted, but immovable pods still count as "evicted" toward the PTS algorithm.
 			// So, a better selection heuristic could improve performance.
-			if !node.PodFitsAnyOtherNode(getPodsAssignedToNode, aboveToEvict[k], nodes) {
+
+			if !node.PodFitsAnyOtherNode(getPodsAssignedToNode, aboveToEvict[k], nodesBelowIdealAvg) {
 				klog.V(2).InfoS("ignoring pod for eviction as it does not fit on any other node", "pod", klog.KObj(aboveToEvict[k]))
 				continue
 			}
@@ -316,6 +318,25 @@ func balanceDomains(
 		sortedDomains[j].pods = sortedDomains[j].pods[:len(sortedDomains[j].pods)-movePods]
 		sortedDomains[i].pods = append(sortedDomains[i].pods, aboveToEvict...)
 	}
+}
+
+// filterNodesBelowIdealAvg will return nodes that have fewer pods matching topology domain than the idealAvg count.
+// the desired behavior is to not consider nodes in a given topology domain that are already packed.
+func filterNodesBelowIdealAvg(nodes []*v1.Node, sortedDomains []topology, topologyKey string, idealAvg float64) []*v1.Node {
+	topologyNodesMap := make(map[string][]*v1.Node, len(sortedDomains))
+	for _, node := range nodes {
+		if topologyDomain, ok := node.Labels[topologyKey]; ok {
+			topologyNodesMap[topologyDomain] = append(topologyNodesMap[topologyDomain], node)
+		}
+	}
+
+	var nodesBelowIdealAvg []*v1.Node
+	for _, domain := range sortedDomains {
+		if float64(len(domain.pods)) < idealAvg {
+			nodesBelowIdealAvg = append(nodesBelowIdealAvg, topologyNodesMap[domain.pair.value]...)
+		}
+	}
+	return nodesBelowIdealAvg
 }
 
 // sortDomains sorts and splits the list of topology domains based on their size
--- a/pkg/descheduler/strategies/topologyspreadconstraint_test.go
+++ b/pkg/descheduler/strategies/topologyspreadconstraint_test.go
@@ -936,6 +936,127 @@ func TestTopologySpreadConstraint(t *testing.T) {
 			},
 			namespaces: []string{"ns1"},
 		},
+		{
+			name: "3 domains, sizes [2,3,4], maxSkew=1, NodeFit is enabled, and not enough cpu on zoneA; nothing should be moved",
+			nodes: []*v1.Node{
+				test.BuildTestNode("n1", 250, 2000, 9, func(n *v1.Node) { n.Labels["zone"] = "zoneA" }),
+				test.BuildTestNode("n2", 1000, 2000, 9, func(n *v1.Node) { n.Labels["zone"] = "zoneB" }),
+				test.BuildTestNode("n3", 1000, 2000, 9, func(n *v1.Node) { n.Labels["zone"] = "zoneC" }),
+			},
+			pods: createTestPods([]testPodList{
+				{
+					count:       2,
+					node:        "n1",
+					labels:      map[string]string{"foo": "bar"},
+					constraints: getDefaultTopologyConstraints(1),
+				},
+				{
+					count:       3,
+					node:        "n2",
+					labels:      map[string]string{"foo": "bar"},
+					constraints: getDefaultTopologyConstraints(1),
+				},
+				{
+					count:       4,
+					node:        "n3",
+					labels:      map[string]string{"foo": "bar"},
+					constraints: getDefaultTopologyConstraints(1),
+				},
+			}),
+			expectedEvictedCount: 0,
+			strategy: api.DeschedulerStrategy{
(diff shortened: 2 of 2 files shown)


JaneLiuL (commented):
/lgtm

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -284,7 +291,7 @@ func balanceDomains(
 			// In other words, PTS can perform suboptimally if some of its chosen pods don't fit on other nodes.
 			// This is because the chosen pods aren't sorted, but immovable pods still count as "evicted" toward the PTS algorithm.
 			// So, a better selection heuristic could improve performance.
-			if !node.PodFitsAnyOtherNode(getPodsAssignedToNode, aboveToEvict[k], nodes) {
+			if !node.PodFitsAnyOtherNode(getPodsAssignedToNode, aboveToEvict[k], topologyNodesMap[sortedDomains[i].pair.value]) {
```
ingvagabund, line 294: This limits the check to nodes with the same topology domain value. Which makes the solution less suboptimal. Example:
Lets have the following distribution of pods among domains:
- nodeA1{zone=A, pods capacity=1}: 1
- nodeA2{zone=A, pods capacity=4}: 4
- nodeB1{zone=B, pods capacity=2}: 2
- nodeB2{zone=B, pods capacity=2}: 1

After getting processed:
```
sortedDomains = [
	topology{
		pair={key=zone,value=B},
		pods=[6,7,8],
	},
	topology{
		pair={key=zone,value=A},
		pods=[1,2,3,4,5],
	},
]
topologyNodesMap = {
	"A": [nodeA1, nodeA2],
	"B": [nodeB1, nodeB2],
}
skew=2 (5-3)
```

To get skew=1 one can move a pod from nodeA2 to nodeB2 since nodeA1 does not have resources for 2 pods. With the change applied in this PR `node.PodFitsAnyOtherNode` will be false as with `j=1` `topologyNodesMap[sortedDomains[1].pair.value] = topologyNodesMap["A"] = [nodeA1, nodeA2]`.

Did you consider this case?

a7i: ping

ingvagabund: @a7i apologies for the delay. I was about to share the comment yesterday thought it was too late in the night to provide a failing unit test.

ingvagabund: @damemi also when running the unit tests ^^, [namespaceTopologySpreadConstraints](https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/descheduler/strategies/topologyspreadconstraint.go#L114) collects 30 items even though there's only one distinct constraint. Worth adding a check if a constraint already exists in `namespaceTopologySpreadConstraints`.

a7i: Thanks for the details, that makes sense.

Also, regarding the 30 items, it is due to LabelSelector being a pointer to a struct, where the comparison is on the memory address rather than the content of the selector. I can look into patching this up.
namespaceTopologySpreadConstraints

knelasevero: I plan to take care of this namespaceTopologySpreadConstraints issue after this discussion unfolds :) I will keep an eye on it

a7i: fyi, the test case with 6 domains seems to be a separate issue (fails on current and this new code). There are scenarios where the `j` position has less than `j-1` so it seems to me that we need to re-sort at the end of each iteration to ensure that index `j` is the largest in size.

ingvagabund: What about instead of passing `topologyNodesMap[sortedDomains[i].pair.value]` to pass all `topologyNodesMap[sortedDomains[i].pair.value]` for `i=0` until `i=l`, where `l` is the last domain that has size < `idealAvg`?

> There are scenarios where the j position has less than j-1 so it seems to me that we need to re-sort at the end of each iteration to ensure that index j is the largest in size.

I wonder if it is related to:
```
// skew = actual difference between the domains
skew := float64(len(sortedDomains[j].pods) - len(sortedDomains[i].pods))

// if k and j are within the maxSkew of each other, move to next belowOrEqualAvg
if int32(skew) <= constraint.MaxSkew {
	i++
	continue
}
```

I'd expect the condition to rather be:
```
if len(sortedDomains[i].pods) >= idealAvg {
	i++
	continue
}
```

Checking the code forth and back I don't see a need to re-sort the domains.

a7i: Thanks for the hint 🥇 , I was thinking something along those lines. Will report back soon

ingvagabund: > we need to re-sort at the end of each iteration to ensure that index j is the largest in size.

I see it now. In the 6 domains test case these are the movement in the domains:
- len(sortedDomains[0].pods)=1, len(sortedDomains[5].pods=9
- len(sortedDomains[0].pods)=4, len(sortedDomains[5].pods=6
- len(sortedDomains[1].pods)=2, len(sortedDomains[5].pods=6
- len(sortedDomains[1].pods)=3, len(sortedDomains[5].pods=5
- len(sortedDomains[1].pods)=3, len(sortedDomains[4].pods=8
- len(sortedDomains[1].pods)=5, len(sortedDomains[4].pods=6
- len(sortedDomains[2].pods)=4, len(sortedDomains[4].pods=6
- len(sortedDomains[3].pods)=6, len(sortedDomains[4].pods=6

The third line causes another eviction from the last domain even though 6 pods in the last domain leads to "correct" re-distribution of pods. The domains need to be re-sorted as you say so a domain with the highest number of pods is always the one giving pods.

ingvagabund: After sorting the domains after each cycle the final distribution of pods is calculated as: [5, 4, 4, 5, 6, 6]. Without the sorting bits it is `[4, 5, 4, 6, 6, 5]`. With the same number of evictions.

I can't find an example where the final distribution of pods is wrong while keeping the domains unsorted. Not sure if sorting the pods will actually make any difference.

a7i: I noticed that I needed to sort both i and j

ingvagabund: Experimenting with three options:
- no sorting
- sorting only the higher domains (when j-th domain's size gets decreased, move the domain to the left until all domains to the right have higher number of pods)
- sorting all the domains (when j-th domain's size gets decreased, move the domain to the left until all domains to the right have higher number of pods; when the i-th domain size gets increased, move the domain to the right until all domains to the left have lower number of pods)

Final distribution of all the options with the number of iterations of the `for i < j` cycle and evictions:
- [4, 5, 4, 6, 6, 5], iterations=7, evictions=6
- [5, 4, 4, 5, 6, 6], iterations=8, evictions=6
- [4, 4, 4, 6, 6, 6], iterations=7, evictions=5

Sequences:
- no sorting
  ```
  [1, 2, 4, 6, 8, 9]
  #[4, 2, 4, 6, 8, 6], evicted=3
  [4, 2, 4, 6, 8, 6]
  [4, 2, 4, 6, 8, 6]
  #[4, 3, 4, 6, 8, 5], evicted=1
  [4, 3, 4, 6, 8, 5]
  #[4, 5, 4, 6, 6, 5], evicted=2
  [4, 5, 4, 6, 6, 5]
  [4, 5, 4, 6, 6, 5]
  [4, 5, 4, 6, 6, 5]
  iterations: 7, all_evictions: 6, movements: 3
  ```
- the second half getting sorted
  ```
  [1, 2, 4, 6, 8, 9]
  #[4, 2, 4, 6, 6, 8], evicted=3
  [4, 2, 4, 6, 6, 8]
  #[5, 2, 4, 6, 6, 7], evicted=1
  [5, 2, 4, 6, 6, 7]
  [5, 2, 4, 6, 6, 7]
  #[5, 4, 4, 5, 6, 6], evicted=2
  [5, 4, 4, 5, 6, 6]
  [5, 4, 4, 5, 6, 6]
  [5, 4, 4, 5, 6, 6]
  [5, 4, 4, 5, 6, 6]
  iterations: 8, all_evictions: 6, movements: 3
  ```
- all sorted
  ```
  [1, 2, 4, 6, 8, 9]
  #[2, 4, 4, 6, 6, 8], evicted=3
  [2, 4, 4, 6, 6, 8]
  #[4, 4, 4, 6, 6, 6], evicted=2
  [4, 4, 4, 6, 6, 6]
  [4, 4, 4, 6, 6, 6]
  [4, 4, 4, 6, 6, 6]
  [4, 4, 4, 6, 6, 6]
  [4, 4, 4, 6, 6, 6]
  iterations: 7, all_evictions: 5, movements: 2
  ```

ingvagabund: @damemi was sorting the pods after each [movement of pods](https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/descheduler/strategies/topologyspreadconstraint.go#L311-L312) something you considered/abandoned when designing the algorithm? At this point it looks like an interesting improvement to the algorithm. Would be great to generate random sets of domains and measure the number of iterations/movements/evictions of the algorithm to see if there's any larger improvement.

a7i: Very nice, thank you. I was trying to find out how to track which pods get evicted. We definitely need that assertion as part of our unit tests.

a7i: Would you PTAL again? @ingvagabund @damemi @knelasevero 

- Balance domains: Ensure only nodes with pods below average count are considered as candidates for fit
- Test: Ensure specific pods were evicted (I did not do the assertion for ALL of the existing tests but we can do so after the [movements of pods](https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/descheduler/strategies/topologyspreadconstraint.go#L311-L312) changes)

knelasevero: It looks good to me so far. Very cool to now have specific pod evictions being checked. I want to go over it a bit more consciously and circle back here.

a7i: We have been running this in production and seeing a lot of improvements. Any chance this can be a patch candidate for 0.24?

damemi: Not to sound lazy, but due to the work involved with patch releases and the timing for the next release, non-critical fixes are generally not worth the effort. I'm not opposed to it if there is a lot of demand though.

a7i: Understood and makes sense. This was really urgent for us but we recently decided to run this fork for now.
eviction-loop-same-node.

damemi (commented):
/approve

JaneLiuL (commented):
/lgtm
thanks your hard working @a7i

PR #840, opened 2022-06-07
nodeFit: fix docs spacing for placement of the field

a7i:


README.md
```diff
@@ -760,7 +760,7 @@ strategies:
           "cpu": 50
           "memory": 50
           "pods": 50
```
damemi, line 762: are these 3 indented wrong too?

a7i: interesting.. doesn't look that way in my IDE

damemi: @a7i not to nit about this but maybe the necessary indentation will be more readable if you move the `nodefit` param up above the thresholds in this example

Issue #837, opened 2022-06-06
Descheduling framework: define framework types


ingvagabund:
Based on comments from https://github.com/kubernetes-sigs/descheduler/issues/753:
- descheduler configuration with profiles (internal/v1alpha2) with conversion/validation/defaulting
- handle for injecting clientset, evictor, shared informer factory, etc.
- plugin extension point interfaces
- example strategy arguments + defaulting/validation

Use cases to keep in mind:
- **The descheduling framework can be configured**: A user needs to be allowed to specify which extension points are enabled/disabled, alongside specifying plugin configuration. Including ability to configure multiple descheduling profiles.
- **The descheduling framework configuration can be converted into an internal representation (e.g. v1alpha2 -> internal)**: To provide ability to introduce/remove features by increasing configuration versions, yet still having path to convert all configuration versions into a single internal so the framework internals are kept as stable as possible
- **The descheduling framework configuration gets defaulted**: To reduce need to specify value for every possible configuration, also defaulting serves as a recommended/opinionated settings for the plugins
- **The descheduling framework needs to provide basic handle for plugins**: The framework needs to set responsibility for/ownership of lifecycle of processes/workflows, i.e. the framework is responsible for initializing clientset, node lister, pod lister, evictor, etc. The framework handle types need to provide basic handles while still keeping the abstraction level high.
- **The descheduling plugin interface is correctly specified**: A plugin is an encapsulation of a user eviction decision making logic. It defines a boundary/communication protocol. Thus, it is important to clearly but simply define how the plugin is communicating with the framework. E.g. specifying the extension points.
- **The descheduling plugins are available to the framework through plugin registry**: The framework needs to provide an easy way of registering in-tree/out-of-tree plugins.
- **The descheduling framework code hierarchy is flexible**: The framework code needs to be properly divided so it is clear, easily extendable and mocked/testable.
- **Easy to read examples on how to build a plugin**: Provide few examples of plugin implementations

In overall, the task in question here is to define the data model of the framework. The functionality itself is part of the next issue(s).

**Pre-requisities**:
- [x] split the `PodEvictor` into the evictor and filter: https://github.com/kubernetes-sigs/descheduler/pull/847
- [x] simplify the pod evictor so it can be simply injected and invoked inside strategies: https://github.com/kubernetes-sigs/descheduler/pull/846

**Strategies for migration**:
- [x] RemoveDuplicates
- [x] LowNodeUtilization
- [x] HighNodeUtilization
[...]

PR #846, opened 2022-06-09
Prepare pod evictor for the descheduling framework plugin

ingvagabund:
- Pass the strategy name into evictor through context to reduce the list of parameters to miminum
- Drop node parameter as it can be retrieved from the pod object
- EvictPod: stop returning error and have higher processes decide when is the right time to exit a strategy loop

Pre-requisite for https://github.com/kubernetes-sigs/descheduler/issues/837

### DIFF
--- a/pkg/descheduler/descheduler.go
+++ b/pkg/descheduler/descheduler.go
@@ -302,7 +302,7 @@ func RunDeschedulerStrategies(ctx context.Context, rs *options.DeschedulerServer
 						continue
 					}
 					evictorFilter := evictions.NewEvictorFilter(nodes, getPodsAssignedToNode, evictLocalStoragePods, evictSystemCriticalPods, ignorePvcPods, evictBarePods, evictions.WithNodeFit(nodeFit), evictions.WithPriorityThreshold(thresholdPriority))
-					f(ctx, rs.Client, strategy, nodes, podEvictor, evictorFilter, getPodsAssignedToNode)
+					f(context.WithValue(ctx, "strategyName", string(name)), rs.Client, strategy, nodes, podEvictor, evictorFilter, getPodsAssignedToNode)
 				}
 			} else {
 				klog.ErrorS(fmt.Errorf("unknown strategy name"), "skipping strategy", "strategy", name)
--- a/pkg/descheduler/evictions/evictions.go
+++ b/pkg/descheduler/evictions/evictions.go
@@ -102,11 +102,18 @@ func (pe *PodEvictor) TotalEvicted() uint {
 	return total
 }
 
-// EvictPod returns non-nil error only when evicting a pod on a node is not
-// possible (due to maxPodsToEvictPerNode constraint). Success is true when the pod
-// is evicted on the server side.
-// eviction reason can be set through the ctx's evictionReason:STRING pair
-func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod) (bool, error) {
+// NodeLimitExceeded checks if the number of evictions for a node was exceeded
+func (pe *PodEvictor) NodeLimitExceeded(node *v1.Node) bool {
+	if pe.maxPodsToEvictPerNode != nil {
+		return pe.nodepodCount[node.Name] == *pe.maxPodsToEvictPerNode
+	}
+	return false
+}
+
+// EvictPod evicts a pod while exercising eviction limits.
+// Returns true when the pod is evicted on the server side.
+// Eviction reason can be set through the ctx's evictionReason:STRING pair
+func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod) bool {
 	strategy := ""
 	if ctx.Value("strategyName") != nil {
 		strategy = ctx.Value("strategyName").(string)
@@ -121,7 +128,8 @@ func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod) (bool, error) {
 			if pe.metricsEnabled {
 				metrics.PodsEvicted.With(map[string]string{"result": "maximum number of pods per node reached", "strategy": strategy, "namespace": pod.Namespace, "node": pod.Spec.NodeName}).Inc()
 			}
-			return false, fmt.Errorf("Maximum number %v of evicted pods per %q node reached", *pe.maxPodsToEvictPerNode, pod.Spec.NodeName)
+			klog.ErrorS(fmt.Errorf("Maximum number of evicted pods per node reached"), "limit", *pe.maxPodsToEvictPerNode, "node", pod.Spec.NodeName)
+			return false
 		}
 	}
 
@@ -129,7 +137,8 @@ func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod) (bool, error) {
 		if pe.metricsEnabled {
 			metrics.PodsEvicted.With(map[string]string{"result": "maximum number of pods per namespace reached", "strategy": strategy, "namespace": pod.Namespace, "node": pod.Spec.NodeName}).Inc()
 		}
-		return false, fmt.Errorf("Maximum number %v of evicted pods per %q namespace reached", *pe.maxPodsToEvictPerNamespace, pod.Namespace)
+		klog.ErrorS(fmt.Errorf("Maximum number of evicted pods per namespace reached"), "limit", *pe.maxPodsToEvictPerNamespace, "namespace", pod.Namespace)
+		return false
 	}
 
 	err := evictPod(ctx, pe.client, pod, pe.policyGroupVersion)
@@ -139,7 +148,7 @@ func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod) (bool, error) {
 		if pe.metricsEnabled {
 			metrics.PodsEvicted.With(map[string]string{"result": "error", "strategy": strategy, "namespace": pod.Namespace, "node": pod.Spec.NodeName}).Inc()
 		}
-		return false, nil
+		return false
 	}
 
 	if pod.Spec.NodeName != "" {
@@ -161,7 +170,7 @@ func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod) (bool, error) {
 		r := eventBroadcaster.NewRecorder(scheme.Scheme, v1.EventSource{Component: "sigs.k8s.io.descheduler"})
 		r.Event(pod, v1.EventTypeNormal, "Descheduled", fmt.Sprintf("pod evicted by sigs.k8s.io/descheduler%s", reason))
 	}
-	return true, nil
+	return true
 }
 
 func evictPod(ctx context.Context, client clientset.Interface, pod *v1.Pod, policyGroupVersion string) error {
--- a/pkg/descheduler/evictions/evictions.go
+++ b/pkg/descheduler/evictions/evictions.go
@@ -44,7 +44,7 @@ const (
 )
 
 // nodePodEvictedCount keeps count of pods evicted on node
-type nodePodEvictedCount map[*v1.Node]uint
+type nodePodEvictedCount map[string]uint
 type namespacePodEvictCount map[string]uint
 
 type PodEvictor struct {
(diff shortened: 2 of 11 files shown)


ingvagabund:
CC @knelasevero

pkg/descheduler/evictions/evictions.go
```diff
@@ -118,56 +117,68 @@ func (pe *PodEvictor) TotalEvicted() uint {
...
 }
 
-// EvictPod returns non-nil error only when evicting a pod on a node is not
-// possible (due to maxPodsToEvictPerNode constraint). Success is true when the pod
-// is evicted on the server side.
-func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod, node *v1.Node, strategy string, reasons ...string) (bool, error) {
-	reason := strategy
-	if len(reasons) > 0 {
-		reason += " (" + strings.Join(reasons, ", ") + ")"
+// EvictPod evicts a pod.
```
a7i, line 120: seems like a useless comment

pkg/descheduler/strategies/duplicates.go
```diff
@@ -210,10 +210,7 @@ func RemoveDuplicatePods(
 				// It's assumed all duplicated pods are in the same priority class
 				// TODO(jchaloup): check if the pod has a different node to lend to
 				for _, pod := range pods[upperAvg-1:] {
-					if _, err := podEvictor.EvictPod(ctx, pod, nodeMap[nodeName], "RemoveDuplicatePods"); err != nil {
-						klog.ErrorS(err, "Error evicting pod", "pod", klog.KObj(pod))
-						break
-					}
+					podEvictor.EvictPod(ctx, pod)
```
a7i, line 213: It appears that functionality is being changed here? In other words, before it would break on a single "failed" eviction but now it loops through all pods.

Should it be preserved?
```suggestion
					if !podEvictor.EvictPod(ctx, pod) {
						break
					}

**Comment by ingvagabund** [comment]
https://github.com/kubernetes-sigs/descheduler/pull/847 needs to be merged first.

File: pkg/descheduler/strategies/duplicates.go (hunk 2 above)
Comment (ingvagabund) [reply pkg/descheduler/strategies/duplicates.go:213]: In the past there was only a single error issued (exceeding limit of number of pods allowed to be evicted per node). So it made sense to break and move to another node. However, few months back we added a [limit](https://github.com/kubernetes-sigs/descheduler/pull/658) for the number of pods evicted per namespace. So when the namespace limit gets exceeded we can still continue evicting.

Though, I plan to introduce a check for the node limit exceeded. I will incorporate it in this PR once https://github.com/kubernetes-sigs/descheduler/pull/847 is merged.

Comment (ingvagabund) [reply pkg/descheduler/strategies/duplicates.go:203]: Just introduced `NodeLimitExceeded` method for performing the check.

File: pkg/descheduler/evictions/evictions.go [hunk 3]
```diff
@@ -118,56 +117,68 @@ func (pe *PodEvictor) TotalEvicted() uint {
...
+// Eviction reason can be set through the ctx's evictionReason:STRING pair
+func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod) bool {
+	strategy := ""
+	if ctx.Value("strategyName") != nil {
+		strategy = ctx.Value("strategyName").(string)
 	}
-	if pe.maxPodsToEvictPerNode != nil && pe.nodepodCount[node]+1 > *pe.maxPodsToEvictPerNode {
+	reason := ""
+	if ctx.Value("evictionReason") != nil {
+		strategy = ctx.Value("evictionReason").(string)
```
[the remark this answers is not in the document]
ingvagabund: Thx, done

pkg/descheduler/evictions/evictions.go
```diff
@@ -103,56 +102,76 @@ func (pe *PodEvictor) TotalEvicted() uint {
...
+	strategy := ""
+	if ctx.Value("strategyName") != nil {
+		strategy = ctx.Value("strategyName").(string)
+	}
+	reason := ""
+	if ctx.Value("evictionReason") != nil {
+		reason = ctx.Value("evictionReason").(string)
+	}
+
+	if pod.Spec.NodeName == "" {
```
a7i, line 126: PodLifeTime strategy allows Pods in `Pending` state. I assume there could be a state where the Pod has not been scheduled yet.

[/b2418ef481298c6caf185a5f88dd0bb6ddc1cdbf/pkg/descheduler/strategies/pod_lifetime.go#L42](https://github.com/kubernetes-sigs/descheduler/blob/b2418ef481298c6caf185a5f88dd0bb6ddc1cdbf/pkg/descheduler/strategies/pod_lifetime.go#L42)

pkg/descheduler/strategies/node_taint.go
```diff
@@ -106,9 +107,9 @@ func RemovePodsViolatingNodeTaints(ctx context.Context, client clientset.Interfa
 				taintFilterFnc,
 			) {
 				klog.V(2).InfoS("Not all taints with NoSchedule effect are tolerated after update for pod on node", "pod", klog.KObj(pods[i]), "node", klog.KObj(node))
-				if _, err := podEvictor.EvictPod(ctx, pods[i], node, "NodeTaint"); err != nil {
-					klog.ErrorS(err, "Error evicting pod")
-					break
+				podEvictor.EvictPod(ctx, pods[i])
+				if podEvictor.NodeLimitExceeded(node) {
+					continue loop
```
a7i, line 112:
```suggestion
					continue
```

pkg/descheduler/evictions/evictions.go
ingvagabund: I will alter the code to take this case into account. Thanks for noticing!

a7i: On second thought... after scanning the code, it seems that it only lists Pods on Nodes. We should probably discuss changing that but not part of this PR.

ingvagabund: We might remove condition in https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/descheduler/pod/pods.go#L129-L131 to have pending pods included in `getPodsAssignedToNode` function. Considering an empty nodename as a special case.

pkg/descheduler/evictions/evictions.go
```diff
@@ -103,56 +102,75 @@ func (pe *PodEvictor) TotalEvicted() uint {
...
-	if pe.maxPodsToEvictPerNode != nil && pe.nodepodCount[node]+1 > *pe.maxPodsToEvictPerNode {
-		if pe.metricsEnabled {
-			metrics.PodsEvicted.With(map[string]string{"result": "maximum number of pods per node reached", "strategy": strategy, "namespace": pod.Namespace, "node": node.Name}).Inc()
+	return false
+}
+
+// EvictPod evicts a pod while exercising eviction limits.
+// Returns true when the pod is evicted on the server side.
+// Eviction reason can be set through the ctx's evictionReason:STRING pair
+func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod) bool {
```
a7i, line 116: I'm curious to know why the return signature was changed to not return the error.

Previously, we had the flexibility to return an error for namespace limit reached, node limit reached, and the potential for some other limit in the future. But now it's up to each strategy to handle limit exceeded logic.

a7i: Is it because each plugin should now be responsible for deciding if it should continue/abort?

a7i, line 116: Can we continue to leave the `v1.Node` parameter? See https://github.com/kubernetes-sigs/descheduler/pull/859 for use-case

pkg/descheduler/strategies/node_affinity.go
```diff
@@ -74,6 +74,7 @@ func RemovePodsViolatingNodeAffinity(ctx context.Context, client clientset.Inter
 
 		switch nodeAffinity {
 		case "requiredDuringSchedulingIgnoredDuringExecution":
```
JaneLiuL, line 76: lgtm actually, but i just wonder why this name `requiredDuringSchedulingIgnoredDuringExecution` so long...

JaneLiuL (commented):
/lgtm

pkg/descheduler/evictions/evictions.go
ingvagabund: > Is it because each plugin should now be responsible for deciding if it should continue/abort?

There are several limits now. Not all of them require to stop processing the current node. E.g. when a namespace limit is reached, a plugin can continue processing another pod on the same node. Thus yes, a plugin is allowed to decide if the right course of action is to stop processing a node and continue with another one. Or, just skipping a pod and taking another on the same node

pkg/descheduler/strategies/node_affinity.go
ingvagabund: Chosen by the designers: https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#node-affinity

a7i (approved):

pkg/descheduler/strategies/node_affinity.go
JaneLiuL: get it. look good to me now~
/lgtm

pkg/descheduler/descheduler.go
```diff
@@ -302,7 +302,7 @@ func RunDeschedulerStrategies(ctx context.Context, rs *options.DeschedulerServer
 						continue
 					}
 					evictorFilter := evictions.NewEvictorFilter(nodes, getPodsAssignedToNode, evictLocalStoragePods, evictSystemCriticalPods, ignorePvcPods, evictBarePods, evictions.WithNodeFit(nodeFit), evictions.WithPriorityThreshold(thresholdPriority))
-					f(ctx, rs.Client, strategy, nodes, podEvictor, evictorFilter, getPodsAssignedToNode)
+					f(context.WithValue(ctx, "strategyName", string(name)), rs.Client, strategy, nodes, podEvictor, evictorFilter, getPodsAssignedToNode)
```
damemi, line 305: is this intended to be the final approach or just a work in progress step? (passing values through the context)

damemi:
/hold
just had 2 questions before this auto-merges

pkg/descheduler/strategies/duplicates.go
```diff
@@ -192,16 +192,17 @@ func RemoveDuplicatePods(
 		}
 
 		upperAvg := int(math.Ceil(float64(ownerKeyOccurence[ownerKey]) / float64(len(targetNodes))))
+	loop:
```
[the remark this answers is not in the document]
ingvagabund: This was a quick change. +1 to refactor the loops during the plugin migration

pkg/descheduler/descheduler.go
damemi: I think a better approach would be to define an "options" struct to pass as an optional param to `EvictPod`. that struct can have `strategy` and `reasons` fields to start with, and if we decide to add more options in the future then we don't need to change `EvictPod`'s signature. 

what do you think about something like that? As this is, I don't think it's an appropriate use of `context`

damemi: that said, I don't need to block on this right now since this PR has been open for a while (and that's my bad for not getting around to reviewing until now). If my idea sounds good I'll take it as a follow up refactor and we can unhold this PR

ingvagabund: We might as well move the strategy name from `EvictPod`. It's used mostly for metrics. Then two lines for logging which can be logged without the strategy name (the same for the reason). Instead, printing both the strategy name and the reason (if there's any) outside of `EvictPod` through a wrapper which will get introduced after the framework primitives are implemented.

My reasoning is to use `EvictPod` only for the actual eviction. The method does not need to know anything about the strategy/plugin/reason/etc. in order to evict a pod. We can have higher invokers to log the additional information.

> If my idea sounds good I'll take it as a follow up refactor and we can unhold this PR

+1 for refactoring the code more in the follow up PR.

damemi: I think the idea of passing additional info/options to `EvictPod` makes sense. It's already a wrapper for a private `evictPod` function (not sure why though), but as the public interface it seems like a good spot to expose a handle for customization/logging/metrics.

PR #847, opened 2022-06-13
Split pod evictor and evictor filter

ingvagabund:
The evictor filter is planned to be moved under a plugin. In order to initialize the evictor filter outside of the strategies, the `PodEvictor` is split into a pod evictor (responsible for actually evicting pods) and an evictor filter (responsible for deciding whether a pod can be evicted).

Pre-requisite for https://github.com/kubernetes-sigs/descheduler/issues/837

### DIFF
--- a/pkg/descheduler/descheduler.go
+++ b/pkg/descheduler/descheduler.go
@@ -45,6 +45,7 @@ import (
 	podutil "sigs.k8s.io/descheduler/pkg/descheduler/pod"
 	"sigs.k8s.io/descheduler/pkg/descheduler/strategies"
 	"sigs.k8s.io/descheduler/pkg/descheduler/strategies/nodeutilization"
+	"sigs.k8s.io/descheduler/pkg/utils"
 )
 
 func Run(ctx context.Context, rs *options.DeschedulerServer) error {
@@ -87,7 +88,7 @@ func Run(ctx context.Context, rs *options.DeschedulerServer) error {
 	return runFn()
 }
 
-type strategyFunction func(ctx context.Context, client clientset.Interface, strategy api.DeschedulerStrategy, nodes []*v1.Node, podEvictor *evictions.PodEvictor, getPodsAssignedToNode podutil.GetPodsAssignedToNodeFunc)
+type strategyFunction func(ctx context.Context, client clientset.Interface, strategy api.DeschedulerStrategy, nodes []*v1.Node, podEvictor *evictions.PodEvictor, evictorFilter *evictions.EvictorFilter, getPodsAssignedToNode podutil.GetPodsAssignedToNodeFunc)
 
 func cachedClient(
 	realClient clientset.Interface,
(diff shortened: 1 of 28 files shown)


ingvagabund:
CC @knelasevero

pkg/descheduler/descheduler.go
```diff
@@ -283,18 +284,25 @@ func RunDeschedulerStrategies(ctx context.Context, rs *options.DeschedulerServer
...
 		for name, strategy := range deschedulerPolicy.Strategies {
 			if f, ok := strategyFuncs[name]; ok {
 				if strategy.Enabled {
-					f(ctx, rs.Client, strategy, nodes, podEvictor, getPodsAssignedToNode)
+					nodeFit := false
+					if name != "PodLifeTime" {
+						if strategy.Params != nil {
+							nodeFit = strategy.Params.NodeFit
+						}
+					}
```
a7i, line 298: Ideally, in the future we should have the strategy initialized where the interface provides `NodeFit()` similar to [this](https://github.com/kubernetes-sigs/descheduler/pull/706/files#diff-67816189b07b98460b57a10b41280403c53a4dfda42fdd8f54475812cb99f56eR83)

ingvagabund: This is only a temporary hack to transition into the new v1alpha2 configuration where we can implement a builder/factory and run a validation to make sure the nodeFit gets set only where it makes sense.

damemi:
Is there a technical reason why we are choosing to start with migrating the PodEvictor? From https://github.com/kubernetes-sigs/descheduler/issues/837#issuecomment-1149127005 it seemed like we were going to start framework changes with the API and strategy updates

ingvagabund:
The PodEvictor's Evictable (future eviction filter) is initiliazed inside the strategies. Which disallows to migrate the strategies into plugins as every plugin is given a handle which has the eviction filter as one of its methods (`d.handle.Evictor().Filter`). I need to decouple the Evictable method from the strategies first.

pkg/descheduler/strategies/node_affinity_test.go
```diff
@@ -222,15 +222,25 @@ func TestRemovePodsViolatingNodeAffinity(t *testing.T) {
...
 				false,
 				false,
+				evictions.WithNodeFit(nodeFit),
 			)
 
-			RemovePodsViolatingNodeAffinity(ctx, fakeClient, tc.strategy, tc.nodes, podEvictor, getPodsAssignedToNode)
+			RemovePodsViolatingNodeAffinity(ctx, fakeClient, tc.strategy, tc.nodes, podEvictor, evictorFilter, getPodsAssignedToNode)
 			actualEvictedPodCount := podEvictor.TotalEvicted()
 			if actualEvictedPodCount != tc.expectedEvictedPodCount {
 				t.Errorf("Test %#v failed, expected %v pod evictions, but got %v pod evictions\n", tc.description, tc.expectedEvictedPodCount, actualEvictedPodCount)
```
JaneLiuL, line 246: replace `t.Errorf("Test %#v failed` to `t.Errorf("Test %v failed`

ingvagabund: There are other places where `Test %#v failed` string is used. After the migration to plugins all the unit tests will need to be updated. We can improve the error strings in there.

JaneLiuL: yet, lgtm now

JaneLiuL (commented):
/approve
/lgtm

pkg/descheduler/descheduler.go
damemi: could this check still be in the strategy code?

pkg/descheduler/evictions/evictions.go
```diff
@@ -318,7 +308,7 @@ func (pe *PodEvictor) Evictable(opts ...func(opts *Options)) *evictable {
 }
 
 // IsEvictable decides when a pod is evictable
-func (ev *evictable) IsEvictable(pod *v1.Pod) bool {
+func (ef *EvictorFilter) Filter(pod *v1.Pod) bool {
```
damemi, line 311: so eventually, as a plugin users will customize this by implementing their own `EvictorFilter.Filter()` function?

pkg/descheduler/descheduler.go
damemi: ah, is the answer no since `nodeFit` is being baked into the `evictorFilter` now? This will eventually be the type of thing that's in the strategy code though, right?

pkg/descheduler/evictions/evictions.go
```diff
@@ -230,21 +215,26 @@ func WithLabelSelector(labelSelector labels.Selector) func(opts *Options) {
...
-// The method accepts a list of options which allow to extend constraints
-// which decides when a pod is considered evictable.
-func (pe *PodEvictor) Evictable(opts ...func(opts *Options)) *evictable {
+func NewEvictorFilter(
+	nodes []*v1.Node,
+	nodeIndexer podutil.GetPodsAssignedToNodeFunc,
+	evictLocalStoragePods bool,
+	evictSystemCriticalPods bool,
+	ignorePvcPods bool,
+	evictFailedBarePods bool,
```
damemi, line 228: as we move forward, I'd like to think about how we can generalize passing/configuring all of these options for better customization.

damemi (commented):
@ingvagabund thanks for working on this, looks like a good start to me. A few questions but other than that nothing big blocking this  imo

pkg/descheduler/descheduler.go
ingvagabund: Yeah. The `nodeFit` option/flag is going to be transform into the `DefaultEvictor` plugin argument. Anyone can then implement their own evictor `Filter` extension point. Making the descheduler unaware of the `nodeFit` option.

pkg/descheduler/evictions/evictions.go
ingvagabund: The goal is to transform the `NewEvictorFilter` into `DefaultEvictor` plugin (the name can be different at the end). Making `evictLocalStoragePods`, `evictSystemCriticalPods`, `ignorePvcPods`, `evictFailedBarePods` (and possible other) as arguments to the plugin.

pkg/descheduler/evictions/evictions.go
ingvagabund: Yes. `Filter` (alongside `Sort`) is going to be a new extension point.

damemi (commented):
/approve
this is an exciting first step in implementing the framework. Good work @ingvagabund!

PR #848, opened 2022-06-13
Check existing constraints before assigning

knelasevero:
We need to check v1.TopologySpreadConstraint deepEquality because two v1.TopologySpreadConstraints are not the same (even with same content) since they have one field that is a pointer (LabelSelector). So when we do the `namespaceTopologySpreadConstraints[constraint] = struct{}{}` we are creating a new key every time. Now that we check their deep equality, v1.TopologySpreadConstraints with same content will be inserted as the same key in the assignment.

### DIFF
--- a/pkg/descheduler/strategies/topologyspreadconstraint.go
+++ b/pkg/descheduler/strategies/topologyspreadconstraint.go
@@ -19,6 +19,7 @@ package strategies
 import (
 	"context"
 	"math"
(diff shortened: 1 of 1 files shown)


ingvagabund:
@knelasevero would you mind providing a simple unit test?

knelasevero:
@ingvagabund Sure! will do :)

PR #852, opened 2022-06-14
Check existing constraints before assigning

knelasevero:
We need to check v1.TopologySpreadConstraint deepEquality because two v1.TopologySpreadConstraints are not the same (even with same content) since they have one field that is a pointer (LabelSelector). So when we do the namespaceTopologySpreadConstraints[constraint] = struct{}{} we are creating a new key every time. Now that we check their deep equality, v1.TopologySpreadConstraints with same content will be inserted as the same key in the assignment.

(recreated it, again)

### DIFF
--- a/pkg/descheduler/strategies/topologyspreadconstraint.go
+++ b/pkg/descheduler/strategies/topologyspreadconstraint.go
@@ -19,6 +19,7 @@ package strategies
 import (
 	"context"
 	"math"
+	"reflect"
 	"sort"
 
 	v1 "k8s.io/api/core/v1"
@@ -111,6 +112,12 @@ func RemovePodsViolatingTopologySpreadConstraint(
 				if constraint.WhenUnsatisfiable == v1.ScheduleAnyway && (strategy.Params == nil || !strategy.Params.IncludeSoftConstraints) {
 					continue
 				}
+				// Need to check v1.TopologySpreadConstraint deepEquality because
+				// v1.TopologySpreadConstraint.LabelSelector is a pointer
+				// and assigning new constrains would always add more keys
+				if hasIdenticalConstraints(constraint, namespaceTopologySpreadConstraints) {
+					continue
+				}
 				namespaceTopologySpreadConstraints[constraint] = struct{}{}
 			}
 		}
@@ -181,6 +188,16 @@ func RemovePodsViolatingTopologySpreadConstraint(
 			break
 		}
 	}
+}
+
+// hasIdenticalConstraints checks if already had an identical TopologySpreadConstraint in namespaceTopologySpreadConstraints map
+func hasIdenticalConstraints(newConstraint v1.TopologySpreadConstraint, namespaceTopologySpreadConstraints map[v1.TopologySpreadConstraint]struct{}) bool {
+	for constraint := range namespaceTopologySpreadConstraints {
+		if reflect.DeepEqual(newConstraint, constraint) {
+			return true
+		}
+	}
+	return false
 }
 
 // topologyIsBalanced checks if any domains in the topology differ by more than the MaxSkew
--- a/pkg/descheduler/strategies/topologyspreadconstraint.go
+++ b/pkg/descheduler/strategies/topologyspreadconstraint.go
@@ -105,7 +105,7 @@ func RemovePodsViolatingTopologySpreadConstraint(
 		}
 
 		// ...where there is a topology constraint
-		namespaceTopologySpreadConstraints := make(map[v1.TopologySpreadConstraint]struct{})
+		namespaceTopologySpreadConstraints := []v1.TopologySpreadConstraint{}
 		for _, pod := range namespacePods.Items {
 			for _, constraint := range pod.Spec.TopologySpreadConstraints {
 				// Ignore soft topology constraints if they are not included
@@ -113,12 +113,12 @@ func RemovePodsViolatingTopologySpreadConstraint(
 					continue
 				}
 				// Need to check v1.TopologySpreadConstraint deepEquality because
-				// v1.TopologySpreadConstraint.LabelSelector is a pointer
-				// and assigning new constrains would always add more keys
+				// v1.TopologySpreadConstraint has pointer fields
+				// and we don't need to go over duplicated constraints later on
 				if hasIdenticalConstraints(constraint, namespaceTopologySpreadConstraints) {
 					continue
 				}
-				namespaceTopologySpreadConstraints[constraint] = struct{}{}
+				namespaceTopologySpreadConstraints = append(namespaceTopologySpreadConstraints, constraint)
 			}
 		}
 		if len(namespaceTopologySpreadConstraints) == 0 {
@@ -126,7 +126,7 @@ func RemovePodsViolatingTopologySpreadConstraint(
 		}
 
 		// 2. for each topologySpreadConstraint in that namespace
-		for constraint := range namespaceTopologySpreadConstraints {
+		for _, constraint := range namespaceTopologySpreadConstraints {
 			constraintTopologies := make(map[topologyPair][]*v1.Pod)
 			// pre-populate the topologyPair map with all the topologies available from the nodeMap
 			// (we can't just build it from existing pods' nodes because a topology may have 0 pods)
@@ -190,9 +190,9 @@ func RemovePodsViolatingTopologySpreadConstraint(
 	}
 }
 
-// hasIdenticalConstraints checks if already had an identical TopologySpreadConstraint in namespaceTopologySpreadConstraints map
-func hasIdenticalConstraints(newConstraint v1.TopologySpreadConstraint, namespaceTopologySpreadConstraints map[v1.TopologySpreadConstraint]struct{}) bool {
-	for constraint := range namespaceTopologySpreadConstraints {
+// hasIdenticalConstraints checks if we already had an identical TopologySpreadConstraint in namespaceTopologySpreadConstraints slice
+func hasIdenticalConstraints(newConstraint v1.TopologySpreadConstraint, namespaceTopologySpreadConstraints []v1.TopologySpreadConstraint) bool {
+	for _, constraint := range namespaceTopologySpreadConstraints {
 		if reflect.DeepEqual(newConstraint, constraint) {
 			return true
 		}
--- a/pkg/descheduler/strategies/topologyspreadconstraint_test.go
+++ b/pkg/descheduler/strategies/topologyspreadconstraint_test.go
@@ -1041,3 +1041,50 @@ func getDefaultTopologyConstraints(maxSkew int32) []v1.TopologySpreadConstraint 
 		},
 	}
 }
+
+func TestCheckIdenticalConstraints(t *testing.T) {
+	newConstrainSame := v1.TopologySpreadConstraint{
+		MaxSkew:           2,
+		TopologyKey:       "zone",
+		WhenUnsatisfiable: v1.DoNotSchedule,
+		LabelSelector:     &metav1.LabelSelector{MatchLabels: map[string]string{"foo": "bar"}},
+	}
+	newConstrainDifferent := v1.TopologySpreadConstraint{
+		MaxSkew:           3,
+		TopologyKey:       "node",
+		WhenUnsatisfiable: v1.DoNotSchedule,
+		LabelSelector:     &metav1.LabelSelector{MatchLabels: map[string]string{"foo": "bar"}},
+	}
+	namespaceTopologySpreadConstraint := make(map[v1.TopologySpreadConstraint]struct{})
+	namespaceTopologySpreadConstraint[v1.TopologySpreadConstraint{
+		MaxSkew:           2,
+		TopologyKey:       "zone",
+		WhenUnsatisfiable: v1.DoNotSchedule,
+		LabelSelector:     &metav1.LabelSelector{MatchLabels: map[string]string{"foo": "bar"}},
+	}] = struct{}{}
+	testCases := []struct {
+		name                               string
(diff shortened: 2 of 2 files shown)


pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -183,6 +190,16 @@ func RemovePodsViolatingTopologySpreadConstraint(
 	}
 }
 
+// checkIdenticalConstraints checks if already had an identical TopologySpreadConstraint in namespaceTopologySpreadConstraints map
+func checkIdenticalConstraints(newConstraint v1.TopologySpreadConstraint, namespaceTopologySpreadConstraints map[v1.TopologySpreadConstraint]struct{}) bool {
```
ingvagabund, line 194: `s/checkIdenticalConstraints/hasConstraint` so it is clear from the function name what is getting checked, also given `namespaceTopologySpreadConstraints` is expected to be seen as a set prefixing the function name with `has` makes it more "set-ish".

pkg/descheduler/strategies/topologyspreadconstraint_test.go
```diff
@@ -1041,3 +1041,50 @@ func getDefaultTopologyConstraints(maxSkew int32) []v1.TopologySpreadConstraint
...
+		MaxSkew:           3,
+		TopologyKey:       "node",
+		WhenUnsatisfiable: v1.DoNotSchedule,
+		LabelSelector:     &metav1.LabelSelector{MatchLabels: map[string]string{"foo": "bar"}},
+	}
+	// Unique constrains, but with same content (except for different LabelSelector addresses)
+	uniqueConstrains := make(map[v1.TopologySpreadConstraint]struct{})
+	for i := 1; i < 5; i++ {
+		uniqueConstrains[v1.TopologySpreadConstraint{
+			MaxSkew:           2,
```
ingvagabund, line 1062: Should the index `i` be used here? Or, as part of the `LabelSelector`? Otherwise, all constraints are identical.

pkg/descheduler/strategies/topologyspreadconstraint_test.go
```diff
@@ -1041,3 +1041,50 @@ func getDefaultTopologyConstraints(maxSkew int32) []v1.TopologySpreadConstraint
...
+			LabelSelector:     &metav1.LabelSelector{MatchLabels: map[string]string{"foo": "bar"}},
+		}] = struct{}{}
+	}
+	testCases := []struct {
+		name                               string
+		namespaceTopologySpreadConstraints map[v1.TopologySpreadConstraint]struct{}
+		newConstraint                      v1.TopologySpreadConstraint
+		expectedResult                     bool
+	}{
+		{name: "multiple duplicated constraints, new one is identical",
```
ingvagabund, line 1074: Could you format the items as:
```
{
	name: "multiple duplicated constraints, new one is identical",
	namespaceTopologySpreadConstraints: uniqueConstrains,
	...
},
```
? So we maintain the same structure in all the tests.

ingvagabund (changes requested):

pkg/descheduler/strategies/topologyspreadconstraint_test.go
ingvagabund: Running `gofmt -s` should do the trick

ingvagabund:
From https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_descheduler/852/pull-descheduler-verify-master/1536619373073010688:
```
./hack/verify-govet.sh
./hack/verify-spelling.sh
./hack/verify-gofmt.sh
!!! 'gofmt -s' needs to be run on the following files: 
./pkg/descheduler/strategies/topologyspreadconstraint.go
make: *** [Makefile:107: verify-gofmt] Error 1
```

pkg/descheduler/strategies/topologyspreadconstraint_test.go
knelasevero: For some reason I wanted a bunch of identical Constrains (any number), but tbh we just need one, and try to add another one. I did that

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -111,6 +112,12 @@ func RemovePodsViolatingTopologySpreadConstraint(
 				if constraint.WhenUnsatisfiable == v1.ScheduleAnyway && (strategy.Params == nil || !strategy.Params.IncludeSoftConstraints) {
 					continue
 				}
+				// Need to check v1.TopologySpreadConstraint deepEquality because
+				// v1.TopologySpreadConstraint.LabelSelector is a pointer
+				// and assigning new constrains would always add more keys
+				if hasIdenticalConstraints(constraint, namespaceTopologySpreadConstraints) {
```
a7i, line 118: At this point, why even use a map? we can just convert to a slice

knelasevero: that's true...

What if we made the label selector be always the same, but only changing its contents, like with:
```
func setBaseLabelSelector(constraint v1.TopologySpreadConstraint, baseLabelSelector *metav1.LabelSelector ) v1.TopologySpreadConstraint {
	baseLabelSelector.MatchExpressions = constraint.LabelSelector.MatchExpressions
	baseLabelSelector.MatchLabels = constraint.LabelSelector.MatchLabels
	constraint.LabelSelector = baseLabelSelector
	return constraint
}
```

a7i: I like your approach of `DeepEqual` because soon we will have to add support for [MinDomains](https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/3022-min-domains-in-pod-topology-spread) and it would just work. And later down the road additional fields will be added and with the current implementation, we won't have to make any additional changes.

ingvagabund: @knelasevero label selectors of two constraints do not have to be always equal. You can have two constraints in the same namespace just with different label selectors. E.g. color=red and color=blue.

knelasevero: @a7i ok, makes sense!

@ingvagabund , but in that code snippet I'm grabbing the inner values that are different. I'm just keeping that "same label selector" object so it has the same address, but the values are changing to the ones that come from the constraint.

knelasevero: Changed it to slice, let me know what you think

ingvagabund: Would not all the constraints have the same label selector equal to the label selector of the last constraint given all the constraint share the same address of the baseLabelSelector?

pkg/descheduler/strategies/topologyspreadconstraint_test.go
```diff
@@ -1041,3 +1041,50 @@ func getDefaultTopologyConstraints(maxSkew int32) []v1.TopologySpreadConstraint
 		},
 	}
 }
+
+func TestCheckIdenticalConstraints(t *testing.T) {
+	newConstrainSame := v1.TopologySpreadConstraint{
```
ingvagabund, line 1046: newConstrainSame -> newConstraintSame

pkg/descheduler/strategies/topologyspreadconstraint_test.go
```diff
@@ -1041,3 +1041,50 @@ func getDefaultTopologyConstraints(maxSkew int32) []v1.TopologySpreadConstraint
...
+		MaxSkew:           2,
+		TopologyKey:       "zone",
+		WhenUnsatisfiable: v1.DoNotSchedule,
+		LabelSelector:     &metav1.LabelSelector{MatchLabels: map[string]string{"foo": "bar"}},
+	}
+	newConstrainDifferent := v1.TopologySpreadConstraint{
```
ingvagabund, line 1052: newConstrainDifferent -> newConstraintDifferent

pkg/descheduler/strategies/topologyspreadconstraint_test.go
```diff
@@ -1041,3 +1041,50 @@ func getDefaultTopologyConstraints(maxSkew int32) []v1.TopologySpreadConstraint
...
+		{
+			name:                               "new constraint is different",
+			namespaceTopologySpreadConstraints: namespaceTopologySpreadConstraint,
+			newConstraint:                      newConstrainDifferent, expectedResult: false,
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			isIdentitcal := hasIdenticalConstraints(tc.newConstraint, tc.namespaceTopologySpreadConstraints)
```
ingvagabund, line 1084: isIdentitcal -> isIdentical

pkg/descheduler/strategies/topologyspreadconstraint_test.go
```diff
@@ -1041,3 +1041,50 @@ func getDefaultTopologyConstraints(maxSkew int32) []v1.TopologySpreadConstraint
...
+	testCases := []struct {
+		name                               string
+		namespaceTopologySpreadConstraints []v1.TopologySpreadConstraint
+		newConstraint                      v1.TopologySpreadConstraint
+		expectedResult                     bool
+	}{
+		{
+			name:                               "new constraint is identical",
+			namespaceTopologySpreadConstraints: namespaceTopologySpreadConstraint,
+			newConstraint:                      newConstrainSame, expectedResult: true},
```
ingvagabund, line 1074: Can you move `expectedResult: true` on its own line?

pkg/descheduler/strategies/topologyspreadconstraint_test.go
```diff
@@ -1041,3 +1041,50 @@ func getDefaultTopologyConstraints(maxSkew int32) []v1.TopologySpreadConstraint
...
+		{
+			name:                               "new constraint is different",
+			namespaceTopologySpreadConstraints: namespaceTopologySpreadConstraint,
+			newConstraint:                      newConstrainDifferent, expectedResult: false,
```
ingvagabund, line 1078: Can you move `expectedResult: false` on its own line?

ingvagabund (changes requested):

pkg/descheduler/strategies/topologyspreadconstraint.go
```diff
@@ -183,6 +190,16 @@ func RemovePodsViolatingTopologySpreadConstraint(
 	}
 }
 
+// hasIdenticalConstraints checks if we already had an identical TopologySpreadConstraint in namespaceTopologySpreadConstraints slice
+func hasIdenticalConstraints(newConstraint v1.TopologySpreadConstraint, namespaceTopologySpreadConstraints []v1.TopologySpreadConstraint) bool {
+	for constraint := range namespaceTopologySpreadConstraints {
```
ingvagabund, line 195: `for _, constraint := range namespaceTopologySpreadConstraints`

constraint is an index in here otherwise

knelasevero: Good catch... I would never find this with the test being green haha

pkg/descheduler/strategies/topologyspreadconstraint_test.go
```diff
@@ -1041,3 +1041,50 @@ func getDefaultTopologyConstraints(maxSkew int32) []v1.TopologySpreadConstraint
...
+		LabelSelector:     &metav1.LabelSelector{MatchLabels: map[string]string{"foo": "bar"}},
+	}
+	newConstrainDifferent := v1.TopologySpreadConstraint{
+		MaxSkew:           3,
+		TopologyKey:       "node",
+		WhenUnsatisfiable: v1.DoNotSchedule,
+		LabelSelector:     &metav1.LabelSelector{MatchLabels: map[string]string{"foo": "bar"}},
+	}
+	namespaceTopologySpreadConstraint := []v1.TopologySpreadConstraint{}
+	namespaceTopologySpreadConstraint = append(namespaceTopologySpreadConstraint, v1.TopologySpreadConstraint{
```
ingvagabund, line 1059:
```
namespaceTopologySpreadConstraint := []v1.TopologySpreadConstraint{
		v1.TopologySpreadConstraint{
			MaxSkew:           2,
			TopologyKey:       "zone",
			WhenUnsatisfiable: v1.DoNotSchedule,
			LabelSelector:     &metav1.LabelSelector{MatchLabels: map[string]string{"foo": "bar"}},
		},
	}
```
works as well

pkg/descheduler/strategies/topologyspreadconstraint_test.go
```diff
@@ -1041,3 +1041,50 @@ func getDefaultTopologyConstraints(maxSkew int32) []v1.TopologySpreadConstraint
...
+			newConstraint:                      newConstrainSame, expectedResult: true},
+		{
+			name:                               "new constraint is different",
+			namespaceTopologySpreadConstraints: namespaceTopologySpreadConstraint,
+			newConstraint:                      newConstrainDifferent, expectedResult: false,
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
```
ingvagabund, line 1083: This is actually causing the `tc.expectedResult` to be false for both unit cases. Can you remove the line?

ingvagabund:
@knelasevero worth backporting the fix into 1.24 and 1.23 releases

@damemi FYI

knelasevero:
PRs up here:

[#868](https://github.com/kubernetes-sigs/descheduler/pull/868) & [#869](https://github.com/kubernetes-sigs/descheduler/pull/869)

Issue #829, opened 2022-06-01
Helm Tests + Kustomize


jklaw90:
If using kustomize to render the chart and add extra transformations you can't disable the tests with the normal --skip-tests, see [issue](https://github.com/kubernetes-sigs/kustomize/issues/3831).

We could add a helm value to get around this case ([kind of odd](https://github.com/kubernetes-sigs/descheduler/pull/804)), but in general i don't see much value form the test provided. 
I just tested it locally and it "passed" while the image wasn't able to pull so i do wondering if we should remove the helm test or improve it to be more useful?

a7i:
@jklaw90 Curious why not use [this directory](https://github.com/kubernetes-sigs/descheduler/tree/master/kubernetes) for kustomize?

Agreed on improving the test. 👍🏼

PR #854, opened 2022-06-14
Improving helm setup

knelasevero:
Main thing here is to have a way to see if changes to charts are at least valid in the PR. 

I see ./test/run-helm-tests.sh, but I am not sure how this is being called. In the docs it says that this is being used by the CI, but  I don't see it. Can we remove it?

### DIFF
--- a/.github/ci/ct.yaml
+++ b/.github/ci/ct.yaml
@@ -1,0 +1,5 @@
+chart-dirs:
+  - ./charts
+helm-extra-args: "--timeout=5m"
+check-version-increment: false
+target-branch: master
--- a/.github/workflows/helm.yaml
+++ b/.github/workflows/helm.yaml
@@ -1,0 +1,44 @@
+name: Helm
+
+on:
+  push:
+    branches:
+      - master
+      - release-*
+    paths:
+      - 'charts/**'
+  pull_request:
+    paths:
+      - 'charts/**'
+
+jobs:
+  lint-and-test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Set up Helm
+        uses: azure/setup-helm@v2.1
+        with:
+          version: v3.4.2
+
+      - uses: actions/setup-python@v3.1.2
+        with:
+          python-version: 3.7
+
+      - name: Set up chart-testing
+        uses: helm/chart-testing-action@v2.2.1
+
+      - name: Run chart-testing (list-changed)
+        id: list-changed
+        run: |
+          changed=$(ct list-changed --config=.github/ci/ct.yaml)
+          if [[ -n "$changed" ]]; then
+            echo "::set-output name=changed::true"
+          fi
+
+      - name: Run chart-testing (lint)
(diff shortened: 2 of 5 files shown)


damemi (commented):
We configure CI through a prow setup in https://github.com/kubernetes/test-infra. The helm test is run [here](https://github.com/kubernetes/test-infra/blob/5ed7f26/config/jobs/kubernetes-sigs/descheduler/descheduler-presubmits-master.yaml#L167). 

So, we shouldn't need a new github workflow for this. But there is some work that needs to be done to improve our helm tests as found in https://github.com/kubernetes-sigs/descheduler/issues/829

ingvagabund:
@knelasevero https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-sigs/descheduler/descheduler-presubmits-master.yaml#L146-L170

ingvagabund:
So based on https://github.com/kubernetes-sigs/descheduler/blob/master/charts/descheduler/templates/tests/test-descheduler-pod.yaml#L30 the helm test only checks the descheduler job completed. @damemi I wonder what's the scope of the helm test? To "just" verify the descheduler got "properly" deployed? Does not seem we are testing any functionality here.

knelasevero:
@ingvagabund it doesn't test that also... It seems to pass even when some stuff fails.

@damemi I saw that chart releaser was being used here on GitHub actions, and there are quite a bunch of ready to use GitHub actions for helm, so we don't have to reinvent the wheel. I wonder if we can't leave everything helm related on actions, and everything else on prow [?]

knelasevero:
@damemi  Makes sense! Thanks for pointing me in the direction of the prow test :)

knelasevero:
Ok, I was having a look over other sig sponsored projects that have anything with helm charts, and it seems that people only check lint and sometimes the ct install. Some use gh-actions, some use prow (most just check linting):

- ingress nginx on prow, but only linting: https://github.com/kubernetes/test-infra/blob/5ed7f2609f61a8ddfb5ee2ecb51f098e6a895edc/config/jobs/kubernetes/sig-network/ingress-nginx/ingress-nginx-presubmit.yaml#L93

- external-dns on gh-actions (same approach that I have added here, and same that we have added for [external-secrets](https://github.com/external-secrets/external-secrets/blob/main/.github/workflows/helm.yml)): https://github.com/kubernetes-sigs/external-dns/blob/master/.github/workflows/lint-test-chart.yaml

I think it would make sense to remove the one from prow and leave it here. Would require a PR to test-infra, and then I would remote the script from here. So let me know if I should pursue that.

damemi:
@knelasevero sounds good to me, just wanted to make sure we weren't doing something totally unprecedented. Thanks for offering to handle the work for that

knelasevero:
Started by removing the test from prow: https://github.com/kubernetes/test-infra/pull/26630

After that I will continue removing stuff from here, and then I will go forward with improving the helm checks on gh-actions.

knelasevero:
Hey @damemi, I will remove wip from this PR then. So we can merge the helm workflow and check its results. (as per your comment on https://github.com/kubernetes/test-infra/pull/26630#issuecomment-1161912089).

Then after that I can followup with removals in k/test-infra and also removals of the scripts here.

PR #859, opened 2022-06-22
provide configuration to include custom node and pod labels on metrics

a7i:
/cc @jklaw90 @damemi 

In order to improve our observability/monitoring, we would like to include certain pod labels in our metrics.

use-cases:
1. We have compliance tags to identify cost-center/teams/projects (e.g. `team`, `project`) and we need these associated with eviction metrics
2. We would like to include labels such as `toplogy.kubernetes.io/zone` as it relates to our reliability/resilience efforts

I realize that in the new descheduler framework efforts, `v1.Node` is no longer being passed in `EvictPod`. If this is a deal breaker, we can just continue with `podLabels` but flatten out the `metricsConfig` structure.


damemi (commented):
Thanks for the work @a7i. This is an interesting idea. I do have some concerns, though.

As I understand it, prometheus metrics are strict in their definitions such that 2 metrics with different label sets are incompatible. I think this means that, with something like this, we could technically never mark these metrics as stable.

I think if there is a case for adding some known k8s pod/node labels to our metrics, we should just add them, or introduce them as an entirely new metric. If there is another example or precedent for dynamic re-labeling, that would reassure me. But this approach is pretty new and unique to me so I am just a little unsure about it.

For custom arbitrary pod labels, I think there is probably a better way to extract those into your metrics server without piping them through the descheduler. It might be something you're better off instrumenting in the applications themselves. Then, with the info already exported from descheduler's metrics, you can get enough to correlate the data. Have you tried something like that?

ingvagabund:
Hey @a7i. Thank You for providing detailed description and an example for better understanding.

> As I understand it, prometheus metrics are strict in their definitions such that 2 metrics with different label sets are incompatible. I think this means that, with something like this, we could technically never mark these metrics as stable.

From https://kubernetes.io/blog/2021/04/23/kubernetes-release-1.21-metrics-stability-ga/#what-are-stable-metrics-and-why-do-we-need-them:
```
A stable metric is one which, from a consumption point of view, can be reliably consumed across a number of Kubernetes versions without risk of ingestion failure.
```

I don't think there's a way to say "this set of labels is guaranteed, the rest is not". So rather than extending the current one I agree with @damemi on creating a new set of metrics instead. Which we might indefinitely keep in the alpha level. Something we might call a customizable metric for custom consumption. Though I'd like to have a conformation from the metrics community that this approach is considered acceptable so we do not end up maintaining something that's not recommended by the community.

> I realize that in the new descheduler framework efforts, v1.Node is no longer being passed in EvictPod.

Given the v1.Node object would be only required by the customizable metrics, we might just move the metrics out of the `EvictPod`. I.e. have the `EvictPod` return a `Status` struct with an eviction result (e.g. "maximum number of pods per node reached") and an eviction state (evicted, eviction rejected). Then have higher logic to read the `Status` and record the metrics.

Once all the strategies are migrated, we can move the metrics out. Then extend them as needed after we have convincing discussion with the monitoring community. @a7i what's your though on this approach?

PR #857, opened 2022-06-15
Migrate RemovePodsViolatingNodeTaints to a plugin

ingvagabund:
Built on top of https://github.com/kubernetes-sigs/descheduler/pull/846
/hold
until https://github.com/kubernetes-sigs/descheduler/pull/846 gets merged

PR #860, opened 2022-06-23
Migrate RemovePodsViolatingNodeAffinity to plugin

knelasevero:
Built on top of https://github.com/kubernetes-sigs/descheduler/pull/857
/hold
until https://github.com/kubernetes-sigs/descheduler/pull/846 and https://github.com/kubernetes-sigs/descheduler/pull/857 (both) get merged

I will be rebasing review changes from those into here.


knelasevero:
TBH probably best if @ingvagabund reworks that method to reduce cyclomatic complexity on his PR so we all get those changes to work on top of it. Then we don´t have to cherry pick stuff? WDYT?

a7i:
> TBH probably best if @ingvagabund reworks that method to reduce cyclomatic complexity on his PR so we all get those changes to work on top of it. Then we don´t have to cherry pick stuff? WDYT?

I agree. I'm also having the same issue and it's going to be some nasty merge conflicts if we all did it.

pkg/framework/plugins/removepodsviolatingnodeaffinity/node_affinity.go
```diff
@@ -0,0 +1,120 @@
...
+						Err: fmt.Errorf("error listing pods on a node: %v", err),
+					}
+				}
+
+				for _, pod := range pods {
+					if pod.Spec.Affinity != nil && pod.Spec.Affinity.NodeAffinity != nil && pod.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution != nil {
+						klog.V(1).InfoS("Evicting pod", "pod", klog.KObj(pod))
+						d.handle.Evictor().Evict(ctx, pod, evictions.EvictOptions{})
+						if d.handle.Evictor().NodeLimitExceeded(node) {
+							continue loop
```
a7i, line 110: Should it not just break and move on to the next node? Seems like this is completely skipping the iteration of nodes

pkg/framework/plugins/removepodsviolatingnodeaffinity/node_affinity.go
```diff
@@ -0,0 +1,120 @@
...
+					node.Name,
+					d.handle.GetPodsAssignedToNodeFunc(),
+					podutil.WrapFilterFuncs(d.podFilter, func(pod *v1.Pod) bool {
+						return d.handle.Evictor().Filter(pod) &&
+							!nodeutil.PodFitsCurrentNode(d.handle.GetPodsAssignedToNodeFunc(), pod, node) &&
+							nodeutil.PodFitsAnyNode(d.handle.GetPodsAssignedToNodeFunc(), pod, nodes)
+					}),
+				)
+				if err != nil {
+					return &framework.Status{
```
a7i, line 100: To preserve existing behavior, this should log an error message and continue processing next node.

pkg/apis/componentconfig/types_pluginargs.go
```diff
@@ -32,3 +32,15 @@ type RemovePodsViolatingNodeTaintsArgs struct {
...
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// RemovePodsViolatingNodeAffinityArgs holds arguments used to configure RemovePodsViolatingNodeAffinity plugin.
+type RemovePodsViolatingNodeAffinityArgs struct {
+	metav1.TypeMeta
+
+	Namespaces              *api.Namespaces
+	LabelSelector           *metav1.LabelSelector
+	IncludePreferNoSchedule bool
```
ingvagabund, line 44: The `RemovePodsViolatingNodeAffinity` strategy does not use `IncludePreferNoSchedule` param.

pkg/framework/plugins/removepodsviolatingnodeaffinity/node_affinity.go
ingvagabund: I wonder whether getting `err != nil` for the first time when running `podutil.ListPodsOnANode` will only mean there's gonna be `err != nil` the second, the third and the next time `podutil.ListPodsOnANode` gets invoked again. Could be some temporary issue (e.g. connection getting refused) or some hard issue (misconfiguration of the plugin/framework).

Both cases of continuing/not continuing make sense.

pkg/framework/plugins/removepodsviolatingnodeaffinity/node_affinity_test.go
```diff
@@ -233,18 +236,33 @@ func TestRemovePodsViolatingNodeAffinity(t *testing.T) {
...
-				tc.nodes,
-				getPodsAssignedToNode,
-				false,
-				false,
-				false,
-				false,
-				evictions.WithNodeFit(nodeFit),
+			plugin, err := New(
+				&componentconfig.RemovePodsViolatingNodeAffinityArgs{
+					NodeAffinityType: tc.strategy.Params.NodeAffinityType,
```
ingvagabund, line 257: Would you please remove the `tc.strategy.Params` bits entirely and replace it with `RemovePodsViolatingNodeAffinityArgs`? After all the strategies are migrated the `api.StrategyParameters` type will be deprecated/removed.

ingvagabund (changes requested):

pkg/framework/plugins/removepodsviolatingnodeaffinity/node_affinity.go
ingvagabund: The `loop` label refers to `for _, node := range nodes` loop. All good here.

pkg/framework/plugins/removepodsviolatingnodeaffinity/node_affinity_test.go
knelasevero: My bad, rebase + copy mistake

pkg/framework/plugins/removepodsviolatingnodeaffinity/node_affinity.go
knelasevero: Do we keep it like this then?

ingvagabund (changes requested):

pkg/apis/componentconfig/validation/validation_pluginargs.go
```diff
@@ -48,3 +48,17 @@ func validateCommonArgs(namespaces *api.Namespaces, labelSelector *metav1.LabelS
...
 }
+
+// ValidateRemovePodsViolatingNodeAffinityArgs validates RemovePodsViolatingNodeAffinity arguments
+func ValidateRemovePodsViolatingNodeAffinityArgs(args *componentconfig.RemovePodsViolatingNodeAffinityArgs) error {
+	// At most one of include/exclude can be set
+	if args.Namespaces != nil && len(args.Namespaces.Include) > 0 && len(args.Namespaces.Exclude) > 0 {
+		return fmt.Errorf("only one of Include/Exclude namespaces can be set")
+	}
+
+	if args == nil || len(args.NodeAffinityType) == 0 {
```
ingvagabund, line 59: Would you please move the check before `if args.Namespaces != nil`? Given `args == nil` is checked after `args.Namespaces != nil`, the condition is already satisfied.

ingvagabund (changes requested):

pkg/framework/plugins/removepodsviolatingnodeaffinity/node_affinity.go
ingvagabund: > Should it not just break and move on to the next node? Seems like this is completely skipping the iteration of nodes

@a7i I get it now :) The code got updated removing the continue in favor of the simple break.

Issue #864, opened 2022-07-01
LowNodeUtilization: cpu utilization don't precise.


Beherenowhjy:
What version of descheduler are you using?

descheduler version: v0.24.1

Does this issue reproduce with the latest release?
yes

Which descheduler CLI options are you using?

   
          
![image](https://user-images.githubusercontent.com/59044653/176817314-4a063968-533a-4e44-8963-6d3ad9c8101e.png)

Please provide a copy of your descheduler policy config file

![image](https://user-images.githubusercontent.com/59044653/176817535-c006a580-4343-4202-8e81-c38bd64b4367.png)

What k8s version are you using (kubectl version)?

kubectl version Output
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2", GitCommit:"c97fe5036ef3df2967d086711e6c0c405941e14b", GitTreeState:"clean", BuildDate:"2019-10-15T19:18:23Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}
What did you do?

I get the limit cpu info by kubectl. @damemi say: " the descheduler determines usage based only on pod requests/limits",but the result i got is wrong.the shell file content is :
![image](https://user-images.githubusercontent.com/59044653/176835379-63c48aab-4caf-4f65-b8fe-084ff634e735.png)
the result of cpu limit sum in my server 49-15-01 is 164. but the server 49-15-01 only have 80 cores actually.
![image](https://user-images.githubusercontent.com/59044653/176836114-583fef67-bda8-4e63-9495-fd394d7402c9.png)
actually the cpu limit sum is over 100%,but i get usagePercentage of cpu is 90%. that is a wrong result.
`I0701 06:20:21.308940       1 nodeutilization.go:224] "Node is overutilized" node="n49-15-01" usage=map[cpu:72662m memory:178608144385 pods:37] usagePercentage=map[cpu:90.8275 memory:88.81546469920734 pods:33.63636363636363]`
why i get a wrong result? do u need something more information?
#730 

What did you expect to see?

What did you see instead?

damemi:
Sorry, the nodeutilization calculation is based only on the *requests* of the pods ([see here](https://github.com/kubernetes-sigs/descheduler/blob/d8897635b0767b8e63176dc578d215b882102d14/pkg/descheduler/node/node.go#L266)). (We should double check if this is consistent with the scheduler)

Can you run your script based on requests to see if the values are more in line with what you expect?

[...]

PR #872, opened 2022-07-06
fix log-file and log-dir issue

JaneLiuL:
fix https://github.com/kubernetes-sigs/descheduler/issues/864
as check from https://github.com/kubernetes/klog/blob/main/README.md and https://github.com/kubernetes/klog/blob/main/klog.go#424, we need to update our code and readme to fix the log dir and log file issue


damemi (commented):
/approve
Thanks @JaneLiuL. A couple of notes:
1. Because these flags are already marked as deprecated, we should be fine with this update without having to do anything to keep supporting the old flags.
2. We should revisit the deprecation notice in these flags. They are referring to the KEP which removes logging flags from all k8s components (https://github.com/kubernetes/enhancements/tree/master/keps/sig-instrumentation/2845-deprecate-klog-specific-flags-in-k8s-components). The reasoning in that KEP makes sense, and we should follow suit at some point for the same motivations outlined there (and consistency with other components)
3. As I understand it, we are able to support this for now thanks to https://github.com/kubernetes-sigs/descheduler/commit/e6314d2c7e277c180926b526465c7df7e218e93b which moved us to importing klog flags directly, rather than implicitly getting them from the upstream k8s libraries

@JaneLiuL if you want, these are some good follow up items to this PR

damemi:
@JaneLiuL no, my point was just to call out that even though the k8s deprecation doesn't technically affect us anymore, we should still plan on removing these flags ourselves eventually. For now, this fix is fine imo
/lgtm

PR #875, opened 2022-07-07
Revert "Existing contraints fix backport 1.24"

ingvagabund:
Reverts kubernetes-sigs/descheduler#868

Taking back as we don't have sufficient proof to show the backport significantly improves the memory consumption


a7i:
If not memory consumption, do we know what problem this was trying to solve? I cannot seem to replicate a situation where unnecessary/incorrect eviction occurs.

PR #876, opened 2022-07-07
Use lease resource name from Helm value

iijimakazuyuki:
This PR makes the lease name of Helm value and the ClusterRole's lease resource name consistent.

### DIFF
--- a/charts/descheduler/templates/clusterrole.yaml
+++ b/charts/descheduler/templates/clusterrole.yaml
@@ -30,7 +30,7 @@ rules:
   verbs: ["create", "update"]
 - apiGroups: ["coordination.k8s.io"]
   resources: ["leases"]
-  resourceNames: ["descheduler"]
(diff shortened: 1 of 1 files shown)


damemi (commented):
/ok-to-test

JaneLiuL (commented):
/lgtm

damemi:
thought of this too late but @iijimakazuyuki: right now the [default value for this is commented out](https://github.com/kubernetes-sigs/descheduler/blob/v0.24.1/charts/descheduler/values.yaml#L55). Should we uncomment this? I'm thinking this will break now if users don't set a value

damemi:
> thought of this too late but @iijimakazuyuki: right now the [default value for this is commented out](https://github.com/kubernetes-sigs/descheduler/blob/v0.24.1/charts/descheduler/values.yaml#L55). Should we uncomment this? I'm thinking this will break now if users don't set a value

Opened https://github.com/kubernetes-sigs/descheduler/pull/881 to do this

PR #881, opened 2022-07-08
Set default value for lease resource name

damemi:
Following https://github.com/kubernetes-sigs/descheduler/pull/876

PR #882, opened 2022-07-08
Add default lease resource name in Helm chart's ClusterRole

iijimakazuyuki:
Ref. #876 #881
This PR adds the default value for the lease resource name in Helm chart's ClusterRole.

### DIFF
--- a/charts/descheduler/templates/clusterrole.yaml
+++ b/charts/descheduler/templates/clusterrole.yaml
@@ -30,7 +30,7 @@ rules:
   verbs: ["create", "update"]
 - apiGroups: ["coordination.k8s.io"]
   resources: ["leases"]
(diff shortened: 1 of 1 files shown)


damemi (commented):
lgtm, @iijimakazuyuki could you just confirm if you've tested this to make sure it works when no value is set in `values.yml`?

iijimakazuyuki:
@damemi Yes, I tested using `values.yaml` as following:

```
kind: Deployment
leaderElection:
  enabled: true
```

And I confirmed the lease resource name is set in ClusterRole as following:

```
- apiGroups:
  - coordination.k8s.io
  resourceNames:
  - descheduler
  resources:
  - leases
  verbs:
  - get
  - patch
  - delete
```

damemi:
@iijimakazuyuki awesome, thank you!
/approve

PR #887, opened 2022-07-21
Stop redeclaring and re-starting structured logging

### LINKED ISSUE #884: Memory consumption of the descheduler
Hey,

we had a memory leak reported internally. and this made us start an investigation into the descheduler memory consumption with various strategies with some scales of nodes and pods. I want to use this issue to both walk everyone through our findings and also report the memory leak itself.

The leak did not seem to appear in very small setups, or at least it was slow enough to be hard to reproduce, so I went to a bigger setup to get results faster:

### First scenario - LifecycleAndUtilization:

Number of nodes in control plane: 3
Number of worker nodes: 15
Cluster built using openshift-installer for aws with default instance types.

- Started the test with 500 pods
- Installed the kube-descheduler-operator from operatorhub
- Applied a KubeDescheduler resource with only LifecycleAndUtilization profile
- Started monitoring memory consumption in grafana
- RAM oscilates quite a bit, but does not accumulate
- goes from 16mb, 40mb, back to 20... and so on
- I bumped number of pods to 700
- Some pods get evicted by descheduler at some cycles, but memory seems to oscilate at the same range
- I bump to 1000, wait a bit, then to 1250 and to 1500 and memory seems to be always in the same range.

Here is the graph:

image

### Second scenario - AffinityAndTaints:

- Started with 500 pods
- Memory consumption around 20mb
- Bumped to 1000 pods
- Same consumption basically
- Threw a taint of NoSchedule in one of the nodes
- Memory consumption went up to 50mb
- Let it roll like that for a bit  to see
- Bumped pods to 1250, waited a bunch, then to 1500
- No change in memory with number of pods alone
- Introduced another taint in another node
- Around 93~95mb consumtion
- Third taint in
- Went to 114mb consumption

Here is the graph:

image

This is still not the memory leak that I am looking for, but it is interesting expected behavior for users to be aware.

### Third scenario - TopologyAndDuplicates:

- I already had some previous knowledge of suspicious behavior on this one and wait to 750 pods already
- Memory on 80mb consumption initially
- Increasing by itself, going to 100mb at 11:10 (graph below)
- Increased number of pods to 1000 to see if it would go up faster (11:20)
- Got to 500mb at 11:26 and with that, OOM killed (different timezone in grafana and terminal)

image

image

- At 11:36, back to 200 pods in the cluster to see if the curve goes slower
- It does
- I just wait for another OOM kill

image

### Forth scenario - TopologyAndDuplicates - Less nodes:

Here I changed my setup:

Number of nodes in control plane: 3

Number of worker nodes: 3

- Wanted to see if number of nodes would impact on something.
- So, initially with 200 pods, 12:54
- Memory oscilating around 25mb, 23mb, 27 mb...
- Left it like that for some minutes, nothing much happening
- Upped it to 500 pods
- Graph shows it going down (13:02), but not restarts are seen, probably something with the monitoring instrumentation pods
- Memory around 38 - 36, but no accumulation
- Upped it to 750 pods
- Again a drop in the graph that is not from a kill or restart, something with the monitoring pods
- Memory went up to ~104mb
- And started to clearly grow
- 123mb -> 140mb -> 175...
- Started to have some general pressure on multiple components... Also since it was a smaller cluster, api started to timeout and whatnot.

Here is the graph:

image

We confirm that this descheduler operator profile has the memory leak, and also, we confirm that it only actually is apparent with more than 750 pods in general. I don't think number of nodes impact that much here.

### Fifth scenario - RemoveDuplicates strategy:

- Back to 15 workers setup so I don't have to worry about it not being able to handle it
- Stopped using the operator and started deploying descheduler directly so I could enable single strategies
- Enabled RemoveDuplicates
- Already started with 750 pods
- Nothing much happening in memory (blue in graph)
- 14:57 bump to 1000 pods
- Same
- Realized descheduling-interval was bigger than I wanted, it was 5minutes or something like that
- Reduced to 7s, still 1000 pods (changes of color means new pod, because I applied with new arguments)
- 15:02 72 mb
- 15:04 144mb
- 15:07 Noticed things were maybe too fast [?]. Descheduler were evicting pods, they would get in ContainerCreating Status, and Descheduler would be evicting more pods before the previous ones would be Running still. So I assumed things were piggybacking on each other, and memory was going up fast.
- Increased to descheduling-interval 20s
- Still too fast
- 15:11 increased to 40s. Still too fast
- Then to 2m, which seemed fine. Gave pods a change to be running before a cycle comes around
- Every run it evics around 49~80 pods (since most come from the same deployment, bunch of duplicates in nodes)
- ~57:21 84mb, 15:25 105 mb
- Memory growing even with a good space between my evictions and these same pods being running again. I think we have a Leak here
- 15:33 135mb
- 15:34 1250 pods

Graph here:

image

### Sixth scenario - TopologySpreadConstrain strategy:

- Already with 1000 pods
- Starts growing wich makes me think we will see another leak
- It seems to stabilize and just oscilate
- 16:46 1250 pods
- 16:50 1500 pods
- 16:59 1750 pods
- Left it running for longer 

image

.

image

- Recreated with small refresh since this one is not actually evicting anything
- Not much change
- More pods makes it use more memory, so there is a limitation here, and around 10kpods I think we hit OOM kills from time to time, even though we don't have a leak here

Another graph: 

image

Wanted to create this issue already, before I would forget and would reach a point where I could not make sense of my notes. I will investigate the RemoveDuplicates leak specifically further next and circle back here.

knelasevero:
Fixes #884

Did a bit more investigation and this is the graph of the same setup without this change (green) and with this change (yellow):

image

So removing eventBroadcaster + StartStructuredLogging from the loop fixes the leak. (specially StartStructuredLogging() call)

I am not sure I should be passing the Broadcaster as a new field of PodEvictor as I am doing right now, please let me know.

It should affect other strategies, but it affected only RemoveDuplicates in my case because of the scenario that I had (bunch of deployments with a bunch of replicas), so it had evictions every time at every cycle. (I think that any user with a lot of eveitions at every cycle, in any strategy, should be saying memory going up because of this issue...)

### DIFF
--- a/pkg/descheduler/descheduler.go
+++ b/pkg/descheduler/descheduler.go
@@ -30,6 +30,7 @@ import (
 	clientset "k8s.io/client-go/kubernetes"
 	fakeclientset "k8s.io/client-go/kubernetes/fake"
 	core "k8s.io/client-go/testing"
+	"k8s.io/client-go/tools/record"
 	"k8s.io/klog/v2"
 
 	corev1informers "k8s.io/client-go/informers/core/v1"
@@ -275,6 +276,8 @@ func RunDeschedulerStrategies(ctx context.Context, rs *options.DeschedulerServer
 		} else {
 			podEvictorClient = rs.Client
 		}
+		eventBroadcaster := record.NewBroadcaster()
+		eventBroadcaster.StartStructuredLogging(3)
 
 		klog.V(3).Infof("Building a pod evictor")
 		podEvictor := evictions.NewPodEvictor(
@@ -285,6 +288,7 @@ func RunDeschedulerStrategies(ctx context.Context, rs *options.DeschedulerServer
 			deschedulerPolicy.MaxNoOfPodsToEvictPerNamespace,
 			nodes,
 			!rs.DisableMetrics,
+			eventBroadcaster,
 		)
 
 		for name, strategy := range deschedulerPolicy.Strategies {
--- a/pkg/descheduler/evictions/evictions.go
+++ b/pkg/descheduler/evictions/evictions.go
@@ -28,7 +28,6 @@ import (
 	"k8s.io/apimachinery/pkg/util/errors"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
-	clientcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/descheduler/metrics"
@@ -57,6 +56,7 @@ type PodEvictor struct {
 	nodepodCount               nodePodEvictedCount
 	namespacePodCount          namespacePodEvictCount
 	metricsEnabled             bool
+	eventBroadcaster           record.EventBroadcaster
 }
 
 func NewPodEvictor(
@@ -67,6 +67,7 @@ func NewPodEvictor(
 	maxPodsToEvictPerNamespace *uint,
 	nodes []*v1.Node,
 	metricsEnabled bool,
+	eventBroadcaster record.EventBroadcaster,
 ) *PodEvictor {
 	var nodePodCount = make(nodePodEvictedCount)
 	var namespacePodCount = make(namespacePodEvictCount)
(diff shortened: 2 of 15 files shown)


knelasevero:
nevermind, just noticed StartRecordingToSink was missing. Back to plotting

PR #888, opened 2022-07-21
fix: events memory leak. Using new events implementation and take recorder out of EvictPod

### LINKED ISSUE #884: Memory consumption of the descheduler

knelasevero:
Fixes #884 

I was trying something around #887 but that did not do it. There I just mistakenly removed `StartRecordingToSink` which when reintroduced would cause the leak again. So you can just ignore that. (just mentioning to document attempts)

Shutting down eventBroadcaster seems to do the trick. If we look back to the [pprof top report for in use objects](https://github.com/kubernetes-sigs/descheduler/issues/884#issuecomment-1180547263) we see that we have something getting accumulated in k8s.io/apimachinery/pkg/watch.(*Broadcaster).Watch. I am guessing every time we call StartRecordingToSink that is when we start accumulating it.

Shutting down the eventBroadcaster right after we publish the event seems to stop this accumulation and it also seems to not affect the publication of the event itself (I kept monitoring some evicted pods in terminating state to be sure):

```
Events:
  Type    Reason          Age   From                     Message
  ----    ------          ----  ----                     -------
  Normal  Scheduled       2m    default-scheduler        Successfully assigned default/nginx-deployment-85d945584d-zvcps to ip-10-0-177-50.ec2.internal
  Normal  AddedInterface  116s  multus                   Add eth0 [10.130.2.231/23] from openshift-sdn
  Normal  Pulled          115s  kubelet                  Container image "k8s.gcr.io/pause:3.1" already present on machine
  Normal  Created         111s  kubelet                  Created container nginx
  Normal  Started         111s  kubelet                  Started container nginx
  Normal  Descheduled     29s   sigs.k8s.io.descheduler  pod evicted by sigs.k8s.io/descheduler
```

Here is new graph of memory consumption:

image

### DIFF
--- a/kubernetes/base/rbac.yaml
+++ b/kubernetes/base/rbac.yaml
@@ -4,7 +4,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: descheduler-cluster-role
 rules:
-- apiGroups: [""]
+- apiGroups: ["events.k8s.io"]
   resources: ["events"]
   verbs: ["create", "update"]
 - apiGroups: [""]
--- a/pkg/descheduler/descheduler.go
+++ b/pkg/descheduler/descheduler.go
@@ -53,11 +53,12 @@ import (
 func Run(ctx context.Context, rs *options.DeschedulerServer) error {
 	metrics.Register()
 
-	rsclient, err := client.CreateClient(rs.KubeconfigFile)
+	rsclient, eventClient, err := createClients(rs.KubeconfigFile)
 	if err != nil {
 		return err
 	}
 	rs.Client = rsclient
+	rs.EventClient = eventClient
 
 	deschedulerPolicy, err := LoadPolicyConfig(rs.PolicyConfigFile)
 	if err != nil {
@@ -287,6 +288,16 @@ func RunDeschedulerStrategies(ctx context.Context, rs *options.DeschedulerServer
 		ignorePvcPods = *deschedulerPolicy.IgnorePVCPods
 	}
 
+	var eventClient clientset.Interface
+	if rs.DryRun {
+		eventClient = fakeclientset.NewSimpleClientset()
+	} else {
+		eventClient = rs.Client
+	}
+
+	eventBroadcaster, eventRecorder := utils.GetRecorderAndBroadcaster(ctx, eventClient)
+	defer eventBroadcaster.Shutdown()
+
 	wait.NonSlidingUntil(func() {
 		nodes, err := nodeutil.ReadyNodes(ctx, rs.Client, nodeInformer, nodeSelector)
 		if err != nil {
@@ -340,6 +351,7 @@ func RunDeschedulerStrategies(ctx context.Context, rs *options.DeschedulerServer
 			deschedulerPolicy.MaxNoOfPodsToEvictPerNamespace,
 			nodes,
 			!rs.DisableMetrics,
+			eventRecorder,
 		)
 
 		for name, strategy := range deschedulerPolicy.Strategies {
@@ -412,4 +424,18 @@ func RunDeschedulerStrategies(ctx context.Context, rs *options.DeschedulerServer
 	}, rs.DeschedulingInterval, ctx.Done())
 
 	return nil
+}
+
+func createClients(kubeconfig string) (clientset.Interface, clientset.Interface, error) {
+	kClient, err := client.CreateClient(kubeconfig, "descheduler")
+	if err != nil {
+		return nil, nil, err
+	}
+
+	eventClient, err := client.CreateClient(kubeconfig, "")
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return kClient, eventClient, nil
 }
--- a/pkg/descheduler/evictions/evictions.go
+++ b/pkg/descheduler/evictions/evictions.go
@@ -27,9 +27,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/errors"
 	clientset "k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/kubernetes/scheme"
-	clientcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
-	"k8s.io/client-go/tools/record"
+	"k8s.io/client-go/tools/events"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/descheduler/metrics"
 	nodeutil "sigs.k8s.io/descheduler/pkg/descheduler/node"
@@ -57,6 +55,7 @@ type PodEvictor struct {
 	nodepodCount               nodePodEvictedCount
 	namespacePodCount          namespacePodEvictCount
 	metricsEnabled             bool
+	eventRecorder              events.EventRecorder
 }
 
 func NewPodEvictor(
@@ -67,6 +66,7 @@ func NewPodEvictor(
 	maxPodsToEvictPerNamespace *uint,
 	nodes []*v1.Node,
 	metricsEnabled bool,
+	eventRecorder events.EventRecorder,
 ) *PodEvictor {
 	var nodePodCount = make(nodePodEvictedCount)
 	var namespacePodCount = make(namespacePodEvictCount)
@@ -85,6 +85,7 @@ func NewPodEvictor(
 		nodepodCount:               nodePodCount,
 		namespacePodCount:          namespacePodCount,
 		metricsEnabled:             metricsEnabled,
+		eventRecorder:              eventRecorder,
 	}
 }
 
@@ -166,11 +167,14 @@ func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod, opts EvictOptio
 		klog.V(1).InfoS("Evicted pod in dry run mode", "pod", klog.KObj(pod), "reason", opts.Reason, "strategy", strategy, "node", pod.Spec.NodeName)
 	} else {
 		klog.V(1).InfoS("Evicted pod", "pod", klog.KObj(pod), "reason", opts.Reason, "strategy", strategy, "node", pod.Spec.NodeName)
-		eventBroadcaster := record.NewBroadcaster()
-		eventBroadcaster.StartStructuredLogging(3)
-		eventBroadcaster.StartRecordingToSink(&clientcorev1.EventSinkImpl{Interface: pe.client.CoreV1().Events(pod.Namespace)})
-		r := eventBroadcaster.NewRecorder(scheme.Scheme, v1.EventSource{Component: "sigs.k8s.io.descheduler"})
-		r.Event(pod, v1.EventTypeNormal, "Descheduled", fmt.Sprintf("pod evicted by sigs.k8s.io/descheduler%s", opts.Reason))
+		reason := opts.Reason
+		if len(reason) == 0 {
+			reason = strategy
+			if len(reason) == 0 {
+				reason = "NotSet"
+			}
+		}
+		pe.eventRecorder.Eventf(pod, nil, v1.EventTypeNormal, reason, "Descheduled", "pod evicted by sigs.k8s.io/descheduler")
 	}
 	return true
 }
--- a/pkg/descheduler/strategies/duplicates_test.go
+++ b/pkg/descheduler/strategies/duplicates_test.go
@@ -27,6 +27,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes/fake"
+	"k8s.io/client-go/tools/events"
 
 	"sigs.k8s.io/descheduler/pkg/api"
 	"sigs.k8s.io/descheduler/pkg/descheduler/evictions"
@@ -313,6 +314,8 @@ func TestFindDuplicatePods(t *testing.T) {
 			sharedInformerFactory.Start(ctx.Done())
 			sharedInformerFactory.WaitForCacheSync(ctx.Done())
 
+			eventRecorder := &events.FakeRecorder{}
+
 			podEvictor := evictions.NewPodEvictor(
 				fakeClient,
 				"v1",
@@ -321,6 +324,7 @@ func TestFindDuplicatePods(t *testing.T) {
 				nil,
 				testCase.nodes,
 				false,
+				eventRecorder,
 			)
 
 			nodeFit := false
@@ -751,6 +755,8 @@ func TestRemoveDuplicatesUniformly(t *testing.T) {
 			sharedInformerFactory.Start(ctx.Done())
 			sharedInformerFactory.WaitForCacheSync(ctx.Done())
 
+			eventRecorder := &events.FakeRecorder{}
+
 			podEvictor := evictions.NewPodEvictor(
 				fakeClient,
 				policyv1.SchemeGroupVersion.String(),
@@ -759,6 +765,7 @@ func TestRemoveDuplicatesUniformly(t *testing.T) {
 				nil,
 				testCase.nodes,
 				false,
+				eventRecorder,
 			)
 
 			evictorFilter := evictions.NewEvictorFilter(
--- a/pkg/descheduler/strategies/node_affinity_test.go
+++ b/pkg/descheduler/strategies/node_affinity_test.go
@@ -26,6 +26,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes/fake"
+	"k8s.io/client-go/tools/events"
 
 	"sigs.k8s.io/descheduler/pkg/api"
 	"sigs.k8s.io/descheduler/pkg/descheduler/evictions"
@@ -215,6 +216,8 @@ func TestRemovePodsViolatingNodeAffinity(t *testing.T) {
 			sharedInformerFactory.Start(ctx.Done())
 			sharedInformerFactory.WaitForCacheSync(ctx.Done())
 
+			eventRecorder := &events.FakeRecorder{}
+
 			podEvictor := evictions.NewPodEvictor(
 				fakeClient,
 				policyv1.SchemeGroupVersion.String(),
@@ -223,6 +226,7 @@ func TestRemovePodsViolatingNodeAffinity(t *testing.T) {
 				tc.maxNoOfPodsToEvictPerNamespace,
 				tc.nodes,
 				false,
+				eventRecorder,
 			)
 
 			nodeFit := false
--- a/test/e2e/e2e_duplicatepods_test.go
+++ b/test/e2e/e2e_duplicatepods_test.go
@@ -26,6 +26,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/client-go/tools/events"
 	"k8s.io/utils/pointer"
 	deschedulerapi "sigs.k8s.io/descheduler/pkg/api"
 	"sigs.k8s.io/descheduler/pkg/descheduler/evictions"
@@ -138,6 +139,9 @@ func TestRemoveDuplicates(t *testing.T) {
 			if err != nil || len(evictionPolicyGroupVersion) == 0 {
 				t.Fatalf("Error creating eviction policy group %v", err)
 			}
+
+			eventRecorder := &events.FakeRecorder{}
+
 			podEvictor := evictions.NewPodEvictor(
 				clientSet,
 				evictionPolicyGroupVersion,
@@ -146,6 +150,7 @@ func TestRemoveDuplicates(t *testing.T) {
 				nil,
 				nodes,
 				false,
+				eventRecorder,
 			)
 
 			t.Log("Running DeschedulerStrategy strategy")
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -36,6 +36,7 @@ import (
 	"k8s.io/client-go/informers"
 	coreinformers "k8s.io/client-go/informers/core/v1"
 	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/events"
 	v1qos "k8s.io/kubectl/pkg/util/qos"
 	"k8s.io/utils/pointer"
 
@@ -107,7 +108,7 @@ func RcByNameContainer(name, namespace string, replicas int32, labels map[string
 }
 
 func initializeClient(t *testing.T) (clientset.Interface, coreinformers.NodeInformer, podutil.GetPodsAssignedToNodeFunc, chan struct{}) {
-	clientSet, err := client.CreateClient(os.Getenv("KUBECONFIG"))
+	clientSet, err := client.CreateClient(os.Getenv("KUBECONFIG"), "")
 	if err != nil {
 		t.Errorf("Error during client creation with %v", err)
 	}
@@ -193,6 +194,8 @@ func runPodLifetimeStrategy(
 		t.Fatalf("Failed to get threshold priority from strategy's params")
 	}
 
+	eventRecorder := &events.FakeRecorder{}
+
 	strategies.PodLifeTime(
 		ctx,
 		clientset,
@@ -206,6 +209,7 @@ func runPodLifetimeStrategy(
 			maxPodsToEvictPerNamespace,
 			nodes,
 			false,
+			eventRecorder,
 		),
 		evictions.NewEvictorFilter(
 			nodes,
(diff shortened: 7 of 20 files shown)


damemi (commented):
/ok-to-test

pkg/descheduler/evictions/evictions.go
```diff
@@ -171,6 +171,7 @@ func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod, opts EvictOptio
 		eventBroadcaster.StartRecordingToSink(&clientcorev1.EventSinkImpl{Interface: pe.client.CoreV1().Events(pod.Namespace)})
 		r := eventBroadcaster.NewRecorder(scheme.Scheme, v1.EventSource{Component: "sigs.k8s.io.descheduler"})
 		r.Event(pod, v1.EventTypeNormal, "Descheduled", fmt.Sprintf("pod evicted by sigs.k8s.io/descheduler%s", opts.Reason))
+		eventBroadcaster.Shutdown()
```
ingvagabund, line 174: I wonder whether we actually broadcast the events properly. Checking the KS code [1]:
```
cc.EventBroadcaster.StartRecordingToSink(ctx.Done())
defer cc.EventBroadcaster.Shutdown()
```
and [2]:
```
// Prepare kube clients.
client, eventClient, err := createClients(kubeConfig)
if err != nil {
	return nil, err
}

c.EventBroadcaster = events.NewEventBroadcasterAdapter(eventClient)
```

`StartRecordingToSink` and `Shutdown` is invoked only once.

[1] https://github.com/kubernetes/kubernetes/blob/master/cmd/kube-scheduler/app/server.go#L160-L161
[2] https://github.com/kubernetes/kubernetes/blob/master/cmd/kube-scheduler/app/options/options.go#L280-L286

damemi: I would think the eventbroadcaster should be started at startup (followed by a defer), and then just passed everywhere it's used. I agree it seems weird we're starting and stopping it right here

ingvagabund: From https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/tools/events/interfaces.go#L42:
```
Eventf(regarding runtime.Object, related runtime.Object, eventtype, reason, action, note string, args ...interface{})
```

The descheduler would then invoke `eventBroadcaster.Eventf(...)`

ingvagabund: Worth refactoring the code and moving the bits up in https://github.com/kubernetes-sigs/descheduler/blob/d5e66ab62e3683dfce663417ba46f957b5b29b59/pkg/descheduler/descheduler.go#L169. `EvictPod` should be really invoking only a single command to record an event.

knelasevero: Looking at it now I see that we probably are dealing with events in a non optimal way here. I will have a proper look at what kube-scheduler is doing there and reproduce it here

ingvagabund: @knelasevero much appreciate the time and effort in investigating the issue and figuring out where the leak is

kubernetes/base/rbac.yaml
```diff
@@ -4,7 +4,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: descheduler-cluster-role
 rules:
-- apiGroups: [""]
+- apiGroups: ["events.k8s.io"]
```
knelasevero, line 7: now that I am using client-go/tools/events, had to specify this apiGroup so our service account can create events

pkg/descheduler/evictions/evictions.go
```diff
@@ -166,11 +170,11 @@ func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod, opts EvictOptio
...
 	} else {
 		klog.V(1).InfoS("Evicted pod", "pod", klog.KObj(pod), "reason", opts.Reason, "strategy", strategy, "node", pod.Spec.NodeName)
-		eventBroadcaster := record.NewBroadcaster()
-		eventBroadcaster.StartStructuredLogging(3)
-		eventBroadcaster.StartRecordingToSink(&clientcorev1.EventSinkImpl{Interface: pe.client.CoreV1().Events(pod.Namespace)})
-		r := eventBroadcaster.NewRecorder(scheme.Scheme, v1.EventSource{Component: "sigs.k8s.io.descheduler"})
-		r.Event(pod, v1.EventTypeNormal, "Descheduled", fmt.Sprintf("pod evicted by sigs.k8s.io/descheduler%s", opts.Reason))
+		reason := opts.Reason
+		if len(reason) == 0 {
+			reason = "NotSet"
```
knelasevero, line 175: We are not setting Reasons yet, but if we try to emit events with Event and a empty Reason, it fails. I am leaving this here but we can remove when we start setting them in all strategies

pkg/descheduler/descheduler.go
```diff
@@ -275,6 +277,10 @@ func RunDeschedulerStrategies(ctx context.Context, rs *options.DeschedulerServer
 		} else {
 			podEvictorClient = rs.Client
 		}
+		eventBroadcaster := events.NewEventBroadcasterAdapter(rs.EventClient)
+		eventBroadcaster.StartRecordingToSink(ctx.Done())
+		eventRecorder := eventBroadcaster.NewRecorder("sigs.k8s.io.descheduler")
```
knelasevero, line 282: I am creating the Recorder already here and passing it as another field in the PodEvictor struct, but if we look at how KS does it, it had a Recorder Factory with some other abstractions there: https://github.com/kubernetes/kubernetes/blob/8e14585d8700d9cf7861be66943f023d11d41075/cmd/kube-scheduler/app/server.go#L285

But I guess we should go that way only when we finish setting up the descheduler framework here

knelasevero: Hmm, looking at the failing e2e tests, maybe I need the factory

pkg/descheduler/evictions/evictions.go
a7i: we used to [set](https://github.com/kubernetes-sigs/descheduler/blob/v0.24.1/pkg/descheduler/evictions/evictions.go#L125) reason to strategy name if it wasn't set.

pkg/descheduler/strategies/node_affinity_test.go
```diff
@@ -215,6 +216,11 @@ func TestRemovePodsViolatingNodeAffinity(t *testing.T) {
 			sharedInformerFactory.Start(ctx.Done())
 			sharedInformerFactory.WaitForCacheSync(ctx.Done())
 
+			eventBroadcaster := events.NewEventBroadcasterAdapter(fakeClient)
+			eventBroadcaster.StartRecordingToSink(ctx.Done())
+			eventRecorder := eventBroadcaster.NewRecorder("sigs.k8s.io.descheduler")
+			defer eventBroadcaster.Shutdown()
```
a7i, line 222: can we move these to a common utility for the tests? Or perhaps `events.go` in [utils](https://github.com/kubernetes-sigs/descheduler/tree/master/pkg/utils) so descheduler and tests can call it.

pkg/descheduler/descheduler.go
```diff
@@ -51,11 +52,12 @@ import (
 func Run(ctx context.Context, rs *options.DeschedulerServer) error {
 	metrics.Register()
 
-	rsclient, err := client.CreateClient(rs.KubeconfigFile)
+	rsclient, eventClient, err := createClients(rs.KubeconfigFile)
```
a7i, line 55: for my understanding, why do we need separate clients?

ingvagabund: The only difference between both clients is the userAgent. Which helps to separate e.g. descheduler audit logs from the events audit logs. Checking other components:
- [KCM has a single client](https://github.com/kubernetes/kubernetes/blob/a9593d634c6a053848413e600dadbf974627515f/cmd/kube-controller-manager/app/controllermanager.go#L184
)
- [kubelet has a single client](https://github.com/kubernetes/kubernetes/blob/a9593d634c6a053848413e600dadbf974627515f/cmd/kubelet/app/server.go#L450)
- [kube-proxy has both](https://github.com/kubernetes/kubernetes/blob/a9593d634c6a053848413e600dadbf974627515f/cmd/kube-proxy/app/server.go#L589)

Although in all the cases the event broadcaster is actually using only a single client for recording events:
KCM:
```go
eventBroadcaster := record.NewBroadcaster()
eventRecorder := eventBroadcaster.NewRecorder(clientgokubescheme.Scheme, v1.EventSource{Component: 
KubeControllerManagerUserAgent})
c.EventBroadcaster.StartStructuredLogging(0)
c.EventBroadcaster.StartRecordingToSink(&v1core.EventSinkImpl{Interface: c.Client.CoreV1().Events("")})
defer c.EventBroadcaster.Shutdown()
```
kubelet:
```go
eventBroadcaster := record.NewBroadcaster()
kubeDeps.Recorder = eventBroadcaster.NewRecorder(legacyscheme.Scheme, v1.EventSource{Component: componentKubelet, Host: string(nodeName)})
eventBroadcaster.StartStructuredLogging(3)
eventBroadcaster.StartRecordingToSink(&v1core.EventSinkImpl{Interface: kubeDeps.EventClient.Events("")})
```
kube-proxy: it creates `eventClient` but still injects `client.EventsV1()` into the event broadcaster
```go
eventClient, err := clientset.NewForConfig(kubeConfig)
eventBroadcaster := events.NewBroadcaster(&events.EventSinkImpl{Interface: client.EventsV1()})
recorder := eventBroadcaster.NewRecorder(scheme.Scheme, "kube-proxy")
...
if s.Broadcaster != nil && s.EventClient != nil {
	stopCh := make(chan struct{})
	s.Broadcaster.StartRecordingToSink(stopCh)
}
...
```

Checking other places under `k8s.io/kubernetes/staging/src/k8s.io`:
```bash
$ grep -rn "\.StartRecordingToSink" --include "*.go" --exclude "*_test.go"
sample-controller/controller.go:101:	eventBroadcaster.StartRecordingToSink(&typedcorev1.EventSinkImpl{Interface: kubeclientset.CoreV1().Events("")})
client-go/tools/events/event_broadcaster.go:381:		e.eventsv1Broadcaster.StartRecordingToSink(stopCh)
client-go/tools/events/event_broadcaster.go:384:		e.coreBroadcaster.StartRecordingToSink(&typedv1core.EventSinkImpl{Interface: e.coreClient.Events("")})
legacy-cloud-providers/aws/aws.go:1423:	c.eventBroadcaster.StartRecordingToSink(&v1core.EventSinkImpl{Interface: c.kubeClient.CoreV1().Events("")})
legacy-cloud-providers/azure/azure.go:674:	az.eventBroadcaster.StartRecordingToSink(&v1core.EventSinkImpl{Interface: az.KubeClient.CoreV1().Events("")})
legacy-cloud-providers/gce/gce.go:655:	g.eventBroadcaster.StartRecordingToSink(&v1core.EventSinkImpl{Interface: g.client.CoreV1().Events("")})
cloud-provider/controllers/service/controller.go:114:	broadcaster.StartRecordingToSink(&v1core.EventSinkImpl{Interface: kubeClient.CoreV1().Events("")})
cloud-provider/controllers/node/node_controller.go:119:	eventBroadcaster.StartRecordingToSink(&v1core.EventSinkImpl{Interface: kubeClient.CoreV1().Events("")})
cloud-provider/controllers/route/route_controller.go:109:		rc.broadcaster.StartRecordingToSink(&v1core.EventSinkImpl{Interface: rc.kubeClient.CoreV1().Events("")})
cloud-provider/controllers/nodelifecycle/node_lifecycle_controller.go:78:	eventBroadcaster.StartRecordingToSink(&v1core.EventSinkImpl{Interface: kubeClient.CoreV1().Events("")})
cloud-provider/app/controllermanager.go:148:	c.EventBroadcaster.StartRecordingToSink(&v1core.EventSinkImpl{Interface: c.Client.CoreV1().Events("")})
```

Looks like kube-scheduler is the only one creating a separate event client.

ingvagabund: @knelasevero I will leave it up to you to make the decision whether we use a single client or separate ones. I don't have any objections here.

pkg/descheduler/evictions/evictions.go
knelasevero: any reason we stopped doing that?

knelasevero: Ok, did something similar now

pkg/descheduler/descheduler.go
knelasevero: Since this project is closely related to KS in heart and soul, I would slightly prefer to keep it similar to what is being done in there. But also no strong preference from my side.

pkg/descheduler/descheduler.go
knelasevero: @damemi @ingvagabund @a7i I tried a few things on my local to get e2e passing, but I am not sure why e2e tests are failing here :thinking:

I initially thought I was shutting broadcaster down prematurely in e2e tests. But it seems it wasnt it. Also tried having new recorders every-time, also did not help.

Any ideas here?

pkg/descheduler/strategies/node_affinity_test.go
knelasevero: Moved to events.go in utils. We still need to defer Shutdown in the upper level though, so I just created a function that returns both the broadcaster and the recorder.

test/e2e/e2e_test.go
```diff
@@ -1424,6 +1429,11 @@ func initPodEvictorOrFail(t *testing.T, clientSet clientset.Interface, getPodsAs
 	if err != nil || len(evictionPolicyGroupVersion) == 0 {
 		t.Fatalf("Error creating eviction policy group: %v", err)
 	}
+
+	ctx := context.Background()
+	eventBroadcaster, eventRecorder := utils.GetRecorderAndBroadcaster(ctx, clientSet)
+	defer eventBroadcaster.Shutdown()
```
ingvagabund, line 1435: `defer eventBroadcaster.Shutdown()` needs to be invoked outside of `initPodEvictorOrFail`

knelasevero: I tried removing all defer shutdowns from e2e tests just to quickly to see if it would make e2e pass, but it didn't :(

I will make that change anyways :)

knelasevero: Hmm, maybe I imagined that, one of the e2e passed now

pkg/descheduler/evictions/evictions.go
```diff
@@ -57,6 +55,8 @@ type PodEvictor struct {
 	nodepodCount               nodePodEvictedCount
 	namespacePodCount          namespacePodEvictCount
 	metricsEnabled             bool
+	eventBroadcaster           events.EventBroadcasterAdapter
```
ingvagabund, line 58: `eventBroadcaster` is not used anywhere in the code, the parameter can be dropped

pkg/descheduler/strategies/duplicates_test.go
```diff
@@ -313,6 +313,9 @@ func TestFindDuplicatePods(t *testing.T) {
 			sharedInformerFactory.Start(ctx.Done())
 			sharedInformerFactory.WaitForCacheSync(ctx.Done())
 
+			eventBroadcaster, eventRecorder := utils.GetRecorderAndBroadcaster(ctx, fakeClient)
```
ingvagabund, line 316: Given the unit tests do not check the events recorded during testing the `eventRecorder` can be replaced by [NewFakeRecorder](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/tools/events/fake.go#L41) which does not need any `eventBroadcaster.Shutdown()`.

test/e2e/e2e_duplicatepods_test.go
```diff
@@ -138,6 +139,10 @@ func TestRemoveDuplicates(t *testing.T) {
 			if err != nil || len(evictionPolicyGroupVersion) == 0 {
 				t.Fatalf("Error creating eviction policy group %v", err)
 			}
+
+			eventBroadcaster, eventRecorder := utils.GetRecorderAndBroadcaster(ctx, clientSet)
```
ingvagabund, line 143: Given the e2e test does not check the events recorded during testing the `eventRecorder` can be replaced by [NewFakeRecorder](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/tools/events/fake.go#L41) which does not need any `eventBroadcaster.Shutdown()`.

ingvagabund (changes requested):

test/e2e/e2e_duplicatepods_test.go
ingvagabund: Benefit of the fake recorder is no go routine inside `Eventf`.

pkg/descheduler/descheduler.go
```diff
@@ -330,6 +331,8 @@ func RunDeschedulerStrategies(ctx context.Context, rs *options.DeschedulerServer
 		} else {
 			podEvictorClient = rs.Client
 		}
+		eventBroadcaster, eventRecorder := utils.GetRecorderAndBroadcaster(ctx, podEvictorClient)
```
ingvagabund, line 334: No need to create utils.GetRecorderAndBroadcaster in every cycle. The event recorder/broadcaster can be created only once and used throughout the entire run of the descheduler. E.g. https://github.com/kubernetes-sigs/descheduler/pull/891/commits/ca96f8387ab76c39eb26dbab091d928bdf6feaaf#diff-67816189b07b98460b57a10b41280403c53a4dfda42fdd8f54475812cb99f56eR291-R300

knelasevero: Nice!

knelasevero:
All green :heavy_check_mark: :)

ingvagabund (changes requested):
There are still few places in unit tests that have:
```
eventBroadcaster, eventRecorder := utils.GetRecorderAndBroadcaster(ctx, fakeClient)
defer eventBroadcaster.Shutdown()
```
These are potential places for writes into closed channels.

pkg/descheduler/descheduler.go
knelasevero: :heavy_check_mark:

knelasevero:
> There are still few places in unit tests that have:
> 
> ```
> eventBroadcaster, eventRecorder := utils.GetRecorderAndBroadcaster(ctx, fakeClient)
> defer eventBroadcaster.Shutdown()
> ```
> 
> These are potential places for writes into closed channels.

I think I got all now. Just pushed :)

pkg/descheduler/strategies/duplicates_test.go
```diff
@@ -751,6 +755,8 @@ func TestRemoveDuplicatesUniformly(t *testing.T) {
 			sharedInformerFactory.Start(ctx.Done())
 			sharedInformerFactory.WaitForCacheSync(ctx.Done())
 
+			eventRecorder := events.NewFakeRecorder(utils.FakeRecorderSize)
```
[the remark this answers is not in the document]
knelasevero: With lower sizes it was hanging forever and reaching timeouts. I had to keep increasing this size to get tests to pass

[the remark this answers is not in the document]
knelasevero: Went with &events.FakeRecorder{} :heavy_check_mark:

ingvagabund:
/hold
to give some time to other reviewers to take a look
/lgtm
feel free to unhold after a day or two

knelasevero:
> @knelasevero @ingvagabund it looks like there was a lot of iteration on this, is the main PR description still up to date with the changes? (or could we get a summary of your findings)

Basically I went with what we discussed here:  [comment](https://github.com/kubernetes-sigs/descheduler/pull/888/#discussion_r926863909)

- Started using new events implementation. 
- Mimicked some of what is being done in KS, so using different clientSet for **events** and for the actual **descheduler** actions. 
- Used &events.FakeRecorder{} in tests instead of the real thing.
- Instantiated the broadcaster and recorder only one time outside NonSlidingUntil


I think that's it. Sorry, I should have waited a bit more to squash the commits :)

JaneLiuL (commented):
/lgtm

PR #893, opened 2022-07-28
HighNodeUtilization: support node rebalance when high utilized nodes empty

Jan-dai:
background: I'm using descscheduler highnodeutilization mode in production to improve machine utilization and reduce unnecessary waste. Highnodeutilization is to set a threshold, divide the nodes into two, and migrate the pods on the nodes with low utilization to the nodes with high utilization. However, there are some issue in this mode. When all nodes have low utilization, and the highnodeutilization mode will not work. However, the pods on the node with low utilization can be migrated to the node with low utilization to achieve the effect.

### DIFF
--- a/pkg/descheduler/strategies/nodeutilization/highnodeutilization.go
+++ b/pkg/descheduler/strategies/nodeutilization/highnodeutilization.go
@@ -29,7 +29,47 @@ import (
 	"sigs.k8s.io/descheduler/pkg/descheduler/evictions"
 	nodeutil "sigs.k8s.io/descheduler/pkg/descheduler/node"
 	podutil "sigs.k8s.io/descheduler/pkg/descheduler/pod"
+	"sigs.k8s.io/descheduler/pkg/utils"
 )
+
+// balanceNodes balance nodes from low utilized nodes to high utilized nodes. A corner case: All nodes are underutilized
+func balanceNodes(sourceNodes, destinationNodes []NodeInfo) ([]NodeInfo, []NodeInfo) {
+	if len(sourceNodes) == 0 || len(destinationNodes) > 0 {
+		return sourceNodes, destinationNodes
+	}
+	numOfBalanced := 0
+	partition := int(float64(len(sourceNodes)) * 0.5)
+	src := make([]NodeInfo, 0, partition)
+	dst := destinationNodes
+	hasImportantPods := func(node *NodeInfo) bool {
+		for _, pod := range node.allPods {
+			if utils.IsMirrorPod(pod) {
+				return true
+			}
+			if utils.IsPodWithPVC(pod) {
+				return true
+			}
+			if utils.IsCriticalPriorityPod(pod) {
+				return true
+			}
+		}
+		return false
+	}
+	for _, node := range sourceNodes {
+		if nodeutil.IsNodeUnschedulable(node.node) {
+			src = append(src, node)
+			continue
+		}
+
+		if numOfBalanced < partition && !hasImportantPods(&node) {
+			numOfBalanced++
+			dst = append(dst, node)
+		} else {
+			src = append(src, node)
+		}
+	}
+	return src, dst
+}
 
 // HighNodeUtilization evicts pods from under utilized nodes so that scheduler can schedule according to its strategy.
 // Note that CPU/Memory requests are used to calculate nodes' utilization and not the actual resource usage.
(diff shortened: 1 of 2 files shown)


ingvagabund:
Hi @Jan-dai, thank You for the strategy improvement. Would you please update the PR description with more details? E.g. what benefits this change has, what use cases does this improves, etc.

Given this changes the default functionality, we might consider enabling this support through a new flag in the strategy configuration.

pkg/descheduler/strategies/nodeutilization/highnodeutilization.go
```diff
@@ -29,8 +29,48 @@ import (
 	"sigs.k8s.io/descheduler/pkg/descheduler/evictions"
 	nodeutil "sigs.k8s.io/descheduler/pkg/descheduler/node"
 	podutil "sigs.k8s.io/descheduler/pkg/descheduler/pod"
+	"sigs.k8s.io/descheduler/pkg/utils"
 )
 
+// balanceNodes balance nodes from low utilized nodes to high utilized nodes. A corner case: All nodes are underutilized
+func balanceNodes(sourceNodes, destinationNodes []NodeInfo) ([]NodeInfo, []NodeInfo) {
```
ingvagabund, line 36: Based on the
```
if len(highNodes) == 0 {
	sourceNodes, highNodes = balanceNodes(sourceNodes, highNodes)
}
```
condition len(destinationNodes) will always be 0. Is `balanceNodes` function expected to be used on other places in the future?

pkg/descheduler/strategies/nodeutilization/highnodeutilization.go
```diff
@@ -29,8 +29,48 @@ import (
...
+	if len(sourceNodes) == 0 || len(destinationNodes) > 0 {
+		return sourceNodes, destinationNodes
+	}
+	numOfBalanced := 0
+	partition := int(float64(len(sourceNodes)) * 0.5)
```
ingvagabund, line 41: Worth making `0.5` param configurable.

pkg/descheduler/strategies/nodeutilization/highnodeutilization.go
```diff
@@ -29,8 +29,48 @@ import (
...
+	src := make([]NodeInfo, 0, partition)
+	dst := destinationNodes
+	hasImportantPods := func(node *NodeInfo) bool {
```
ingvagabund, line 44: `hasImportantPods` might be a generic function. @damemi @a7i @knelasevero worth discussing definition of what an important pod is.

pkg/descheduler/strategies/nodeutilization/highnodeutilization.go
```diff
@@ -29,8 +29,48 @@ import (
...
+				return true
+			}
+			if utils.IsCriticalPriorityPod(pod) {
+				return true
+			}
+		}
+		return false
+	}
+	for _, node := range sourceNodes {
+		if nodeutil.IsNodeUnschedulable(node.node) {
```
ingvagabund, line 59: +1 unschedulable nodes seems like good candidates for scaling down

pkg/descheduler/strategies/nodeutilization/highnodeutilization.go
```diff
@@ -29,8 +29,48 @@ import (
...
+			src = append(src, node)
+			continue
+		}
+
+		if numOfBalanced < partition && !hasImportantPods(&node) {
```
ingvagabund, line 64: With `!hasImportantPods(&node)` as a condition is the goal here to protect nodes that have important pods from attracting pods from the source nodes? Given the kube-scheduler does not know anything about hasImportantPods property, this might not actually work as expected. Also, when all the nodes have at least one important node, dst will be empty.

pkg/descheduler/strategies/nodeutilization/highnodeutilization.go
Jan-dai: No, it will only be used in corner case of HighNodeUtilization.

pkg/descheduler/strategies/nodeutilization/highnodeutilization.go
Jan-dai: a good suggestion

pkg/descheduler/strategies/nodeutilization/highnodeutilization.go
Jan-dai: I'm not sure how to split the low utilization nodes into two is the best way

pkg/descheduler/strategies/nodeutilization/highnodeutilization.go
knelasevero: What we have here seems like a good start I think. Mirror, with PVC, or Critical.

Issue #897, opened 2022-08-02
The descheduler trivy security scanning is reporting critical CVE-2022-1996


rsvalerio:
**What version of descheduler are you using?**
The version v0.24.1

**Does this issue reproduce with the latest release?**
Yes

**Which descheduler CLI options are you using?**
Runtime is not needed as we can scan the binary directly without running it

**Please provide a copy of your descheduler policy config file**
Runtime is not needed as we can scan the binary directly without running it

**What k8s version are you using (`kubectl version`)?**
Runtime is not needed as we can scan the binary directly without running it

**What did you do?**

- Installed [trivy binary](https://aquasecurity.github.io/trivy/v0.30.4/getting-started/installation/)

- Run the image scanning

```bash
trivy image --ignore-unfixed --severity CRITICAL k8s.gcr.io/descheduler/descheduler:v0.24.1
```

**What did you expect to see?**
```txt
=====================
Total: 0 (CRITICAL: 0)
```

**What did you see instead?**
```txt
==========================
Total: 1 (CRITICAL: 1)

+--------------------------------+------------------+----------+---------------------+---------------+--------------------------------------+
|            LIBRARY             | VULNERABILITY ID | SEVERITY |  INSTALLED VERSION  | FIXED VERSION |                TITLE                 |
+--------------------------------+------------------+----------+---------------------+---------------+--------------------------------------+
| github.com/emicklei/go-restful | CVE-2022-1996    | CRITICAL | v2.9.5+incompatible | v3.8.0        | go-restful: Authorization Bypass     |
|                                |                  |          |                     |               | Through User-Controlled Key          |
|                                |                  |          |                     |               | -->avd.aquasec.com/nvd/cve-2022-1996 |
+--------------------------------+------------------+----------+---------------------+---------------+--------------------------------------+
```

First of all, thanks for this great tool!

I'll be happy to help with this if I can.

Cheers.

damemi:
Thanks @rsvalerio, this seems like a transitive dependency, probably from one of the k8s libraries if I had to guess. Let's see if bumping those brings in the fixed version and if so we can backport it through the supported descheduler versions. If you'd like to help with that please feel free to self-assign

rsvalerio:
yes, you are right! 

I've bumped versions to the latest, but still using old transitive dependency for go-restful.
```txt
	github.com/client9/misspell v0.3.4
	github.com/spf13/cobra v1.5.0
	github.com/spf13/pflag v1.0.5
[...]

PR #898, opened 2022-08-05
add security scan into gh-actions

JaneLiuL:
part of https://github.com/kubernetes-sigs/descheduler/issues/897
summarize here:

This workflow scans:

Every push to the main and master branch
The default branch every day at 0:30 UTC

security result can be see:
For scheduler result is on GitHub Security tab.

### DIFF
--- a/.github/workflows/security.yaml
+++ b/.github/workflows/security.yaml
@@ -1,0 +1,47 @@
+name: "Security"
+
+on:
+  push:
+    branches:
+      - main
+      - master
+      - release-*
+  schedule:
+    - cron: '30 1 * * 0'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Build image
+        run: |
+          IMAGE_REPO=${HELM_IMAGE_REPO:-descheduler}
+          IMAGE_TAG=${HELM_IMAGE_TAG:-security-test}
+          VERSION=security-test make image
+      - name: Run Trivy vulnerability scanner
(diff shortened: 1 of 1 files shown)


JaneLiuL:
as we see the here: https://github.com/kubernetes-sigs/descheduler/pull/898/checks?check_run_id=7687777891
we can scan the security issue.


image

ingvagabund:
Will the `Code scanning results / Trivy ` test be blocking or informative only?

JaneLiuL:
will be integrate with pull request, each pull request will build image and use trivy scan, if fail, just like above, will block the pr.

damemi:
Is there any way to make it non-blocking, or otherwise not part of the PR workflow? I've seen other github repos where scans like this just appear as a notice. 

Even though security fixes like this are important, sometimes it is resolved in the next bump or there is other work going on around it. In those cases we shouldn't block all other work from merging and it would be very noisy to have this in every PR

JaneLiuL:
hi, so sorry for reply so late, i was vacation several days.
is it better to only alert on each pr and not block it?

.github/workflows/security.yaml
```diff
@@ -0,0 +1,34 @@
+name: security
+
+on:
+  schedule:
+    - cron: '30 1 * * 0'
```
damemi, line 5: we probably don't need to run it as often as once every half hour, that's pretty frequent. Maybe once a day at minimum?

JaneLiuL: yet, now it is  runs every sunday at 1:30 UTC. i will change to scan every day :).
the result if you integrate with every pr, it will on each pr checks button.
now we skip scan every pr, and the result is in GitHub Security tab.
i cope the descheduler code to my repository, and you can see result here: https://github.com/JaneLiuL/security-test-descheduler/security/code-scanning


but i still suggest that we need to **integrate with each pr and each merge to main or master and release branch**. we can not fail the pr or the pipeliens , because we can easily to check which pr import a new critical or high security issue and not fail the pr.

JaneLiuL: sure, update done now :)

PR #861, opened 2022-06-24
Migrate RemoveFailedPods to plugin

a7i:

PR #899, opened 2022-08-05
separate args validation for better reuse

a7i:
1. Separate validations for reuse
2. Aggregate all errors as opposed to stopping on first validation error

### DIFF
--- a/pkg/apis/componentconfig/validation/validation_pluginargs.go
+++ b/pkg/apis/componentconfig/validation/validation_pluginargs.go
@@ -18,6 +18,7 @@ package validation
 
 import (
 	"fmt"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/descheduler/pkg/api"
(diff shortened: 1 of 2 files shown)


damemi (commented):
Thanks for doing this work @a7i, but this change wasn't really my intent with the comment in https://github.com/kubernetes-sigs/descheduler/pull/861#discussion_r937811130. The point I'm trying to get at more is that our default plugins shouldn't be any different from theoretical third-party plugins. So, not dependent on any internal functions like this.

In my other comment, I meant more that Namespace and Label arg validation should be public, common functions that we export as part of the framework (to be used by any plugins). Does that make sense?

a7i:
> 
> In my other comment, I meant more that Namespace and Label arg validation should be public, common functions that we export as part of the framework (to be used by any plugins). Does that make sense?

So basically move these to a new file and make them public. Do you have a place in mind that will fit this?
same directory i.e. `pkg/apis/componentconfig/validation`  seems good?

damemi:
@a7i I think for now they can stay where they are, I just wanted to make a note of my intent for these eventually. Breaking them into individual functions works toward that goal so this PR looks fine to me. Sorry for the confusion
/lgtm

PR #895, opened 2022-07-29
WIP: add helm test step to gh-actions

knelasevero:
Also, remove local helm testing in favor of ct in PRs.

Example successful run can be checked here (from my fork to my fork): https://github.com/knelasevero/descheduler/pull/2/checks

Example fail: https://github.com/knelasevero/descheduler/runs/7564334471?check_suite_focus=true

We already get events, describe, and logs of the deployed chart pod in the CI directly. 

I am already removing the scripts and references to previous helm testing approach. 

So /hold this one until I normalize and merge https://github.com/kubernetes/test-infra/pull/26630

PR #904, opened 2022-08-08
add helm ct install. 

knelasevero:
First step, see https://github.com/kubernetes-sigs/descheduler/pull/895#issuecomment-1203608848

Ater this PR is merged we can check it prow considers gh-actions when deciding to merge. After that we can go forward with the other changes from https://github.com/kubernetes-sigs/descheduler/pull/895

### DIFF
--- a/.github/workflows/helm.yaml
+++ b/.github/workflows/helm.yaml
@@ -7,9 +7,11 @@ on:
       - release-*
     paths:
       - 'charts/**'
+      - '.github/workflows/helm.yaml'
   pull_request:
     paths:
       - 'charts/**'
+      - '.github/workflows/helm.yaml'
 
 jobs:
   lint-and-test:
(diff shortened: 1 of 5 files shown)


knelasevero:
It wont show meaningful info in this CI since https://github.com/kubernetes-sigs/descheduler/pull/895/files#r934597809 was removed and we don't have chart changes.

.github/workflows/helm.yaml
```diff
@@ -39,6 +43,22 @@ jobs:
...
           fi
-
+      
       - name: Run chart-testing (lint)
         run: ct lint --config=.github/ci/ct.yaml --validate-maintainers=false
+      
+      # Need a multi node cluster so descheduler runs until evictions
+      - name: Create multi node Kind cluster
+        run: |
+          kind create cluster --name kind --config ./hack/kind_config.yaml --wait 2m
```
a7i, line 53: @knelasevero what step install the `kind` cli? and how can I control the verison?

knelasevero: Not sure if kind comes pre-baked in this gh-actions image or if it is installed by one of the previous actions :thinking:

PR #907, opened 2022-08-09
introduce template error on purpose to test PR

knelasevero:
I am not adding WIP/draft since we want tests to run and check how prow behaves with this thing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -143,13 +143,13 @@ ifndef HAS_HELM
 endif
 
 lint-chart: ensure-helm-install
-	helm lint ./charts/descheduler
+	ls
 
 build-helm:
 	helm package ./charts/descheduler --dependency-update --destination ./bin/chart
 
 test-helm: ensure-helm-install
-	./test/run-helm-tests.sh
+	ls
 
 kind-multi-node:
 	kind create cluster --name kind --config ./hack/kind_config.yaml --wait 2m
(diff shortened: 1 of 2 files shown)


knelasevero:
@ingvagabund how would we properly test this (as we discussed, checking if prow would avoid merging also with gh-actions failing) without the risk of the actual merge?

knelasevero:
> fyi I think WIP will still run tests (it just adds the hold label), but draft PRs don't run tests

Thanks!

@ingvagabund I don't think we can follow what you described in https://github.com/kubernetes-sigs/descheduler/pull/895#issuecomment-1203608848

If we don't remove the test from test-infra, and try to introduce an error on purpose, it will fail there anyways... and then we can't test if prow will consider gh-actions steps

knelasevero:
I have asked a clarification question on slack on #prow channel: https://kubernetes.slack.com/archives/CDECRSC5U/p1660078225162069

ingvagabund:
> If we don't remove the test from test-infra, and try to introduce an error on purpose, it will fail there anyways... and then we can't test if prow will consider gh-actions steps. Also we have a helm lint check on pull-descheduler-verify-master step

Updating the Makefile and temporarily skipping both the `test-helm` script and the `lint-chart` might help.

knelasevero:
> Updating the Makefile and temporarily skipping both the test-helm script and the lint-chart might help. We can always quickly revert this PR if it merges by accident.

Nice suggestion. I think we can test already. Prow folks already told me prow would hold, but does not hurt to test :D

knelasevero:
based on

> We can always quickly revert this PR if it merges by accident.

I am going to

/unhold

ingvagabund:
/approve
/lgtm
Fingers crossed

knelasevero:
Closing since this is validated

image

Issue #873, opened 2022-07-07
RemovePodsViolatingNodeAffinity should not rely on node capacity when check if pod fits on current node


seleznev:
**What version of descheduler are you using?**

descheduler version: 0.24.1

**Does this issue reproduce with the latest release?**

Yes.

**Which descheduler CLI options are you using?**

```
- --policy-config-file
- /policy-dir/policy.yaml
- --v
- "4"
```

**Please provide a copy of your descheduler policy config file**

```
apiVersion: "descheduler/v1alpha1"
kind: "DeschedulerPolicy"
evictLocalStoragePods: true
ignorePvcPods: true
maxNoOfPodsToEvictPerNamespace: 1
maxNoOfPodsToEvictPerNode: 1
strategies:
  LowNodeUtilization:
    enabled: false
    params:
      nodeResourceUtilizationThresholds:
        targetThresholds:
          cpu: 50
          memory: 50
          pods: 50
        thresholds:
          cpu: 20
          memory: 20
          pods: 20
  RemoveDuplicates:
    enabled: false
  RemovePodsViolatingInterPodAntiAffinity:
    enabled: false
  RemovePodsViolatingNodeAffinity:
    enabled: true
    params:
      nodeAffinityType:
      - requiredDuringSchedulingIgnoredDuringExecution
  RemovePodsViolatingNodeTaints:
    enabled: false
```

**What k8s version are you using (`kubectl version`)?**

<details><summary><code>kubectl version</code> Output</summary><br><pre>
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.14", GitCommit:"57a3aa3f13699cf3db9c52d228c18db94fa81876", GitTreeState:"clean", BuildDate:"2021-12-15T14:52:33Z", GoVersion:"go1.15.15", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.14", GitCommit:"57a3aa3f13699cf3db9c52d228c18db94fa81876", GitTreeState:"clean", BuildDate:"2021-12-15T14:47:10Z", GoVersion:"go1.15.15", Compiler:"gc", Platform:"linux/amd64"}
</pre></details>

**What did you do?**

Steps to reproduce:

1. Create 2 k8s nodes with same labels and same capacity.
[...]

PR #916, opened 2022-08-16
NodeFit: do not check whether node fitsRequest when a pod is already assigned to the node

ingvagabund:
SSIA

Fixes: https://github.com/kubernetes-sigs/descheduler/issues/873


damemi (commented):
/approve
a lot simpler solution than I had in https://github.com/kubernetes-sigs/descheduler/issues/873#issuecomment-1177652294, there's no real reason to be checking against the pod's current node in NodeFit

PR #918, opened 2022-08-18
Bump to k8s 1.25-rc.0

JaneLiuL:
as title


JaneLiuL:
@ingvagabund  it seems that we miss to delete RemovePodsViolatingNodeTaints codes, see https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/descheduler/descheduler_test.go#L33
could i delete in this pr?

JaneLiuL:
@damemi  @ingvagabund 
i finger out k8s 1.25.0-rc.0 have some issues with  fakeclient EvictV1 pod. please see the code i just raise in this pr. see the function here
https://github.com/JaneLiuL/descheduler/blob/master/pkg/descheduler/descheduler_test.go#176

i use fakeclient to evict pod, but the pod number still the same.
that's why ut jobs always fail on `TestTaintsUpdated`.

JaneLiuL:
/cc @damemi  @ingvagabund 
when bump to k8s 1.25-rc.0, fakeclient create two pods, and fakeclient evict one pod, but the pod number still two.
and the ut case `TestTaintsUpdated ` will fail.
and other ut case since they use `podEvictor.TotalEvicted()` so will not fail.
what do you think for this situation?

damemi:
@JaneLiuL I played with that test a little, and it's failing for me too. Is the taint actually being applied to the node? We should check all the errors that are being handled too, maybe something is exiting before descheduler actually runs. Or maybe it is a side effect of the refactoring (@ingvagabund maybe you could take a look?)

That test has some pretty short timeouts and intervals too. I'm not sure how well the descheduler runs with an interval of only 100ms. Maybe play around with those

@JaneLiuL it looks like you also have some vendor updates that are missing based on the CI, but we also need to update those too so maybe it's CI that's off

JaneLiuL:
thanks @damemi  i update the vendor now.
i try extend `time.After(1 * time.Minute)`, and still fail on `TestTaintsUpdated ` as below.
```
=== RUN   TestTaintsUpdated
    descheduler_test.go:102: Unable to evict pod, node taint did not get propagated to descheduler strategies timed out waiting for the condition
--- FAIL: TestTaintsUpdated (61.01s)
```

@ingvagabund would you help to have a look? much thanks for your help :)

damemi:
@JaneLiuL have you tried any other debugging steps? We need to figure out why the test is not seeing the pods evicted. That could be a few things:
* the taints are not being applied to the node
* the fakeclient is not being correctly called to evict the pod
* the descheduler strategy is broken

just a few ideas, could you try to see what's actually causing the failure?

JaneLiuL:
@damemi  i just update the code above, you can check i add one more unit test case call `TestDuplicate`, but still can not been evict with fakeclient .

damemi:
@JaneLiuL ok, we need to figure out what is breaking because this is blocking the 1.25 release.

Are we sure the fake client is being properly passed to the strategy? Have you tried any other debugging approaches?

We can't just close this PR, because we need to update to 1.25 anyway. So this test is a priority to figure out

a7i:
I can help debug later today.

damemi:
thanks @a7i, I don't have a lot of time to dedicate to this right now, so I appreciate the help!

JaneLiuL:
Thanks @a7i , thanks very much for your help.

a7i:
I debugged through the test and the eviction client is Evicting the Pod, but when we do a List, the same Pod is returned.

I filed an [Issue](https://github.com/kubernetes/kubernetes/issues/112168) upstream (and have a suspicion that it was caused by [this](https://github.com/kubernetes/kubernetes/pull/110425/files#diff-5e284010fb4c5f529014f3aad4e3d05969574187d878fb9cd3802b3147cda9c3L108) which is part of 1.25)

I continue to debug and hopefully contribute upstream if I find the root cause.

a7i:
Also @JaneLiuL I have a few changes that I can push up:
1. A few v1beta1 sticking around (both import and comments)
2. Remove all references to Pod Security Policy (since it's removed in 1.25)

I can push these up tomorrow.

damemi:
Thanks for the investigation @a7i, will follow that issue to see what the resolution is.

I wonder if updating our CI to k8s 1.25 would fix it? We could run the unit tests against a local k8s 1.25 cluster to check (and a 1.24 cluster to verify that it fails)

JaneLiuL:
@damemi  hi , unit test is not depend on the k8s version. i still believe this is the bug of kubernetes fakeclient. let's wait about 1 more weeks to see k8s team response, if still not fix, i suggest maybe we can annotation the fail unit test first and release descheduler.
when k8s fix or confirm this issue, and we go back to fix it.

a7i:
So turns out that we weren't really asserting that a pod is no longer returned. Instead, we were looking for an error [here](https://github.com/kubernetes-sigs/descheduler/blob/1eade5bf915e8921ca3267a0275975ef8398830b/pkg/descheduler/descheduler_test.go#L94) 🙈 

I am in the process of massaging the test to use a Reactor, stay tuned and I will have the changes in an hour or so.

PR #928, opened 2022-08-27
clarify which version PodLifeTime introduced states parameter and deprecated podStatusPhases

a7i:
Signed-off-by: Amir Alavi <amiralavi7@gmail.com>

Add a ~footnote~ column to indicate which version `states` will be supported and `podStatusPhases` was deprecated in.

Feedback: https://github.com/kubernetes-sigs/descheduler/pull/834#discussion_r956608971


damemi (commented):
/approve
/lgtm

PR #931, opened 2022-08-30
remove TODO comments for cronjob v1beta1 support

a7i:
- Descheduler promises 3 version skew
- The batch/v1beta1 API version of CronJob will no longer be served in v1.25.

### DIFF
--- a/charts/descheduler/values.yaml
+++ b/charts/descheduler/values.yaml
@@ -28,7 +28,7 @@ fullnameOverride: ""
 # labels that'll be applied to all resources
 commonLabels: {}
 
-cronJobApiVersion: "batch/v1"  # Use "batch/v1beta1" for k8s version < 1.21.0. TODO(@7i) remove with 1.23 release
+cronJobApiVersion: "batch/v1"
 schedule: "*/2 * * * *"
 suspend: false
 # startingDeadlineSeconds: 200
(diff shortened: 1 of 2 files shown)


charts/descheduler/values.yaml
```diff
@@ -28,7 +28,7 @@ fullnameOverride: ""
 # labels that'll be applied to all resources
 commonLabels: {}
 
-cronJobApiVersion: "batch/v1"  # Use "batch/v1beta1" for k8s version < 1.21.0. TODO(@7i) remove with 1.23 release
+cronJobApiVersion: "batch/v1"
```
damemi, line 31: should we remove this helm value entirely if we're no longer supporting k8s versions with batch/v1beta1? I don't know the rules for helm chart compatibility, so that's something to check (ie, if we can even remove it)

a7i: I thought about that but was concerned with compatibility.

I assume that depending on the change, I would need to bump the Chart Version

damemi: we will bump the chart version for the new release anyway, but are there any compatibility conventions around field removal for helm charts?

damemi: we can look into removing the field later, for now this is fine

damemi (commented):
/approve
/lgtm

PR #937, opened 2022-09-03
remove e2e tests from helm github action

a7i:
Rationale:
1. e2e tests are not executed against a Descheduler install. If anything, we could end up with a race condition between a Descheduler run and the e2e tests. For example, notice [here](https://github.com/kubernetes-sigs/descheduler/blob/master/test/e2e/e2e_leaderelection_test.go#L105) that the test runs Descheduler itself.
2. We already run e2e tests against 3 Kubernetes Versions

### DIFF
--- a/.github/workflows/helm.yaml
+++ b/.github/workflows/helm.yaml
@@ -60,10 +60,3 @@ jobs:
       # helm-extra-set-args only available after ct 3.6.0
       - name: Run chart-testing (install)
         run: ct install --config=.github/ci/ct.yaml --helm-extra-set-args='--set=kind=Deployment'
-
-      - name: E2E after chart install
-        env:
-          KUBERNETES_VERSION: "v1.26.0"
-          KIND_E2E: true
-          SKIP_INSTALL: true
-        run: make test-e2e


JaneLiuL (commented):
/lgtm

damemi (commented):
does running the e2e after install just serve as a way to verify that the helm chart was properly installed and the descheduler is actually working?

Maybe there is a simpler test we could run to verify that, but the general purpose for having some functionality test post-install makes sense to me

a7i:
> does running the e2e after install just serve as a way to verify that the helm chart was properly installed and the descheduler is actually working?
> 
> Maybe there is a simpler test we could run to verify that, but the general purpose for having some functionality test post-install makes sense to me

AFAIK, the e2e tests do not rely on the Descheduler install. If anything, we could see a race condition between the Descheduler helm install and the e2e Descheduler run.

knelasevero:
Yeah, I saw the other issue mentioning the additional button step. https://github.com/kubernetes-sigs/descheduler/issues/946

Having these a bit separate was what made me consider the e2e tests in the helm workflow as useful: https://github.com/kubernetes-sigs/descheduler/issues/946

If we have a contribution that is not from an org member, and only has changes in the chart, we could avoid running /ok-to-test for the 3 versions, and just press the button and only get the feedback that is relevante (1 lint and chart install test + 1 latest e2e). [from non-org-members that are not first time contributors helm workflow would just run, and /ok-to-test would not be required]

But since we want to unify that (based on comments of https://github.com/kubernetes-sigs/descheduler/issues/946) my rationale wouldn't  make sense indeed, and e2e tests in helm workflow becomes redundant.

And yes, it does not run against the pod installed, `ct install` creates a ns, installs everything, checks results, and deletes the namespace. To make it run against an install we would have to change how e2e tests are implemented.

knelasevero:
Just sharing an update on this, I talked with @cpanato and ct install does not have a skipp cleanup option, but he thinks it is a valid new feature if we want to contribute to the chart test repository. That would be useful for doing that simple tests mentioned here with the pod running and not deleted after ct install.

ingvagabund:
> But we should add a follow up to add some small test to verify descheduler is actually working after helm install

I wonder whether we can create a new CI job that will build a descheduler image from the current PR. An image which we can then use in all the `pull-descheduler-test-e2e-k8s-BRANCH-VERSION` jobs as an e.g. env which will get injected into `make e2e` so right before running individial e2e a descheduler is deployed as a pod. Having each e2e to only change the descheduler configuration, wait until a descheduler is restarted and running again and then running a test.

knelasevero:
> > But we should add a follow up to add some small test to verify descheduler is actually working after helm install
> 
> I wonder whether we can create a new CI job that will build a descheduler image from the current PR. An image which we can then use in all the `pull-descheduler-test-e2e-k8s-BRANCH-VERSION` jobs as e.g. an env which will get injected into `make e2e`. Deploying a descheduler as a pod right before running individial e2e. Having each e2e to only change the descheduler configuration, wait until a descheduler is restarted and running again and then running a test.
> 
> Helm e2e would "just" replace the step of deploying a descheduler as a pod. So all e2e would be interacting with the helm-based deployment of the descheduler.
> 
> When it comes to https://github.com/kubernetes-sigs/descheduler/blob/master/test/e2e/e2e_leaderelection_test.go#L105 the test would reuse the running instance instead of creating two new instances.

In External Secrets Operator we programaticaly install the chart during our e2e tests, but we have some hard to read setup there... https://github.com/external-secrets/external-secrets/blob/main/e2e/framework/addon/chart.go

And as you say, there we also only do the chart linting, because the chart install is being tested in our e2e tests altogether.

ingvagabund:
Checking logs of `pull-kubernetes-e2e-kind` and `pull-kubernetes-e2e-gce-ubuntu-containerd` jobs in kubernetes/kubernetes repo each job builds its own kubernetes [kind] image. Meaning each job we run would have to build its own descheduler image and deploy it.

knelasevero:
Oh, if we want to keep it still in gh-actions: @cpanato just added that feature 😃❤️: https:/github.com/helm/chart-testing/pull/478#issuecomment-1260652448

https://github.com/helm/chart-testing/pull/478/commits/0e8be346b20fbc09800f4bc7d4e28a62ed46f88b

knelasevero:
Let's merge this. @a7i would you please rebase?

PR #938, opened 2022-09-04
remove dependency on kubectl

a7i:
one less dependency we have to worry about


damemi (commented):
/approve
thanks @a7i, looks like there's just an unused variable

test/e2e/e2e_test.go
```diff
@@ -1325,7 +1325,12 @@ func computeCPUMemFraction(t *testing.T, cs clientset.Interface, node *v1.Node,
 	}
 	for _, pod := range allpods.Items {
 		if pod.Spec.NodeName == node.Name {
-			req, _ := utils.PodRequestsAndLimits(&pod)
+			req, lim := utils.PodRequestsAndLimits(&pod)
```
damemi, line 1328: this is why ci is failing:
```
# sigs.k8s.io/descheduler/test/e2e [sigs.k8s.io/descheduler/test/e2e.test]
test/e2e/e2e_test.go:1328:9: lim declared but not used
```

damemi:
Looks like we need a 1.19 bump similar to https://github.com/kubernetes-sigs/descheduler/pull/801
working on a pr

a7i:
> Looks like we need a 1.19 bump similar to #801 working on a pr

This is actually done in the 0.25 PR but I guess it makes sense to keep the two separate?

damemi:
Ah okay, I missed that it was done there. It can be easier to keep them separate (or at least in separate commits) but this is fine

PR #947, opened 2022-09-08
feat: change DeschedulerVersion and GitVersion labels

eminaktas:
This commit changes build_info metric labels
- AppVersion label will show major+minor version for example 0.24.1

It will be seen like this
```bash
# HELP descheduler_build_info [ALPHA] Build info about descheduler, including Go version, Descheduler version, Git SHA, Git branch
# TYPE descheduler_build_info gauge
descheduler_build_info{AppVersion="0.24.1",GitBranch="metric-label-fix",GitSha1="0317be1b7672c511f5e2e53acaea1dd5cb74fd77",DeschedulerVersion="v20220909-v0.24.1-188-g0317be1b7",GoVersion="go1.19"} 0
```

Signed-off-by: eminaktas <eminaktas34@gmail.com>

### DIFF
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -43,7 +43,7 @@ var (
 			Subsystem:      DeschedulerSubsystem,
 			Name:           "build_info",
 			Help:           "Build info about descheduler, including Go version, Descheduler version, Git SHA, Git branch",
-			ConstLabels:    map[string]string{"GoVersion": version.Get().GoVersion, "DeschedulerVersion": version.Get().GitVersion, "GitBranch": version.Get().GitBranch, "GitSha1": version.Get().GitSha1},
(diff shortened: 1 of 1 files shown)


eminaktas:
cc @damemi

damemi (commented):
/ok-to-test
since this is an alpha metric, I think we can technically change what the labels are reporting. But it may still be confusing that the `DeschedulerVersion` label is different now to anyone who was using that.

Maybe add this to a new label (instead of updating the existing one), like `AppVersion`?

eminaktas:
> /ok-to-test since this is an alpha metric, I think we can technically change what the labels are reporting. But it may still be confusing that the `DeschedulerVersion` label is different now to anyone who was using that.
> 
> Maybe add this to a new label (instead of updating the existing one), like `AppVersion`?

You are right. I just updated the PR.

damemi (commented):
/approve

a7i:
/lgtm

thank you @eminaktas 🏆

PR #953, opened 2022-09-18
set helm-extra-set-args as inline command arg

### LINKED ISSUE #949: Helm test with ct install is installing as CronJob and not Deployment
While looking at https://github.com/kubernetes-sigs/descheduler/issues/946 I noticed that in the gh-actions output https://github.com/kubernetes-sigs/descheduler/pull/945/checks ct install is installing descheduler in CronJob mode, which does not get useful events and logs back to us (even though we have `helm-extra-set-args: "--set=kind=Deployment"` set in https://github.com/kubernetes-sigs/descheduler/blob/master/.github/ci/ct.yaml) . 

This worked at some point since we had this during reviews: https://github.com/kubernetes-sigs/descheduler/pull/895#pullrequestreview-1055672165 :thinking: 

Installing it as Deployment will be better to check info and possibly add subsequent simple tests against that pod.

a7i:
Based on the docs, it seems that `helm-extra-set-args` is only available as [an inline arg](https://github.com/helm/chart-testing/blob/main/doc/ct_install.md#options) but not as a [config](https://github.com/helm/chart-testing/blob/main/pkg/config/config.go#L44).

Closes #949

### DIFF
--- a/.github/ci/ct.yaml
+++ b/.github/ci/ct.yaml
@@ -2,5 +2,4 @@ chart-dirs:
   - charts
 helm-extra-args: "--timeout=5m"
 check-version-increment: false
-helm-extra-set-args: "--set=kind=Deployment"
 target-branch: master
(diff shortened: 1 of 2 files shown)


damemi (commented):
/lgtm

Issue #924, opened 2022-08-25
Convert current evictor filter into a plugin


ingvagabund:
Converting the [evictor filter](https://github.com/kubernetes-sigs/descheduler/blob/0aa233415e6834ec063d5ca71d9faa2f5c790f87/pkg/descheduler/evictions/evictions.go#L246) into a plugin will allow users to either replace the plugin with a different list of filters. Or, register additional plugins to extend the list of available filters. Also, new `preEvictFilter` extension point can be implement to provide strategies ability to protect pods against eviction right before the eviction itself. E.g. LowNodeUtilization plugin needs to operate over all pods to take into account the overall resource consumption. Thus, the only place where pods can be filter out from eviction is right after pods were marked for eviction.

The same process for creating a new plugin can be followed. Including creating plugin arguments (in the plugin package this time). The original `NewEvictorFilter` function will be then replaced with the plugin constructor with passing the plugin arguments.

knelasevero:
![image](https://user-images.githubusercontent.com/2432275/188159104-20ebf6b3-eed0-4d54-b2e9-71069942ab24.png)

Correct me if I am wrong, but right now this is somewhat the plan that we have (just at a high level, I don't wanna focus on details). The idea behind naming the filter that comes right before the eviction PreEvictionFilter makes sense until we think about people coming to the project without much context and having to understand why PreEvictionFilter, with "pre" prefix in it's name, is coming after the "normal" Filter. :thinking: 

I was thinking about a name that could just mean that it is right next to the eviction, and "Near" comes to mind.

So something like

NearEvictionFilter
NearFilterSorter... ?

What do you think?

![image](https://user-images.githubusercontent.com/2432275/188160625-42190472-0086-475f-a7db-ab179d8c6fc5.png)

damemi:
I think `PreEvictionFilter` is fine, because it's referring to something happening Pre-Eviction. If it was just called PreFilter, that would be confusing. But imo this sounds like a second Filter phase that's specific to Eviction, which makes sense to me.

ingvagabund:
I am not against changing `preEvictFilter`, resp. `PreEvictionFilter` into more suitable name that will reflect the fact a pod filtering is taking place right before eviction.

knelasevero:
I went with PreEvictionFilter for now, we can discuss the naming in the PR: https://github.com/kubernetes-sigs/descheduler/pull/955

ingvagabund:
[...]

Issue #753, opened 2022-03-02
Descheduler Framework Proposal


damemi:
Hi all,

I'm creating this issue to gauge interest in our longstanding idea of creating a "Descheduler Framework". The general idea is to abstract the Descheduler internals into a library that third-party developers can use to create their own custom Deschedulers.

I think this is relevant now because we have a lot of big, ongoing tasks that I think would fit well under the umbrella of this effort.  For example, each of these features presents a potential "extension point" for future custom descheduler developers (so, if we choose to take on a framework, it would be worth considering the implementation of these features with that extensibility in mind):
* Supporting out-of-tree strategies ([#679](https://github.com/kubernetes-sigs/descheduler/issues/679), [#557](https://github.com/kubernetes-sigs/descheduler/issues/557))
* Allowing custom options for pod evictions / custom `PodEvictor` implementations ([#690](https://github.com/kubernetes-sigs/descheduler/issues/690), [#725](https://github.com/kubernetes-sigs/descheduler/issues/725))
* Defining an implementable interface for Strategies ([#586](https://github.com/kubernetes-sigs/descheduler/issues/586), [#706](https://github.com/kubernetes-sigs/descheduler/pull/706))
* Refactoring the API to be more flexible and extensible ([#587](https://github.com/kubernetes-sigs/descheduler/pull/587), [#486](https://github.com/kubernetes-sigs/descheduler/issues/486))
* Allowing event-triggered strategy runs ([#488](https://github.com/kubernetes-sigs/descheduler/pull/488), [#696](https://github.com/kubernetes-sigs/descheduler/issues/696))

In addition, the Descheduler has seen a lot of growth and requests for unique features to support custom use cases. Enabling users to build their own Deschedulers in a preferred way will free up our maintainers to focus on larger tasks like those above, while also steering the Descheduler toward a stable project.

I would like to propose an outline for this effort, but first I want to know if this is something people think is worth tackling as a project. I think this also will be a good chance for a lot of contributors to make a significant impact on the project as a whole.

So the main question is: do folks think it would be worth the effort to re-focus some of the above tasks under the umbrella goal of the Descheduler Framework?

---

Update: Based on feedback I put together some more organized thoughts into a doc: https://docs.google.com/document/d/1fr5tfSaCkqGZVXUMJA9pUkIy6GGRwJumPN5MeZ8oewo/edit?usp=sharing

Please comment there or here to help flesh out the design goals for if/when we pursue this

[...]

PR #955, opened 2022-09-20
add PreEvictionFilter extension to DefaultEvictor Plugin

knelasevero:
As per https://github.com/kubernetes-sigs/descheduler/issues/640#issuecomment-1248167067 I am just moving nodefit check to the preEvictionFilter extension point of the DefaultEvictor plugin. Follow ups would include those other options of that issue (and namespace selectors, etc).

Let me know if the evocation of the extension point shouldn't be done in the way that I did.

Also related to https://github.com/kubernetes-sigs/descheduler/issues/924, https://github.com/kubernetes-sigs/descheduler/issues/753#issuecomment-1227422028,

### DIFF
--- a/pkg/framework/plugins/defaultevictor/defaultevictor.go
+++ b/pkg/framework/plugins/defaultevictor/defaultevictor.go
@@ -44,7 +44,9 @@ type constraint func(pod *v1.Pod) error
 // This plugin is only meant to customize other actions (extension points) of the evictor,
 // like filtering, sorting, and other ones that might be relevant in the future
 type DefaultEvictor struct {
+	args        runtime.Object
 	constraints []constraint
+	handle      framework.Handle
 }
 
 // IsPodEvictableBasedOnPriority checks if the given pod is evictable based on priority resolved from pod Spec.
@@ -66,6 +68,8 @@ func New(args runtime.Object, handle framework.Handle) (framework.Plugin, error)
 	}
 
 	ev := &DefaultEvictor{}
+	ev.handle = handle
+	ev.args = defaultEvictorArgs
 
 	if defaultEvictorArgs.EvictFailedBarePods {
 		klog.V(1).InfoS("Warning: EvictFailedBarePods is set to True. This could cause eviction of pods without ownerReferences.")
@@ -125,18 +129,6 @@ func New(args runtime.Object, handle framework.Handle) (framework.Plugin, error)
 			return nil
 		})
 	}
-	if defaultEvictorArgs.NodeFit {
-		ev.constraints = append(ev.constraints, func(pod *v1.Pod) error {
-			nodes, err := nodeutil.ReadyNodes(context.TODO(), handle.ClientSet(), handle.SharedInformerFactory().Core().V1().Nodes(), defaultEvictorArgs.NodeSelector)
-			if err != nil {
-				return fmt.Errorf("could not list nodes when processing NodeFit")
-			}
-			if !nodeutil.PodFitsAnyOtherNode(handle.GetPodsAssignedToNodeFunc(), pod, nodes) {
-				return fmt.Errorf("pod does not fit on any other node because of nodeSelector(s), Taint(s), or nodes marked as unschedulable")
-			}
-			return nil
-		})
-	}
 	if defaultEvictorArgs.LabelSelector != nil && !defaultEvictorArgs.LabelSelector.Empty() {
 		ev.constraints = append(ev.constraints, func(pod *v1.Pod) error {
 			if !defaultEvictorArgs.LabelSelector.Matches(labels.Set(pod.Labels)) {
@@ -152,6 +144,23 @@ func New(args runtime.Object, handle framework.Handle) (framework.Plugin, error)
 // Name retrieves the plugin name
 func (d *DefaultEvictor) Name() string {
 	return PluginName
+}
+
+func (d *DefaultEvictor) PreEvictionFilter(pod *v1.Pod) bool {
+	defaultEvictorArgs := d.args.(*DefaultEvictorArgs)
+	if defaultEvictorArgs.NodeFit {
+		nodes, err := nodeutil.ReadyNodes(context.TODO(), d.handle.ClientSet(), d.handle.SharedInformerFactory().Core().V1().Nodes(), defaultEvictorArgs.NodeSelector)
+		if err != nil {
+			klog.V(1).ErrorS(fmt.Errorf("Pod fails the following checks"), "pod", klog.KObj(pod))
+			return false
+		}
+		if !nodeutil.PodFitsAnyOtherNode(d.handle.GetPodsAssignedToNodeFunc(), pod, nodes) {
+			klog.V(1).ErrorS(fmt.Errorf("pod does not fit on any other node because of nodeSelector(s), Taint(s), or nodes marked as unschedulable"), "pod", klog.KObj(pod))
+			return false
+		}
+		return true
+	}
+	return true
 }
 
 func (d *DefaultEvictor) Filter(pod *v1.Pod) bool {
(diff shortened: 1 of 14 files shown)


pkg/framework/plugins/nodeutilization/nodeutilization.go
```diff
@@ -286,36 +288,42 @@ func evictPods(
...
-					if name == v1.ResourcePods {
-						nodeInfo.usage[name].Sub(*resource.NewQuantity(1, resource.DecimalSI))
-						totalAvailableUsage[name].Sub(*resource.NewQuantity(1, resource.DecimalSI))
-					} else {
-						quantity := utils.GetResourceRequestQuantity(pod, name)
-						nodeInfo.usage[name].Sub(quantity)
-						totalAvailableUsage[name].Sub(quantity)
+			shouldEvict, err := podEvictor.PreEvictionFilter(pod)
+			if err != nil {
+				klog.V(1).ErrorS(fmt.Errorf("could not run eviction filter plugin for pod"), "pod", klog.KObj(pod))
```
knelasevero, line 293: Not sure what I should do here since we are not returning errors in this function. So I am just logging it

pkg/framework/plugins/defaultevictor/defaultevictor.go
```diff
@@ -154,6 +146,21 @@ func (d *DefaultEvictor) Name() string {
 	return PluginName
 }
 
+func (d *DefaultEvictor) PreEvictionFilter(pod *v1.Pod) (bool, error) {
```
ingvagabund, line 149: I still prefer to return only bool in the same fashion as `Filter` and log any error the same way. Later on we can decide whether it's more preferable to return a status instead of a bool.

pkg/framework/plugins/defaultevictor/defaultevictor_test.go
```diff
@@ -627,7 +627,9 @@ func TestDefaultEvictorFilter(t *testing.T) {
 			}
 
 			result := evictorPlugin.(framework.EvictorPlugin).Filter(test.pods[0])
-			if result != test.result {
+			// Not handling errors here since we expect them when, for example, all nodes are not available for nodefit
+			resultPreEvictionFilter, _ := evictorPlugin.(framework.EvictorPlugin).PreEvictionFilter(test.pods[0])
```
ingvagabund, line 631: Sooner or later it will be practical to create a separate unit test for `PreEvictionFilter`. Easier to do it now :)

pkg/framework/plugins/podlifetime/pod_lifetime.go
```diff
@@ -124,8 +124,16 @@ func (d *PodLifeTime) Deschedule(ctx context.Context, nodes []*v1.Node) *framewo
 	podutil.SortPodsBasedOnAge(podsToEvict)
 
 	for _, pod := range podsToEvict {
-		if !d.handle.Evictor().NodeLimitExceeded(nodeMap[pod.Spec.NodeName]) {
-			d.handle.Evictor().Evict(ctx, pod, evictions.EvictOptions{})
+		shouldEvict, err := d.handle.Evictor().PreEvictionFilter(pod)
```
ingvagabund, line 127: With `PreEvictionFilter` returning only bool, `PreEvictionFilter` can be combined with `podFilter` since there's no difference between invoking `PreEvictionFilter` right after `Filter` or right before `Evict` for this strategy. The same holds for other strategies like failedpods, toomanyrestarts, etc.

ingvagabund (changes requested):

pkg/framework/plugins/defaultevictor/defaultevictor_test.go
knelasevero: On it :smile:

pkg/framework/plugins/defaultevictor/defaultevictor_test.go
```diff
@@ -30,16 +30,320 @@ import (
...
+		evictFailedBarePods     bool
+		evictLocalStoragePods   bool
+		evictSystemCriticalPods bool
+		priorityThreshold       *int32
+		nodeFit                 bool
+		result                  bool
+	}
+
+	testCases := []testCase{
+		{
```
ingvagabund, line 54: Worth adding a new test with `nodeFit: false`.

pkg/framework/plugins/defaultevictor/defaultevictor_test.go
```diff
@@ -30,16 +30,320 @@ import (
...
+
+			getPodsAssignedToNode, err := podutil.BuildGetPodsAssignedToNodeFunc(podInformer)
+			if err != nil {
+				t.Errorf("Build get pods assigned to node function error: %v", err)
+			}
+
+			sharedInformerFactory.Start(ctx.Done())
+			sharedInformerFactory.WaitForCacheSync(ctx.Done())
+
+			// var opts []func(opts *FilterOptions)
```
ingvagabund, line 303: This comment can be removed.

pkg/framework/plugins/defaultevictor/defaultevictor_test.go
```diff
@@ -347,225 +651,6 @@ func TestDefaultEvictorFilter(t *testing.T) {
 			evictSystemCriticalPods: true,
 			priorityThreshold:       &lowPriority,
 			result:                  true,
-		}, {
```
ingvagabund, line 350: Worth introducing a new test that has `nodeFit: true` to prove setting the arg has no effect on the `Filter` ep functionality.

pkg/framework/plugins/defaultevictor/defaultevictor_test.go
```diff
@@ -627,7 +712,8 @@ func TestDefaultEvictorFilter(t *testing.T) {
 			}
 
 			result := evictorPlugin.(framework.EvictorPlugin).Filter(test.pods[0])
-			if result != test.result {
+			// Not handling errors here since we expect them when, for example, all nodes are not available for nodefit
```
ingvagabund, line 715: Is this comment still valid?

pkg/framework/plugins/podlifetime/pod_lifetime.go
```diff
@@ -124,9 +125,7 @@ func (d *PodLifeTime) Deschedule(ctx context.Context, nodes []*v1.Node) *framewo
 	podutil.SortPodsBasedOnAge(podsToEvict)
 
 	for _, pod := range podsToEvict {
-		if !d.handle.Evictor().NodeLimitExceeded(nodeMap[pod.Spec.NodeName]) {
-			d.handle.Evictor().Evict(ctx, pod, evictions.EvictOptions{})
-		}
+		d.handle.Evictor().Evict(ctx, pod, evictions.EvictOptions{})
```
ingvagabund, line 128: `!d.handle.Evictor().NodeLimitExceeded(nodeMap[pod.Spec.NodeName])` condition is still needed here

pkg/framework/plugins/removepodsviolatinginterpodantiaffinity/pod_antiaffinity.go
```diff
@@ -58,6 +58,7 @@ func New(args runtime.Object, handle framework.Handle) (framework.Plugin, error)
 	}
 
 	podFilter, err := podutil.NewOptions().
+		WithFilter(handle.Evictor().PreEvictionFilter).
```
ingvagabund, line 61: `Deschedule` method is using `d.handle.Evictor().Filter`. `handle.Evictor().PreEvictionFilter` needs to be combine there:
```
for i := 0; i < totalPods; i++ {
	if checkPodsWithAntiAffinityExist(pods[i], pods) && d.handle.Evictor().Filter(pods[i]) {
		if d.handle.Evictor().Evict(ctx, pods[i], evictions.EvictOptions{}) {
```

ingvagabund (changes requested):

PR #956, opened 2022-09-21
fix sharedInformerFactory register

xiaoanyunfei:
sharedInformerFactory should register resource before call `sharedInformerFactory.Start()`

I get logs `Node lister returned empty list, now fetch directly`


damemi (commented):
/ok-to-test

ingvagabund:
/approve
/lgtm

@xiaoanyunfei thanks for noticing this.

Issue #951, opened 2022-09-17
Proposal: OpenTelemetry Tracing Per-Plugin 


Dentrax:
**Is your feature request related to a problem? Please describe.**

Tracing is one of the important observability topics about continuously watch overall health and performance of the system. Especially with the new descheduler framework, it would perfectly fit for implementing tracing per-plugin to observe such as `Deschedule()`, `Balance()`, etc. functions using spans.

**Describe the solution you'd like**

Implement [OpenTelemetry](https://opentelemetry.io/) `(go.opentelemetry.io)` client in Descheduler. [Here is an example](https://github.com/kyverno/kyverno/blob/9b41e2e0174c10b3548a5364e205911d6a0047c7/pkg/tracing/tracing.go) source file how Kyverno implemented OpenTracing in their codebase.

Introduce a new OpenTelemetry client in `framework.Handle` _interface_:
```
type Handle interface {
  ...
  TraceClient() *TracerProvider
}
```

Introduce a new boolean flag to `PodEvictor` struct to enable tracing:

```
type PodEvictor struct {
  ...
  tracingEnabled bool
}
```

Wrap all the following functions in a `DoInSpan` func if `tracingEnabled`, otherwise it should be as is:

Plugin initialization:
```
tracing.DoInSpan(context.Background(), "PLUGIN_NAME", "new", func(ctx context.Context) {
		pg, err := PLUGIN_NAME.New(args, handle)
})
```

Deschedule:
```
tracing.DoInSpan(context.Background(), "PLUGIN_NAME", "deschedule", func(ctx context.Context) {
		status = pg.(framework.DeschedulePlugin).Deschedule(ctx, nodes)
})
```

Balance:
```
tracing.DoInSpan(context.Background(), "PLUGIN_NAME", "balance", func(ctx context.Context) {
		status = pg.(framework.BalancePlugin).Balance(ctx, nodes)
})	
```

Tracing internal metrics of a custom plugin would be straightforward. Here is some function list that would consider worth to trace:

* `classifyNodes`
* `evictPodsFromSourceNodes`
* `getTargetNodes`
* observe end-start timing for entire `for i := 0; i < totalPods; i++ {` loops
* observe end-start timing for entire `for _, pod := range podsToEvict` loops
* observe end-start timing for entire `for _, node := range nodes` loops
* etc.

**Describe alternatives you've considered**
`-`

**What version of descheduler are you using?**
`-`

**Additional context**
`-`

[...]

PR #958, opened 2022-09-23
added otel capabilities for descheduler plugin

harshanarayana:
## Note

Early Draft of the PR to get feedbacks on. This is not yet ready for full review. 

Fixes: #951


pkg/descheduler/strategy_migration.go
```diff
@@ -263,9 +238,24 @@ var pluginsMap = map[string]func(ctx context.Context, nodes []*v1.Node, params *
...
+	},
+}
+
+// RunFunc represents a framework.Plugin function.
+// - (framework.BalancePlugin).Balance
+// - (framework.DeschedulePlugin).Deschedule
+type RunFunc func(ctx context.Context, nodes []*v1.Node) *framework.Status
+
+// RunPlugin runs a plugin.
+func RunPlugin(plugin framework.Plugin, runFn RunFunc, operationName, pluginName string) func(ctx context.Context, nodes []*v1.Node) {
```
damemi, line 251: rather than wrapping the plugin's function in a span, why not start and close the span within the plugin's `Deschedule` function?

I would prefer if traces are implemented by the plugins themselves rather than the framework. The framework can provide the mechanisms (like registering a global traceprovider) but I would leave the actual spans up to plugin developers to implement

Dentrax: My initial attempt was to put Span login within each plugin individually. And then I thought it could hard to maintain if we manage tracing stuff separately:
* it would be easy to forget tracing logic in the function
* it would bring code duplication (1 line tracing, 1 line defer _times_ plugin count)
* it would be easy to forget to add `span.RecordError(status.Err)` function in all error cases
* should plugins really know the existence of tracing? (why plugin developers should duplicate this logic?)

These are just my concern, I'm OK to go with your way anyway. Wanted to learn more about your concern. 🤞

damemi: > should plugins really know the existence of tracing? (why plugin developers should duplicate this logic?)

One benefit of a framework model is detangling the plugin logic from the surrounding framework internals, so in that sense the answer is yes.

On that end, it makes sense for us to start a parent trace in our internals (somewhere around [here](https://github.com/kubernetes-sigs/descheduler/blob/v0.25.1/pkg/descheduler/descheduler.go#L355)), and plugins can use the context to add whatever additional spans they want. But the responsibilities of good design you mention in your other bullet points are kept to the individual plugin developers

harshanarayana: > On that end, it makes sense for us to start a parent trace in our internals (somewhere around [here](https://github.com/kubernetes-sigs/descheduler/blob/v0.25.1/pkg/descheduler/descheduler.go#L355)),

+1 from me for this. I think we can start from this for now and see how it goes. Enforcing the tracing related code can be done mostly during the code review of a new plugin getting added anyway.  The current code in the repo already lets us create the parent trace and pass the context down to the individual plugin/strategies so that they can trace the required bit of the code in that accordingly. 

This will also leave the plugin developer with freedom of being able to trace some of their internal helper/util functions as well if required.

damemi: @harshanarayana yup, that's exactly what I'm thinking

harshanarayana: Screenshot 2022-10-04 at 4 18 24 PM

pkg/apis/componentconfig/v1alpha1/types.go
```diff
@@ -22,6 +22,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	componentbaseconfig "k8s.io/component-base/config"
 	registry "k8s.io/component-base/logs/api/v1"
+	"sigs.k8s.io/descheduler/pkg/apis/componentconfig"
```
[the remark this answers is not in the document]
harshanarayana: /Ack.  I am working on getting this changed. Will publish a copy of this by tomorrow.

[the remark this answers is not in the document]
harshanarayana: Changes done.

harshanarayana:
@damemi  @damemi 

https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_descheduler/958/pull-descheduler-verify-master/1580431724137943040

Unable to figure out why this issue on the CICD showing up. 

```bash
❯ hack/verify-conversions.sh
hack/uGenerated conversions verified.

```
if I run the verify locally, it works just fine!

damemi:
@harshanarayana I checked out your branch locally and it fails when I run it with the same output as the CI. Could you try running `make clean` first, or perhaps in a container?

harshanarayana:
@mikedanese That is definitely weird. I did that already and I see not delta in the commits or files! :(

damemi:
@harshanarayana I think it has something to do with https://github.com/kubernetes-sigs/descheduler/pull/982, but I can't quite figure it out. Posted a more detailed explanation there

harshanarayana:
> 

Definitely looks like the same. Let me run the same generation code in a different machine I have and see if actually makes a difference. Do you think it matters where my code resides though? Mine doesn't exists under the `$GOPATH/src/<x>/<y>` path

damemi:
@harshanarayana could you rebase and try running the scripts again to see if there's any changes now?
/retest

harshanarayana:
> @harshanarayana could you rebase and try running the scripts again to see if there's any changes now? /retest

Doing that right aways. Thanks @mikedanese

harshanarayana:
> @harshanarayana could you rebase and try running the scripts again to see if there's any changes now? /retest

Thanks a lot @mikedanese Seem to have done the right thing now. I was not able to get it to work when I rebased and ran on the machine directly. had to build the update command inside the container for it to work. #sigh

harshanarayana:
```bash
❯ make verify
./hack/verify-govet.sh
# sigs.k8s.io/descheduler/pkg/apis/componentconfig
pkg/apis/componentconfig/zz_generated.deepcopy.go:75:10: cannot use c (variable of type *TracingConfiguration) as type runtime.Object in return statement:
	*TracingConfiguration does not implement runtime.Object (missing GetObjectKind method)
```
@mikedanese  This is where it got stuck now with new changes. Looks like the deep copy thing is not fully generated.

damemi:
@harshanarayana is there a reason for the [deepcopy gen tag](https://github.com/kubernetes-sigs/descheduler/blob/d33137b47340deeda22191f17456b96ecdd5b58f/pkg/apis/componentconfig/v1alpha1/types.go#L68) on `TracingConfiguration`? (also in the unversioned type) that might be the problem, can you try without it?

(btw, please be careful with autocompleting my github name, it's @damemi, but sometimes github picks up a different name -- sorry @mikedanese!)

harshanarayana:
> @harshanarayana is there a reason for the [deepcopy gen tag](https://github.com/kubernetes-sigs/descheduler/blob/d33137b47340deeda22191f17456b96ecdd5b58f/pkg/apis/componentconfig/v1alpha1/types.go#L68) on `TracingConfiguration`? (also in the unversioned type) that might be the problem, can you try without it?

Done. Let me see how that goes.  Don't think the deep copy bit was really required for that

Dentrax:
Rebased. I think we are ready to review! 😊

---

I'm not sure if we correctly used the context in the following line:
```go
parentCtx, _, parentCloser := tracing.StartSpan(context.Background(), "descheduler_parent", "deschedule")
```

We initiated new context and passed to plugins; or should we use the `ctx` here?

damemi:
> We initiated new context and passed to plugins; or should we use the `ctx` here?

We should use the same context that's already available

Dentrax:
> We should use the same context that's already available

I did think so, makes sense! Updated:

```diff
- parentCtx, _, parentCloser := tracing.StartSpan(context.Background(), "descheduler_parent", "deschedule")
+ parentCtx, _, parentCloser := tracing.StartSpan(ctx, "descheduler_parent", "deschedule")
```

FYI @harshanarayana

Dentrax:
Kindly ping @damemi. 🤞

harshanarayana:
Rebasing them and pushing a new set by tomorrow.

harshanarayana:
```bash
1 chart(s) linted, 0 chart(s) failed
./hack/verify-toc.sh
./hack/verify-conversions.sh
Generated output differs:
diff -Naupr pkg/apis/componentconfig/v1alpha1/zz_generated.conversion.go ./hack/../_tmp/kube-verify.Z6K2v6/descheduler/pkg/apis/componentconfig/v1alpha1/zz_generated.conversion.go
--- pkg/apis/componentconfig/v1alpha1/zz_generated.conversion.go	2023-01-03 06:54:10.555176506 +0000
+++ ./hack/../_tmp/kube-verify.Z6K2v6/descheduler/pkg/apis/componentconfig/v1alpha1/zz_generated.conversion.go	2023-01-03 06:56:33.674897486 +0000
@@ -2,7 +2,7 @@
 // +build !ignore_autogenerated
 
 /*
-Copyright 2022 The Kubernetes Authors.
+Copyright 2023 The Kubernetes Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
Generated conversions verify failed. Please run ./hack/update-generated-conversions.sh
make: *** [Makefile:133: verify-gen] Error 1

```

@damemi  Looks like one more odd behavior with verify. I am not even sure what the diff is from that error. And if I run the update hack script, it generates no diff. Any idea what I am doing wrong ?

damemi:
@harshanarayana the diff is in the copyright year for the generated headers (maybe from merging the 1.26 GA libraries last week).

Since it's happening in CI but not your local, try rebasing on master and running the generator locally again.

harshanarayana:
@mikedanese thought the same an re-ran the `hack/update-generated-conversions.sh` after rebasing, but there is no diff generated from it.

damemi:
@harshanarayana I checked out your branch and ran `hack/update-generated-conversions.sh` and got the expected diff:
```
$ git fetch git@github.com:harshanarayana/descheduler.git feature/enable-otel-tracing
remote: Enumerating objects: 276, done.
remote: Counting objects: 100% (276/276), done.
remote: Compressing objects: 100% (107/107), done.
remote: Total 276 (delta 155), reused 253 (delta 150), pack-reused 0
Receiving objects: 100% (276/276), 169.01 KiB | 856.00 KiB/s, done.
Resolving deltas: 100% (155/155), completed with 87 local objects.
From github.com:harshanarayana/descheduler
 * branch                feature/enable-otel-tracing -> FETCH_HEAD
 * 
$ git checkout FETCH_HEAD
HEAD is now at 222fc09ef update vendoring to match the upstream branches
        
mikedame[(HEAD detached at 222fc09ef)] ~/go/src/sigs.k8s.io/descheduler$ hack/update-generated-conversions.sh 

mikedame[(HEAD detached at 222fc09ef) !] ~/go/src/sigs.k8s.io/descheduler$ git diff
diff --git a/pkg/apis/componentconfig/v1alpha1/zz_generated.conversion.go b/pkg/apis/componentconfig/v1alpha1/zz_generated.conversion.go
index 924a81590..5087a6369 100644
--- a/pkg/apis/componentconfig/v1alpha1/zz_generated.conversion.go
+++ b/pkg/apis/componentconfig/v1alpha1/zz_generated.conversion.go
@@ -2,7 +2,7 @@
 // +build !ignore_autogenerated
 
 /*
-Copyright 2022 The Kubernetes Authors.
+Copyright 2023 The Kubernetes Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
```

Maybe it is a Go version issue? Or try running in a container?

harshanarayana:
@mikedanese  /ack. Trying that in a container now.

harshanarayana:
@mikedanese  PTAL when possible? I think most of the changes are done. Would really appreciate if you can check this and give a go ahead with regards to the changes here if all is well

damemi:
@harshanarayana thanks for reminding me about this, will take another look

cmd/descheduler/app/server.go
```diff
@@ -94,6 +96,15 @@ func NewDeschedulerCommand(out io.Writer) *cobra.Command {
 				return
 			}
 
+			tracerProvider, err := tracing.NewTracerProvider(s.Tracing.OtelCollector, s.Tracing.OtelTransportCert, s.Tracing.OtelTraceName, s.Tracing.OtelTraceNamespace)
+			if err != nil {
+				klog.Fatalf("failed to setup tracing provider", "err", err)
+				return
+			}
+			traceCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+			defer tracing.Shutdown(traceCtx, tracerProvider)
```
damemi, line 105: we should be re-using the `ctx` that's already here where available rather than creating a new one (noticed this throughout the PR)

pkg/apis/componentconfig/types.go
```diff
@@ -54,10 +54,29 @@ type DeschedulerConfiguration struct {
...
 	// Logging specifies the options of logging.
 	// Refer [Logs Options](https://github.com/kubernetes/component-base/blob/master/logs/api/v1/options.go) for more information.
 	Logging registry.LoggingConfiguration
 }
+
+// TracingConfiguration contains tracing options.
+type TracingConfiguration struct {
+	// OtelCollector is the address of the OpenTelemetry collector.
+	// If not specified, tracing will be used NoopTraceProvider.
+	OtelCollector string
```
damemi, line 72: maybe `OtelCollectorEndpoint`?

pkg/apis/componentconfig/types.go
```diff
@@ -54,10 +54,29 @@ type DeschedulerConfiguration struct {
...
+	// OtelTransportCert is the path to the certificate file for the OpenTelemetry collector.
+	// If not specified, provider will start in insecure mode.
+	OtelTransportCert string
+	// OtelTraceName is the name of the service to be used in the OpenTelemetry collector.
+	// If not specified, the default value is "descheduler_traces".
+	OtelTraceName string
```
damemi, line 78: how about `ServiceName`? I also don't think all of these need the `Otel` prefix. That gets especially confusing for fields like this (is this the descheduler's service name? or the name of the collector's service?)

pkg/apis/componentconfig/v1alpha1/types.go
```diff
@@ -54,10 +54,29 @@ type DeschedulerConfiguration struct {
...
+	OtelCollector string
+	// OtelTransportCert is the path to the certificate file for the OpenTelemetry collector.
+	// If not specified, provider will start in insecure mode.
+	OtelTransportCert string
+	// OtelTraceName is the name of the service to be used in the OpenTelemetry collector.
+	// If not specified, the default value is "descheduler_traces".
+	OtelTraceName string
+	// OtelTraceNamespace is the namespace of the service to be used in the OpenTelemetry collector.
+	// If not specified, tracing will be used default namespace.
+	OtelTraceNamespace string
```
damemi, line 81: same comments as for the internal types :)

pkg/descheduler/descheduler.go
```diff
@@ -318,6 +319,8 @@ func RunDeschedulerStrategies(ctx context.Context, rs *options.DeschedulerServer
 			eventRecorder,
 		)
 
+		parentCtx, _, parentCloser := tracing.StartSpan(ctx, "descheduler_parent", "deschedule")
```
damemi, line 322: can this just re-assign to the existing `ctx` instead of creating a new variable?

pkg/descheduler/descheduler.go
```diff
@@ -398,7 +401,7 @@ func RunDeschedulerStrategies(ctx context.Context, rs *options.DeschedulerServer
 			// handle or function which the Evictor has access to. For migration/in-progress framework
 			// work, we are currently passing this via context. To be removed
 			// (See discussion thread https://github.com/kubernetes-sigs/descheduler/pull/885#discussion_r919962292)
-			childCtx := context.WithValue(ctx, "strategyName", pg.Name())
+			childCtx := context.WithValue(parentCtx, "strategyName", pg.Name())
```
damemi, line 404: same question with these two, rather than creating a new `childCtx` this could just reassign to `ctx`. Though I also don't think you need to manually use `context.WithValue` here, but something like span attributes would be better

pkg/descheduler/evictions/evictions.go
```diff
@@ -113,6 +116,17 @@ type EvictOptions struct {
 // EvictPod evicts a pod while exercising eviction limits.
 // Returns true when the pod is evicted on the server side.
 func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod, opts EvictOptions) bool {
+	var err error
+	evictCtx, evictSpan, evictCloser := tracing.StartSpanWithOptions(ctx, tracing.EvictionOption, []trace.TracerOption{}, "EvictPod", []attribute.KeyValue{
```
damemi, line 120: same question re:new `evictCtx` variable

pkg/framework/plugins/nodeutilization/highnodeutilization.go
```diff
@@ -71,7 +72,14 @@ func (h *HighNodeUtilization) Name() string {
 }
 
 // Balance extension point implementation for the plugin
-func (h *HighNodeUtilization) Balance(ctx context.Context, nodes []*v1.Node) *framework.Status {
+func (h *HighNodeUtilization) Balance(ctx context.Context, nodes []*v1.Node) (status *framework.Status) {
+	pluginCtx, span, spanCloser := tracing.StartSpan(ctx, tracing.BalanceOperation, h.Name())
```
damemi, line 76: same question re: `pluginCtx` throughout all plugins

pkg/tracing/tracing.go
```diff
@@ -0,0 +1,128 @@
...
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	semconv "go.opentelemetry.io/otel/semconv/v1.12.0"
+	"go.opentelemetry.io/otel/trace"
+	"google.golang.org/grpc/credentials"
+	"k8s.io/klog/v2"
+)
+
+const (
+	// DefaultServiceName is the default service name used for tracing.
+	DefaultServiceName = "descheduler_traces"
```
damemi, line 23:
```suggestion
	DefaultServiceName = "descheduler"
```

pkg/tracing/tracing.go
```diff
@@ -0,0 +1,128 @@
...
+func StartSpan(ctx context.Context, tracerName, operationName string) (context.Context, trace.Span, func()) {
+	return StartSpanWithOptions(ctx, tracerName, []trace.TracerOption{}, operationName, []attribute.KeyValue{})
+}
+
+func StartSpanWithOptions(ctx context.Context, tracerName string, traceOpts []trace.TracerOption, operationName string, attributes []attribute.KeyValue) (context.Context, trace.Span, func()) {
+	spanCtx, span := otel.Tracer(tracerName, traceOpts...).Start(ctx, operationName, trace.WithAttributes(attributes...))
+	return spanCtx, span, func() {
+		span.End()
+	}
+}
```
damemi, line 114: are these functions necessary? They seem to just be wrapping the otel functions

damemi (commented):
thanks for your patience @harshanarayana, this is going to be really great to have.

my main questions right now are around the new context variables. ideally we don't have to keep track of additional contexts just for tracing, so that they can be passed around and capture spans from other functions and processes without much work.

I'm also not sure you need the new `tracing` package you added for wrappers around the otel functionality. Left some comments there too.

Thanks again for working on this!

pkg/tracing/tracing.go
```diff
@@ -0,0 +1,128 @@
...
+		}
+		resource, err := sdkresource.New(ctx, resourceOpts...)
+		if err != nil {
+			klog.ErrorS(err, "failed to create traceable resource")
+			return nil, err
+		}
+
+		spanProcessor := sdktrace.NewBatchSpanProcessor(exporter)
+		provider = sdktrace.NewTracerProvider(
+			sdktrace.WithSampler(sdktrace.AlwaysSample()),
```
damemi, line 86: This should be configurable, and probably `ParentBased` by default

harshanarayana:
@damemi  Thanks a lot for the detailed review. I will get these comments addressed in a day or two and update you once ready for review.

harshanarayana:
@damemi  PTAL when possible. Comments have been addressed.

a7i:
Would you please update the doc [here](https://github.com/kubernetes-sigs/descheduler/blob/master/docs/user-guide.md#cli-options)

I believe if you just run --help and paste the result there, it should be good enough. Unfortunately there's not an auto-gen command to do it.

harshanarayana:
> Would you please update the doc [here](https://github.com/kubernetes-sigs/descheduler/blob/master/docs/user-guide.md#cli-options)
> 
> I believe if you just run --help and paste the result there, it should be good enough. Unfortunately there's not an auto-gen command to do it.

Sure..

pkg/tracing/tracing.go
Dentrax: Actually, it's not. But pretty lifesaver since these funcs prevents code-duplication during the implementation. And provides more readable structure since we abstract some of the unnecessary parameters and function calls (i.e., Start(), WithAttributes(), etc.)

Dentrax:
> Would you please update the doc [here](https://github.com/kubernetes-sigs/descheduler/blob/master/docs/user-guide.md?rgh-link-date=2023-01-23T17%3A26%3A07Z#cli-options)

Done!

harshanarayana:
> Unfortunately there's not an auto-gen command to do it.

Not anymore. That capability has been enabled. I can split that into a different PR if that is easier to track ?

harshanarayana:
> > Unfortunately there's not an auto-gen command to do it.
> 
> Not anymore. That capability has been enabled. I can split that into a different PR if that is easier to track ?

Never mind. Doing that as a different PR itself.

pkg/tracing/tracing.go
```diff
@@ -0,0 +1,144 @@
...
+			klog.ErrorS(err, "failed to create an instance of the trace exporter")
+			return nil, err
+		}
+		if name == "" {
+			klog.V(5).InfoS("no name provided, using default service name for tracing", "name", DefaultServiceName)
+			name = DefaultServiceName
+		}
+		resourceOpts := []sdkresource.Option{sdkresource.WithAttributes(semconv.ServiceNameKey.String(name)), sdkresource.WithSchemaURL(semconv.SchemaURL)}
+		if namespace != "" {
+			klog.V(5).InfoS("no namespace provided, using default namespace")
```
a7i, line 91: this is a case where namespace is actually provided

cmd/descheduler/app/server.go
```diff
@@ -94,6 +96,15 @@ func NewDeschedulerCommand(out io.Writer) *cobra.Command {
 				return
 			}
 
+			tracerProvider, err := tracing.NewTracerProvider(s.Tracing.CollectorEndpoint, s.Tracing.TransportCert, s.Tracing.ServiceName, s.Tracing.ServiceNamespace, s.Tracing.SampleRate)
+			if err != nil {
+				klog.Fatalf("failed to setup tracing provider", "err", err)
```
a7i, line 101: should we really fail if we can't set up the tracing provider?

pkg/descheduler/evictions/evictions.go
```diff
@@ -124,7 +138,7 @@ func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod, opts EvictOptio
 			if pe.metricsEnabled {
 				metrics.PodsEvicted.With(map[string]string{"result": "maximum number of pods per node reached", "strategy": strategy, "namespace": pod.Namespace, "node": pod.Spec.NodeName}).Inc()
 			}
-			klog.ErrorS(fmt.Errorf("Maximum number of evicted pods per node reached"), "limit", *pe.maxPodsToEvictPerNode, "node", pod.Spec.NodeName)
+			klog.ErrorS(fmt.Errorf("Maximum number of evicted pods per node reached"), "Eviction limit reached", "limit", *pe.maxPodsToEvictPerNode, "node", pod.Spec.NodeName)
```
a7i, line 141: Why is this being added? It is not really an eviction (API) limit, but rather limits set in the policy by the user.

pkg/framework/plugins/nodeutilization/highnodeutilization.go
```diff
@@ -71,7 +72,14 @@ func (h *HighNodeUtilization) Name() string {
 }
 
 // Balance extension point implementation for the plugin
-func (h *HighNodeUtilization) Balance(ctx context.Context, nodes []*v1.Node) *framework.Status {
+func (h *HighNodeUtilization) Balance(ctx context.Context, nodes []*v1.Node) (status *framework.Status) {
+	ctx, span, spanCloser := tracing.StartSpan(ctx, tracing.BalanceOperation, h.Name())
```
a7i, line 76: I'm debating whether to have this one level up, so it doesn't have to be repeated in every plugin and if we add a new one we won't forget to add this.

If it's moved up, does it mean that we lose some context?

a7i (commented):
Good work! A few nits/comments/questions.

Other than that, LGTM once resolved.

pkg/tracing/tracing.go
damemi: This would make more sense if there was something we intentionally wanted to inject on every span, but besides that it's not really recommended to add your own wrapper package for otel

pkg/tracing/tracing.go
```diff
@@ -0,0 +1,144 @@
...
+	return
+}
+
+// StartSpan starts a new span with the given operation name and attributes.
+func StartSpan(ctx context.Context, tracerName, operationName string) (context.Context, trace.Span, func()) {
+	return StartSpanWithOptions(ctx, tracerName, []trace.TracerOption{}, operationName, []attribute.KeyValue{})
+}
+
+func StartSpanWithOptions(ctx context.Context, tracerName string, traceOpts []trace.TracerOption, operationName string, attributes []attribute.KeyValue) (context.Context, trace.Span, func()) {
+	spanCtx, span := otel.Tracer(tracerName, traceOpts...).Start(ctx, operationName, trace.WithAttributes(attributes...))
```
damemi, line 126: It looks like you are passing a different `tracerName` for each type of operation (ie, `descheduler_parent`, `balance`, `evict`) right? We can just use a single tracer for all of descheduler. Attributes like that should come from the spans themselves

We could actually store one tracer at startup and pass that variable through everywhere, maybe in the framework handle. By calling `otel.Tracer()` each time, you are making a new allocation (that's more a performance improvement though)

pkg/descheduler/evictions/evictions.go
```diff
@@ -113,6 +116,17 @@ type EvictOptions struct {
 // EvictPod evicts a pod while exercising eviction limits.
 // Returns true when the pod is evicted on the server side.
 func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod, opts EvictOptions) bool {
+	var err error
+	ctx, evictSpan, evictCloser := tracing.StartSpanWithOptions(ctx, tracing.EvictionOption, []trace.TracerOption{}, "EvictPod", []attribute.KeyValue{
+		{Key: "pod-name", Value: attribute.StringValue(pod.GetName())},
+		{Key: "pod-namespace", Value: attribute.StringValue(pod.GetNamespace())},
```
damemi, line 122: we could pass these simpler with the [attribute](https://pkg.go.dev/go.opentelemetry.io/otel/attribute#String) package helpers (example in k8s: https://github.com/kubernetes/kubernetes/blob/d29e3bd7aa4978f10a01b9dfdebb2a647634f8fe/staging/src/k8s.io/apiserver/pkg/storage/etcd3/store.go#L367-L371)

pkg/descheduler/descheduler.go
```diff
@@ -323,6 +324,9 @@ func RunDeschedulerStrategies(ctx context.Context, rs *options.DeschedulerServer
 			eventRecorder,
 		)
 
+		var parentCloser func()
+		ctx, _, parentCloser = tracing.StartSpan(ctx, "descheduler_parent", "deschedule")
+		defer parentCloser()
```
damemi, line 329: this context already has a `cancel()` function that is already deferred, so I don't think `parentCloser()` is necessary

damemi: also (commented below re:this), but we can just use 1 named tracer, so `descheduler_parent` could just be `descheduler`. It might be better to actually name our tracer something namespaced like `sigs.k8s.io/descheduler` but I'll double check that

damemi (commented):
a couple comments, thanks for your patience

harshanarayana:
@a7i @damemi Sorry for the delay and responding back to this. I will get this addressing your comments today and push the changes accordingly.

cmd/descheduler/app/server.go
harshanarayana: Good question. I am not entirely sure what would be the right thing to do. One thing we can do is, switch back to a NoOp provider if we run into error and log that switch back here.

cmd/descheduler/app/server.go
```diff
@@ -94,6 +96,15 @@ func NewDeschedulerCommand(out io.Writer) *cobra.Command {
...
+				return
+			}
+			ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
+			defer tracing.Shutdown(ctx, tracerProvider)
+			defer cancel()
```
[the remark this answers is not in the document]
harshanarayana: Done

pkg/descheduler/evictions/evictions.go
harshanarayana: This was added because the logger is not using proper klog arg counts for invoking them. The message has `<MISSING>` showing up because of the arg length. Nothing to do wiht the current PR but this was a "fix if you find something" kind :)

pkg/framework/plugins/nodeutilization/highnodeutilization.go
harshanarayana: That was the original idea. But @damemi  suggested that it would be fine to let the individual plugins handle this on their own. I am open to moving this to a different location if that is the need.

pkg/tracing/tracing.go
harshanarayana: > maybe in the framework handle.

Do you mean one plugin will get one Tracer with the name set for it ? Or for the entire runtime, there will be one tracer and every plugin will just use the same tracer ?

pkg/tracing/tracing.go
harshanarayana: I am fine with moving this out and using the native calls directly. Sure it won't be that much of an extra duplication.

pkg/descheduler/evictions/evictions.go
harshanarayana: This brings up an interesting thing. Instead of us wrting some of these handlers ourself. should we just use teh `component-base/tracing` directly for all the use cases ?

harshanarayana: But then again, they still don't support Certs and others on the trace endpoints
https://github.com/kubernetes/component-base/blob/ae9be4dda9dabe6503eff5dd6162f48247880eea/tracing/api/v1/types.go#L20

harshanarayana:
@a7i @damemi Thanks for your patience on this. I have been on again and off again due to some personal commitments. Getting back to this again. Left a few comments. Need help with those so that I can close the pending comments.

pkg/tracing/tracing.go
damemi: one tracer for the entire runtime

harshanarayana:
@damemi I've addressed most of the comments already. Let me tenses again and push changes.. I'm fine with moving this to 1.28

pkg/descheduler/descheduler.go
damemi: @harshanarayana does this sound good to you?

pkg/tracing/tracing.go
damemi: cc @harshanarayana

pkg/tracing/tracing.go
damemi: cc @harshanarayana

damemi:
@harshanarayana I bumped the comments that I noticed that still need to be fixed from when we left this off

pkg/framework/plugins/nodeutilization/highnodeutilization.go
damemi: We could just do this trace around the calls to the plugins, rather than in the plugin itself (I think that's what you mean by `have this one level up` @a7i right?) I didn't mean to imply that each plugin *has* to handle this on their own, sorry if that was confusing.

That would make this PR a lot smaller, much easier to review for a first pass. And we can always add spans inside the plugins later if we want.

a7i:
/remove-lifecycle stale

@harshanarayana apologies for the delay on this review. would be good to include this as part of next release. would you be able to address the issues above? I can also address if you are ok with that.

harshanarayana:
@a7i  I will take care of this. I might just close and re-create a new PR entirely with a clean slate and keep a reference for this to track the history of discussion. I will have a PR ready for you in a day or two.

harshanarayana:
Closing this in favor of #1189

Issue #960, opened 2022-09-26
includeSoftConstraints configuration param not being respected for TopologySpreadConstraint


vholer:
**What version of descheduler are you using?**

descheduler version: 0.25.0

**Does this issue reproduce with the latest release?**

yes

**Which descheduler CLI options are you using?**

`--policy-config-file /policy-dir/policy.yaml --descheduling-interval 30s --v 5`

**Please provide a copy of your descheduler policy config file**

```
apiVersion: "descheduler/v1alpha1"
kind: "DeschedulerPolicy"
strategies:
  LowNodeUtilization:
    enabled: false
  PodLifeTime:
    enabled: false
  RemoveDuplicates:
    enabled: false
  RemovePodsHavingTooManyRestarts:
    enabled: false
  RemovePodsViolatingInterPodAntiAffinity:
    enabled: false
  RemovePodsViolatingNodeAffinity:
    enabled: false
  RemovePodsViolatingNodeTaints:
    enabled: false
  RemovePodsViolatingTopologySpreadConstraint:
    enabled: true
    params:
      includeSoftConstraints: true
      nodeFit: true
```

**What k8s version are you using (`kubectl version`)?**

<details><summary><code>kubectl version</code> Output</summary><br><pre>
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"24", GitVersion:"v1.24.4", GitCommit:"95ee5ab382d64cfe6c28967f36b53970b8374491", GitTreeState:"archive", BuildDate:"2022-08-18T00:00:00Z", GoVersion:"go1.18.5", Compiler:"gc", Platform:"linux/amd64"}
Kustomize Version: v4.5.4
Server Version: version.Info{Major:"1", Minor:"24", GitVersion:"v1.24.3", GitCommit:"aef86a93758dc3cb2c658dd9657ab4ad4afc21cb", GitTreeState:"clean", BuildDate:"2022-07-13T14:23:26Z", GoVersion:"go1.18.3", Compiler:"gc", Platform:"linux/amd64"}
</pre></details>

**What did you do?**

I have deployed 6 pods across 2 nodes with topologySpreadConstraints:

```
      topologySpreadConstraints:
        - maxSkew: 1
          topologyKey: kubernetes.io/hostname
          whenUnsatisfiable: ScheduleAnyway
          labelSelector:
            matchLabels:
              role: app
              size: small
```

The pods are deployed/balanced nicely:

```
$ kubectl get nodes
[...]

PR #961, opened 2022-09-26
includeSoftConstraints not being respected for TopologySpreadConstraint

vholer:
Issue #960

Signed-off-by: Vlastimil Holer <vh@fortrabbit.com>


damemi (commented):
/ok-to-test
@ingvagabund or @knelasevero could we add an e2e for this?

ingvagabund:
> could we add an e2e for this?

The affected code will get removed eventually. The code under strategy_migration.go file will get re-implemented. So any sort of testing will be only temporary. I will pass this.

a7i:
@damemi @ingvagabund Unfortunately our e2e tests are calling the strategy directly (and therefore bypassing the strategy migration). In an ideal world, e2e would be end-to-end testing that would catch it.

I'll think about this some more on how we can modify our e2e but if you have any ideas, let me know.

PR #964, opened 2022-09-27
Update Helm chart to v0.25.1

damemi:


damemi (commented):
/cc @ingvagabund @knelasevero 
After this merges, we'll promote the image and publish the new release

Issue #391, opened 2020-09-02
Add 'namespace' parameter support to LowNodeUtilization strategy


korjek:
Hello.

I think that would be great if 'namespace' parameter is supported with LowNodeUtilization strategy.

seanmalloy:
@korjek thanks for opening this issue. I'm assuming you are referring to the [namespace filtering](https://github.com/kubernetes-sigs/descheduler#namespace-filtering) feature that was just added in the v0.19.0 release.

When the namespace filtering enhancement was initially discussed it was determined that it would not make sense to enable this for all descheduler strategies. For example for the `LowNodeUtilization` strategy to be effective it needs to be able to process as many nodes as possible.

Can you provide any additional details on your use case?

seanmalloy:
@korjek are the pods running in this namespace running on a different set of nodes? Could you maybe filter using a node selector?

korjek:
@seanmalloy no, they are using the same nodes pool. 

But what's the reason to do not have namespace filtering for LowNodeUtilization strategy?

damemi:
LowNodeUtilization needs to consider as many pods as possible to accurately calculate each node's resource consumption. Only considering a subset of these for eviction would hurt its effectiveness at balancing cluster resource usage.

Can you give some more details on what your use case is? Perhaps there is another approach with the scheduler/descheduler that would fit better

korjek:
@damemi but why the descheduler needs to calculate node's resource consumptions if that info is available from node's status?

How I see this: when LowNodeUtilization strategy generates a list of pods for eviction it should take into account namespace filtering (ie when evaluating pods for eviction, skip pods that are filtered by namespace).  The rest of the logic shouldn't be changed.

Here is my usecase https://github.com/kubernetes-sigs/descheduler/issues/391#issuecomment-687212687

damemi:
If you limit the pods that are available for eviction with LowNodeUtilization, the strategy won't work very well. The point of it is to balance entire node usage, so unless the namespace you're restricting it to makes up a large portion of your cluster's consumption then there isn't much point to running this strategy in this namespace.

By your use case, I meant why do you specifically want to run LowNodeUtilization in one namespace? I understand that you want to restrict evictions to certain idempotent pods, but are you just trying to keep these pods spread out between nodes? Are they actually a large portion of your node's usage?

[...]

PR #967, opened 2022-09-28
add namespace filter to nodeutilization

### LINKED ISSUE #966: preEvictionFilter also filtering namespaces
Follow up from https://github.com/kubernetes-sigs/descheduler/issues/924, so users can also filter namespaces in strategies like LowNodeUtilization.

knelasevero:
Closes #966

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,6 +243,7 @@ actual usage metrics. Implementing metrics-based descheduling is currently TODO 
 |`thresholdPriority`|int (see [priority filtering](#priority-filtering))|
 |`thresholdPriorityClassName`|string (see [priority filtering](#priority-filtering))|
 |`nodeFit`|bool (see [node fit filtering](#node-fit-filtering))|
+|`evictableNamespaces`|(see [namespace filtering](#namespace-filtering))|
 
 **Example:**
 
@@ -315,6 +316,7 @@ actual usage metrics. Implementing metrics-based descheduling is currently TODO 
 |`thresholdPriority`|int (see [priority filtering](#priority-filtering))|
 |`thresholdPriorityClassName`|string (see [priority filtering](#priority-filtering))|
 |`nodeFit`|bool (see [node fit filtering](#node-fit-filtering))|
+|`evictableNamespaces`|(see [namespace filtering](#namespace-filtering))|
 
 **Example:**
 
@@ -613,6 +615,7 @@ The following strategies accept a `namespaces` parameter which allows to specify
 * `RemoveDuplicates`
 * `RemovePodsViolatingTopologySpreadConstraint`
 * `RemoveFailedPods`
+* `LowNodeUtilization` and `HighNodeUtilization` (Only filtered right before eviction, parameter named `evictableNamespaces`)
 
 For example:
 
--- a/pkg/framework/plugins/nodeutilization/nodeutilization.go
+++ b/pkg/framework/plugins/nodeutilization/nodeutilization.go
@@ -24,6 +24,7 @@ import (
 
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/descheduler/pkg/descheduler/evictions"
 	"sigs.k8s.io/descheduler/pkg/descheduler/node"
@@ -212,6 +213,7 @@ func classifyNodes(
 // TODO: @ravig Break this function into smaller functions.
 func evictPodsFromSourceNodes(
 	ctx context.Context,
+	evictableNamespaces *api.Namespaces,
 	sourceNodes, destinationNodes []NodeInfo,
 	podEvictor framework.Evictor,
 	podFilter func(pod *v1.Pod) bool,
@@ -265,13 +267,14 @@ func evictPodsFromSourceNodes(
 		klog.V(1).InfoS("Evicting pods based on priority, if they have same priority, they'll be evicted based on QoS tiers")
 		// sort the evictable Pods based on priority. This also sorts them based on QoS. If there are multiple pods with same priority, they are sorted based on QoS tiers.
 		podutil.SortPodsBasedOnPriorityLowToHigh(removablePods)
-		evictPods(ctx, removablePods, node, totalAvailableUsage, taintsOfDestinationNodes, podEvictor, continueEviction)
+		evictPods(ctx, evictableNamespaces, removablePods, node, totalAvailableUsage, taintsOfDestinationNodes, podEvictor, continueEviction)
 
 	}
 }
 
 func evictPods(
 	ctx context.Context,
+	evictableNamespaces *api.Namespaces,
 	inputPods []*v1.Pod,
 	nodeInfo NodeInfo,
 	totalAvailableUsage map[v1.ResourceName]*resource.Quantity,
@@ -280,6 +283,11 @@ func evictPods(
 	continueEviction continueEvictionCond,
 ) {
 
+	var excludedNamespaces sets.String
+	if evictableNamespaces != nil {
+		excludedNamespaces = sets.NewString(evictableNamespaces.Exclude...)
+	}
+
 	if continueEviction(nodeInfo, totalAvailableUsage) {
 		for _, pod := range inputPods {
 			if !utils.PodToleratesTaints(pod, taintsOfLowNodes) {
@@ -287,7 +295,16 @@ func evictPods(
 				continue
 			}
 
-			if podEvictor.PreEvictionFilter(pod) {
+			preEvictionFilterWithOptions, err := podutil.NewOptions().
+				WithFilter(podEvictor.PreEvictionFilter).
+				WithoutNamespaces(excludedNamespaces).
+				BuildFilterFunc()
+			if err != nil {
+				klog.V(1).ErrorS(err, "could not build preEvictionFilter with namespace exclusion")
+				continue
+			}
+
+			if preEvictionFilterWithOptions(pod) {
 				if podEvictor.Evict(ctx, pod, evictions.EvictOptions{}) {
 					klog.V(3).InfoS("Evicted pods", "pod", klog.KObj(pod))
 
--- a/pkg/framework/plugins/nodeutilization/types.go
+++ b/pkg/framework/plugins/nodeutilization/types.go
@@ -28,6 +28,10 @@ type LowNodeUtilizationArgs struct {
 	Thresholds             api.ResourceThresholds
 	TargetThresholds       api.ResourceThresholds
 	NumberOfNodes          int
+	// Naming this one differently since namespaces are still
+	// considered while considering resoures used by pods
+	// but then filtered out before eviction
+	EvictableNamespaces *api.Namespaces
 }
 
 // +k8s:deepcopy-gen=true
@@ -38,4 +42,8 @@ type HighNodeUtilizationArgs struct {
 
 	Thresholds    api.ResourceThresholds
 	NumberOfNodes int
+	// Naming this one differently since namespaces are still
+	// considered while considering resoures used by pods
+	// but then filtered out before eviction
+	EvictableNamespaces *api.Namespaces
 }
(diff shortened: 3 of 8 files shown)


pkg/framework/plugins/nodeutilization/nodeutilization.go
```diff
@@ -280,14 +283,30 @@ func evictPods(
...
 				klog.V(3).InfoS("Skipping eviction for pod, doesn't tolerate node taint", "pod", klog.KObj(pod))
 				continue
 			}
 
-			if podEvictor.PreEvictionFilter(pod) {
+			preEvictionFilterWithOptions, err := podutil.NewOptions().
+				WithFilter(podEvictor.PreEvictionFilter).
+				WithNamespaces(includedNamespaces).
+				WithoutNamespaces(excludedNamespaces).
+				BuildFilterFunc()
```
knelasevero, line 303: Do we want to use these, or do we want to throw this functionality into the PreEvictionFilter directly?

damemi:
We chose not to allow namespace filtering for NodeUtilization because it ultimately only hinders the strategy (see https://github.com/kubernetes-sigs/descheduler/issues/391 for a longer discussion on it)

tl;dr is that NodeUtilization is designed to balance resource consumption for an entire node. The idea of balanced resources across specific namespaces doesn't have a clear use case. Most of the use cases that seem like they fit are actually better served by a different strategy (like topology spreading).

Just curious is there more detail on where this is coming from for these strategies?

knelasevero:
> We chose not to allow namespace filtering for NodeUtilization because it ultimately only hinders the strategy (see #391 for a longer discussion on it)
> 
> tl;dr is that NodeUtilization is designed to balance resource consumption for an entire node. The idea of balanced resources across specific namespaces doesn't have a clear use case, and we can't guarantee it with this strategy anyway. Most of the use cases that seem like they fit are actually better served by a different strategy (like topology spreading).
> 
> Just curious is there more detail on where this is coming from for these strategies?

In the discussion https://github.com/kubernetes-sigs/descheduler/issues/391#issuecomment-790554717

I saw this message:

> +1 for allowing to set a list of excluded namespaces only yet clearly expressing risks of misusing it. Also stressing the setting applies in eviction phase only.

So, for NodeUtilization all namespaces need to be considered for balancing, and here in this PR we are only deciding not to evict a pod from an excluded namespace right before the eviction call (still considering all namespaces while processing the balancing).

Internally we had a client requesting this because they have some pods in kube-system namespace that not necessarily have system critical priority class, and because of that, they want to enforce exclusion explicitly in the descheduler config.

damemi:
> So, for NodeUtilization all namespaces need to be considered for balancing, and here in this PR we are only deciding not to evict a pod from an excluded namespace right before the eviction call (still considering all namespaces while processing the balancing).

I think this will still create some confusing situations where the descheduler is reporting over/underutilized nodes and not doing anything about it because the namespaces are excluded.

@ingvagabund's point of clearly explaining the risks helps with that, but it won't stop people from misusing it. My concern is that the downsides to this outweigh the potential benefits

ingvagabund:
> I think this will still create some confusing situations where the descheduler is reporting over/underutilized nodes and not doing anything about it because the namespaces are excluded.

The same situation can already occur when the over/underutilized nodes are full of pods with system critical priority class set. Or pods with local storage or pvc.

> but it won't stop people from misusing it. My concern is that the downsides to this outweigh the potential benefits

This is one of those design decisions which are not easy to make. We need a mechanism to protect namespaces but to also guarantee sufficient degree of balancing nodes. Two goals going against each other under certain condition.

We might hardcode a protection mechanism calculating whether pods from excluded namespaces occupy "too many" nodes. If true, the strategy would refuse any eviction until the pods are re-distributed.

knelasevero:
> We might hardcode a protection mechanism calculating whether pods from excluded namespaces occupy "too many" nodes. In which case the strategy would refuse any eviction until the pods are re-distributed.

Hmm :thinking: I personally think this would be even more confusing. 

I think that if you explicitly choose a namespace to be ignored, and you are using a strategy that re-balances pods, you would expect the behavior that we described (where the descheduler is reporting over/underutilized nodes and not doing anything about it because the namespaces are excluded). It is something that if well documented would also be easy to respond to if users get confused (maybe not only docs, but also reporting things in lower log levels of these re-balancing strategies, maybe something like "we found pods that should be evicted to proceed with rebalancing, but they are system-critical or part of a filtered namespace").

pkg/framework/plugins/nodeutilization/nodeutilization.go
ingvagabund: Into the PreEvictionFilter directly

ingvagabund:
@damemi I would like to get agreement on this. Either get this moving or decide to look for a different approach for namespace inclusion/exclusion. Alternative is to allow to specify only a list of excluded protected namespaces. Users will need to explicitly and willingly enumerate a list of excluded namespaces.

damemi:
Overall, I'm personally against this because it doesn't have a clear use case and is designed to detriment an existing plugin. I think doing this will encourage users to use lownodeutilization when their use case is actually better fit for something like pod topology spread. Adding a big warning not to use this just adds confusion (why add a feature that we're warning most people not to use?)

Having voiced my concerns, I do see that I'm in the minority on this so it's unfair of me to block it. So, thank you for hearing me out

> Alternative is to allow to specify only a list of excluded protected namespaces. Users will need to explicitly and willingly enumerate a list of excluded namespaces.

I would prefer this. I'm thinking most clusters will have a large number of overall namespaces with a smaller fraction of high-priority namespaces which they want to exclude. The alternative, where there is only a small fraction of namespaces to *include*, is much worse for the strategy's effectiveness. So discouraging that is probably best

knelasevero:
@damemi @ingvagabund do we mean disallowing the Include field of namespaces? Or using another struct for this?

ingvagabund:
yes, making use of `included` invalid/forbidden in the preEvictionFilter EP

damemi:
> yes, making use of `included` invalid/forbidden in the preEvictionFilter EP

to clarify, we are only talking about NodeUtilization strategies right? Others can still use include

If making include available to NodeUtilization is a technical requirement for it to be available to all strategies, then we don't have much choice there besides adding it to this strategy's validation

knelasevero:
Can't  we use nodeutilization validation (https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/framework/plugins/nodeutilization/validation.go) to disallow this only for this plugin like @damemi suggested? We don't have to change anything on other plugins.

damemi:
> As providers of the upstream plugins with default setting and validation for all "approved" plugins we can make sure the namespace inclusion/exclusion is limited for specific strategies. However, we can not guarantee the same quality of protection will be enforced globally for custom plugins.

This is totally fine for me, we're the code owners for this plugin, so it's up to us what options to allow. Having Include/Exclude be available for every plugin makes sense (the framework is providing the mechanism to use this feature), but plugins like this one should choose to disallow it.

Do you think that's a reasonable flow for developers?

knelasevero:
> I mean the expectation is to combine the preEvictionFilter with the namespace inclusion/exclusion. Namespace inclusion/exclusion is set by each strategy explicitly. Though, there's no way of detecting whether a custom plugin implementing the preEvictionFilter extension point is performing its own namespace inclusion/exclusion check. As providers of the upstream plugins with default setting and validation for all "approved" plugins we can make sure the namespace inclusion/exclusion is limited for specific strategies. However, we can not guarantee the same quality of protection will be enforced globally for custom plugins.

I think that's fine, we can have guidelines for people creating custom plugins, and also we should be able to review/accept new plugins at a "registry" repo like it is done for the kube-scheduler. We just need to make sure this is part of the review process.

> and validation for all "approved" plugins we can make sure the namespace inclusion/exclusion is limited for specific strategies

> same quality of protection

So, about `making sure`, and `quality of protection`, are we talking about the validation.go of those plugins still?

ingvagabund:
> but plugins like this one should choose to disallow it.

A plugin only sees the preEvictionFilter extension point. It does not see what's invoked inside. So the only explicit mechanism for disallowing the node utilization plugins to get the node inclusion is to enforce the `DefaultEvictor` plugin for the `preEvictionFilter` extension point during the profile validation. Which is also acceptable.

ingvagabund:
> So, about making sure, and quality of protection, are we talking about the validation.go of those plugins still?

There's no way to validate the behavior which we do not control/own unless we enforce use of specific extension points for specific plugins. Which might be too constraining for users. In which point we can introduce a new descheduler option saying `--disable-strict-validation` or similar.

damemi:
Maybe it would be good to see a poc change with that kind of enforcement applied? I'm getting a little confused so seeing the alternative implementation could help inform a better decision. @ingvagabund would you mind throwing together a rough approach to show what you mean?

knelasevero:
I just noticed that having v1alpha2 and the changes to policyconfig.go of https://github.com/kubernetes-sigs/descheduler/pull/974 would be kind of necessary here to work on that :sweat:. Since this is validation at the profile level and not at the plugin level.

ingvagabund:
@damemi locking a profile for the node utilization strategies. E.g. whenever a profile configures LowNodeUtilization plugin, the validation would make sure `preEvictionFilter` is always set to `enabled = [DefaultEvictor]`. To avoid enabling another preEvictionFilter extension point of a different plugin which would include/exclude namespaces.
```yaml
apiVersion: descheduler/v1alpha2
kind: DeschedulerPolicy
profiles:
- name: ProfileName
  pluginConfig:
  - name: DefaultEvictor
    args:
      ...
  - name: LowNodeUtilization
    args:
      ...
  plugins:
    balance:
      enabled:
      - LowNodeUtilization
    preEvictionFilter:
      enabled:
      - DefaultEvictor
```

damemi:
@ingvagabund ah, I see. That seems like a reasonable approach for plugin developers to take imo

Instead of enforcing it against a specific evictor, could it enforce that the evictor's arguments don't use namespace filtering? So you don't get stuck with DefaultEvictor

knelasevero:
So, as per latest discussions of the last community meeting, should we merge this one? I added docs and that simple validation.

I wanted to land https://github.com/kubernetes-sigs/descheduler/pull/974 before merging this one, but since I am kind of stuck with decoding/conversions there, I wanted to push this one forward already.

README.md
```diff
@@ -243,6 +243,7 @@ actual usage metrics. Implementing metrics-based descheduling is currently TODO
 |`thresholdPriority`|int (see [priority filtering](#priority-filtering))|
 |`thresholdPriorityClassName`|string (see [priority filtering](#priority-filtering))|
 |`nodeFit`|bool (see [node fit filtering](#node-fit-filtering))|
+|`evictionNamespaces`|(see [namespace filtering](#namespace-filtering))|
```
ingvagabund, line 246: evictableNamespaces

pkg/framework/plugins/nodeutilization/types.go
```diff
@@ -28,6 +28,10 @@ type LowNodeUtilizationArgs struct {
 	Thresholds             api.ResourceThresholds
 	TargetThresholds       api.ResourceThresholds
 	NumberOfNodes          int
+	// Naming this one differently since namespaces are still
+	// considered while considering resoures used by pods
+	// but then filtered out before eviction
+	EvictionNamespaces *api.Namespaces
```
ingvagabund, line 34: EvictableNamespaces

pkg/framework/plugins/nodeutilization/types.go
```diff
@@ -38,4 +42,8 @@ type HighNodeUtilizationArgs struct {
 
 	Thresholds    api.ResourceThresholds
 	NumberOfNodes int
+	// Naming this one differently since namespaces are still
+	// considered while considering resoures used by pods
+	// but then filtered out before eviction
+	EvictionNamespaces *api.Namespaces
```
ingvagabund, line 48: EvictableNamespaces

pkg/framework/plugins/nodeutilization/nodeutilization.go
ingvagabund: The namespace is a parameter of the strategies so it needs to stay outside of PreEvictionFilter extension point

ingvagabund (changes requested):

ingvagabund:
Few nits. Looks good in overall.

knelasevero:
> The namespace is a parameter of the strategies so it needs to stay outside of PreEvictionFilter extension point

I couldn't understand this one.

ingvagabund:
Related to

> Do we want to use these, or do we want to throw this functionality into the PreEvictionFilter directly?

The former.

ingvagabund:
@knelasevero there are still few mentions of the `evictionNamespaces` and https://github.com/kubernetes-sigs/descheduler/pull/967#discussion_r1003104994. Otherwise lgtm.

PR #968, opened 2022-09-29
Pod security context fix for Openshift

knelasevero:
Related to the Downstreamed descheduler and openshift specific requirements. But adding Security contexts here would not hurt, and avoids conflicts for us when downstreaming changes there: https://github.com/openshift/descheduler/pull/80

### DIFF
--- a/test/e2e/e2e_duplicatepods_test.go
+++ b/test/e2e/e2e_duplicatepods_test.go
@@ -38,6 +38,8 @@ import (
 )
 
 func TestRemoveDuplicates(t *testing.T) {
+	runAsUser := true
+	ape := false
 	ctx := context.Background()
 
 	clientSet, sharedInformerFactory, _, getPodsAssignedToNode, stopCh := initializeClient(t)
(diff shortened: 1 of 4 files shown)


a7i:
@madeelibm would you please sign the CNCF CLA?

damemi (commented):
I don't think it is a great idea to upstream vendor-specific changes just for convenience. If this is something that's only needed for OpenShift's tests, then I think a carry patch in the downstream fork is the right approach

knelasevero:
> I don't think it is a great idea to upstream vendor-specific changes just for convenience. If this is something that's only needed for OpenShift's tests, then I think a carry patch in the downstream fork is the right approach

> @knelasevero hopefully those changes can be made compatible for both Kubernetes and OpenShift

But I think @damemi is also against that? It makes the pod spec in tests more verbose and it is only a requirement for k8s flavors where security context are required (Namely only Openshift I think).

@damemi That is the rationale, right?

ingvagabund:
This exercise is only relevant when the same change is compatible with both Kubernetes and OpenShift. If not, we will carry downstream patch.

knelasevero:
> I admit I'm not an expert in security contexts, so I'm just commenting from an oss point of view. But if this is a valid improvement for users besides OpenShift's CI, then it's fine by me

It is best practice to define security context, and that's why this is enforced in Openshift, but, tbh, I don't think it brings a lot of value to pods in e2e tests (with some exceptions at the end of this comment). 

The downsides that I can see here are:

- verbosity of declaring the pods in tests
- this is yet another thing that can fail in tests, that are not related to what we want to test. (if we change how the image is built and for some reason if invalidates the security context and it fails for that we would have to be aware and also change things in tests)

The pros are:

- e2e tests closer to what people should run in real life [?]
- easier to downstream when this is a requirement
- this is safer for people wanting to re-run e2e tests themselves in their environments, in their CIs, if they re-build things (maybe gets more relevant when we think about people using the framework to build their own deschedulers)

madeelibm:
> @madeelibm would you please sign the CNCF CLA?

I have started the process, it will take some time.

knelasevero:
Closing so original Author can take care :)

#975

PR #969, opened 2022-09-30
run: `gofumpt -w -extra .`

Dentrax:
Enable gofumpt in golangci

Signed-off-by: Furkan <furkan.turkal@trendyol.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,6 +29,9 @@ LDFLAGS=-ldflags "-X ${LDFLAG_LOCATION}.version=${VERSION} -X ${LDFLAG_LOCATION}
 GOLANGCI_VERSION := v1.49.0
 HAS_GOLANGCI := $(shell ls _output/bin/golangci-lint 2> /dev/null)
 
+GOFUMPT_VERSION := v0.4.0
+HAS_GOFUMPT := $(shell command -v gofumpt 2> /dev/null)
+
 # REGISTRY is the container registry to push
 # into. The default is to push to the staging
 # registry, not production.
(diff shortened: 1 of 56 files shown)


Dentrax:
> lgtm, but could we add it to the makefile so people can easily update when their PRs fail?

Make sense! Added `make fmt` and updated the doc.

Makefile
```diff
@@ -137,6 +140,12 @@ ifndef HAS_GOLANGCI
 endif
 	./_output/bin/golangci-lint run
 
+fmt:
+ifndef HAS_GOFUMPT
```
a7i, line 144: thoughts on adding it to the script used by `verify-gofmt`?

Dentrax: actually golangci-lint already verifies it. should we also add a secondary check anyway?

a7i: I don't think that's necessary. @damemi ?

damemi: It looks like `make verify` already runs `lint`, so we don't need a second check then. As long as there is a local way to verify that this has been updated

Dentrax: please also notice that this `make fmt` is to run gofumpt formatter on every `**/*.go` file. (excludes ./vendor by default)

a7i: @damemi are you suggesting that we should remove `make fmt` since `lint` already runs this?

a7i:
@Dentrax sorry for the delay on this PR. would you please rebase so we can move forward on the review?

Dentrax:
Rebased.

damemi (commented):
/approve

Issue #970, opened 2022-10-05
Namespace fillter is not shown as a reason the pod shouldn't be evicted


migueleliasweb:
**What version of descheduler are you using?**

descheduler version: v0.25.1

**Does this issue reproduce with the latest release?**

Yes

**Which descheduler CLI options are you using?**

```
  - args:
    - --policy-config-file
    - /policy-dir/policy.yaml
    - --descheduling-interval
    - 5m
    - --v
    - "4"
    command:
    - /bin/descheduler
```

**Please provide a copy of your descheduler policy config file**

```yaml
apiVersion: "descheduler/v1alpha1"
kind: "DeschedulerPolicy"
evictLocalStoragePods: true
maxNoOfPodsToEvictPerNode: 1
strategies:
  "PodLifeTime":
    enabled: true
    params:
      podLifeTime:
        maxPodLifeTimeSeconds: 3600
      labelSelector:
        matchLabels:
          app: foo
      namespaces:
        include:
        - foo
```

**What k8s version are you using (`kubectl version`)?**

<details><summary><code>kubectl version</code> Output</summary><br><pre>
$ kubectl version

Client: v1.23
Server: v1.22

</pre></details>

**What did you do?**

Deployed the descheduler with the above policy + use verbosity lv 4.

**What did you expect to see?**

The default evictor seems to be unaware of the namespace filter thus is doesn't mention it in the checks list.

See: https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/framework/plugins/defaultevictor/defaultevictor.go#L166

Also, the pod lifetime plugin wouldn't be able to output that information as it filters for the namespace early on: https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/framework/plugins/podlifetime/pod_lifetime.go#L61

```bash
#from the logs
defaultevictor.go:188] "Pod fails the following checks" pod="kube-system/foobar" checks="[pod is a DaemonSet pod, pod has system critical priority, pod has higher priority than specified priority class threshold, po
d has local storage and descheduler is not configured with evictLocalStoragePods]"
```

**What did you see instead?**

Something mentioning the pod wouldn't be evicted due to it's namespace didn't match the filtering requirements.

damemi:
[...]

PR #976, opened 2022-10-07
Add error output for namespace filters

migueleliasweb:
Related to: https://github.com/kubernetes-sigs/descheduler/issues/970

- Added log information for pods skipping eviction due to namespace filtering (include or exclude).
- Added unittests

As mentioned in the #970 issue, there are at least a couple of ways to achieve this but changing the the `ListPodsOnANode` seemed too much as it would require more extensive changes. The benefit of the approach I took is that all pods are shown in the logs which makes it easier to troubleshoot any missed pods due to filtering or other configs.

Please let me know if further changes are required. :+1:

### DIFF
--- a/pkg/framework/plugins/defaultevictor/defaultevictor.go
+++ b/pkg/framework/plugins/defaultevictor/defaultevictor.go
@@ -22,6 +22,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/klog/v2"
 	nodeutil "sigs.k8s.io/descheduler/pkg/descheduler/node"
 	podutil "sigs.k8s.io/descheduler/pkg/descheduler/pod"
@@ -71,6 +72,32 @@ func New(args runtime.Object, handle framework.Handle) (framework.Plugin, error)
 	ev.handle = handle
 	ev.args = defaultEvictorArgs
 
+	ev.constraints = append(ev.constraints, func(pod *v1.Pod) error {
+		if defaultEvictorArgs.StrategyParameters != nil &&
+			defaultEvictorArgs.StrategyParameters.Namespaces != nil &&
+			defaultEvictorArgs.StrategyParameters.Namespaces.Include != nil {
+			if !sets.NewString(defaultEvictorArgs.StrategyParameters.Namespaces.Include...).Has(pod.Namespace) {
+				return fmt.Errorf(
+					"pod is not present on included namespaces: %s",
+					defaultEvictorArgs.StrategyParameters.Namespaces.Include,
+				)
+			}
+		}
+		return nil
+	})
+	ev.constraints = append(ev.constraints, func(pod *v1.Pod) error {
+		if defaultEvictorArgs.StrategyParameters != nil &&
+			defaultEvictorArgs.StrategyParameters.Namespaces != nil &&
+			defaultEvictorArgs.StrategyParameters.Namespaces.Exclude != nil {
+			if sets.NewString(defaultEvictorArgs.StrategyParameters.Namespaces.Exclude...).Has(pod.Namespace) {
+				return fmt.Errorf(
+					"pod is present on excluded namespaces: %s",
+					defaultEvictorArgs.StrategyParameters.Namespaces.Exclude,
+				)
+			}
+		}
+		return nil
+	})
 	if defaultEvictorArgs.EvictFailedBarePods {
 		klog.V(1).InfoS("Warning: EvictFailedBarePods is set to True. This could cause eviction of pods without ownerReferences.")
 		ev.constraints = append(ev.constraints, func(pod *v1.Pod) error {
@@ -151,7 +178,7 @@ func (d *DefaultEvictor) PreEvictionFilter(pod *v1.Pod) bool {
 	if defaultEvictorArgs.NodeFit {
 		nodes, err := nodeutil.ReadyNodes(context.TODO(), d.handle.ClientSet(), d.handle.SharedInformerFactory().Core().V1().Nodes().Lister(), defaultEvictorArgs.NodeSelector)
 		if err != nil {
-			klog.V(1).ErrorS(fmt.Errorf("Pod fails the following checks"), "pod", klog.KObj(pod))
+			klog.V(1).ErrorS(fmt.Errorf("pod fails the following checks"), "pod", klog.KObj(pod))
 			return false
 		}
 		if !nodeutil.PodFitsAnyOtherNode(d.handle.GetPodsAssignedToNodeFunc(), pod, nodes) {
--- a/pkg/descheduler/descheduler.go
+++ b/pkg/descheduler/descheduler.go
@@ -388,6 +388,7 @@ func RunDeschedulerStrategies(ctx context.Context, rs *options.DeschedulerServer
 						PriorityThreshold: &api.PriorityThreshold{
 							Value: &thresholdPriority,
 						},
+						StrategyParameters: params,
 					}
 
 					evictorFilter, _ := defaultevictor.New(
--- a/pkg/framework/plugins/defaultevictor/defaultevictor_test.go
+++ b/pkg/framework/plugins/defaultevictor/defaultevictor_test.go
@@ -374,6 +374,7 @@ func TestDefaultEvictorFilter(t *testing.T) {
 		priorityThreshold       *int32
 		nodeFit                 bool
 		result                  bool
+		strategyParameters      *api.StrategyParameters
 	}
 
 	testCases := []testCase{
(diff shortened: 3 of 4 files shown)


damemi (commented):
/ok-to-test

migueleliasweb:
```
$ ./hack/update-generated-deep-copies.sh
$ git status
On branch output-errors-for-namespace-filters
Untracked files:
  (use "git add <file>..." to include in what will be committed)
        k8s.io/
        sigs.k8s.io/
```

Running the script the output says to run creates these folders but I wasn't expect the change to be this big. Did I do something wrong?

```
$ git status -u                                                                              
On branch output-errors-for-namespace-filters                                                                                                     
Untracked files:                                                                                                                                  
  (use "git add <file>..." to include in what will be committed)                                                                                  
        k8s.io/api/admission/v1/zz_generated.deepcopy.go                                                                                          
        k8s.io/api/admission/v1beta1/zz_generated.deepcopy.go                                                                                     
        k8s.io/api/admissionregistration/v1/zz_generated.deepcopy.go                                                                              
        k8s.io/api/admissionregistration/v1beta1/zz_generated.deepcopy.go                                                                         
        k8s.io/api/apiserverinternal/v1alpha1/zz_generated.deepcopy.go                                                                            
        k8s.io/api/apps/v1/zz_generated.deepcopy.go                                                                                               
        k8s.io/api/apps/v1beta1/zz_generated.deepcopy.go                                                                                          
        k8s.io/api/apps/v1beta2/zz_generated.deepcopy.go                                                                                          
        k8s.io/api/authentication/v1/zz_generated.deepcopy.go                                                                                     
        k8s.io/api/authentication/v1beta1/zz_generated.deepcopy.go                                                                                
        k8s.io/api/authorization/v1/zz_generated.deepcopy.go                                                                                      
        k8s.io/api/authorization/v1beta1/zz_generated.deepcopy.go                                                                                 
        k8s.io/api/autoscaling/v1/zz_generated.deepcopy.go                                                                                        
        k8s.io/api/autoscaling/v2/zz_generated.deepcopy.go                                                                                        
        k8s.io/api/autoscaling/v2beta1/zz_generated.deepcopy.go                                                                                   
        ....

        # the list goes on for a total of 90 files generated, not changed (which is what I was expecting)
```

pkg/framework/plugins/defaultevictor/defaultevictor.go
```diff
@@ -71,6 +72,32 @@ func New(args runtime.Object, handle framework.Handle) (framework.Plugin, error)
 	ev.handle = handle
 	ev.args = defaultEvictorArgs
 
+	ev.constraints = append(ev.constraints, func(pod *v1.Pod) error {
```
ingvagabund, line 75: The default evictor is not responsible for filtering out pods. Only the performing checks whether a pod can be evicted. Printing any information about excluding/including pods from processing by a strategy needs to be printed by the filtering code.

ingvagabund: Also, assuming the case of a cluster with thousands of pods getting excluded, there would be thousands of log lines produced by each strategy saying "pod 'POD' excluded from eviction by 'filter' extension point".

ingvagabund: @migueleliasweb each strategy can have its own custom filters which may be excluding excessive number of pods. Even with namespace exclusion getting logged, there may be other unreported places of pods getting filter out.

ingvagabund: Filters are meant to be invoked multiple times. Whereas eviction only once. That's why it is practical to log reasons why a pod can not be evicted. Whereas filters are intended to reduce the number of pods getting processed by each strategy.

ingvagabund: Also, if a pod that's excluded/included based on a namespace gets here, something's wrong. The only exception arethe utilization strategies which currently do not perform the namespace inclusion/exclusion check.

migueleliasweb: Thanks for the feedback, @ingvagabund .

> Also, assuming the case of a cluster with thousands of pods getting excluded, there would be thousands of log lines produced by each strategy saying "pod 'POD' excluded from eviction by 'filter' extension point".

> Also, if a pod that's excluded/included based on a namespace gets here, something's wrong. The only exception are the utilization strategies which currently do not perform the namespace inclusion/exclusion check.

I reckon that might be already happening as for each 'cycle check' from the descheduler. At least that's what I'm seeing when I run it on my cluster that has around 1k pods running. As you can see, from the initial issue I oppened (https://github.com/kubernetes-sigs/descheduler/issues/970), there is a namespace constraint (redacted to `foo`) but pods from `kube-system` are shown in the logs. That means the `defaultevictor` is picking up **all pods** and outputting information about whether they shuold be evicted or not.

The core problem here is that there must be a balance between providing useful information regarding **all the pods** in the cluster and why they are or aren't being considered for eviction and the amount of log that is generated by the `Descheduler`. That's what I'm trying to solve here.

A orthogonal idea would be to provide a way to query the `Descheduler` about a specific POD and wether it matches any of the configured policies, instead of outputting tons of logs about all pods in every cycle.

ingvagabund: > As you can see, from the initial issue I oppened (https://github.com/kubernetes-sigs/descheduler/issues/970), there is a namespace constraint (redacted to foo) but pods from kube-system are shown in the logs. That means the defaultevictor is picking up all pods and outputting information about whether they shuold be evicted or not.

I am not able to reproduce this scenario with the following policy:
```
kind: "DeschedulerPolicy"
strategies:
  "PodLifeTime":
    enabled: true
    params:
      podLifeTime:
        maxPodLifeTimeSeconds: 10
      labelSelector:
        matchLabels:
          role: web
      namespaces:
        include:
        - foo
```

and following RS created in kube-system and foo namespaces:
```
apiVersion: apps/v1
kind: ReplicaSet
metadata:
  name: web
  labels:
    env: dev
    role: web
spec:
  replicas: 5
  selector:
    matchLabels:
      role: web
  template:
    metadata:
      labels:
        role: web
    spec:
      nodeSelectorTerms:
      topologySpreadConstraints:
      - maxSkew: 1
        topologyKey: "topology.kubernetes.io/zone"
        whenUnsatisfiable: DoNotSchedule
        labelSelector:
          matchLabels:
            role: web
      containers:
      - name: nginx
        image: gcr.io/google-samples/hello-app:2.0
```

Always getting only pods from foo NS evicted. Pods in the kube-system NS are untouched.

@migueleliasweb would you please provide exact steps on how to reproduce the situation with full descheduler logs?

ingvagabund: > Since I saw the defaultevictor was being used to provide debugging logs about the running pods, I thought I could just add the log to it.

The changes you performed in this PR are located in the `Filter` extension point which can be disabled if a user decides to provide a custom implementation of the extension point.

I am not opposed to adding more logs (with higher log level) printing all excluded pods. Nevertheless, the code changes needs to go into [BuildGetPodsAssignedToNodeFunc](https://github.com/kubernetes-sigs/descheduler/blob/c10e6d6c129b0db29415fb2043e4b8b07532ecab/pkg/descheduler/pod/pods.go#L151) where the individual pods are included. You will not be able to tell whether a pod was excluded due to namespace exclusion/inclusion. Though, you will be able to see a pod was filter out before it was processed by a strategy.

migueleliasweb: > Always getting only pods from foo NS evicted. Pods in the kube-system NS are untouched.

I think this might have been a misunderstanding. I was just saying the `descheduler` outputs logs from **all pods** but not necesarrily touching/evicting any of them. That's why I added the log lines to the same place (`defaultevictor`).

I will have a look at [BuildGetPodsAssignedToNodeFunc](https://github.com/kubernetes-sigs/descheduler/blob/c10e6d6c129b0db29415fb2043e4b8b07532ecab/pkg/descheduler/pod/pods.go#L151) to see how this could be implemented. It could be interesting to add an even higher level of verbosity for this.

migueleliasweb: Hummm, that's interesting. Yeah you're right. It can be called multiple times. I just noticed there are specific calls for each "plugin". I reckon if we add the plugin information to the option, we could then then output smarter log lines for specific plugins.

From https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/framework/plugins/podlifetime/pod_lifetime.go#L59
```
	WithFilter(podutil.WrapFilterFuncs(handle.Evictor().Filter, handle.Evictor().PreEvictionFilter)).
	WithNamespaces(includedNamespaces).
	WithoutNamespaces(excludedNamespaces).
	WithLabelSelector(podLifeTimeArgs.LabelSelector).
	WithPlugin(PluginName, plugin) # <-- this
	BuildFilterFunc()
if err != nil {
	return nil, fmt.Errorf("error initializing pod filter function: %v", err)
}
```

So the log line could contain something like: 

```
pod <foo> is excluded from namespace <bar> for plugin <taz>
```

With that, we wouldn't have to implement caches or any other fancy ways to get around the fact the `FilterFunc` is called multiple times per cycle.

ingvagabund: Even a single plugin can have multiple invocations of the filter.

migueleliasweb: Wouldn't that be fine? I guess we still want to output per pod.

ingvagabund: When the filter is run over the same pod multiple times, the filter will produce duplicated log lines. Also, the namespace filtering is as [third in the list](https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/descheduler/pod/pods.go#L101-L115). So when either `o.labelSelector` or `o.filter` (which is a combination of other filters) gives false, there's not going to be any log saying a pod was excluded from eviction because of the namespace filtering constraint.

If the overall goal is to improve visibility of the decision making to a point of getting complete information about why a pod was excluded from eviction, the `BuildFilterFunc` is not the suitable place for it.

PR #954, opened 2022-09-19
Plugin arguments defaulting + moving plugin arguments to each corresponding plugin #925

spike-liu:
Please feel free to review.

Issue #313, opened 2020-06-03
Hack scripts assume "sigs.k8s.io/descheduler" project root unless PRJ_PREFIX is set


damemi:
When running hack scripts (or `make gen`, which calls them) to generate conversions, etc, outside a project root that matches what is set in [`hack/lib/init.sh`](https://github.com/kubernetes-sigs/descheduler/blob/bd412bf87f72132efad2a07833666463cb909969/hack/lib/init.sh#L46) the scripts will silently fail, generating no output, with no indication why. This can be easily fixed by setting `PRJ_PREFIX`, but this is not documented anywhere. It would be even nicer if the scripts could identify when the project is located outside that default directory and output a message alerting the user that this may cause it to silently fail.

/kind documentation

damemi:
cc @pmundt thanks for helping find this!

pravarag:
@damemi @seanmalloy as mentioned here, I've started working on this issue and will raise a new PR for the same. So, below is what I've understood w.r.t this issue,
So let' say I've two clones of `kubernetes-sigs/descheduler` one is the upstream repo under (`$GOPATH/src/sigs.k8s.io/descheduler`) and another one is the working repo which could be under `$GOPATH/src/github.com/pravarag/descheduler`. Now when I run either `make gen` or `./hack/update-generated-*.sh` from my current working directory i.e. (`$GOPATH/src/github.com/pravarag/descheduler`), then it basically updates the packages under directory `$GOPATH/src/sigs.k8s.io/descheduler` ( as `PRJ_PREFIX` is set to that path). But, what we want to achieve is if `make gen` or any of `./hack/update-*.sh` scripts are run from outside the `$GOPATH/sigs.k8s.io/descheduler` path, then it should update the packages within the current working directory and not the packages under clone of upstream? Please correct me if I'm wrong here.

seanmalloy:
> @damemi @seanmalloy as mentioned here, I've started working on this issue and will raise a new PR for the same. So, below is what I've understood w.r.t this issue,
[...]

PR #982, opened 2022-10-13
Update verify scripts to use the generation scripts directly

damemi:
In https://github.com/kubernetes-sigs/descheduler/pull/958 we are noticing odd behavior with the `verify` tools for our generator scripts, specifically conversions. CI is [failing to verify the conversion-gen](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_descheduler/958/pull-descheduler-verify-master/1580431724137943040) saying there are changes missing. But, running these generators on some local environments (and containers) produces no diff, while other environments do produce the generated output that the tests are expecting.

https://github.com/kubernetes-sigs/descheduler/pull/954/commits/96c03a3f97b6440c1622a9e09606785fdbb4de87 and https://github.com/kubernetes-sigs/descheduler/pull/954/commits/2b1746cda8883c199329510ac021b1c6badf4984 update the generator scripts (deepcopies, conversion, and defaulters) to dynamically parse which files to run on using the function [`find_dirs_containing_comment_tags`](https://github.com/kubernetes-sigs/descheduler/blob/2b1746cda8883c199329510ac021b1c6badf4984/hack/lib/generator-help.sh#L15). But, it didn't update the verify scripts to parse similarly (they are still hardcoded).

Updating the verify scripts to call the generators directly is a good idea to keep our generators and changes in line. Doing so fixed the verify failures (at least for me locally). But, I don't know if this is actually fixing it, or if there is a bug in the new generator command with `find_dirs_containing_comment_tags` and this is just copying the bug over to the verify scripts as well. 

What looks suspicious to me is [this line](https://github.com/kubernetes-sigs/descheduler/blob/2b1746cda8883c199329510ac021b1c6badf4984/hack/lib/generator-help.sh#L22) which replaces `.` in the path names with `${PRJ_PREFIX}`. So on my local for example, the output of this function for conversion-gen is:
```
sigs.k8s.io/descheduler/pkg/api/v1alpha1,sigs.k8s.io/descheduler/pkg/apis/componentconfig/v1alpha1
```
as opposed to
```
./pkg/api/v1alpha1,./pkg/apis/componentconfig/v1alpha1
```

This stands out because `verify-conversions.sh` specifically [currently has the latter](https://github.com/kubernetes-sigs/descheduler/pull/982/files#diff-7318ba5bb4b500e18e147fd0db7ccb39d26c47690dd9ac70ba5d4bb7e6fcc917L23) as its input dirs. But, the generator is using the *former* as its input dirs.

The fact that this shows up differently on different environments makes me think it may be an issue with how `PRJ_PREFIX` is parsed. Maybe similar to https://github.com/kubernetes-sigs/descheduler/issues/313. Marking this WIP for now to continue work on it

See also a similar weird output from the deepcopy script in https://github.com/kubernetes-sigs/descheduler/pull/976#issuecomment-1272207052

### DIFF
--- a/hack/verify-conversions.sh
+++ b/hack/verify-conversions.sh
@@ -16,12 +16,7 @@ git archive --format=tar --prefix=descheduler/ "$(git write-tree)" | (cd "${_des
 _deschedulertmp="${_deschedulertmp}/descheduler"
 
 pushd "${_deschedulertmp}" > /dev/null 2>&1
-go build -o "${OS_OUTPUT_BINPATH}/conversion-gen" "k8s.io/code-generator/cmd/conversion-gen"
-
-${OS_OUTPUT_BINPATH}/conversion-gen \
-		--go-header-file "hack/boilerplate/boilerplate.go.txt" \
-		--input-dirs "./pkg/apis/componentconfig/v1alpha1,./pkg/api/v1alpha1" \
-		--output-file-base zz_generated.conversion
+./hack/update-generated-conversions.sh
 popd > /dev/null 2>&1
 
 pushd "${DESCHEDULER_ROOT}" > /dev/null 2>&1
--- a/hack/verify-deep-copies.sh
+++ b/hack/verify-deep-copies.sh
@@ -16,12 +16,7 @@ git archive --format=tar --prefix=descheduler/ "$(git write-tree)" | (cd "${_des
 _deschedulertmp="${_deschedulertmp}/descheduler"
 
 pushd "${_deschedulertmp}" > /dev/null 2>&1
-go build -o "${OS_OUTPUT_BINPATH}/deepcopy-gen" "k8s.io/code-generator/cmd/deepcopy-gen"
-
-${OS_OUTPUT_BINPATH}/deepcopy-gen \
-                --go-header-file "hack/boilerplate/boilerplate.go.txt" \
-                --input-dirs "./pkg/apis/componentconfig,./pkg/apis/componentconfig/v1alpha1,./pkg/api,./pkg/api/v1alpha1,./pkg/framework/plugins/defaultevictor/" \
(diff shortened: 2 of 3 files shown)


spike-liu:
> ping @spike-liu, are you able to check [`find_dirs_containing_comment_tags`](https://github.com/kubernetes-sigs/descheduler/blob/2b1746cda8883c199329510ac021b1c6badf4984/hack/lib/generator-help.sh#L15) to see if there is an issue there? It looks like the generator is giving some weird output in [#976 (comment)](https://github.com/kubernetes-sigs/descheduler/pull/976#issuecomment-1272207052) as well

Yes, of course. Let me do some research.

spike-liu:
@damemi Yes, you are right. Previously replacing . with ${PRJ_PREFIX} in find_dirs_containing_comment_tags is for the consistency with the original hardcode. Verified in local test, this line could be removed safely and verify-deep-copies.sh could be run successfully within this PR as below:
image

As for the issue in #976, I found some golang files under _tmp folder generated in verify-xxx.sh would be found mistakenly by find_dirs_containing_comment_tags(). However we could fix with excluding _tmp folders as below:
image

Please feel free to let me know if PR was needed.

damemi:
@spike-liu thanks for debugging this! yes, if you wouldn't mind opening a PR with your fixes it would be really helpful

spike-liu:
> @spike-liu thanks for debugging this! yes, if you wouldn't mind opening a PR with your fixes it would be really helpful

Done with https://github.com/kubernetes-sigs/descheduler/pull/988

Please feel free to review.

knelasevero:
Nice! This have been bugging me for some time. Great stuff

/lgtm

JaneLiuL (approved):
/lgtm

PR #988, opened 2022-10-17
Fix wired behavior during code generating

spike-liu:
- exclude files & folders under _tmp folder during code-generator;
- remove redundant path replacement to keep the consistent logic with verify-xx.sh;


spike-liu:
This PR try to fix code generating issues mentioned in https://github.com/kubernetes-sigs/descheduler/pull/982 and https://github.com/kubernetes-sigs/descheduler/pull/976

damemi (commented):
/ok-to-test

Issue #991, opened 2022-10-17
Can't change log level using --v cli option


poojasj:
**What version of descheduler are you using?**

descheduler version:
v0.24.1

**Which descheduler CLI options are you using?**
```  
          containers:
            - args:
              - --policy-config-file
              - /policy-dir/policy.yaml
              - --logging-format
              - json
              - --v
              - "4"
```
**Please provide a copy of your descheduler policy config file**

**What k8s version are you using (`kubectl version`)?**

<details><summary><code>kubectl version</code> Output</summary><br><pre>
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.4", GitCommit:"b695d79d4f967c403a96986f1750a35eb75e75f1", GitTreeState:"clean", BuildDate:"2021-11-17T15:48:33Z", GoVersion:"go1.16.10", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"21+", GitVersion:"v1.21.14-eks-6d3986b", GitCommit:"8877a3e28d597e1184c15e4b5d543d5dc36b083b", GitTreeState:"clean", BuildDate:"2022-07-20T22:05:32Z", GoVersion:"go1.16.15", Compiler:"gc", Platform:"linux/amd64"}
</pre></details>

**What did you do?**

 I was trying to get pod eviction logs of descheduler, so I increased the log level to 4. But that doesn't seem to work, there were only 3 log events in each descheduler pod logs.
 We recently upgraded to descheduler v0.24 and started using `--logging-format=json` option. So I tried removing `--logging-format=json` from the cronjob and I started seeing correct logs as per level set in the cli option.

**What did you expect to see?**
Info level descheduler logs, in json format

**What did you see instead?**
Logs:
<pre>
{"ts":1666025401338.376,"caller":"server/secure_serving.go:210","msg":"Serving securely on [::]:10258\n","v":0}
{"ts":1666025401338.4546,"caller":"dynamiccertificates/tlsconfig.go:240","msg":"Starting DynamicServingCertificateController","v":0}
{"ts":1666025405001.4187,"caller":"dynamiccertificates/tlsconfig.go:255","msg":"Shutting down DynamicServingCertificateController","v":0}
</pre>

poojasj:
@damemi That doesn't work.
I changed the arguments like this:
<pre>            - args:
              - --policy-config-file
              - /policy-dir/policy.yaml
[...]

Issue #965, opened 2022-09-27
Allow Logging format to be changed


gallowaystorm:
**Is your feature request related to a problem? Please describe.**

Yes it is related to a problem.  When sending the logs to Datadog, all logs show as errors because I cannot switch them to a format that Datadog excepts.  This means that all logs are errors even if they are just info level logs.

**Describe the solution you'd like**

The ability to change the format of the logs.

**Describe alternatives you've considered**

none

**What version of descheduler are you using?**

descheduler version: Helm 0.24.1

**Additional context**

damemi:
Hi @gallowaystorm, is the [`--logging-format` flag](https://github.com/kubernetes-sigs/descheduler/blob/0317be1b7672c511f5e2e53acaea1dd5cb74fd77/docs/user-guide.md#cli-options) possibly helpful? That supports text and json logging output.

gallowaystorm:
How can I use that with the Helm chart?

harshanarayana:
```bash
helm install ...... --set cmdOptions.logging-format=text
```
@gallowaystorm  This should get you going 

```bash
❯ helm install test . --dry-run --set cmdOptions.logging-format=text | grep -A2 logging
                - "--logging-format"
                - "text"
                - "--v"

```

gallowaystorm:
Thank you I will try this!

gallowaystorm:
Also that did not help the issue in Datadog.  It did change the log format however

damemi:
@gallowaystorm could you give more details on what the issue in Datadog is? Some sample output or screenshots would help, I don't know if they are looking for a specific format or what that format you need is

damemi:
We just use the [klog](https://github.com/kubernetes/klog) library to set our logs, which should also handle the json formatting, so I'm not sure why it wouldn't be passing the severity.

Could you try setting `--skip-headers` as well? I found [this issue](https://github.com/kubernetes/klog/issues/145) and it looks like maybe the klog headers are in your logs... still don't see the severity in the JSON format, but this could be a start

harshanarayana:
@damemi Might be related to the discussion on https://github.com/kubernetes-sigs/descheduler/issues/811 ?

gallowaystorm:
@damemi   I tried to set the skip headers via our helm release in Terraform (how we do helm releases) and it does not seem possible to do so:

```
resource "helm_release" "descheduler" {
  chart      = local.service_name
  name       = "${local.service_name}-${var.environment}"
[...]

PR #993, opened 2022-10-20
correct JSON log format

wakeful:
today I upgraded to `0.25.2` version and noticed that the `json` format is not working
after a quick scan of the code I noticed that `factory` is always `nil`.

some related problems #991 and #965 

I'm sure there is a much more elegant solution to this problem.

### DIFF
--- a/cmd/descheduler/app/server.go
+++ b/cmd/descheduler/app/server.go
@@ -34,6 +34,7 @@ import (
 	"k8s.io/apiserver/pkg/server/mux"
 	restclient "k8s.io/client-go/rest"
 	registry "k8s.io/component-base/logs/api/v1"
+	jsonLog "k8s.io/component-base/logs/json"
 	_ "k8s.io/component-base/logs/json/register"
 	"k8s.io/component-base/metrics/legacyregistry"
 	"k8s.io/klog/v2"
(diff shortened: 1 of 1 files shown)


ingvagabund:
The last change in the formatting was performed in https://github.com/kubernetes-sigs/descheduler/commit/c9b0fbe46742b4e00a39336d30f98ffaec20a77f. @wakeful would undoing the change help?

wakeful:
@ingvagabund I believe so but then the external dep upgrade will also be rollback,

```diff
-
-			factory, _ := registry.LogRegistry.Get(s.Logging.Format)
+			var factory registry.LogFormatFactory
 			if factory == nil {
 				klog.ClearLogger()
```

looking at the above patch I suspect the underlying API has change and the code was just `adjusted` to compile 
as this project is lacking tests around flags it went unnoticed.

damemi:
Thank you for finding this @wakeful

> personally I think we should fix it without rollback / revert, add test(s) and push out a patch version

+1

damemi (commented):
/ok-to-test

cmd/descheduler/app/server.go
```diff
@@ -62,13 +63,19 @@ func NewDeschedulerCommand(out io.Writer) *cobra.Command {
 				return
 			}
 			var factory registry.LogFormatFactory
+
+			switch s.Logging.Format {
+			case "json":
```
ingvagabund, line 68: @wakeful are there any other cases to consider? Based on https://kubernetes.io/blog/2020/09/04/kubernetes-1-19-introducing-structured-logs/ there's currently only json. Also from
```
      --logging-format string                    Sets the log format. Permitted formats: "text", "json". Non-default formats don't honor these flags: --add-dir-header, --alsologtostderr, --log-backtrace-at, --log_dir, --log_file, --log_file_max_size, --logtostderr, --skip-headers, --skip-log-headers, --stderrthreshold, --log-flush-frequency.\nNon-default choices are currently alpha and subject to change without warning. (default "text")
```
Only `json` and `text` values are valid. Using simple `if s.Logging.Format == "json"` is more preferable atm.

wakeful: done

damemi:
@ingvagabund wdyt about another patch release for this?

ingvagabund:
@wakeful is releasing a new descheduler version paramount for your business? Or, can you build the descheduler image from the latest master:HEAD and run in meantime?

@damemi unless there's a strong need for it, I prefer to wait until more fixes are merged.

wakeful:
hey @ingvagabund for now we are running a custom build from top of master :)

PR #999, opened 2022-11-07
Separate migration of strategy configuration from the constructing and running the plugins

ingvagabund:
Moving the code around to simplify implementation of conversion from v1alpha1 -> v1alpha2.

With a goal to separate the code responsible for processing strategy arguments from the rest of the code. Keeping assumption the only v1alpha2 policy configuration is produced from v1alpha1. Once the conversion is done, the main loop can be extended to take into account more reach expressiveness of v1alpha2. E.g. defaulting a list of plugins, pluginConfig and enabled/disabled extension points.


knelasevero:
strategyParamsToPluginArgs map is way cleaner than all those conversion funcs that I had in the WIP PR, thanks for showing me that :D

I wonder if we could have some tests for the policyconfig.go code. Than we can make next changes easier to verify with changing tests.

pkg/descheduler/policyconfig.go
```diff
@@ -46,10 +50,150 @@ func LoadPolicyConfig(policyConfigFile string) (*api.DeschedulerPolicy, error) {
...
+	ignorePvcPods := false
+	if deschedulerPolicy.IgnorePVCPods != nil {
+		ignorePvcPods = *deschedulerPolicy.IgnorePVCPods
+	}
+
+	var profiles []api.Profile
+
+	// Build profiles
+	for name, strategy := range deschedulerPolicy.Strategies {
+		if _, ok := strategyFuncs[name]; ok {
```
knelasevero, line 111: If we are using this map (`strategyFuncs[name]`) only to check if the strategy name exist, shouldn't we change this map name to something like `strategies` or maybe just drop this check

pkg/descheduler/policyconfig.go
```diff
@@ -46,10 +50,150 @@ func LoadPolicyConfig(policyConfigFile string) (*api.DeschedulerPolicy, error) {
...
 	return internalPolicy, nil
 }
+
+func v1alpha2v1alpha2(
+	client clientset.Interface,
+	deschedulerPolicy *v1alpha1.DeschedulerPolicy,
+) (*api.DeschedulerPolicy, error) {
+
+	strategyFuncs := map[v1alpha1.StrategyName]interface{}{
+		"RemoveDuplicates":                            nil,
```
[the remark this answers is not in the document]
ingvagabund: Those are strategy names which are not 1:1 to the plugin names.

pkg/descheduler/policyconfig.go
```diff
@@ -46,10 +50,150 @@ func LoadPolicyConfig(policyConfigFile string) (*api.DeschedulerPolicy, error) {
...
+	// TODO(jchaloup): replace this with v1alpha1 -> v1alpha2 conversion
+	internalPolicy, err := v1alpha2v1alpha2(client, versionedPolicy)
+	if err != nil {
 		return nil, fmt.Errorf("failed converting versioned policy to internal policy version: %v", err)
 	}
 
 	return internalPolicy, nil
 }
+
+func v1alpha2v1alpha2(
```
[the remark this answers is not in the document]
ingvagabund: Renamed to v1alpha1Tov1alpha2

pkg/descheduler/policyconfig.go
ingvagabund: Renamed. This is the only place where the strategy name validation is performed.

ingvagabund:
> I wonder if we could have some tests for the policyconfig.go code. Than we can make next changes easier to verify with changing tests.

`v1alpha1Tov1alpha2` is a temporary function so I wonder what would be the right code to test here. Probably valid conversion from v1alpha1 to v1alpha2.

knelasevero:
> > I wonder if we could have some tests for the policyconfig.go code. Than we can make next changes easier to verify with changing tests.
> 
> `v1alpha1Tov1alpha2` is a temporary function so I wonder what would be the right code to test here. Probably valid conversion from v1alpha1 to v1alpha2.

Yeah, valid conversion from v1alpha1 to v1alpha2

knelasevero:
Just to give context to anyone reading and wanting to help review.

Consider https://github.com/kubernetes-sigs/descheduler/pull/974/files#diff-0129f73a80ed727af36e1b6007786fd044191cafed0900875f5a90b38ea33739 to be a **POC** for the whole v1alpha2 introduction + conversion. And consider what Jan is starting here as the actual incremental changes going in that direction but changing whatever makes sense while doing that.

pkg/descheduler/descheduler.go
```diff
@@ -358,74 +317,82 @@ func RunDeschedulerStrategies(ctx context.Context, rs *options.DeschedulerServer
...
-						EvictFailedBarePods:     evictBarePods,
-						NodeFit:                 nodeFit,
-						PriorityThreshold: &api.PriorityThreshold{
-							Value: &thresholdPriority,
-						},
+		// Build plugins
+		for _, profile := range deschedulerPolicy.Profiles {
+			pc := getPluginConfig(defaultevictor.PluginName, profile.PluginConfigs)
+			if pc == nil {
+				klog.ErrorS(fmt.Errorf("unable to get plugin config"), "skipping plugin", "plugin", defaultevictor.PluginName)
```
a7i, line 327: thoughts on including profile name in all structured logs?
```suggestion
				klog.ErrorS(fmt.Errorf("unable to get plugin config"), "skipping plugin", "plugin", defaultevictor.PluginName, "profile", profile.Name)
```

pkg/descheduler/descheduler.go
```diff
@@ -358,74 +317,82 @@ func RunDeschedulerStrategies(ctx context.Context, rs *options.DeschedulerServer
...
+			// Assuming only a list of enabled extension points.
+			// Later, when a default list of plugins and their extension points is established,
+			// compute the list of enabled extension points as (DefaultEnabled + Enabled - Disabled)
+			for _, plugin := range append(profile.Plugins.Deschedule.Enabled, profile.Plugins.Balance.Enabled...) {
+				pc := getPluginConfig(plugin, profile.PluginConfigs)
+				if pc == nil {
+					klog.ErrorS(fmt.Errorf("unable to get plugin config"), "skipping plugin", "plugin", plugin)
+					continue
+				}
+				pg := pluginsMap[plugin](pc.Args, handle)
```
a7i, line 360: should we guard against invalid plugin names in a profile?

```suggestion
				pgFunc, ok := pluginsMap[plugin]
				if !ok {
					klog.ErrorS(fmt.Errorf("invalid plugin name in profile"), "skipping plugin", "plugin", plugin, "profile", profile.Name)
					continue
				}
				pg := pgFunc(pc.Args, handle)
```

a7i: Also, should we explicitly set the log level?
`klog.V(1).ErrorS`

pkg/descheduler/policyconfig.go
```diff
@@ -46,10 +50,150 @@ func LoadPolicyConfig(policyConfigFile string) (*api.DeschedulerPolicy, error) {
...
+			}
+		} else {
+			klog.ErrorS(fmt.Errorf("unknown strategy name"), "skipping strategy", "strategy", name)
+		}
+	}
+
+	return &api.DeschedulerPolicy{
+		Profiles:                       profiles,
+		NodeSelector:                   deschedulerPolicy.NodeSelector,
+		MaxNoOfPodsToEvictPerNode:      deschedulerPolicy.MaxNoOfPodsToEvictPerNamespace,
```
a7i, line 196:
```suggestion
		MaxNoOfPodsToEvictPerNode:      deschedulerPolicy.MaxNoOfPodsToEvictPerNode,
```

pkg/descheduler/descheduler.go
```diff
@@ -358,74 +317,82 @@ func RunDeschedulerStrategies(ctx context.Context, rs *options.DeschedulerServer
...
+		// Execute extension points
+		for _, pg := range enabledDeschedulePlugins {
+			// TODO: strategyName should be accessible from within the strategy using a framework
+			// handle or function which the Evictor has access to. For migration/in-progress framework
+			// work, we are currently passing this via context. To be removed
+			// (See discussion thread https://github.com/kubernetes-sigs/descheduler/pull/885#discussion_r919962292)
+			childCtx := context.WithValue(ctx, "strategyName", pg.Name())
+			status := pg.Deschedule(childCtx, nodes)
+			if status != nil && status.Err != nil {
+				klog.V(1).ErrorS(err, "plugin finished with error", "pluginName", pg.Name())
```
a7i, line 383:
```suggestion
				klog.V(1).ErrorS(status.Err, "plugin finished with error", "pluginName", pg.Name())
```

pkg/descheduler/descheduler.go
```diff
@@ -358,74 +317,82 @@ func RunDeschedulerStrategies(ctx context.Context, rs *options.DeschedulerServer
...
-				klog.ErrorS(fmt.Errorf("unknown strategy name"), "skipping strategy", "strategy", name)
+		for _, pg := range enabledBalancePlugins {
+			// TODO: strategyName should be accessible from within the strategy using a framework
+			// handle or function which the Evictor has access to. For migration/in-progress framework
+			// work, we are currently passing this via context. To be removed
+			// (See discussion thread https://github.com/kubernetes-sigs/descheduler/pull/885#discussion_r919962292)
+			childCtx := context.WithValue(ctx, "strategyName", pg.Name())
+			status := pg.Balance(childCtx, nodes)
+			if status != nil && status.Err != nil {
+				klog.V(1).ErrorS(err, "plugin finished with error", "pluginName", pg.Name())
```
a7i, line 395:
```suggestion
				klog.V(1).ErrorS(status.Err, "plugin finished with error", "pluginName", pg.Name())
```

pkg/descheduler/descheduler.go
```diff
@@ -358,74 +317,82 @@ func RunDeschedulerStrategies(ctx context.Context, rs *options.DeschedulerServer
...
-						clientSet:                 rs.Client,
-						getPodsAssignedToNodeFunc: getPodsAssignedToNode,
-						sharedInformerFactory:     sharedInformerFactory,
-						evictor: &evictorImpl{
-							podEvictor:    podEvictor,
-							evictorFilter: evictorFilter.(framework.EvictorPlugin),
-						},
-					}
+		// Execute extension points
+		for _, pg := range enabledDeschedulePlugins {
```
a7i, line 375: splitting it to `enabledDeschedulePlugins` and `enabledBalancePlugins` causes it to lose the sorting capability, right? Meaning that I can never have a Balance plugin run before a Deschedule plugin

a7i (commented):
Looking good, very exciting 🏆

pkg/descheduler/descheduler.go
ingvagabund: Actually, we should call the ErrorS without any log level since all errors are expected to be seen right away. I wonder how did this happen :) It got introduced by accident and copy-pasted all around.

pkg/descheduler/descheduler.go
ingvagabund: Yes. This is a breaking change. The Balance extension points are always expected to be run after all Deschedule extension points.

pkg/descheduler/descheduler.go
ingvagabund: > should we guard against invalid plugin names in a profile?

Added the check

pkg/descheduler/policyconfig.go
```diff
@@ -46,10 +50,150 @@ func LoadPolicyConfig(policyConfigFile string) (*api.DeschedulerPolicy, error) {
...
+	// TODO(jchaloup): replace this with v1alpha1 -> v1alpha2 conversion
+	internalPolicy, err := V1alpha1ToInternal(client, versionedPolicy)
+	if err != nil {
 		return nil, fmt.Errorf("failed converting versioned policy to internal policy version: %v", err)
 	}
 
 	return internalPolicy, nil
 }
+
+func V1alpha1ToInternal(
```
knelasevero, line 63: I renamed this to V1alpha1ToInternal, since we don't actually really have the v1alpha2 type here. I will work it out on the next PRs

pkg/descheduler/policyconfig.go
```diff
@@ -46,10 +50,150 @@ func LoadPolicyConfig(policyConfigFile string) (*api.DeschedulerPolicy, error) {
...
+	if deschedulerPolicy.IgnorePVCPods != nil {
+		ignorePvcPods = *deschedulerPolicy.IgnorePVCPods
+	}
+
+	var profiles []api.Profile
+
+	// Build profiles
+	for name, strategy := range deschedulerPolicy.Strategies {
+		if _, ok := validStrategyNames[name]; ok {
+			if strategy.Enabled {
```
knelasevero, line 111: We are only generating profiles for enabled strategies here. I think this is fine, but I wanted to check if this is what we want to do. I notice while writing tests and not getting the disabled profiles populated

pkg/descheduler/policyconfig_test.go
```diff
@@ -0,0 +1,444 @@
...
+					// 	Name: nodeutilization.LowNodeUtilizationPluginName,
+					// 	PluginConfigs: []api.PluginConfig{
+					// 		{
+					// 			Name: nodeutilization.LowNodeUtilizationPluginName,
+					// 			Args: &nodeutilization.LowNodeUtilizationArgs{
+					// 				Thresholds: api.ResourceThresholds{
+					// 					"cpu":    api.Percentage(20),
+					// [...]
+					// [...]
+					// },
```
knelasevero, line 128: Related to https://github.com/kubernetes-sigs/descheduler/pull/999/files#r1019286039

pkg/descheduler/policyconfig.go
ingvagabund: Yes, only for enabled strategies for the moment. Logic for the list of disabled plugins will get implemented later.

pkg/descheduler/strategy_migration.go
```diff
@@ -38,123 +39,98 @@ import (
 // without any wiring. Keeping the wiring here so the descheduler can still use
 // the v1alpha1 configuration during the strategy migration to plugins.
 
-var pluginsMap = map[string]func(ctx context.Context, nodes []*v1.Node, params *api.StrategyParameters, handle *handleImpl){
-	"RemovePodsViolatingNodeTaints": func(ctx context.Context, nodes []*v1.Node, params *api.StrategyParameters, handle *handleImpl) {
+var strategyParamsToPluginArgs = map[string]func(params *v1alpha1.StrategyParameters) (*api.PluginConfig, error){
```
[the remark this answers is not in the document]
ingvagabund: I don't think it's necessary given the map will be dropped eventually. Though, if you believe there's a value in it, by all means :)

knelasevero:
/lgtm

:smile:

pkg/descheduler/policyconfig_test.go
```diff
@@ -0,0 +1,443 @@
...
+	defaultEvictorPluginSet := api.PluginSet{
+		Enabled: []string{defaultevictor.PluginName},
+	}
+	type testCase struct {
+		description string
+		policy      *v1alpha1.DeschedulerPolicy
+		err         error
+		result      *api.DeschedulerPolicy
+	}
+	testCases := []testCase{
```
ingvagabund, line 59: You will not like me but would you please create another test which wires all existing strategy params into their corresponding arguments? So we can avoid https://github.com/kubernetes-sigs/descheduler/pull/961.

pkg/descheduler/policyconfig.go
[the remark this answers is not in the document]
damemi: We should eventually converge on a 1:1 alignment so that we can refer to constants in code

pkg/descheduler/policyconfig_test.go
knelasevero: Would be easier to write the tests and it would be less clunky onto the upper abstraction. We could test exactly what is being passed and what is coming out from params to args

pkg/descheduler/policyconfig.go
damemi: using an internal type is fine, we may want to even leave it like that if we are planning to rapidly iterate and release more api versions

pkg/descheduler/strategy_migration.go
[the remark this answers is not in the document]
damemi: it's temporary, but it is holding things together for a few releases. So testing this would be good to make sure we don't break anything in transition

pkg/descheduler/strategy_migration.go
```diff
@@ -166,106 +142,214 @@ var pluginsMap = map[string]func(ctx context.Context, nodes []*v1.Node, params *
...
 	},
-	"HighNodeUtilization": func(ctx context.Context, nodes []*v1.Node, params *api.StrategyParameters, handle *handleImpl) {
+	"HighNodeUtilization": func(params *v1alpha1.StrategyParameters) (*api.PluginConfig, error) {
+		if params.NodeResourceUtilizationThresholds == nil {
+			params.NodeResourceUtilizationThresholds = &v1alpha1.NodeResourceUtilizationThresholds{}
+		}
 		args := &nodeutilization.HighNodeUtilizationArgs{
-			Thresholds:    params.NodeResourceUtilizationThresholds.Thresholds,
-			NumberOfNodes: params.NodeResourceUtilizationThresholds.NumberOfNodes,
+			EvictableNamespaces: v1alpha1NamespacesToInternal(params.Namespaces),
```
knelasevero, line 195: I have added this way for people to already use preEviction namespace filtering for node utilization even before v1alpha2 is totally here.

pkg/descheduler/strategy_migration.go
```diff
@@ -166,106 +142,214 @@ var pluginsMap = map[string]func(ctx context.Context, nodes []*v1.Node, params *
...
 	},
-	"LowNodeUtilization": func(ctx context.Context, nodes []*v1.Node, params *api.StrategyParameters, handle *handleImpl) {
+	"LowNodeUtilization": func(params *v1alpha1.StrategyParameters) (*api.PluginConfig, error) {
+		if params.NodeResourceUtilizationThresholds == nil {
+			params.NodeResourceUtilizationThresholds = &v1alpha1.NodeResourceUtilizationThresholds{}
+		}
 		args := &nodeutilization.LowNodeUtilizationArgs{
-			Thresholds:             params.NodeResourceUtilizationThresholds.Thresholds,
-			TargetThresholds:       params.NodeResourceUtilizationThresholds.TargetThresholds,
+			EvictableNamespaces:    v1alpha1NamespacesToInternal(params.Namespaces),
```
knelasevero, line 213: same as https://github.com/kubernetes-sigs/descheduler/pull/999/files#r1020336399

pkg/descheduler/strategy_migration.go
knelasevero: It is relevant in this PR because without this we don't get the conversion that we expect on the other side with all params being translated.

README.md
```diff
@@ -243,7 +243,7 @@ actual usage metrics. Implementing metrics-based descheduling is currently TODO
 |`thresholdPriority`|int (see [priority filtering](#priority-filtering))|
 |`thresholdPriorityClassName`|string (see [priority filtering](#priority-filtering))|
 |`nodeFit`|bool (see [node fit filtering](#node-fit-filtering))|
-|`evictableNamespaces`|(see [namespace filtering](#namespace-filtering))|
+|`Namespaces`|(see [namespace filtering](#namespace-filtering))|
```
knelasevero, line 246: The way this was written before did not make sense since this field was not available. Now it makes sense since we translate this Namespaces into the internal arg evictableNamespaces for the nodeUtilization plugins

pkg/descheduler/strategy_migration.go
knelasevero: Always good to write tests and notice these things :point_right: https://github.com/kubernetes-sigs/descheduler/pull/999/files#r1020344073

README.md
knelasevero: (while we are doing conversions with v1alpha1 around)

knelasevero:
/unhold

unholding and waiting for another lgtm :)

ingvagabund:
LGTM

knelasevero:

Not sure this is done like this, let me check :)

PR #1003, opened 2022-11-22
use plugin registry and prepare for conversion

knelasevero:
Before moving on to https://github.com/kubernetes-sigs/descheduler/pull/974 on the spirit of incremental changes:

Decided to make the plugin registry available. I think it can help some other framework inner initiatives in parallel, not only conversion. I will continue with the conversion after this gets merged

### DIFF
--- a/cmd/descheduler/descheduler.go
+++ b/cmd/descheduler/descheduler.go
@@ -20,16 +20,13 @@ import (
 	"os"
 
 	"k8s.io/component-base/cli"
-	"k8s.io/klog/v2"
 	"sigs.k8s.io/descheduler/cmd/descheduler/app"
-	"sigs.k8s.io/descheduler/pkg/framework/plugins/pluginbuilder"
+	"sigs.k8s.io/descheduler/pkg/descheduler"
 )
 
 func init() {
-	klog.SetOutput(os.Stdout)
-	klog.InitFlags(nil)
-	pluginbuilder.PluginRegistry = pluginbuilder.NewRegistry()
-	pluginbuilder.RegisterDefault(pluginbuilder.PluginRegistry)
+	app.SetupLogs()
+	descheduler.SetupPlugins()
 }
 
 func main() {
--- a/cmd/descheduler/descheduler.go
+++ b/cmd/descheduler/descheduler.go
@@ -22,11 +22,14 @@ import (
 	"k8s.io/component-base/cli"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/descheduler/cmd/descheduler/app"
+	"sigs.k8s.io/descheduler/pkg/framework/plugins/pluginbuilder"
 )
 
 func init() {
 	klog.SetOutput(os.Stdout)
 	klog.InitFlags(nil)
+	pluginbuilder.PluginRegistry = pluginbuilder.NewRegistry()
+	pluginbuilder.RegisterDefault(pluginbuilder.PluginRegistry)
 }
 
 func main() {
--- a/pkg/descheduler/descheduler_test.go
+++ b/pkg/descheduler/descheduler_test.go
@@ -74,7 +74,7 @@ func TestTaintsUpdated(t *testing.T) {
 	var evictedPods []string
 	client.PrependReactor("create", "pods", podEvictionReactionFuc(&evictedPods))
 
-	internalDeschedulerPolicy, err := V1alpha1ToInternal(client, dp)
+	internalDeschedulerPolicy, err := V1alpha1ToInternal(client, dp, pluginbuilder.PluginRegistry)
 	if err != nil {
 		t.Fatalf("Unable to convert v1alpha1 to v1alpha2: %v", err)
 	}
@@ -136,7 +136,7 @@ func TestDuplicate(t *testing.T) {
 	var evictedPods []string
 	client.PrependReactor("create", "pods", podEvictionReactionFuc(&evictedPods))
 
-	internalDeschedulerPolicy, err := V1alpha1ToInternal(client, dp)
+	internalDeschedulerPolicy, err := V1alpha1ToInternal(client, dp, pluginbuilder.PluginRegistry)
 	if err != nil {
 		t.Fatalf("Unable to convert v1alpha1 to v1alpha2: %v", err)
 	}
--- a/pkg/descheduler/descheduler_test.go
+++ b/pkg/descheduler/descheduler_test.go
@@ -23,7 +23,7 @@ import (
 
 func TestTaintsUpdated(t *testing.T) {
 	pluginbuilder.PluginRegistry = pluginbuilder.NewRegistry()
-	pluginbuilder.Register(removepodsviolatingnodetaints.PluginName, removepodsviolatingnodetaints.New, &removepodsviolatingnodetaints.RemovePodsViolatingNodeTaintsArgs{}, false, pluginbuilder.PluginRegistry)
+	pluginbuilder.Register(removepodsviolatingnodetaints.PluginName, removepodsviolatingnodetaints.New, &removepodsviolatingnodetaints.RemovePodsViolatingNodeTaintsArgs{}, pluginbuilder.PluginRegistry)
 	ctx := context.Background()
 	n1 := test.BuildTestNode("n1", 2000, 3000, 10, nil)
 	n2 := test.BuildTestNode("n2", 2000, 3000, 10, nil)
@@ -90,7 +90,7 @@ func TestTaintsUpdated(t *testing.T) {
 
 func TestDuplicate(t *testing.T) {
 	pluginbuilder.PluginRegistry = pluginbuilder.NewRegistry()
-	pluginbuilder.Register(removeduplicates.PluginName, removeduplicates.New, &removeduplicates.RemoveDuplicatesArgs{}, false, pluginbuilder.PluginRegistry)
+	pluginbuilder.Register(removeduplicates.PluginName, removeduplicates.New, &removeduplicates.RemoveDuplicatesArgs{}, pluginbuilder.PluginRegistry)
 	ctx := context.Background()
 	node1 := test.BuildTestNode("n1", 2000, 3000, 10, nil)
 	node2 := test.BuildTestNode("n2", 2000, 3000, 10, nil)
--- a/pkg/descheduler/descheduler_test.go
+++ b/pkg/descheduler/descheduler_test.go
@@ -15,10 +15,15 @@ import (
 	"sigs.k8s.io/descheduler/cmd/descheduler/app/options"
 	"sigs.k8s.io/descheduler/pkg/api"
 	"sigs.k8s.io/descheduler/pkg/api/v1alpha1"
+	"sigs.k8s.io/descheduler/pkg/framework/plugins/pluginbuilder"
+	"sigs.k8s.io/descheduler/pkg/framework/plugins/removeduplicates"
+	"sigs.k8s.io/descheduler/pkg/framework/plugins/removepodsviolatingnodetaints"
 	"sigs.k8s.io/descheduler/test"
 )
 
 func TestTaintsUpdated(t *testing.T) {
+	pluginbuilder.PluginRegistry = pluginbuilder.NewRegistry()
+	pluginbuilder.Register(removepodsviolatingnodetaints.PluginName, removepodsviolatingnodetaints.New, &removepodsviolatingnodetaints.RemovePodsViolatingNodeTaintsArgs{}, false, pluginbuilder.PluginRegistry)
 	ctx := context.Background()
 	n1 := test.BuildTestNode("n1", 2000, 3000, 10, nil)
 	n2 := test.BuildTestNode("n2", 2000, 3000, 10, nil)
@@ -84,6 +89,8 @@ func TestTaintsUpdated(t *testing.T) {
 }
 
 func TestDuplicate(t *testing.T) {
+	pluginbuilder.PluginRegistry = pluginbuilder.NewRegistry()
+	pluginbuilder.Register(removeduplicates.PluginName, removeduplicates.New, &removeduplicates.RemoveDuplicatesArgs{}, false, pluginbuilder.PluginRegistry)
 	ctx := context.Background()
 	node1 := test.BuildTestNode("n1", 2000, 3000, 10, nil)
 	node2 := test.BuildTestNode("n2", 2000, 3000, 10, nil)
--- a/pkg/descheduler/policyconfig.go
+++ b/pkg/descheduler/policyconfig.go
@@ -28,6 +28,7 @@ import (
 	"sigs.k8s.io/descheduler/pkg/api"
 	"sigs.k8s.io/descheduler/pkg/api/v1alpha1"
 	"sigs.k8s.io/descheduler/pkg/descheduler/scheme"
+	"sigs.k8s.io/descheduler/pkg/framework"
 	"sigs.k8s.io/descheduler/pkg/framework/plugins/defaultevictor"
 	"sigs.k8s.io/descheduler/pkg/framework/plugins/pluginbuilder"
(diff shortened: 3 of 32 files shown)


pkg/framework/plugins/podlifetime/pod_lifetime.go
```diff
@@ -131,3 +132,15 @@ func (d *PodLifeTime) Deschedule(ctx context.Context, nodes []*v1.Node) *framewo
 
 	return nil
 }
+
+func init() {
```
damemi, line 136: I like have the registry implicitly ([pluginbuilder.go](https://github.com/kubernetes-sigs/descheduler/pull/1003/files#diff-907243a6cfbb3c6f27a53916054eaa5b626cfda6ab368aad8f528769a3bb7451R38)) initialized with `init()` (because it is part of the framework internals), but I'm thinking the plugins should have an explicit function call to register themselves. wdyt?

knelasevero: I think that makes sense as well. But I also like plugins that automatically register themselves if you import them, like I did here.

Using a plugin would then just require a '_' import in a register.go file or something like that. (this makes it super simple for people to use a plugin and only have to change a single file). Not using a plugin means just deleting a line from that list of imports. This is specially interesting when thinking of out-of-tree plugins.

register.go would look like:
```
// packages imported here are registered to the pluginBuilder Plugin Registry.
//nolint:revive
import (
	_ "sigs.k8s.io/descheduler/pkg/framework/plugins/removepodsviolatinginterpodantiaffinity"
	_ "sigs.k8s.io/descheduler-external-plugins/pkg/plugins/custompluginexample"
)
```

damemi: Yeah, it's cool for sure. And it makes sense to use it for the internals. But does it provide value beyond that convenience? Even then, it's only saving 1 line.

I think the user-facing functions should be explicit rather than automatically hooking into framework internals as soon as they're imported. Doing so makes it safer (compile time checks on unused plugins + attempts to use unimported plugins) and clearer about what's going on with less magic for the user to understand.

knelasevero: I think it does more than save one line. Having a new explicit Register function, or even using a similar approach to the KS with the WithPlugin wrapper functions adds complexity to how this works, and I think it works against the UX of the user.

> compile time checks on unused plugins + attempts to use unimported plugins

I don't see how that helps with compile time checks on unused plugins, and attempts to use unimported plugins would trigger similar errors, and we can still treat and have meaningful logs either way.

To me it is as explicit as having a function call, importing or not importing a new plugin. I might be missing something about specifics of the implementation

damemi: > To me it is as explicit as having a function call, importing or not importing a new plugin.

A function to manually register the plugin is much more explicit, and it makes the registration flow easier to follow and debug. Implicitly registering the plugin by importing it means you have to be vigilant about where you import that plugin (for example if 2 plugins share functionality).

Auto-registration is something that could be added later, at a time when we are more comfortable with the framework. Adding it as a day-0 feature is too much in my opinion. I would like to see an end-to-end product first before adding convenience functionality like this.

knelasevero: I see your point! Gonna add the Register func and register all internal plugins by default :)

cmd/descheduler/descheduler.go
```diff
@@ -22,11 +22,14 @@ import (
...
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/descheduler/cmd/descheduler/app"
+	"sigs.k8s.io/descheduler/pkg/framework/plugins/pluginbuilder"
 )
 
 func init() {
 	klog.SetOutput(os.Stdout)
 	klog.InitFlags(nil)
+	pluginbuilder.PluginRegistry = pluginbuilder.NewRegistry()
+	pluginbuilder.RegisterDefault(pluginbuilder.PluginRegistry)
```
knelasevero, line 32: A bit unsure about adding this here

pkg/descheduler/descheduler_test.go
```diff
@@ -15,10 +15,15 @@ import (
...
 	"sigs.k8s.io/descheduler/pkg/api/v1alpha1"
+	"sigs.k8s.io/descheduler/pkg/framework/plugins/pluginbuilder"
+	"sigs.k8s.io/descheduler/pkg/framework/plugins/removeduplicates"
+	"sigs.k8s.io/descheduler/pkg/framework/plugins/removepodsviolatingnodetaints"
 	"sigs.k8s.io/descheduler/test"
 )
 
 func TestTaintsUpdated(t *testing.T) {
+	pluginbuilder.PluginRegistry = pluginbuilder.NewRegistry()
+	pluginbuilder.Register(removepodsviolatingnodetaints.PluginName, removepodsviolatingnodetaints.New, &removepodsviolatingnodetaints.RemovePodsViolatingNodeTaintsArgs{}, false, pluginbuilder.PluginRegistry)
```
knelasevero, line 26: tests like these will have to register the plugin to use/test them

test/e2e/e2e_leaderelection_test.go
```diff
@@ -35,9 +35,12 @@ import (
 	utilpointer "k8s.io/utils/pointer"
 	"sigs.k8s.io/descheduler/cmd/descheduler/app/options"
 	"sigs.k8s.io/descheduler/pkg/descheduler"
+	"sigs.k8s.io/descheduler/pkg/framework/plugins/pluginbuilder"
 )
 
 func TestLeaderElection(t *testing.T) {
+	pluginbuilder.PluginRegistry = pluginbuilder.NewRegistry()
+	pluginbuilder.RegisterDefault(pluginbuilder.PluginRegistry)
```
knelasevero, line 43: When writing e2e tests, sometimes, depending on the test, we will have to remember to register plugins

pkg/framework/plugins/pluginbuilder/pluginbuilder.go
```diff
@@ -25,6 +36,7 @@ type PluginBuilderAndArgsInstance struct {
 	// Just an example instance of this PluginArg so we can avoid having
 	// to deal with reflect Types
 	PluginArgInstance runtime.Object
+	Balance           bool
```
ingvagabund, line 39: This needs to be detected automatically by checking whether a plugin implements a specific interface. Also, a plugin builder does not need to know about any extension points as its purpose is to build plugins.

knelasevero:
@damemi @ingvagabund please have a look when you have some time :)

pkg/descheduler/policyconfig.go
```diff
@@ -163,10 +164,21 @@ func V1alpha1ToInternal(
...
+				// Plugins have either of the two extension points right now
+				// this might change in the future
+				exampleArgs := pluginbuilder.PluginRegistry[string(name)].PluginArgInstance
+				pluginInstance, err := pluginbuilder.PluginRegistry[string(name)].PluginBuilder(exampleArgs, &handleImpl{})
+				if err != nil {
+					klog.ErrorS(fmt.Errorf("could not build plugin"), "plugin build error", "plugin", name)
+					return nil, fmt.Errorf("could not build plugin: %v", name)
+				}
+
+				switch p := pluginInstance.(type) {
```
ingvagabund, line 176: A plugin can implement both BalancePlugin and DeschedulePlugin interfaces. Right now there's no such plugin. Though, later we might introduce one and forget to refactor the lines.

pkg/framework/plugins/pluginbuilder/pluginbuilder.go
```diff
@@ -47,28 +46,27 @@ func NewRegistry() Registry {
 	return Registry{}
 }
 
-func Register(name string, builderFunc PluginBuilder, exampleArg runtime.Object, balance bool, registry Registry) {
+func Register(name string, builderFunc PluginBuilder, exampleArg runtime.Object, registry Registry) {
```
ingvagabund, line 39: Plugin registration does not require passing its arguments given the same plugin builder can be invoked multiple times with different arguments.

ingvagabund (changes requested):

pkg/framework/plugins/pluginbuilder/pluginbuilder.go
knelasevero: > Plugin registration does not require passing its arguments given the same plugin builder can be invoked multiple times with different arguments.

That's why I call them exampleArgs... https://github.com/kubernetes-sigs/descheduler/pull/1003#discussion_r1033529025

They are not meant to be "filled". This is just so when doing manual conversations latter we don't have to deal with raw runtime.Objects and reflect types and we know already what type of args each registered plugin holds. These will be always empty args as an example for comparison.

Ref from the other pr: https://github.com/knelasevero/descheduler/blob/7c4d2960ebeea39cf7ff81474cc4a99e3d2aa22d/pkg/api/v1alpha2/conversion.go#L95

pkg/descheduler/policyconfig.go
```diff
@@ -163,10 +164,21 @@ func V1alpha1ToInternal(
 					},
 				}
 
-				// Plugins have either of the two extension points
-				if pluginbuilder.PluginRegistry[string(name)].Balance {
+				// Plugins have either of the two extension points right now
+				// this might change in the future
+				exampleArgs := pluginbuilder.PluginRegistry[string(name)].PluginArgInstance
+				pluginInstance, err := pluginbuilder.PluginRegistry[string(name)].PluginBuilder(exampleArgs, &handleImpl{})
```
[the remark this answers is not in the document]
knelasevero: In this functions specifically? Or you mean in general? we will always depend on that global registry variable somewhere. I moved the dependency up to LoadPolicyConfig tho now.

pkg/descheduler/policyconfig.go
knelasevero: If we add another type this can become a mess of combinations. I have added inner switches then, let me know if another approach would be better

pkg/framework/plugins/pluginbuilder/pluginbuilder.go
knelasevero: @ingvagabund this exampleArg thing is already making me avoid having to have a big switch case here as well: 				https://github.com/knelasevero/descheduler/blob/5edd5a7bd10276fba1ad077745aa5a1353737e67/pkg/descheduler/policyconfig.go#L167

It is just useful knowing the plugin to Arg type pair registered always, and being able to just pull it up from the registry. To me using the registry to store only New() funcs is a waste (And, again, this is what will make the manual conversion that I plan doing later possible, while no one figures out the native scheme conversion).

knelasevero:
I am using this PR to build the followup PR with the conversions and I will discard #974 when I have it up

pkg/descheduler/policyconfig.go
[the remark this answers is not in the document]
ingvagabund: The higher the global variable gets injected, the better.

pkg/framework/plugins/pluginbuilder/pluginbuilder.go
knelasevero: :raised_hands:

pkg/descheduler/policyconfig.go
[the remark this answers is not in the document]
knelasevero: Ok, moved it all the way to the Run function, so LoadPolicyConfig also does not depend on it

pkg/descheduler/policyconfig.go
knelasevero: separated them so we check each and both checks can happen for same plugin

knelasevero:
@ingvagabund @damemi do we think we can merge? :D

pkg/descheduler/setupplugins.go
```diff
@@ -0,0 +1,50 @@
...
+	"sigs.k8s.io/descheduler/pkg/framework/plugins/removepodshavingtoomanyrestarts"
+	"sigs.k8s.io/descheduler/pkg/framework/plugins/removepodsviolatinginterpodantiaffinity"
+	"sigs.k8s.io/descheduler/pkg/framework/plugins/removepodsviolatingnodeaffinity"
+	"sigs.k8s.io/descheduler/pkg/framework/plugins/removepodsviolatingnodetaints"
+	"sigs.k8s.io/descheduler/pkg/framework/plugins/removepodsviolatingtopologyspreadconstraint"
+)
+
+func SetupPlugins() {
+	pluginbuilder.PluginRegistry = pluginbuilder.NewRegistry()
+	RegisterDefaultPlugins(pluginbuilder.PluginRegistry)
```
damemi, line 35: we should probably decouple registry initialization from default plugin registration. From a third-party dev perspective, I can't build my own descheduler without registering all of the default plugins.

damemi: Ah nevermind, you can call `NewRegistry()` directly. no problem with this as an option then

damemi:
/lgtm
this looks great @knelasevero, thanks again for all of your work

pkg/descheduler/policyconfig.go
```diff
@@ -28,11 +28,13 @@ import (
...
 	"sigs.k8s.io/descheduler/pkg/api/v1alpha1"
 	"sigs.k8s.io/descheduler/pkg/descheduler/scheme"
+	"sigs.k8s.io/descheduler/pkg/framework"
 	"sigs.k8s.io/descheduler/pkg/framework/plugins/defaultevictor"
+	"sigs.k8s.io/descheduler/pkg/framework/plugins/pluginbuilder"
 	"sigs.k8s.io/descheduler/pkg/utils"
 )
 
-func LoadPolicyConfig(policyConfigFile string, client clientset.Interface) (*api.DeschedulerPolicy, error) {
+func LoadPolicyConfig(policyConfigFile string, client clientset.Interface, registry map[string]pluginbuilder.PluginBuilderAndArgsInstance) (*api.DeschedulerPolicy, error) {
```
ingvagabund, line 37: s/map[string]pluginbuilder.PluginBuilderAndArgsInstance/pluginbuilder.Registry

pkg/descheduler/policyconfig.go
```diff
@@ -175,13 +164,16 @@ func V1alpha1ToInternal(
 					},
 				}
 
-				// Plugins have either of the two extension points
-				switch pluginToExtensionPoint[pluginConfig.Name] {
-				case descheduleEP:
-					profile.Plugins.Deschedule.Enabled = []string{pluginConfig.Name}
-				case balanceEP:
-					profile.Plugins.Balance.Enabled = []string{pluginConfig.Name}
+				pluginArgs := pluginbuilder.PluginRegistry[string(name)].PluginArgInstance
```
ingvagabund, line 167: s/pluginbuilder.PluginRegistry/registry

ingvagabund (changes requested):

pkg/descheduler/policyconfig.go
```diff
@@ -63,20 +64,8 @@ func LoadPolicyConfig(policyConfigFile string, client clientset.Interface) (*api
 func V1alpha1ToInternal(
 	client clientset.Interface,
 	deschedulerPolicy *v1alpha1.DeschedulerPolicy,
+	registry map[string]pluginbuilder.PluginBuilderAndArgsInstance,
```
ingvagabund, line 67: s/map[string]pluginbuilder.PluginBuilderAndArgsInstance/pluginbuilder.Registry

knelasevero: done :heavy_check_mark:

pkg/descheduler/policyconfig.go
knelasevero: done :heavy_check_mark:

pkg/descheduler/policyconfig.go
knelasevero: done :heavy_check_mark:

PR #1019, opened 2022-12-13
parallelize docker build for each arch

a7i:
Builds have been timing out since we run them sequentially [here](https://github.com/kubernetes-sigs/descheduler/blob/9f6afd4d2526cc31e1486d25e2f1ecd1d7476de5/Makefile#L89)

Sample failed build: https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-descheduler-push-images/1602480325080911872

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,7 +84,7 @@ push: image
 	$(CONTAINER_ENGINE) tag $(IMAGE) $(IMAGE_GCLOUD)
 	$(CONTAINER_ENGINE) push $(IMAGE_GCLOUD)
 
-push-all: image.amd64 image.arm image.arm64
+push-images:
 	gcloud auth configure-docker
 	for arch in $(ARCHS); do \
 		$(CONTAINER_ENGINE) tag $(IMAGE)-$${arch} $(IMAGE_GCLOUD)-$${arch} ;\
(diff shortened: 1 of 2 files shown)


damemi:
thanks @a7i, is there any way to test this? It's our postsubmit builds that are timing out, right?

cloudbuild.yaml
```diff
@@ -7,14 +7,53 @@ timeout: 1500s
...
+    waitFor: ['-']
+
+  - name: 'gcr.io/k8s-staging-test-infra/gcb-docker-gcloud:v20221007-ad65926f6b'
     entrypoint: make
     env:
     - DOCKER_CLI_EXPERIMENTAL=enabled
     - VERSION=$_GIT_TAG
     - BASE_REF=$_PULL_BASE_REF
     args:
     - push-all
```
damemi, line 50: this is still going to call all 3 builds in this step: https://github.com/kubernetes-sigs/descheduler/blob/9f6afd4d2526cc31e1486d25e2f1ecd1d7476de5/Makefile#L87

maybe we need a new `make` target that decouples `push` from `build`?

damemi (changes requested):

cloudbuild.yaml
damemi: true, when it tries to build them it should go quicker since there won't be any changes. but I think it would be better to not make an unnecessary `build` call when we really just need a push. it makes it clearer what each step is doing and keeps job logs cleaner for debugging. wdyt?

a7i: that's fair, should be coming right up!

Makefile
```diff
@@ -84,7 +84,7 @@ push: image
 	$(CONTAINER_ENGINE) tag $(IMAGE) $(IMAGE_GCLOUD)
 	$(CONTAINER_ENGINE) push $(IMAGE_GCLOUD)
 
-push-all: image.amd64 image.arm image.arm64
+push-all-without-build:
```
damemi, line 87: nit: maybe just call this `push-images`?

damemi (commented):
one nit on the name
/approve

damemi:
/lgtm
thanks @a7i, fingers crossed! 🤞

PR #1023, opened 2022-12-14
increase cloudbuild timeout to 30 minutes until we figure out proper optimizations

a7i:
/cc @damemi


damemi (commented):
/approve
/lgtm
annoying that the parallelization didn't work. It looks like it had no effect. We might need to actually split this into separate jobs if possible rather than just parallel gcb steps

damemi:
Looks like it timed out again even at 30mins: https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-descheduler-push-images/1603073238139670528

Maybe we should ask around in test-infra how to handle multiple simultaneous builds

PR #1033, opened 2023-01-04
pod anti-affinity check among nodes

10hin:
Current RemovePodsViolatingInterPodAntiAffinity policy implementation evicts pods violating Pod anti-affinity only on single Node. But in general context, violation of Pod anti-affinity is not limited to single Node.

This PullRequest(PR) changes to check violation of Pod anti-affinity not only on sigle node but among different nodes.


damemi (commented):
/ok-to-test

pkg/framework/plugins/removepodsviolatinginterpodantiaffinity/pod_antiaffinity.go
```diff
@@ -78,6 +78,13 @@ func (d *RemovePodsViolatingInterPodAntiAffinity) Name() string {
 }
 
 func (d *RemovePodsViolatingInterPodAntiAffinity) Deschedule(ctx context.Context, nodes []*v1.Node) *framework.Status {
+	allPods, err := d.listExistingPods(ctx)
+	if err != nil {
+		return &framework.Status{
+			Err: fmt.Errorf("error listing all pods: %v", err),
+		}
+	}
+	nodeMap := createNodeMap(nodes)
```
damemi, line 87: could `nodeMap` replace the need to call `ListPodsOnANode()` in the loop below?

damemi: or, maybe not nodeMap sorry... but looking at how we call `allPods` above and then call `ListPodsOnANode`, maybe we can cut out one of them

10hin: Thank you for your review and sorry for my late reply.

I removed `ListPodsOnANode` call and keep code to query for all pods. Because listing pods by each node looks not efficient.
But, I'm not sure that my code filtering pods for `podsOnANode` does same as `ListPodsOnANode` call.

pkg/framework/plugins/removepodsviolatinginterpodantiaffinity/pod_antiaffinity.go
```diff
@@ -78,24 +78,36 @@ func (d *RemovePodsViolatingInterPodAntiAffinity) Name() string {
...
+		return &framework.Status{
+			Err: fmt.Errorf("error listing all pods: %v", err),
+		}
+	}
+
+	podsInANamespace := groupByNamespace(podsList)
+
+	runningPodFilter := func(pod *v1.Pod) bool {
+		return pod.Status.Phase != v1.PodSucceeded && pod.Status.Phase != v1.PodFailed
+	}
```
10hin, line 92: Also, I'm afraid that this code duplicating with following:

https://github.com/kubernetes-sigs/descheduler/blob/93a014ef5ebe66d4945f5ad181deb5af9c61ccd8/pkg/descheduler/pod/pods.go#L170-L172

damemi (commented):

10hin:
@damemi @knelasevero 
Thank you for your reviews and actions.
I'm sorry if I am wrong, but, because CI status looks wait only for adding `approved` label (see attached image), is @damemi  forgetting it? Or just wait for something?
If there are anything I can do in this PR, please tell me.

image

damemi (commented):
/approve

Issue #1020, opened 2022-12-13
Helm chart version update to v0.26.0


damemi:

damemi:
note merging this will be blocked on #1021 (https://github.com/kubernetes-sigs/descheduler/blob/master/docs/release-guide.md#flowchart)

damemi:
@JaneLiuL if you're still available for this, we are ready for the helm chart change

JaneLiuL:
sure, raise now

JaneLiuL:
sorry for so late for my home have some network issue

PR #911, opened 2022-08-10
introduce ct for local helm install test

knelasevero:
Updated docs and also removed old approach. To test it locally you can `make kind-multi-node` and then `make ct-helm`

For reference, a failing gh-actions step also prevents merges by prow/tide: https://github.com/kubernetes-sigs/descheduler/pull/907#issuecomment-1210920158

Will need to merge: https://github.com/kubernetes/test-infra/pull/26630

Then merge this PR itself

Following: https://github.com/kubernetes-sigs/descheduler/pull/895#issuecomment-1203608848

PR #1038, opened 2023-01-05
Update helm chart to v0.26.0

damemi:
Fixes https://github.com/kubernetes-sigs/descheduler/issues/1020

### DIFF
--- a/charts/descheduler/Chart.yaml
+++ b/charts/descheduler/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: descheduler
-version: 0.25.0
-appVersion: 0.25.0
+version: 0.26.0
+appVersion: 0.26.0
 description: Descheduler for Kubernetes is used to rebalance clusters by evicting pods that can potentially be scheduled on better nodes. In the current implementation, descheduler does not schedule replacement of evicted pods but relies on the default scheduler for that.
 keywords:
 - kubernetes


damemi:
Ah
```
Failed to pull image "registry.k8s.io/descheduler/descheduler:v0.26.0": rpc error: code = NotFound desc = failed to pull and unpack image "registry.k8s.io/descheduler/descheduler:v0.26.0": failed to resolve reference "registry.k8s.io/descheduler/descheduler:v0.26.0": registry.k8s.io/descheduler/descheduler:v0.26.0: not found
```
So it looks like the updated helm chart version attempts to get installed. I'm thinking maybe this is the behavior now after https://github.com/kubernetes-sigs/descheduler/pull/911? @knelasevero any ideas for that?

Some ways around this for releases:
1. roll back/change #911
2. Build/tag the v0.26.0 image before merging the helm chart change.
3. manually trigger the v0.26.0 image build (if possible? this could be the right way to do things going forward)

We use the helm chart change as the final "no-op" merge to trigger a build with the proper git tag (https://github.com/kubernetes-sigs/descheduler/blob/master/docs/release-guide.md#flowchart). But this step could be replaced with another non functional change such as readme updates. The important part is that functional code is merged *before* tagging the release in Git, followed by another GCB build to pull in the new git tag.

If there's no easy change to #911 we could make now, I'll just open a no-op docs change or something to trigger the build. Or manually trigger it

a7i:
Option 2 doesn't seem so bad.

Another option would be to add a step to the GitHub Action to query the latest image release and use that as the `AppVersion`. In other words, always use the latest published image but it does seem add a bit of confusion as it would not use the AppVersion defined in the Chart.

JaneLiuL (commented):
/lgtm

JaneLiuL:
prefer option 2 (Build/tag the v0.26.0 image before merging the helm chart change)

JaneLiuL:
prefer option 2 (Build/tag the v0.26.0 image before merging the helm chart change)

knelasevero:
> If there's no easy change to #911 we could make now

:heavy_plus_sign: 

> Another option would be to add a step to the GitHub Action to query the latest image release and use that as the AppVersion. In other words, always use the latest published image but it does seem add a bit of confusion as it would not use the AppVersion defined in the Chart.

This is indeed an option, and not that hard to make I guess, but I agree that it is confusing. Probably having a doc change to be the no op step and then the helm chart step coming next is the best friction-less option going forward, and building it locally and pushing is ok to unblock this PR here?

damemi:
We actually have some other changes under the v0.26.0 tag (like https://github.com/kubernetes-sigs/descheduler/pull/995), so for this release that should be able to promote to the public image.

For the next release we'll need to think about this. It also might be good to add a "release-mode" blocker so that PRs don't merge automatically or accidentally while we're doing the release process.

damemi:
Successful image tag, I'll work on updating the process for next release to address this.

PR #1041, opened 2023-01-11
metrics: enable loop duration and strategy duration metrics

### LINKED ISSUE #1037: Run duration metric
When trying to tune the run interval I'm not quite sure how long the descheduler run loop takes on average. 

It'd be nice to have some metrics around the duration of the run. Also could be nice to have some idea of each particular strategy run time too.

Look at logs to determine this.

descheduler version: 1.25.X

**Additional context**

harshanarayana:
This PR enables collection of metrics around the following

1. Amount of time taken for each loop of descheduler operation
2. Amount of time taken for each strategy operation to complete

This represents the values mentioned above as a Histogram with measurements tracked as a factor of seconds in time unit.

Closes #1037

### DIFF
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -48,9 +48,29 @@ var (
 		},
 	)
 
+	DeschedulerLoopDuration = metrics.NewHistogramVec(
+		&metrics.HistogramOpts{
+			Subsystem:      DeschedulerSubsystem,
+			Name:           "descheduler_loop_duration_seconds",
+			Help:           "Time taken to complete a strategy descheduling operation",
+			StabilityLevel: metrics.ALPHA,
+			Buckets:        []float64{0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10, 25, 50, 100, 250, 500},
+		}, []string{})
+
+	DeschedulerStrategyDuration = metrics.NewHistogramVec(
+		&metrics.HistogramOpts{
+			Subsystem:      DeschedulerSubsystem,
+			Name:           "descheduler_strategy_duration_seconds",
+			Help:           "Time taken to complete Each strategy of the descheduling operation",
+			StabilityLevel: metrics.ALPHA,
+			Buckets:        []float64{0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10, 25, 50, 100},
+		}, []string{"strategy", "profile"})
+
 	metricsList = []metrics.Registerable{
 		PodsEvicted,
 		buildInfo,
+		DeschedulerLoopDuration,
+		DeschedulerStrategyDuration,
 	}
 )
 
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -52,7 +52,7 @@ var (
 		&metrics.HistogramOpts{
 			Subsystem:      DeschedulerSubsystem,
 			Name:           "descheduler_loop_duration_seconds",
-			Help:           "Time taken to complete a strategy descheduling operation",
+			Help:           "Time taken to complete a full descheduling cycle",
 			StabilityLevel: metrics.ALPHA,
 			Buckets:        []float64{0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10, 25, 50, 100, 250, 500},
 		}, []string{})
--- a/pkg/descheduler/descheduler.go
+++ b/pkg/descheduler/descheduler.go
@@ -19,6 +19,7 @@ package descheduler
 import (
 	"context"
 	"fmt"
+	"time"
 
 	componentbaseconfig "k8s.io/component-base/config"
 	"k8s.io/klog/v2"
@@ -49,6 +50,16 @@ import (
 	"sigs.k8s.io/descheduler/pkg/framework/plugins/defaultevictor"
 	"sigs.k8s.io/descheduler/pkg/utils"
 )
+
+type enabledDeschedulePluginEntry struct {
+	Plugin  framework.DeschedulePlugin
+	Profile string
+}
+
+type enabledBalancePluginEntry struct {
+	Plugin  framework.BalancePlugin
+	Profile string
+}
 
 func Run(ctx context.Context, rs *options.DeschedulerServer) error {
 	metrics.Register()
@@ -268,6 +279,8 @@ func RunDeschedulerStrategies(ctx context.Context, rs *options.DeschedulerServer
 	defer eventBroadcaster.Shutdown()
 
 	wait.NonSlidingUntil(func() {
+		loopStartDuration := time.Now()
+		defer metrics.DeschedulerLoopDuration.With(map[string]string{}).Observe(time.Since(loopStartDuration).Seconds())
 		nodes, err := nodeutil.ReadyNodes(ctx, rs.Client, nodeLister, nodeSelector)
 		if err != nil {
 			klog.V(1).InfoS("Unable to get ready nodes", "err", err)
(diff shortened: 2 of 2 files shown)


metrics/metrics.go
```diff
@@ -48,9 +48,29 @@ var (
...
+			Help:           "Time taken to complete one full loop of descheduling operation",
+			StabilityLevel: metrics.ALPHA,
+			Buckets:        []float64{0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10, 25},
+		}, []string{})
+
+	DeschedulerStrategyDuration = metrics.NewHistogramVec(
+		&metrics.HistogramOpts{
+			Subsystem:      DeschedulerSubsystem,
+			Name:           "descheduler_strategy_duration_seconds",
+			Help:           "Time taken to complete one full loop of descheduling operation",
```
a7i, line 64:
```suggestion
			Help:           "Time taken to complete a strategy descheduling operation",
```

metrics/metrics.go
```diff
@@ -48,9 +48,29 @@ var (
 		},
 	)
 
+	DeschedulerLoopDuration = metrics.NewHistogramVec(
+		&metrics.HistogramOpts{
+			Subsystem:      DeschedulerSubsystem,
+			Name:           "descheduler_loop_duration_seconds",
+			Help:           "Time taken to complete one full loop of descheduling operation",
+			StabilityLevel: metrics.ALPHA,
+			Buckets:        []float64{0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10, 25},
```
a7i, line 57: is it fair to say that a loop for large clusters can certainly take longer than 25 seconds?
```suggestion
			Buckets:        []float64{0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10, 25, 50, 120},
```

pkg/descheduler/descheduler.go
```diff
@@ -386,8 +389,10 @@ func RunDeschedulerStrategies(ctx context.Context, rs *options.DeschedulerServer
 			// handle or function which the Evictor has access to. For migration/in-progress framework
 			// work, we are currently passing this via context. To be removed
 			// (See discussion thread https://github.com/kubernetes-sigs/descheduler/pull/885#discussion_r919962292)
+			strategyStart := time.Now()
 			childCtx := context.WithValue(ctx, "strategyName", pg.Name())
 			status := pg.Deschedule(childCtx, nodes)
+			metrics.DeschedulerStrategyDuration.With(map[string]string{"strategy": pg.Name()}).Observe(time.Since(strategyStart).Seconds())
```
a7i, line 395: if multiple profiles use the same strategy, then we will have multiple entries for it. I think that's still ok given that we run them all sequentially.

metrics/metrics.go
harshanarayana: I think that would be a fair thing to say based on my previous experience. Will update this accordingly. Should I add a bucket for 250, 500 too ?

metrics/metrics.go
harshanarayana: Ack

pkg/descheduler/descheduler.go
harshanarayana: Actually, good that you brought this up. I think it is better to track the profile along with strategy so that it can be a bit useful in the form of metrics. But I am not sure how drastic the cardinality growth might be,

metrics/metrics.go
```diff
@@ -48,9 +48,29 @@ var (
 		},
 	)
 
+	DeschedulerLoopDuration = metrics.NewHistogramVec(
+		&metrics.HistogramOpts{
+			Subsystem:      DeschedulerSubsystem,
+			Name:           "descheduler_loop_duration_seconds",
+			Help:           "Time taken to complete a strategy descheduling operation",
```
damemi, line 55:
```suggestion
			Help:           "Time taken to complete a full descheduling cycle",
```

metrics/metrics.go
```diff
@@ -48,9 +48,29 @@ var (
...
+		}, []string{})
+
+	DeschedulerStrategyDuration = metrics.NewHistogramVec(
+		&metrics.HistogramOpts{
+			Subsystem:      DeschedulerSubsystem,
+			Name:           "descheduler_strategy_duration_seconds",
+			Help:           "Time taken to complete Each strategy of the descheduling operation",
+			StabilityLevel: metrics.ALPHA,
+			Buckets:        []float64{0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10, 25, 50, 100},
+		}, []string{"strategy", "profile"})
```
damemi, line 67: wondering how useful the `profile` label is here. I suppose different profiles can have different strategy options, affecting the run time of each. In that case though, maybe it could be useful in the overall cycle latency metric as well.

pkg/descheduler/descheduler.go
```diff
@@ -49,6 +50,16 @@ import (
...
 
+type enabledDeschedulePluginEntry struct {
+	Plugin  framework.DeschedulePlugin
+	Profile string
+}
+
+type enabledBalancePluginEntry struct {
+	Plugin  framework.BalancePlugin
+	Profile string
+}
```
damemi, line 61: these types feel redundant, because I would expect the plugin and profile name to both be available at the time the metric is recorded.

from what I can tell, we just don't have access to the profile at plugin runtime. So if we didn't record any `profile` label, would these new structs be necessary?

metrics/metrics.go
harshanarayana: Ack

metrics/metrics.go
harshanarayana: There is one tracked for the entire loop and one for each strategy and profile combo. Do you want to track one for each profile as well ?

pkg/descheduler/descheduler.go
harshanarayana: If we are not to record the metrics around profile or the metrics for strategy is not tagged along with the profile, then yes, these structs are redundant.

metrics/metrics.go
harshanarayana: Right now we track to things in this PR. One that tracks the entire loop itself. And one that tracks the duration for each strategy and profile combination. We can track one for just the profile itself as well may be ? That will give you breakdown of the loop vs profile, and profile vs strategy as well. @damemi

harshanarayana:
@damemi  PTAL when possible. Comments have been addressed.

metrics/metrics.go
damemi: Ah I see. In the future, I think it could be useful to break the full loop duration down by profile as well, but I suppose if you're trying to debug why one profile takes longer than another then you really are looking for the strategy breakdown anyway. This looks good to me

pkg/descheduler/descheduler.go
damemi: I see, @knelasevero or @ingvagabund had we brought up the possibility of making profile name available to strategies?

This approach will work as it's just an implementation detail, but I'm curious if it's refactorable at some point

a7i:
tide/merge-method-squash

a7i:
Thank you @harshanarayana 🥇 
/lgtm

pkg/descheduler/descheduler.go
ingvagabund: From the user perspective it's better to have plugins profile unaware. A profile is a high level concept built on top of plugins. It's ok to pass a profile name (or anything else related to profiles) as long as the values are used only for monitoring, logging or decision making logic around plugins.

harshanarayana: @ingvagabund  Right now that is all that is being done in terms of Profiles. Their names are being used with strategy to form a unique combination that one can query from metrics as these are available as labels on them

PR #1042, opened 2023-01-11
Revert "Split CMD into ENTRYPOINT and CMD for Dockerfile"

a7i:
Reverts kubernetes-sigs/descheduler#995


damemi (commented):
/approve
thanks @a7i for the revert, and @Bobonium for the original change. Sorry we need to revert it for now, but we'll look more at this change for the next release.

PR #1049, opened 2023-01-27
v1alpha2 docs

knelasevero:
This PR is just changing Readme and other places in the doc files to reference the new v1alpha2 configuration. I am keeping the deprecated v1alpha1 part of the readme in a new file that we can remove after we announce the official removal of v1alpha1 altogether.

It actually took me a while because I was trying to figure out how to put our docs in a [mkdocs](https://www.mkdocs.org/) setup with versioned pages/docs (dropdown with versions pointing to previous versions of the software and the docs). But it ended up eating too much of my time and **I gave up** (for now). Also to do that we would have to have access to the settings tab of the project to setup github pages. We can see how to do that later.

Please let me know if I have missed any reference to v1alpha1 or if something is incorrectly explained.


knelasevero:
Removed TOC verification and generation as github has this natively for a while already:

image

README.md
```diff
@@ -26,44 +26,6 @@ Descheduler, based on its policy, finds pods that can be moved and evicts them.
 note, in current implementation, descheduler does not schedule replacement of evicted pods
 but relies on the default scheduler for that.
 
-Table of Contents
```
damemi, line 29: I would prefer to keep the markdown TOC, at least until github's new feature is more widely known. Plaintext links work on all environments

knelasevero: Something else we can do is point to the button in the readme. 

with regards to

> Plaintext links work on all environments

Can you think of an environment where the github native toc would not work?

knelasevero:
In any case, this looks good to me

/lgtm

README.md
knelasevero: Ok, I had a look around other kubernetes-sigs projects to see what they are doing, and I gathered some that fall into the same place as us, so they have these characteristics: 
- kubernetes-sigs or k/ project
- Project is big enough
- Project does not have a website yet, so Readme is heavy/long enough

https://github.com/kubernetes-sigs/external-dns

https://github.com/kubernetes-sigs/metrics-server

https://github.com/kubernetes-sigs/kubefed

https://github.com/kubernetes-sigs/prometheus-adapter

https://github.com/kubernetes/kompose

:point_down:  The only two following above requirements with a TOC that I could find)

https://github.com/kubernetes/client-go 

https://github.com/kubernetes/kube-state-metrics

knelasevero: external-dns has this funny thing "pointing" to the gh-toc:

image


Which I don't like, I think. Maybe writing something like "Click at the [bullet list icon] at the top left corner of the Readme visualization for the github generated table of contents"

damemi: Thanks for checking what other projects are doing for reference, @knelasevero. Maybe it's just a personal preference of mine, so this isn't a blocker for me. We were talking about moving to a dedicated docs site anyway

examples/policy.yaml
```diff
@@ -1,29 +1,39 @@
...
+        includeSoftConstraints: true
+    plugins:
+      evict:
+        enabled:
+          - "DefaultEvictor"
+      deschedule:
+        enabled:
+          - "RemovePodsViolatingInterPodAntiAffinity"
+          - "RemovePodsHavingTooManyRestarts"
+          - ""
```
a7i, line 34:
```suggestion
```

examples/pod-life-time.yml
```diff
@@ -1,11 +1,21 @@
...
         - "PodInitializing"
+    plugins:
+      evict:
+        enabled:
+          - "DefaultEvictor"
+      deschedule:
+        enabled:
+          - "PodLifeTime"
+
+
```
a7i, line 21: nit: not required
```suggestion
```

README.md
```diff
@@ -126,47 +88,84 @@ kustomize build 'github.com/kubernetes-sigs/descheduler/kubernetes/deployment?re
 
 See the [user guide](docs/user-guide.md) in the `/docs` directory.
 
-## Policy and Strategies
+## Policy, Default Evictor and Strategy plugins
+
+**⚠️ v1alpha1 configuration is still suported, but deprecated (and soon will be removed). Please consider migrating to v1alpha2 (described bellow). For previous v1alpha1 documentation go to docs/deprecated/v1alpha1.md**
+
+The Descheduler Policy is configurable and includes default eviction strategy plugins that can be enabled or disabled. It includes a common eviction configuration at the top level, as well as configuration from the Evictor plugin (Default Evictor, if not specified otherwise). Top-level configuration and Evictor plugin configuration are applied to all evictions.
```
ingvagabund, line 95: There's only one default eviction strategy/plugin. By "default eviction strategy plugins", are you referring to various params of the default evictor plugin? Or, are you providing a generic description for custom eviction plugins which might eventually become default plugins?

README.md
```diff
@@ -126,47 +88,84 @@ kustomize build 'github.com/kubernetes-sigs/descheduler/kubernetes/deployment?re
...
 
-Descheduler's policy is configurable and includes strategies that can be enabled or disabled. By default, all strategies are enabled.
+### Top Level configuration
+
+These are top level keys in the Descheduler Policy that you can use to configure all evictions.
+
+| Name | Default Value | Description |
```
ingvagabund, line 101: Let's add another `Type` column saying type of a param. E.g. string, []string, int, ...

README.md
```diff
@@ -126,47 +88,84 @@ kustomize build 'github.com/kubernetes-sigs/descheduler/kubernetes/deployment?re
...
+| `maxNoOfPodsToEvictPerNamespace` | `nil` | maximum number of pods evicted from each namespace (summed through all strategies) |
+
+### Evictor Plugion configuration (Default Evictor)
+
+The Default Evictor Plugin is used by default for filtering pods before processing them in an eviction strategy plugin, or for applying a PreEvictionFilter of pods before eviction. You can also create your own Evictor Plugin or use the Default one provided by Descheduler.  Other uses for the Evictor plugin can be to sort, filter, validate or group pods by different criteria, and that's why this is handled by a plugin and not configured in the top level config.
 
-The policy includes a common configuration that applies to all the strategies:
 | Name | Default Value | Description |
 |------|---------------|-------------|
 | `nodeSelector` | `nil` | limiting the nodes which are processed |
```
ingvagabund, line 113: Worth mentioning the param is used only when `nodeFit=true` and only by the `PreEvictionFilter` EP

README.md
```diff
@@ -126,47 +88,84 @@ kustomize build 'github.com/kubernetes-sigs/descheduler/kubernetes/deployment?re
...
 kind: "DeschedulerPolicy"
-nodeSelector: prod=dev
-evictFailedBarePods: false
-evictLocalStoragePods: true
-evictSystemCriticalPods: true
-maxNoOfPodsToEvictPerNode: 40
-ignorePvcPods: false
-strategies:
-  ...
+nodeSelector: "node1" # you don't need to set this, if not set all will be processed
```
ingvagabund, line 134: I think the smallest valid selector is in a `key=value` form. E.g. `node=node1`

README.md
```diff
@@ -126,47 +88,84 @@ kustomize build 'github.com/kubernetes-sigs/descheduler/kubernetes/deployment?re
...
+          - ...
+      [...]
 ```
 
 The following diagram provides a visualization of most of the strategies to help
 categorize how strategies fit together.
 
 ![Strategies diagram](strategies_diagram.png)
 
+Next sections will go over different eviction strategy plugins. Right now we group them by being Deschedule plugins or Balance plugins (more types could be added).
```
ingvagabund, line 164: "being Deschedule plugins" => "implementing Deschedule extension point"
similar for the Balance

README.md
```diff
@@ -126,47 +88,84 @@ kustomize build 'github.com/kubernetes-sigs/descheduler/kubernetes/deployment?re
...
+
 ### RemoveDuplicates
 
-This strategy makes sure that there is only one pod associated with a ReplicaSet (RS),
+This eviction strategy plugin makes sure that there is only one pod associated with a ReplicaSet (RS),
```
ingvagabund, line 168: The strategy does not have any eviction-based extension point on its own. So the wording needs to stay the same. "strategy" -> "plugin" will do.

README.md
```diff
@@ -178,27 +177,34 @@ has any of these `Kind`s listed as an `OwnerRef`, that pod will not be considere
 pods created by Deployments are considered for eviction by this strategy. The `excludeOwnerKinds` parameter
 should include `ReplicaSet` to have pods created by Deployments excluded.
 
+This is a Balance plugin.
```
ingvagabund, line 180: plugin -> extension point
The same comment applies for the remaining occurrences. Worth mentioning a list of extension points each plugin implements in a separate table maybe?

ingvagabund (changes requested):

README.md
knelasevero: Eviction strategy plugins to me are the stretagies themselves. 

Evictor plugins are the ones configuring filters and sorters and etc (and we currently only have the default evictor as this type)

Maybe what I am calling Eviction strategy plugins we can call just strategy plugins? We need to call them something else than just plugins.

README.md
knelasevero: I removed these phrases and included a table. Extension point is wording that would prefer to avoid in a user facing readme, but I think it is fine in the table like that

README.md
knelasevero: I really don´t wanna just call any plugin just plugin... Specially since we have already different groups of plugins that are used in different ways. I think it is fine if we call them strategy plugins for now (just to differentiate from evictor plugin) in the docs. I agree with removing the word eviction from these strategy plugins, just pushed that. If a better nomenclature comes at some point we can change it then.

README.md
knelasevero: Added the table and expanded this explanation

README.md
knelasevero: :heavy_check_mark:

README.md
knelasevero: :heavy_check_mark:

README.md
knelasevero: :heavy_check_mark:

README.md
ingvagabund: Descheduling plugins/strategies. Eviction plugins corresponds too much with implementation of the Evict extension point.

README.md
ingvagabund: Eviction strategy then

ingvagabund:
@knelasevero you are right. The strategy/plugin/eviction naming is getting overloaded. "Eviction strategy" is the closest term I can think of in the light of the new naming conventions. Feel free to oppose.

README.md
knelasevero: Are you talking about the strategies, or the evictor plugin (default evictor for now)?

knelasevero:
> @knelasevero you are right. The strategy/plugin/eviction naming is getting overloaded. "Eviction strategy" is the closest term I can think of in the light of the new naming conventions. Feel free to oppose.

But now I am not sure you are talking about the strategies or the default evictor.

What we once called strategies, I am now calling "strategy plugins"

and the new plugin for filters/sorters/etc I am calling Evictor plugin, as per the design. 

Is that acceptable?

knelasevero:
Here is the direct visualization of the new readme: https://github.com/kubernetes-sigs/descheduler/blob/963ba698b46b085e259c5270c1fbfcc490d91197/README.md

JaneLiuL (commented):
/lgtm

PR #1051, opened 2023-01-30
Fix v1alpha1 conversion to use universal decoder

gustavomfc:
Remove error handling with string comparison and use UniversalDecoder to convert v1alpha1.

Thanks a lot with all the help @knelasevero

### DIFF
--- a/pkg/descheduler/descheduler_test.go
+++ b/pkg/descheduler/descheduler_test.go
@@ -9,6 +9,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	policy "k8s.io/api/policy/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/conversion"
 	"k8s.io/apimachinery/pkg/runtime"
 	fakeclientset "k8s.io/client-go/kubernetes/fake"
 	core "k8s.io/client-go/testing"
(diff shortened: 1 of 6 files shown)


damemi (commented):
/ok-to-test

knelasevero:
@damemi how do we refresh the cla signing thing on the same PR?

damemi:
@knelasevero looks like the commits aren't properly associated with the github account? Maybe a git or github config setting, the bot links to this page for help: https://docs.github.com/en/pull-requests/committing-changes-to-your-project/troubleshooting-commits/why-are-my-commits-linked-to-the-wrong-user

pkg/descheduler/descheduler_test.go
```diff
@@ -74,7 +91,9 @@ func TestTaintsUpdated(t *testing.T) {
 	var evictedPods []string
 	client.PrependReactor("create", "pods", podEvictionReactionFuc(&evictedPods))
 
-	internalDeschedulerPolicy, err := v1alpha1.V1alpha1ToInternal(client, dp, pluginregistry.PluginRegistry)
+	internalDeschedulerPolicy := &api.DeschedulerPolicy{}
+	scope := scope{}
+	err = v1alpha1.V1alpha1ToInternal(dp, pluginregistry.PluginRegistry, internalDeschedulerPolicy, scope)
```
knelasevero, line 96: Other option is to make dp already be declared in our internal type so the test doesn't have to convert it (since the test is testing something else).

But I guess it is also nice we have the conversion being called at some other places so it is validates in other contexts.

Optional change if you think it makes sense

knelasevero:
Thanks for this fix @gustavomfc :heart:

knelasevero:
@a7i @damemi @ingvagabund @JaneLiuL do you think we can merge this one?

pkg/descheduler/policyconfig.go
```diff
@@ -122,6 +101,17 @@ func setDefaultEvictor(profile api.Profile) api.Profile {
 		profile.Plugins.Evict.Enabled = append(profile.Plugins.Evict.Enabled, defaultevictor.PluginName)
 		profile.PluginConfigs = append(profile.PluginConfigs, newPluginConfig)
 	}
+	var thresholdPriority int32
+	var err error
+	defaultevictorPluginConfig, idx := GetPluginConfig(defaultevictor.PluginName, profile.PluginConfigs)
+	if defaultevictorPluginConfig != nil {
+		thresholdPriority, err = utils.GetPriorityFromStrategyParams(context.TODO(), client, defaultevictorPluginConfig.Args.(*defaultevictor.DefaultEvictorArgs).PriorityThreshold)
```
knelasevero, line 108: Main thing to be able to use universal decoder for both conversions was moving this here, and not needing a client in the conversion.go file

ingvagabund:
I like it very much
/approve

PR #1057, opened 2023-02-15
check pod number

xiaoanyunfei:
`fitsRequest` don't check pod num


ingvagabund:
@xiaoanyunfei thank you for improving the check. Would you also please write a new unit test for it?

xiaoanyunfei:
> @xiaoanyunfei thank you for improving the check. Would you also please write a new unit test for it?

done

damemi (commented):
/ok-to-test

PR #1059, opened 2023-02-16
remove some logic useless

lucming:
After the skipNode, the outer loop is not doing anything, there is no need to execute continue.


damemi (commented):
/ok-to-test

damemi (commented):
/approve

JaneLiuL (commented):
/lgtm
thanks for your contribute~~

PR #1061, opened 2023-02-22
fix:eviction groupversion not support v1beta1

lucming:
question:
when running descheduler in k8s cluster(1.19), we can see some error logs, like this:
```bash
I0222 02:55:05.474234 7 removeduplicates.go:203] "Average occurrence per node" node="stark12.add.bjyt.qihoo.net" ownerKey={namespace:kstone kind:TApp name:test131-etcd imagesHash:quay.io/coreos/etcd:v3.4.13} avg=1

E0222 02:55:05.475172 7 evictions.go:143] "Error evicting pod" err="Eviction in version \"v1\" cannot be handled as a Eviction: no kind \"Eviction\" is registered for version \"policy/v1\" in scheme \"k8s.io/kubernetes/pkg/api/legacyscheme/scheme.go:30\"" pod="kstone/test131-etcd-2" reason=""

I0222 02:55:05.475215 7 descheduler.go:424] "Number of evicted pods" totalEvicted=0
```
and in this scenario, we can found :
截屏2023-02-22 18 45 28

the reason for this is that version 1.19 of k8s does not support the eviction resource whose groupversion is v1

### DIFF
--- a/pkg/descheduler/evictions/evictions.go
+++ b/pkg/descheduler/evictions/evictions.go
@@ -19,9 +19,11 @@ package evictions
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	v1 "k8s.io/api/core/v1"
 	policy "k8s.io/api/policy/v1"
+	policyv1beta1 "k8s.io/api/policy/v1beta1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clientset "k8s.io/client-go/kubernetes"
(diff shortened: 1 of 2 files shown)


damemi (commented):
/hold
@lucming what version of descheduler are you running? For kubernetes 1.19 the latest supported version of descheduler is v0.22 which does [use the v1beta1 eviction api](https://github.com/kubernetes-sigs/descheduler/blob/v0.22.0/pkg/descheduler/evictions/evictions.go#L158)

(see the [compatibility matrix](https://github.com/kubernetes-sigs/descheduler#compatibility-matrix)

lucming:
> /hold @lucming what version of descheduler are you running? For kubernetes 1.19 the latest supported version of descheduler is v0.22 which does [use the v1beta1 eviction api](https://github.com/kubernetes-sigs/descheduler/blob/v0.22.0/pkg/descheduler/evictions/evictions.go#L158)
> 
> (see the [compatibility matrix](https://github.com/kubernetes-sigs/descheduler#compatibility-matrix)

Oh, i see, I made the image using the master branch.
However, different versions of k8s may support multiple versions of the `eviction` resource and others, and although there is only one `storageversion`, is it possible to think about compatibility? If we're not sure which version of the resource to use, perhaps we could consider using `dynamicClient` for that.

Issue #1062, opened 2023-02-22
Add k8s compatibility warnings to logs


damemi:
We document our k8s version compatibility in the Readme, but it would be helpful to users to have this logged directly in the descheduler when it's running. Something like: `Warning: Descheduler version XX is not supported on your version of Kubernetes (YY). See compatibility docs for more info: https://github.com/kubernetes-sigs/descheduler#compatibility-matrix`

Though I don't know if it's possible to self-detect what version of kubernetes we're running on, which is necessary for this

a7i:
Good idea!

We should be able to use the `discoveryClient` like [here](https://github.com/kubernetes/kubernetes/blob/v1.26.1/staging/src/k8s.io/kubectl/pkg/cmd/version/version.go#L139)

And use the skew logic [here](https://github.com/kubernetes/kubernetes/blob/v1.26.1/staging/src/k8s.io/kubectl/pkg/cmd/version/skew_warning.go#L34)

JaneLiuL:
may I have a try on this issue?
/assign

damemi:
Thanks @JaneLiuL!

damemi:
Fixed by https://github.com/kubernetes-sigs/descheduler/pull/1063
/close

PR #1063, opened 2023-02-23
Add k8s compatibility warnings to logs

JaneLiuL:
enhance issue : https://github.com/kubernetes-sigs/descheduler/issues/1062

i have already test on my local kubernetes platform

image

image

### DIFF
--- a/pkg/api/v1alpha1/conversion.go
+++ b/pkg/api/v1alpha1/conversion.go
@@ -242,7 +242,7 @@ func enableProfilePluginsByType(profilePlugins api.Plugins, pluginInstance frame
 func checkBalance(profilePlugins api.Plugins, pluginInstance framework.Plugin, pluginConfig *api.PluginConfig) api.Plugins {
 	_, ok := pluginInstance.(framework.BalancePlugin)
 	if ok {
-		klog.V(3).Info("converting Balance plugin: %s", pluginInstance.Name())
+		klog.V(3).Infof("converting Balance plugin: %s", pluginInstance.Name())
 		profilePlugins.Balance.Enabled = []string{pluginConfig.Name}
 	}
 	return profilePlugins
@@ -251,7 +251,7 @@ func checkBalance(profilePlugins api.Plugins, pluginInstance framework.Plugin, p
 func checkDeschedule(profilePlugins api.Plugins, pluginInstance framework.Plugin, pluginConfig *api.PluginConfig) api.Plugins {
 	_, ok := pluginInstance.(framework.DeschedulePlugin)
 	if ok {
-		klog.V(3).Info("converting Deschedule plugin: %s", pluginInstance.Name())
+		klog.V(3).Infof("converting Deschedule plugin: %s", pluginInstance.Name())
 		profilePlugins.Deschedule.Enabled = []string{pluginConfig.Name}
 	}
 	return profilePlugins
--- a/pkg/descheduler/descheduler.go
+++ b/pkg/descheduler/descheduler.go
@@ -19,6 +19,9 @@ package descheduler
 import (
 	"context"
 	"fmt"
+	"math"
+	"strconv"
+	"strings"
 	"time"
 
 	componentbaseconfig "k8s.io/component-base/config"
@@ -49,6 +52,7 @@ import (
 	"sigs.k8s.io/descheduler/pkg/framework/pluginregistry"
 	"sigs.k8s.io/descheduler/pkg/framework/plugins/defaultevictor"
 	"sigs.k8s.io/descheduler/pkg/utils"
+	"sigs.k8s.io/descheduler/pkg/version"
 )
 
 type enabledDeschedulePluginEntry struct {
@@ -83,6 +87,9 @@ func Run(ctx context.Context, rs *options.DeschedulerServer) error {
 		return fmt.Errorf("deschedulerPolicy is nil")
 	}
 
+	// Add k8s compatibility warnings to logs
+	versionCompatibilityCheck(rs)
+
 	evictionPolicyGroupVersion, err := eutils.SupportEviction(rs.Client)
 	if err != nil || len(evictionPolicyGroupVersion) == 0 {
 		return err
@@ -108,6 +115,29 @@ func Run(ctx context.Context, rs *options.DeschedulerServer) error {
 	}
 
 	return runFn()
+}
+
+func versionCompatibilityCheck(rs *options.DeschedulerServer) {
+	serverVersion, serverErr := rs.Client.Discovery().ServerVersion()
+	if serverErr != nil {
+		klog.V(1).InfoS("Warning: Get Kubernetes server version fail")
+		return
+	}
+
+	deschedulerMinorVersion := strings.Split(version.Get().Minor, ".")[0]
+	deschedulerMinorVersionFloat, err := strconv.ParseFloat(deschedulerMinorVersion, 64)
+	if err != nil {
+		klog.Warning("Warning: Convert Descheduler minor version to float fail")
+	}
+
+	kubernetesMinorVersionFloat, err := strconv.ParseFloat(serverVersion.Minor, 64)
+	if err != nil {
+		klog.Warning("Warning: Convert Kubernetes server minor version to float fail")
+	}
+
+	if math.Abs(deschedulerMinorVersionFloat-kubernetesMinorVersionFloat) > 3 {
+		klog.Warningf("Warning: Descheduler minor version %v is not supported on your version of Kubernetes %v.%v. See compatibility docs for more info: https://github.com/kubernetes-sigs/descheduler#compatibility-matrix", deschedulerMinorVersion, serverVersion.Major, serverVersion.Minor)
+	}
 }
 
 func cachedClient(


pkg/descheduler/descheduler.go
```diff
@@ -107,6 +109,15 @@ func Run(ctx context.Context, rs *options.DeschedulerServer) error {
 		return nil
 	}
 
+	// Add k8s compatibility warnings to logs
+	serverVersion, serverErr := rs.Client.Discovery().ServerVersion()
+	if serverErr != nil {
+		klog.V(1).InfoS("Warning: Get Kubernetes server version fail")
+		return serverErr
+	}
+	if descheduler_minor_version != serverVersion.Minor {
```
a7i, line 118: Shouldn't we look at the version skew? i.e. `For example descheduler v0.18 should work with k8s v1.18, v1.17, and v1.16.`

pkg/descheduler/descheduler.go
```diff
@@ -107,6 +109,15 @@ func Run(ctx context.Context, rs *options.DeschedulerServer) error {
...
+		klog.V(1).InfoS(fmt.Sprintf("Warning: Descheduler minor version %s is not supported on your version of Kubernetes %s.%s. See compatibility docs for more info: https://github.com/kubernetes-sigs/descheduler#compatibility-matrix", descheduler_minor_version, serverVersion.Major, serverVersion.Minor))
```
a7i, line 119: thoughts on klog.Warning?

pkg/descheduler/descheduler.go
```diff
@@ -51,6 +51,8 @@ import (
 	"sigs.k8s.io/descheduler/pkg/utils"
 )
 
+const descheduler_minor_version = "26"
```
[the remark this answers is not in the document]
a7i: Can we get the version from [pkg/version](https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/version/version.go#L56)?

pkg/descheduler/descheduler.go
```diff
@@ -107,6 +109,15 @@ func Run(ctx context.Context, rs *options.DeschedulerServer) error {
 		return nil
 	}
 
+	// Add k8s compatibility warnings to logs
+	serverVersion, serverErr := rs.Client.Discovery().ServerVersion()
+	if serverErr != nil {
+		klog.V(1).InfoS("Warning: Get Kubernetes server version fail")
```
a7i, line 115: Same here. InfoS and then followed by Warning in text. Might as well use klog Warning?

a7i:
Can we add a unit test for this? To make sure that we stay within the skew

pkg/descheduler/descheduler.go
JaneLiuL: but i check the document, it said as below, and at least what I add this logic is after descheduler 0.18 version. so for the judgement, it only check the minor version is it the same
```
“Starting with descheduler release v0.18 the minor version of descheduler matches the minor version of the k8s client packages that it is compiled with.”
```

pkg/descheduler/descheduler.go
JaneLiuL: before i did want to use klog.warning, but i check the code here: https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/descheduler/descheduler.go#L100
it still use `klog.V(1).InfoS("Warning: DryRun is set to True. You need to disable it to use Leader Election.")` , ..that's why..

pkg/descheduler/descheduler.go
JaneLiuL: the same reason as above. in the https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/descheduler/descheduler.go, all of them are using `klog.V(1).InfoS("Warning:`. so i just align.
@damemi  @a7i  please confirm is it i need to replace all the `klog.V(1).InfoS("Warning:` to `klog.Warning()`.

pkg/descheduler/descheduler.go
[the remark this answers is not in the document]
JaneLiuL: hi. yes. i do agree if use const then need to manual update each time release.
i already change as @a7i  mention. please check now. :)

pkg/descheduler/descheduler.go
damemi: We don't need to update all cases of that in this PR, but we can start with just this change using `Warning()` instead of `InfoS`.

If you'd like, it would be nice to open another PR to find other instances of this that can be updated

pkg/descheduler/descheduler.go
damemi: That's right, but we want to show the compatibility/support level which is `At this time descheduler does not have a hard dependency to a specific k8s release. However a particular descheduler release is only tested against the three latest k8s minor versions`

JaneLiuL:
also fix another log issue as below
image

pkg/descheduler/descheduler.go
JaneLiuL: thanks for notice, i already update the logic. please check :)

pkg/descheduler/descheduler.go
JaneLiuL: sure, i already change to `Warning()` instead of `InfoS`

pkg/api/v1alpha1/conversion.go
```diff
@@ -242,7 +242,7 @@ func enableProfilePluginsByType(profilePlugins api.Plugins, pluginInstance frame
 func checkBalance(profilePlugins api.Plugins, pluginInstance framework.Plugin, pluginConfig *api.PluginConfig) api.Plugins {
 	_, ok := pluginInstance.(framework.BalancePlugin)
 	if ok {
-		klog.V(3).Info("converting Balance plugin: %s", pluginInstance.Name())
+		klog.V(3).Info("converting Balance plugin: ", pluginInstance.Name())
```
damemi, line 245: I think these were meant to be `Infof()` with the `%s`?

pkg/descheduler/descheduler.go
```diff
@@ -83,6 +86,28 @@ func Run(ctx context.Context, rs *options.DeschedulerServer) error {
...
+		klog.Warning("Warning: Convert Kubernetes server minor version to int fail")
+	}
+	kubernetesMinorVersionInt, err := strconv.Atoi(serverVersion.Minor)
+	if err != nil {
+		klog.Warning("Warning: Convert Kubernetes server minor version to int fail")
+	}
+	if deschedulerMinorVersion == serverVersion.Minor || deschedulerMinorVersionInt-kubernetesMinorVersionInt < 3 {
+		klog.V(1).InfoS("Info: Descheduler minor version can be support on your version on Kubernetes")
+	} else {
+		klog.Warning(fmt.Sprintf("Warning: Descheduler minor version %s is not supported on your version of Kubernetes %s.%s. See compatibility docs for more info: https://github.com/kubernetes-sigs/descheduler#compatibility-matrix", deschedulerMinorVersion, serverVersion.Major, serverVersion.Minor))
```
a7i, line 108:
```suggestion
		klog.Warningf("Warning: Descheduler minor version %s is not supported on your version of Kubernetes %s.%s. See compatibility docs for more info: https://github.com/kubernetes-sigs/descheduler#compatibility-matrix", deschedulerMinorVersion, serverVersion.Major, serverVersion.Minor)
```

pkg/descheduler/descheduler.go
```diff
@@ -83,6 +86,28 @@ func Run(ctx context.Context, rs *options.DeschedulerServer) error {
...
+	deschedulerMinorVersionInt, err := strconv.Atoi(deschedulerMinorVersion)
+	if err != nil {
+		klog.Warning("Warning: Convert Kubernetes server minor version to int fail")
+	}
+	kubernetesMinorVersionInt, err := strconv.Atoi(serverVersion.Minor)
+	if err != nil {
+		klog.Warning("Warning: Convert Kubernetes server minor version to int fail")
+	}
+	if deschedulerMinorVersion == serverVersion.Minor || deschedulerMinorVersionInt-kubernetesMinorVersionInt < 3 {
+		klog.V(1).InfoS("Info: Descheduler minor version can be support on your version on Kubernetes")
```
a7i, line 106:
```
		klog.V(1).Info("Info: Descheduler minor version can be support on your version on Kubernetes")
```

pkg/descheduler/descheduler.go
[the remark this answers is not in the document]
JaneLiuL: if no comment i will click "resolve conversation" for this part.

pkg/descheduler/descheduler.go
JaneLiuL: if no comment i will click "resolve conversation" for this part.

pkg/api/v1alpha1/conversion.go
JaneLiuL: commit now, please check

pkg/descheduler/descheduler.go
JaneLiuL: commit now, please check

pkg/descheduler/descheduler.go
JaneLiuL: commit now

pkg/descheduler/descheduler.go
```diff
@@ -83,6 +86,28 @@ func Run(ctx context.Context, rs *options.DeschedulerServer) error {
...
+	deschedulerMinorVersion := strings.Split(version.Get().Minor, ".")[0]
+	deschedulerMinorVersionInt, err := strconv.Atoi(deschedulerMinorVersion)
+	if err != nil {
+		klog.Warning("Warning: Convert Kubernetes server minor version to int fail")
+	}
+	kubernetesMinorVersionInt, err := strconv.Atoi(serverVersion.Minor)
+	if err != nil {
+		klog.Warning("Warning: Convert Kubernetes server minor version to int fail")
+	}
+	if deschedulerMinorVersion == serverVersion.Minor || deschedulerMinorVersionInt-kubernetesMinorVersionInt < 3 {
```
damemi, line 105:
```suggestion
	if deschedulerMinorVersionInt-kubernetesMinorVersionInt < 3 {
```
don't need to check if they're equal if we're also checking that their diff is < 3 :)

damemi: actually let's change this to check `if > 3` and drop the V(1) message about descheduler being supported on your version. we only need to log a message/warning if the descheduler version is not supported

JaneLiuL: sure, commit as your suggestion~

damemi (commented):
/approve
thanks @JaneLiuL! This will hopefully help a lot of users

pkg/descheduler/descheduler.go
a7i: may want to make that `Abs`

Descheduler 0.21 is not supported on Kubernetes 1.17 or 1.25

JaneLiuL: thanks a lot~~ good finding~. i will update now

JaneLiuL: update done

JaneLiuL:
@a7i  could you review and give me a /lgtm if possible?
i just back, and so sorry for my mom sick and i leave for one week ago.

a7i:
> * can we move all the logic into its own function? when I look at the `Run` function code, I don't really need to go through all 20+ lines of code. But rather just see a call to "ensure k8s version compatibility"
> * Can we add a test for the new function?

Sorry to hear about that, hope all is well now!

What are your thoughts on the question above? i.e.

>can we move all the logic into its own function? when I look at the Run function code, I don't really need to go through all 20+ lines of code. But rather just see a call to "ensure k8s version compatibility"

>Can we add a test for the new function?

JaneLiuL:
@a7i  i already separate a function, and i do test on local latop for real test done, see the picture as below(the first picture is k8s version 1.25, the second picture is k8s version1.22).  sorry for i don't see any testcase to cover the log in descheduler, maybe when the pr merge, could  you help to add test function for that?
image


image

a7i:
/lgtm

Thank you, I will create a follow-up PR with the test

Issue #1065, opened 2023-02-28
RunAsUser and RunAsGroup not exposed from helm chart deployment


gdasson:
**Is your feature request related to a problem? Please describe.**
The securityContext defined [here](https://github.com/kubernetes-sigs/descheduler/blob/master/charts/descheduler/templates/deployment.yaml#L67-L74) in the helm chart template for deployment doesn't have runAsUser and runAsGroup defined. As a security best practice, it would be great if we could run the processes as non-root. In fact, a user of 1000 is defined in Dockerfile [here](https://github.com/kubernetes-sigs/descheduler/blob/master/Dockerfile#L26) but not defined in the template.

**Describe the solution you'd like**
An ideal solution is to offer the securityContext block from values.yaml instead of being hardcoded in the deployment template

**Describe alternatives you've considered**
Not defining a runAsUser and runAsGroup (which is not ideal in our case)

**What version of descheduler are you using?**
descheduler version: 0.26.0

**Additional context**
None

PR #1066, opened 2023-03-01
expose security context from helm chart

JaneLiuL:
enhance: https://github.com/kubernetes-sigs/descheduler/issues/1065

### DIFF
--- a/charts/descheduler/values.yaml
+++ b/charts/descheduler/values.yaml
@@ -22,6 +22,16 @@ resources:
   #   cpu: 100m
   #   memory: 128Mi
 
+securityContext:
+  allowPrivilegeEscalation: false
+  capabilities:
+    drop:
+      - ALL
+  privileged: false
+  readOnlyRootFilesystem: true
+  runAsNonRoot: true
+  runAsUser: 1000
+
 nameOverride: ""
 fullnameOverride: ""
 
--- a/charts/descheduler/templates/cronjob.yaml
+++ b/charts/descheduler/templates/cronjob.yaml
@@ -81,13 +81,7 @@ spec:
               resources:
                 {{- toYaml .Values.resources | nindent 16 }}
               securityContext:
-                allowPrivilegeEscalation: false
-                capabilities:
-                  drop:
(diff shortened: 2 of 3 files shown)


a7i (approved):
/lgtm

🏆

charts/descheduler/values.yaml
```diff
@@ -22,6 +22,16 @@ resources:
...
 
+securityContext:
+  allowPrivilegeEscalation: false
+  capabilities:
+    drop:
+      - ALL
+  privileged: false
+  readOnlyRootFilesystem: true
+  runAsNonRoot: true
+  runAsUser: 1000
```
a7i, line 33: should we remove the `USER 1000` from the Dockerfile?

JaneLiuL: that's a good question, let's discuss here, i check the the file https://github.com/kubernetes-sigs/descheduler/blob/master/kubernetes/deployment/deployment.yaml#L48 it does not have runasuser, so if we remove from the dockerfile, then i will add runasuser into the deployment/deployment.yaml .

a7i: That makes sense. I personally like the idea of controlling the `runAsUser` in the manifest, if possible.

JaneLiuL: if so, then i will try to remove runasuser from dockerfile, and put it into the manifest.

a7i: @damemi thoughts? Is it too big of a breaking change to do that?

damemi: does setting `runAsUser` override what's in the Dockerfile? The problem I can think of is that our public images are built and published with that, so it could break for users who aren't using our helm chart or default deployment manifests.

So if `runAsUser` overrides the Dockerfile, then I don't think we should remove it from the Dockerfile.

a7i: Yes, it overrides it

a7i: Let's leave it as-is for now and unblock this PR

PR #1068, opened 2023-03-02
Adding descheduler policy API Version option in helm templates

gdasson:
This PR adds the option to specify descheduler policy API version. Currently the version is hardcoded in helm template/manifest and points to deprecated v1alpha1.

This PR fixes the issue: https://github.com/kubernetes-sigs/descheduler/issues/1069

### DIFF
--- a/charts/descheduler/values.yaml
+++ b/charts/descheduler/values.yaml
@@ -60,6 +60,9 @@ leaderElection: {}
 cmdOptions:
   v: 3
 
+#It is recommended to use the latest Policy API version
+deschedulerPolicyAPIVersion: "descheduler/v1alpha2"
+
 deschedulerPolicy:
   # nodeSelector: "key1=value1,key2=value2"
   # maxNoOfPodsToEvictPerNode: 10
--- a/charts/descheduler/values.yaml
+++ b/charts/descheduler/values.yaml
@@ -60,7 +60,7 @@ leaderElection: {}
 cmdOptions:
   v: 3
 
-#It is recommended to use the latest Policy API version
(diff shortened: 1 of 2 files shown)


charts/descheduler/values.yaml
```diff
@@ -60,6 +60,9 @@ leaderElection: {}
 cmdOptions:
   v: 3
 
+#It is recommended to use the latest Policy API version
```
a7i, line 63: nit: 
```suggestion
# Recommended to use the latest Policy API version supported by the Descheduler app version
```

a7i (approved):
/lgtm

/assign @ingvagabund @knelasevero 

Any concerns with making v1alpha2 the default?

ingvagabund:
Before making v1alpha2 as default we need to make sure everything is properly tested and the descheduling framework is finished. +1 for making the api version configurable. Though, let's stick with v1alpha1 for now.

charts/descheduler/values.yaml
```diff
@@ -60,6 +60,9 @@ leaderElection: {}
 cmdOptions:
   v: 3
 
+# Recommended to use the latest Policy API version supported by the Descheduler app version
+deschedulerPolicyAPIVersion: "descheduler/v1alpha2"
```
a7i, line 64: based on @ingvagabund suggestion, let's stick to v1alpha1 for now until we have done thorough testing
```suggestion
deschedulerPolicyAPIVersion: "descheduler/v1alpha1"
```

gdasson: updated as suggested

PR #1093, opened 2023-03-16
Descheduling profile with PoC fake plugin

ingvagabund:
Pre requisite for implementing the descheduling cycle.

- introduce fake plugin with action reactors for some extension points
- introduce descheduling profile
- build the default evictor using the builder, not directly


pkg/framework/profile/profile.go
```diff
@@ -0,0 +1,310 @@
...
+func (hi *handleImpl) SharedInformerFactory() informers.SharedInformerFactory {
+	return hi.sharedInformerFactory
+}
+
+// Evictor retrieves evictor so plugins can filter and evict pods
+func (hi *handleImpl) Evictor() framework.Evictor {
+	return hi.evictor
+}
+
+type profileImpl struct {
```
knelasevero, line 95: Cant we use the profile that we already have in our api? I think it is a bit confusing:

- we read a config file
- we populate our Deschedule policy that in fact has a profile (that in the end has plugins inside).

then we

- move things from the profile of the api into an internal profile that has Run[*]Plugins methods to be able to call those for each profile.

Can't we just the profile that we have in the api? And, if it doesn't deliver something that we need, can't we just change it (so we only have one type with this name inside the code)?

ingvagabund: Do you mean https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/api/types.go#L60? If so, `profileImpl` is actual implementation of handling a single profile configuration. We tend to keep the api definition from anything handling the api. Though, maybe I just misunderstand.

knelasevero: Yes, I mean that one. It is not really that separate, if we think that when we change anything in the api's profile, we would probably also change somethings in this profileImpl. 

Maybe we could rename it in this new place? configuredProfile? It is also about the methods and instantiations. 

maybe [here](https://github.com/ingvagabund/descheduler/blob/2ae5ec8ac37770b5310d9ab8f8bd3e1394d07a32/pkg/descheduler/descheduler.go#L305) prfl could be named configuredProfile, or just profileImpl for the instance as well. 

maybe [methods](https://github.com/ingvagabund/descheduler/blob/2ae5ec8ac37770b5310d9ab8f8bd3e1394d07a32/pkg/framework/profile/profile.go#L167) could also be different so it is also clear this is a different profile, maybe NewConfiguredProfile or something like that

I would also change the name of the package and folder I think.

Just so it is clear that this is a different profile then, that this is a implementation of handling a single profile configuration. I am not sure about the names though. I would not block the PR because of this also

ingvagabund: Good points. The upstream kube-scheduler code is using `KubeSchedulerProfile` name in the API part: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/kube-scheduler/config/v1/types.go#L132. We could use the same naming schema and rename our Profile into DeschedulerProfile. While keeping the rest as it is?

knelasevero: Yeah, that also works I think

pkg/descheduler/descheduler.go
```diff
@@ -366,88 +301,30 @@ func RunDeschedulerStrategies(ctx context.Context, rs *options.DeschedulerServer
...
-				continue
-			}
-			evictorFilter, err := defaultevictor.New(
-				pc.Args,
-				&handleImpl{
-					clientSet:                 rs.Client,
-					getPodsAssignedToNodeFunc: getPodsAssignedToNode,
-					sharedInformerFactory:     sharedInformerFactory,
-				},
+			prfl, err := frwkprofile.NewProfile(
```
a7i, line 305: much easier to read this way:
```suggestion
			currProfile, err := frameworkprofile.NewProfile(
```

pkg/descheduler/descheduler.go
```diff
@@ -366,88 +301,30 @@ func RunDeschedulerStrategies(ctx context.Context, rs *options.DeschedulerServer
...
-			// work, we are currently passing this via context. To be removed
-			// (See discussion thread https://github.com/kubernetes-sigs/descheduler/pull/885#discussion_r919962292)
-			strategyStart := time.Now()
-			childCtx := context.WithValue(ctx, "strategyName", pg.Plugin.Name())
-			status := pg.Plugin.Balance(childCtx, nodes)
-			metrics.DeschedulerStrategyDuration.With(map[string]string{"strategy": pg.Plugin.Name(), "profile": pg.Profile}).Observe(time.Since(strategyStart).Seconds())
+			status = prfl.RunDeschedulePlugins(ctx, nodes)
 			if status != nil && status.Err != nil {
-				klog.ErrorS(status.Err, "plugin finished with error", "pluginName", pg.Plugin.Name())
+				klog.ErrorS(status.Err, "running balance extension point failed with error", "profile", profile.Name)
```
a7i, line 326:
```suggestion
				klog.ErrorS(status.Err, "running deschedule extension point failed with error", "profile", profile.Name)
```

a7i (approved):
/lgtm

/hold

PR #1101, opened 2023-03-25
bump golangci for go 1.20

a7i:
golangci causes oom/execution-error when ran with go 1.20 (fixed in 1.52.0)

https://github.com/golangci/golangci-lint/issues/3538


a7i:
Screenshot 2023-03-25 at 10 59 22 AM
that's fun

PR #1104, opened 2023-04-02
fix default value assignment of EvictLocalStoragePods

yanggangtony:
fix if condition for the right value of EvictLocalStoragePods

PR #1106, opened 2023-04-03
fix default value assignment of EvictLocalStoragePods

a7i:
Patching 0.26 (0.26.1)
https://github.com/kubernetes-sigs/descheduler/pull/1104#issuecomment-1493447747

### DIFF
--- a/pkg/framework/plugins/defaultevictor/defaults.go
+++ b/pkg/framework/plugins/defaultevictor/defaults.go
@@ -28,7 +28,7 @@ func SetDefaults_DefaultEvictorArgs(obj *DefaultEvictorArgs) {
 		obj.NodeSelector = ""
 	}
 	if !obj.EvictLocalStoragePods {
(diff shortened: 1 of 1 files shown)


damemi:
@a7i should I push a v0.26.1 tag to the release branch so we can start the patch release process?

a7i:
@damemi yes, please. how do I go about getting permission to create tags?

damemi:
@a7i I think we update https://github.com/kubernetes/org/blob/b7bd5eb8af25bd3c5917bf5ecee720e84ee7095b/config/kubernetes-sigs/sig-scheduling/teams.yaml#L20-L28

We should really work on a way to automate our release flow instead of having one person manually push a tag right to remote (😬). Something like an issue or PR merge that requires at least a second person to lgtm/approve would be cool. For me, this tag push is the most stressful part of each release because I worry about messing it up.  We could probably set up a github action to do that.

For now, v0.26.1 is pushed

PR #1108, opened 2023-04-03
Update helm chart to v0.26.1

a7i:


damemi:
Helm chart issue is the same as in https://github.com/kubernetes-sigs/descheduler/pull/1038 (our new test uses the tag in the PR, but we rely on this PR to get us a commit SHA that we can use to promote the new image tag)

@a7i do you want to merge https://github.com/kubernetes-sigs/descheduler/pull/1109 first? Then we can promote the image based on the SHA from that PR and that should unblock this

damemi:
@a7i we need to do the image promotion upstream to publish the new 0.26.1 image, that's why the helm test is failing

damemi (commented):
/approve
/lgtm

PR #1111, opened 2023-04-06
Delete "io/ioutil" package.

yanggangtony:
// Deprecated: As of Go 1.16, this function simply calls os.ReadFile.

At the same time , We should not use "io/ioutil" , and replaced by "io" and "os" package.


damemi (commented):
/approve

PR #1119, opened 2023-04-15
Binary tree node patch 1

BinaryTreeNode:


a7i:
@BinaryTreeNode would you please provide some background on this change?

For example, why are OWNERS file deleted?
Why are plugin registrations removed?

PR #1120, opened 2023-04-16
WIP: Drop Evict extension point

ingvagabund:
The Evict extension point is not currently in use. All DefaultEvictor plugin functionality is exposed through Filter and PreEvictionFilter extension points instead.
Thus, no need to limit the number of evictors enabled.


damemi (commented):
/approve

knelasevero:
/lgtm
/hold

feel free to unhold :)

PR #1122, opened 2023-04-17
Add json tags to internal api fields

### LINKED ISSUE #1121: failed decoding policy config : strict decoding error: unknow field \" namesapces.include\"

descheduler version: master branch

```
kind: "DeschedulerPolicy"
profiles:
  - name: ProfileName
    pluginConfig:
    - name: "DefaultEvictor"
    - name: "PodLifeTime"
      args:
        maxPodLifeTimeSeconds: 86400
        namespaces:
          include:
          - "namespace1"
          - "namespace2"
    plugins:
      evict:
        enabled:
          - "DefaultEvictor"
      deschedule:
        enabled:
          - "PodLifeTime"
```

damemi:
Fixes #1121 

Plugin arg objects refer to the internal API type for some settings, but those types don't have json tags which causes an error trying to decode the config. This adds json tags to those types and sets up a new github workflow for testing that the descheduler is deployed in a kind cluster and starts running successfully.

As a follow up, we should expand the test suite to cover all possible plugin configurations (ie, a big config file that uses every setting possible) for both v1alpha1 and v1alpha2.

We also might want to just copy these internal types to the v1alpha2 package and use those, if possible.

### DIFF
--- a/kubernetes/base/configmap.yaml
+++ b/kubernetes/base/configmap.yaml
@@ -6,23 +6,26 @@ metadata:
   namespace: kube-system
 data:
   policy.yaml: |
-    apiVersion: "descheduler/v1alpha1"
+    apiVersion: "descheduler/v1alpha2"
     kind: "DeschedulerPolicy"
-    strategies:
-      "RemoveDuplicates":
-         enabled: true
-      "RemovePodsViolatingInterPodAntiAffinity":
-         enabled: true
-      "LowNodeUtilization":
-         enabled: true
-         params:
-           nodeResourceUtilizationThresholds:
-             thresholds:
-               "cpu" : 20
-               "memory": 20
-               "pods": 20
-             targetThresholds:
-               "cpu" : 50
-               "memory": 50
-               "pods": 50
-
+    profiles:
+      - name: ProfileName
+        pluginConfig:
+        - name: "DefaultEvictor"
+        - name: "RemovePodsViolatingInterPodAntiAffinity"
+        - name: "LowNodeUtilization"
+          args:
+            thresholds:
+              "cpu" : 20
+              "memory": 20
+              "pods": 20
+            targetThresholds:
+              "cpu" : 50
+              "memory": 50
+              "pods": 50
+        plugins:
+          balance:
+            enabled:
+              - "LowNodeUtilization"
+              - "RemovePodsViolatingInterPodAntiAffinity"
+              - "RemoveDuplicates
--- a/test/manifests/v1alpha2/configmap.yaml
+++ b/test/manifests/v1alpha2/configmap.yaml
@@ -1,0 +1,29 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: descheduler-policy-configmap
+  namespace: kube-system
+data:
+  policy.yaml: |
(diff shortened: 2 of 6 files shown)


JaneLiuL (commented):
/lgtm
but if  you merge commits into one commits will be better :)

JaneLiuL (commented):
thanks @damemi  for your contribution~~
/lgtm

test/manifests/v1alpha2/configmap.yaml
```diff
@@ -0,0 +1,32 @@
...
+              value: 10000
+        - name: "PodLifeTime"
+          args:
+            maxPodLifeTimeSeconds: 86400
+            namespaces:
+              include:
+              - "namespace1"
+              - "namespace2"
+        plugins:
+          evict:
```
ingvagabund, line 26: Let's wait for https://github.com/kubernetes-sigs/descheduler/pull/1120 to merge and update the extension point accordingaly

damemi: /hold cancel

damemi:
> I think the test above tests all inner args of the strategy plugins, but now thinking about it you mean like e2e tests that actually tests the file, right (instead of direct go types)? That makes sense

yup, that's what I mean. the yaml->go type decoding can bring up tiny bugs like this

> which internal types?

[api.Namespaces](https://github.com/kubernetes-sigs/descheduler/blob/a7113dab5048c556046fb61addd2ce84f9760094/pkg/api/types.go#L43-L48) and [api.PriorityThreshold](https://github.com/kubernetes-sigs/descheduler/blob/a7113dab5048c556046fb61addd2ce84f9760094/pkg/api/types.go#L55-L58)

ingvagabund:
> Plugin arg objects refer to the internal API type for some settings, but those types don't have json tags which causes an error trying to decode the config.

I wonder why these errors have not been caught by the unit tests. There are some unit tests with the policy configuration as a string.

> As a follow up, we should expand the test suite to cover all possible plugin configurations (ie, a big config file that uses every setting possible) for both v1alpha1 and v1alpha2.
>
> We also might want to just copy these internal types to the v1alpha2 package and use those, if possible.

I don't think we have touched the topic of plugin versioning when it comes to using internal types in the plugin arguments. Internal types can change at any point. For that they should not be directly imported by any versioned type. Given the namespace filtering is plugin-scoped, there's no need to have an internal type for it. On the other hand it is a common field shared across plugin arguments. So we need to make sure the same sets of fields will get used in every in-tree plugin so we can still publish a common API for configuring it. The priority thresholds might be easier to remove from the internal types as they should be used only by the DefaultEvictor.

Anyway, something to follow up later on.

kubernetes/base/configmap.yaml
```diff
@@ -6,23 +6,26 @@ metadata:
...
+            targetThresholds:
+              "cpu" : 50
+              "memory": 50
+              "pods": 50
+        plugins:
+          balance:
+            enabled:
+              - "LowNodeUtilization"
+              - "RemovePodsViolatingInterPodAntiAffinity"
+              - "RemoveDuplicates
```
mueller-ma, line 31: Missing `"` at the end of this line?

PR #1127, opened 2023-04-25
Drop remaining mentions of evict extension point

ingvagabund:
Follow up for https://github.com/kubernetes-sigs/descheduler/pull/1120.

### DIFF
--- a/pkg/descheduler/policyconfig_test.go
+++ b/pkg/descheduler/policyconfig_test.go
@@ -875,12 +875,6 @@ profiles:
         podRestartThreshold: 100
         includingInitContainers: true
     plugins:
-      filter:
-        enabled:
-          - "DefaultEvictor"
-      evict:
-        enabled:
-          - "DefaultEvictor"
       deschedule:
         enabled:
           - "RemovePodsHavingTooManyRestarts"
--- a/README.md
+++ b/README.md
@@ -125,7 +125,7 @@ The Default Evictor Plugin is used by default for filtering pods before processi
 
 ### Example policy
 
-As part of the policy, you will start deciding which top level configuration to use, then which Evictor plugin to use (if you have your own, the Default Evictor if not), followed by deciding the configuration passed to the Evictor Plugin. After that you will enable/disable eviction strategies plugins and configure them properly.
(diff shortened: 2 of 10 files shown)


pkg/descheduler/policyconfig_test.go
```diff
@@ -878,9 +878,6 @@ profiles:
       filter:
         enabled:
           - "DefaultEvictor"
-      evict:
-        enabled:
-          - "DefaultEvictor"
```
damemi, line 883: no `preEvictionFilter` here? Or is that already tested?

damemi (commented):
/approve
though if DefaultEvictor is enabled by default when unset in filter/preEvictionFilter, we can probably just exclude it from the readme+examples. 

One section on how to configure it and explaining that it's on by default is better than adding it to every other example imo. what do you think?

ingvagabund:
@damemi I wanted to be more explicit. However, you are right there's no need to mention it everywhere. I have update the generic description of the policy example. We still don't provide a way to disable any extension point. That will be implement in the next release once the framework gets completed.

pkg/descheduler/policyconfig_test.go
ingvagabund: Dropped the `filter` ep as well. `TestDecodeDefaults` provides testing for the missing eps.

knelasevero:
If we show yamls that are not complete, people will apply those and complain. Unless we explain to them that they are not complete

knelasevero:
I still prefer to leave all config examples with a default evictor included. Also for copy paste value.

damemi:
> If we show yamls that are not complete, people will apply those and complain. Unless we explain to them that they are not complete

As I understood it, these yamls will work even without including the DefaultEvictor since it's enabled by default: https://github.com/kubernetes-sigs/descheduler/blob/a7113dab5048c556046fb61addd2ce84f9760094/pkg/descheduler/policyconfig.go#L110-L117

So it should be copy-pastable either way. But I'm fine following up with this and updating the docs later

knelasevero:
that's true

don't know why I forgot about this. I guess because in all my local yamls I add a default evictor with muscle memory :sweat: 

Disregard my previous comment then

PR #1129, opened 2023-04-26
Do some code cleans.

yanggangtony:
/kind cleanup

Its just some tiny code cleans..


pkg/descheduler/policyconfig.go
```diff
@@ -140,33 +140,21 @@ func validateDeschedulerConfiguration(in api.DeschedulerPolicy, registry pluginr
...
 		}
 	}
-	if errorsInProfiles != nil {
-		return errorsInProfiles
-	}
-	return nil
+	return errorsInProfiles
 }
 
 func setErrorsInProfiles(err error, profileName string, errorsInProfiles error) error {
```
ingvagabund, line 153: `setErrorsInProfiles` is used only once. Worth making it inline.

ingvagabund (changes requested):

pkg/descheduler/policyconfig.go
```diff
@@ -120,10 +120,10 @@ func setDefaultEvictor(profile api.DeschedulerProfile, client clientset.Interfac
 	if defaultevictorPluginConfig == nil {
 		profile.PluginConfigs = append([]api.PluginConfig{newPluginConfig}, profile.PluginConfigs...)
 		defaultevictorPluginConfig = &newPluginConfig
-		idx = 0
+		// when defaultevictorPluginConfig is nil , the idx=0
```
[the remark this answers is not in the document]
yanggangtony: thanks for review.
Will rollback it.

pkg/descheduler/policyconfig.go
yanggangtony: sound good.
update it.

pkg/descheduler/policyconfig.go
```diff
@@ -140,33 +140,16 @@ func validateDeschedulerConfiguration(in api.DeschedulerPolicy, registry pluginr
 				pluginUtilities := registry[pluginConfig.Name]
```
ingvagabund, line 140: We need to fail in case `registry[pluginConfig.Name]` does not exists. E.g.:
```go
if _, ok := registry[pluginConfig.Name]; !ok {
	errorsInProfiles = fmt.Errorf("%w: %s", errorsInProfiles, fmt.Sprintf("in profile %s: plugin %s in pluginConfig not registered", profile.Name, pluginConfig.Name))
	continue
}
pluginUtilities := registry[pluginConfig.Name]
if pluginUtilities.PluginArgValidator == nil {
	continue
}
err := pluginUtilities.PluginArgValidator(pluginConfig.Args)
...
```
This will also reduce the level of nesting by 2.

ingvagabund (changes requested):

pkg/descheduler/policyconfig.go
yanggangtony: Good suggestion.
Update later.

PR #1130, opened 2023-04-26
Add knelasevero to approvers

knelasevero:
As per the discussion on slack, adding myself to the approver list :D


damemi (commented):
+1!
/approve

PR #1139, opened 2023-05-05
use v1alpha2 api for helm chart default values

a7i:
I assume this needs a chart bump?


a7i:
@JaneLiuL would you please review?

arthurk:
Same for me. I installed it yesterday and it works fine but there is big warning in the readme about v1 being deprecated and recommend switching to v2, yet the chart only works with v1.

sherifabdlnaby:
Plus one here!

a7i:
this was communicated by @knelasevero yesterday so we should be merging it soon

a7i:
cc: @jklaw90

PR #1141, opened 2023-05-05
bump log level for processing info

a7i:
we run descheduler at log level 1 but still the "Processing" log lines bring too much noise.
if I wanted to debug why a pod on a node/namespace was not behaving correctly, I would temporarily bump the log level and then revert once done.

### DIFF
--- a/pkg/framework/plugins/podlifetime/pod_lifetime.go
+++ b/pkg/framework/plugins/podlifetime/pod_lifetime.go
@@ -106,7 +106,7 @@ func (d *PodLifeTime) Deschedule(ctx context.Context, nodes []*v1.Node) *framewo
 	nodeMap := make(map[string]*v1.Node, len(nodes))
 
 	for _, node := range nodes {
-		klog.V(1).InfoS("Processing node", "node", klog.KObj(node))
+		klog.V(2).InfoS("Processing node", "node", klog.KObj(node))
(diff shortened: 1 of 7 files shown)


damemi (commented):
/approve
is there anything else that's currently at log level 2 that this will be printed with now?

also maybe we would consider bumping the default log level to 2 (even temporarily) so when people upgrade they don't think descheduler is broken

a7i:
> also maybe we would consider bumping the default log level to 2 (even temporarily) so when people upgrade they don't think descheduler is broken

That's a good point. You want me to do this in the same PR?

damemi:
> That's a good point. You want me to do this in the same PR?

sounds good to me, thanks!

a7i:
> > That's a good point. You want me to do this in the same PR?
> 
> sounds good to me, thanks!

@damemi the manifests and helm chart is already at v3
https://github.com/kubernetes-sigs/descheduler/blob/master/kubernetes/deployment/deployment.yaml#L32

damemi:
@a7i I meant like in the code when we init the flags (https://github.com/kubernetes-sigs/descheduler/blob/master/cmd/descheduler/app/server.go#L119 or somewhere around there). But there might not be a good way to do that since it's just an imported library. If the charts and manifests are already at 3, that's probably good enough

damemi:
> @damemi I tried this via the initFlags but it wasn't very trivial. Let me know if you're okay with bump this to v2

@a7i fair enough, I think since we already ship at v3 this will probably be fine though some people may be confused at first with the change. It shouldn't set off any alarms though
/lgtm

PR #1143, opened 2023-05-10
fix plugin arg conversion when using multiple profiles with same plugin

### LINKED ISSUE #1142: Default evictor nodeFit has incorrect default value in v1alpha2

descheduler version: v0.27.0

**Does this issue reproduce with the latest release?** Yes

```
    kind: "DeschedulerPolicy"
    maxNoOfPodsToEvictPerNamespace: 20
    maxNoOfPodsToEvictPerNode: 10
    profiles:
    - name: nofit
      pluginConfig:
      - args:
          # note that nodeFit is not set here!!
          evictLocalStoragePods: true
          evictSystemCriticalPods: true
          nodeSelector: node-type notin (api,control-plane,etcd,etcd-events)
        name: DefaultEvictor
      - args:
          labelSelector:
            matchExpressions:
            - key: old-pod-evictor.compute.zende.sk/enabled
              operator: In
              values:
              - "true"
          maxPodLifeTimeSeconds: 172800
        name: PodLifeTime
      plugins:
        deschedule:
          enabled:
          - PodLifeTime
```

**What k8s version are you using (`kubectl version`)?** 1.24

$ kubectl version

</pre></details>

**What did you do?**

Per the docs, the default value of nodeFit is `false`. Run Descheduler with the policy attached.

nodeFit to be `false`

**What did you see instead?**
nodeFit is set to `true`

a7i:
If you have multiple profiles and reuse the plugin, the values stump over each other because we were reusing the same `PluginArgInstance`.

closes #1142

### DIFF
--- a/pkg/descheduler/policyconfig.go
+++ b/pkg/descheduler/policyconfig.go
@@ -137,11 +137,6 @@ func validateDeschedulerConfiguration(in api.DeschedulerPolicy, registry pluginr
 	for _, profile := range in.Profiles {
 		for _, pluginConfig := range profile.PluginConfigs {
 			if _, ok := registry[pluginConfig.Name]; ok {
-				if _, ok := registry[pluginConfig.Name]; !ok {
-					errorsInProfiles = fmt.Errorf("%w: %s", errorsInProfiles, fmt.Sprintf("in profile %s: plugin %s in pluginConfig not registered", profile.Name, pluginConfig.Name))
-					continue
-				}
-
 				pluginUtilities := registry[pluginConfig.Name]
 				if pluginUtilities.PluginArgValidator == nil {
 					continue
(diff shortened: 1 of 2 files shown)


pkg/descheduler/policyconfig.go
```diff
@@ -137,11 +137,6 @@ func validateDeschedulerConfiguration(in api.DeschedulerPolicy, registry pluginr
 	for _, profile := range in.Profiles {
 		for _, pluginConfig := range profile.PluginConfigs {
 			if _, ok := registry[pluginConfig.Name]; ok {
-				if _, ok := registry[pluginConfig.Name]; !ok {
```
ingvagabund, line 140: Right :)

This could be actually moved one level up:
```go
for _, pluginConfig := range profile.PluginConfigs {
	if _, ok := registry[pluginConfig.Name]; !ok {
		errorsInProfiles = fmt.Errorf("%w: %s", errorsInProfiles, fmt.Sprintf("in profile %s: plugin %s in pluginConfig not registered", profile.Name, pluginConfig.Name))
		continue
	}
	pluginUtilities := registry[pluginConfig.Name]
	...
```
to remove another level of nesting.

a7i: this is actually unrelated to the fix, but just seemed really odd. I can change it but it will show up as a lot more lines in +/-

PR #1144, opened 2023-05-11
feat: support ScaleDownDeploymentHavingTooManyPodRestarts strategy

### LINKED ISSUE #789: Ability to scale down failing pods (e.g. to zero)

***Use case***: development cluster, microservices.

Some microservices include AI models, that upon startup do heavy lifting (Network, IO, CPU, RAM - all resources). Then service fails. The solution for a problem may not be immediate, with a choice to either to spend human's time and remove Deployment temporarily (and loose "previous logs"), OR to allow higher load on hardware while fix is being made (from minutes - to days).

Ability to do scale pod down (to zero, or via configurable logic) action, instead of eviction. 

```yaml
    RemovePodsHavingTooManyRestarts:
      enabled: true
      params:
        podsHavingTooManyRestarts:
          podRestartThreshold: 50
          includingInitContainers: FALSE
```
- Alternative software to remove failed pods https://github.com/ankilosaurus/kube_remediator
- Bumping Feature Request to "Make CrashLoopBackoff timing tuneable" https://github.com/kubernetes/kubernetes/issues/57291

Both descheduler and kube_remediator do not always solve the problem.

1. Rescheduling failing pod to another node will not solve the problem of a pod failing in most cases.
2. Removing pod covered by Deployment does little, except resetting "restarts" count, as Pod is recreated by default.

descheduler version: 0.23.1

P.S. I hope this feature request lines up with the goals of a project.

fgksgf:
Close #789


fgksgf:
@a7i Would you mind conducting a code review?

PR #1146, opened 2023-05-12
update deprecated sets.String to generic sets

a7i:
[sets#String](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets#String) is now deprecated

Screenshot 2023-05-11 at 10 37 57 PM


damemi (commented):
/approve

Issue #1138, opened 2023-05-04
Pods are not getting evicted on "RemovePodsViolatingTopologySpreadConstraint"


JohnMops:
**What version of descheduler are you using?**

descheduler version: 1.26

**Does this issue reproduce with the latest release?**
yes

**Which descheduler CLI options are you using?**
no CLI

**Please provide a copy of your descheduler policy config file**
```
apiVersion: "descheduler/v1alpha1"
kind: "DeschedulerPolicy"
strategies:
  RemovePodsViolatingTopologySpreadConstraint:
    enabled: true
    params:
      includeSoftConstraints: false
      namespaces:
        include:
        - assets
      nodeFit: false
```

**What k8s version are you using (`kubectl version`)?**
v1.24.10

**What did you do?**

1. Deployment has the following: 

```
topologySpreadConstraints:
    - maxSkew: 1
      topologyKey: topology.kubernetes.io/zone
      whenUnsatisfiable: DoNotSchedule
      labelSelector:
        matchLabels:
          app: nginx
```

2. Also has nodeSelector and tolerations
3. GKE node autoprovisioner is used to scale new nodes according to node selector and tolerations
4. When the AZs are not balanced, descheduler throws: 

"ignoring pod for eviction as it does not fit on any other node" pod="assets/nginx-55b564d9d-77k8d"
even though nodeFit is set to "false"

**What did you expect to see?**
I expect the pod viloating the topology spread to be evicted , rescheduled and for node auto provisioner to pick it up and do its job

**What did you see instead?**
Pods remain untouched with the "ignoring pod for eviction as it does not fit on any other node" pod="assets/nginx-55b564d9d-77k8d" message in the descheduler pod logs.

damemi:
Hi @JohnMops, this message comes from the strategy directly when it checks if the Pod would end up scheduled back onto the same node due to a hard node selector or affinity (see this comment in the code for some more details: https://github.com/kubernetes-sigs/descheduler/blob/eeafdb253a340f1d0fc4f272c304924a2b089e5b/pkg/framework/plugins/removepodsviolatingtopologyspreadconstraint/topologyspreadconstraint.go#L331-L335). It's independent of the nodeFit check since it's directly related to the strategy

Could you share the full Pod yaml? You mentioned it has a nodeSelector and tolerations so that might be the issue.

JohnMops:
@damemi 

Thx for the reply. 

If this is the case then descheduler is not really right for us. 
[...]

PR #1148, opened 2023-05-15
removepodsviolatingtopologyspreadconstraint: implement explicit constraints

### LINKED ISSUE #1123: RemovePodsViolatingTopologySpreadConstraints different setting for constraints

We would to define a different behavior for TopologySpreadConstraints when it's `DoNotSchedule` vs `ScheduleAnyway`.

This is because `DoNotSchedule` triggers a scale-up with cluster autoscaler, but `ScheduleAnyway` does not.
As a result, I do not want to do a `nodeFit` for `DoNotSchedule` but want it enabled for `ScheduleAnyway`.

Ideally we change the Descheduler config API to allow for granular constraints filter.
```
profiles:
  - name: Soft
    pluginConfig:
    - name: "DefaultEvictor"
    - name: "RemovePodsViolatingTopologySpreadConstraint"
      args:
        constraints:
          - "ScheduleAnyway"
        # make sure there is capacity before evicting, otherwise scheduler will just put it back on the same node
        nodeFit: true 
  - name: Hard
    pluginConfig:
    - name: "DefaultEvictor"
    - name: "RemovePodsViolatingTopologySpreadConstraint"
      args:
        constraints:
          - "DoNotSchedule"
        # do not perform a nodeFit to ensure that cluster-autoscaler does a scale-up when it sees failedscheduling
        nodeFit: false
```

We can also deprecate `includeSoftConstraints` arg, enable it when `args.constraints` includes `ScheduleAnyway` and eventually remove it in the next API.

I cannot run two separate profiles or two separate descheduler instances, because there is no way to disable the hard constraint.

descheduler version: 0.24.x but plan on upgrading to 0.26.0

**Additional context**

a7i:
closes #1123 and #1138


knelasevero:
Maybe we should call this internal nodeFit something different? Like we did for `EvictableNamespaces` in the nodeutilization stratedies?

a7i:
> Maybe we should call this internal nodeFit something different? Like we did for EvictableNamespaces in the nodeutilization stratedies?

@knelasevero Yes, I like that idea. But this will be a change in behavior for users going forward, given that we always did "nodeFit" for nodesBelowAverage when balancing domains. Are we ok with that?

knelasevero:
> > Maybe we should call this internal nodeFit something different? Like we did for EvictableNamespaces in the nodeutilization stratedies?
> 
> @knelasevero Yes, I like that idea. Would `topologyBalanceNodeFit` be ok?

Sounds good to me. @damemi @ingvagabund , WDYT?

pkg/framework/plugins/removepodsviolatingtopologyspreadconstraint/types.go
```diff
@@ -28,7 +28,11 @@ import (
...
-	Namespaces             *api.Namespaces       `json:"namespaces"`
-	LabelSelector          *metav1.LabelSelector `json:"labelSelector"`
-	IncludeSoftConstraints bool                  `json:"includeSoftConstraints"`
+	Namespaces    *api.Namespaces       `json:"namespaces"`
+	LabelSelector *metav1.LabelSelector `json:"labelSelector"`
+	Constraints   []string              `json:"constraints"`
+	NodeFit       *bool                 `json:"nodeFit,omitempty"`
+
+	// Deprecated: use Constraints list instead
+	IncludeSoftConstraints bool `json:"includeSoftConstraints"`
```
ingvagabund, line 37: v1alpha2 is not out yet officially so it is ok to drop the arg completely an implement the conversion from v1alpha1 -> v1alpha2

ingvagabund:
> Sounds good to me. @damemi @ingvagabund , WDYT?

Making it configurable is reasonable. The semantics of nodeFit before eviction and right before eviction is not clear here. The original intent of the nodeFit here was to evict pods that can be scheduled to a different node to avoid evicting a pod that lands back on the same node. Thus, helping to improve the skew. The case where a new node can be provisioned as a reaction to eviction was not considered during the implementation. For that, we should not always assume a node pool is static.

> Would topologyBalanceNodeFit be ok?

Both `RemovePodsViolatingTopologySpreadConstraint` and `DefaultEvictor` invoke the same `PodFitsAnyOtherNode` function. Nevertheless, that can change in the future as one can enable a different implementation of `PreEvictionFilter` extension point once the descheduling framework is offcially out. The conditions for when a pod is not suitable for evicting wrt. TopologySpreadConstraint can diverge from what it means for the default evictor to fail to fit another node. E.g. a pod can be evicted to re-balance when an autoscaler is enabled (`RemovePodsViolatingTopologySpreadConstraint`). Vs. a pod can not be evicted since a node can not be autoscaled between 10pm and 6am (custom implementation of the `DefaultEvictor` plugin).

We can probably completely omit the nodeFit keyword part. On the other hand when we expose the possibility to ignore nodeFit conditions, do we know which specific condition is actually causing limitation? Based on https://github.com/kubernetes-sigs/descheduler/issues/1138 there's no existing node because:
- pods does not match a node selector
- taints are not tolerated
- insufficient resources
- no schedulable node
Q is whether we wanna expose this level of detail. Or create groups of conditions that can be enabled/disabled as a whole unit. Providing an argument to completely disable NodeFit seems reasonable. On the other hand what if we get a request to disable/enable a specific condition mentioned in the list? One might want to test for `pods does not match a node selector` and `taints are not tolerated` and ignore `insufficient resources` so the autoscaler can actually kick in only when insufficient resources scenario is on instead of evicting a pod and poke autoscaler just because a taint is not tolerated (for a good reason when it's preferable to pick a different pod for eviction which e.g. still can fit into one of the existing nodes).

a7i:
@ingvagabund thanks for your review.

> The semantics of nodeFit before eviction and right before eviction is not clear here. The original intent of the nodeFit here was to evict pods that can be scheduled to a different node to avoid evicting a pod that lands back on the same node.

The intention of nodeFit when balancing domains is to calculate the spread/skew given the "fit" constraints. This is particularly helpful for a best-effort constraint such as "ScheduleAnyway".


> Both RemovePodsViolatingTopologySpreadConstraint and DefaultEvictor invoke the same PodFitsAnyOtherNode function

RemovePodsViolatingTopologySpreadConstraint only considers *nodesBelowAverage* for NodeFit because when balancing domains, it should only consider nodes that can help the skew. Default Evictor considers *all available* nodes.


> Q is whether we wanna expose this level of detail.

I had created an Issue for NodeFit Specs to be able to provide that level of detail. We can move the discussion there: https://github.com/kubernetes-sigs/descheduler/issues/1149

>  also better to split this PR into two.

Generally, I would agree. However, for my particular use-case, it only helps if both args are available. If it's rolled back, then might as well roll it all back because just one wouldn't satisfy my use cases.

As described in the PR, my use-cases are:
- I need a "Just ScheduleAnyway" that takes nodeFit into account. For this, if nodeFit is not taken into account when balancing domains, then Scheduler will just put it back on the same node.

- I need a just "DoNotSchedule". While this is already possible, nodeFit when balancing domains is not configurable. Hence why I need the nodeFit arg for when balancing domains. i.e. do not perform a nodeFit to force Cluster-Autoscaler to do a scale-up.

knelasevero:
I think we can have a single PR, however, with 2 commits. Not saying we would revert a single commit, and leave it at that. We might revert a single one and still pursue this use case in a different approach taking advantage of the first commit. Or revert both. Still makes sense to separate it.

Would #1149 discussion make this PR obsolete? Maybe we should have a proper design proposal for all our nodeFit "problems"? There are a few issues on that topic

a7i:
@ingvagabund @knelasevero Thank you for your feedback ❤️ 

- Split to two commits
- Changed `nodeFit` to `topologyBalanceNodeFit`
- Removed `includeSoftConstraints` from `v1alpha2` api

Look forward to your feedback

pkg/framework/plugins/removepodsviolatingtopologyspreadconstraint/types.go
```diff
@@ -31,4 +31,5 @@ type RemovePodsViolatingTopologySpreadConstraintArgs struct {
 	Namespaces             *api.Namespaces       `json:"namespaces"`
 	LabelSelector          *metav1.LabelSelector `json:"labelSelector"`
 	IncludeSoftConstraints bool                  `json:"includeSoftConstraints"`
+	TopologyBalanceNodeFit *bool                 `json:"TopologyBalanceNodeFit"`
```
ingvagabund: How about?
```json
RemovePodsViolatingTopologySpreadConstraintArgs:
- TopologyBalanceNodeFit: [] // empty field -> fall back to defaults which is a subject to change
- TopologyBalanceNodeFit: [""] // empty string -> disable the node fit checks
- TopologyBalanceNodeFit: ["Cond1", ..., "CondN"] // arbitrary list of enabled conditions (implemented later)
```

```go
type NodeFitConditions string

var (
	NoCondition NodeFitConditions = ""
)
```

Alternatively, `""` can be replaced with `"NoCheck"` or similar string which if specified will be the only valid item in the list.

a7i: I was under the impression that we will handle NodeFit Spec in a subsequent design/PR. But given that, perhaps we should design/implement that first.

ingvagabund: This is more like a middle step so once the NodeFit is refined the API of `TopologyBalanceNodeFit` does not need to change. It will "just" get extended with additional enums.

ingvagabund:
As discussed on the community meeting, let's keep the new `TopologyBalanceNodeFit` argument for v1alpha2 and deprecate it in v1alpha3.
/approve
/lgtm

MarcPow:
@a7i @damemi - Why is this change being held?  My teams also have the auto-scaler scenario in #1138.  I'd dearly love to get a fix in place for this.

a7i:
> @a7i @damemi - Why is this change being held? My teams also have the auto-scaler scenario in #1138. I'd dearly love to get a fix in place for this.

My apologies, let me address the casing issue real quick:
```
	TopologyBalanceNodeFit *bool                 `json:"TopologyBalanceNodeFit"`
```
to:
```
	TopologyBalanceNodeFit *bool                 `json:"topologyBalanceNodeFit"`
```

a7i:
fyi, we just tested this in our cluster, and this is the config that worked for us:

```
  profiles:
    - name: fit # nodefit enabled plugins
      pluginConfig:
        - name: DefaultEvictor
          args:
            nodeFit: true
          args:
            constraints:
            - ScheduleAnyway
      plugins:
        balance:
          enabled:
            - RemovePodsViolatingTopologySpreadConstraint

    - name: nofit # plugins without nodefit (e.g. force scale-up with cluster-autoscaler)
      pluginConfig:
        - name: "DefaultEvictor"
          args:
            nodeFit: false
          args:
            constraints:
            - DoNotSchedule
            topologyBalanceNodeFit: false
      plugins:
        balance:
          enabled:
            - RemovePodsViolatingTopologySpreadConstraint
```

ingvagabund:
@a7i file conflict :(

PR #1151, opened 2023-05-16
helm: ability to override command and args. set args inline

a7i:
1. ability to override comand and args from helm value. for example we pipe certain commands and as a result, we're using kustomize to patch it since helm doesn't provide the cmd/args values
2. set args inline to make booleans work.
  `$key=$value` as opposed to:
  ```
  - --key
  - value
  ```
  
  Currently this is failing for me:
  ```
  - --dry-run
  - false # this is an issue
  ```
  
  but this works:
  ```
  - --dry-run=false
  ```


damemi (commented):
/approve
this should have a helm chart version bump too

JaneLiuL (commented):
/lgtm

PR #1157, opened 2023-05-25
Fix plugin arg copy

a7i:
merge https://github.com/kubernetes-sigs/descheduler/pull/1143 for v0.27.1 patch


a7i:
/cc @damemi 

should I have only cherry-picked the commit to release-1.27?

PR #1158, opened 2023-05-25
fix plugin arg conversion when using multiple profiles with same plugin

a7i:
cherry-pick commit "fix plugin arg conversion when using multiple profiles with same plugin" to release-1.27


damemi (commented):
/approve

JaneLiuL (commented):
/lgtm
thanks @a7i  for your contribution

PR #1163, opened 2023-06-06
use pod informers for listing pods in removepodsviolatingtopologyspreadconstraint and removepodsviolatinginterpodantiaffinity

a7i:
follow up to #1033 and apply the same in removepodsviolatingtopologyspreadconstraint

### DIFF
--- a/pkg/descheduler/pod/pods.go
+++ b/pkg/descheduler/pod/pods.go
@@ -157,6 +157,20 @@ func BuildGetPodsAssignedToNodeFunc(podInformer cache.SharedIndexInformer) (GetP
 	return getPodsAssignedToNode, nil
 }
 
+// ListPodsOnNodes returns all pods on given nodes.
+func ListPodsOnNodes(nodes []*v1.Node, getPodsAssignedToNode GetPodsAssignedToNodeFunc, filter FilterFunc) ([]*v1.Pod, error) {
+	pods := make([]*v1.Pod, 0)
+	for _, node := range nodes {
+		podsOnNode, err := ListPodsOnANode(node.Name, getPodsAssignedToNode, filter)
+		if err != nil {
+			return nil, err
+		}
+
+		pods = append(pods, podsOnNode...)
+	}
+	return pods, nil
+}
+
 // ListPodsOnANode lists all pods on a node.
 // It also accepts a "filter" function which can be used to further limit the pods that are returned.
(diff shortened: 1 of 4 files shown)


a7i:
With informers, the returned pods are not in order so the assertion is failing. Any suggestions @knelasevero @ingvagabund ?

e.g. one of the tests is expecting this order:
```
expectedEvictedPods:  []string{"pod-5", "pod-6", "pod-7", "pod-8"},
```

damemi:
> With informers, the returned pods are not in order so the assertion is failing. Any suggestions @knelasevero @ingvagabund ?
> 
> e.g. one of the tests is expecting this order:
> 
> ```
> expectedEvictedPods:  []string{"pod-5", "pod-6", "pod-7", "pod-8"},
> ```

I haven't looked at the test, but is there anything semantically different about evicting those different pods? If it's just an ordering issue then you can sort or update the output

ingvagabund:
Tests should not assume a list of pods will always be returned in the same order as inserted. It just happened to be the case :) Sorting the list wherever it makes sense is fine.

ingvagabund:
```
topologyspreadconstraint_test.go:1220: Expected pods [pod-5 pod-6 pod-7 pod-8] to be evicted but [pod-8 pod-5 pod-6 pod-7] were not evicted. Actual pods evicted: [pod-1 pod-2 pod-3 pod-10 pod-11]
```

The test is probably broken. `expectedEvictedCount` is 5 but the list claims it needs only 4 items.

a7i:
> ```
> topologyspreadconstraint_test.go:1220: Expected pods [pod-5 pod-6 pod-7 pod-8] to be evicted but [pod-8 pod-5 pod-6 pod-7] were not evicted. Actual pods evicted: [pod-1 pod-2 pod-3 pod-10 pod-11]
> ```
> 
> The test is probably broken. `expectedEvictedCount` is 5 but the list claims it needs only 4 items. We should probably randomize the order in which the pods are appended into `objs` var that is passed into `fake.NewSimpleClientset` in cases where the order does not matter. At the same time I recall we needed some tests to actually evict a specific list of pods so we can validate the code picks the right pods for eviction.

Thanks @ingvagabund . The test only checks that expected is included in the evicted list, so having only 4 was passing before.

Yes, there are ~3 tests that are expecting a particular order. I can perhaps change them to evicted count per node/topology.

a7i:
fyi - it used to be sored, but then it changed: https://github.com/kubernetes/kubernetes/pull/75730
> "noone should assume the fact that keys are sorted."

PR #1166, opened 2023-06-08
update pause image from 'kubernetes/pause' to 'registry.k8s.io/pause'

a7i:
e2e tests are failing


ingvagabund:
Is there any specific error `kubernetes/pause' produces and 'registry.k8s.io/pause` does not?

knelasevero:
That image is not available anymore:

Error: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_descheduler/1165/pull-descheduler-test-e2e-k8s-master-1-25/1666648699360514048

No tags available: https://hub.docker.com/r/kubernetes/pause/tags


Also registry.k8s.io is the new recommended registry to be used

PR #1186, opened 2023-07-07
fix priority threshold by name alone

knelasevero:
The included test fails on master. If we set priority threshold getting by name alone validation would error out. It should error out only if the user tries to set both the value and the name together.


pkg/descheduler/policyconfig_test.go
```diff
@@ -862,6 +862,79 @@ strategies:
...
+			result: &api.DeschedulerPolicy{
+				Profiles: []api.DeschedulerProfile{
+					{
+						Name: fmt.Sprintf("strategy-%s-profile", podlifetime.PluginName),
+						PluginConfigs: []api.PluginConfig{
+							{
+								Name: "DefaultEvictor",
+								Args: &defaultevictor.DefaultEvictorArgs{
+									PriorityThreshold: &api.PriorityThreshold{
+										Value: utilpointer.Int32(0),
```
damemi, line 890: this is just `0` because in the unit tests there is no `prioritym` class to parse right? In a cluster this would be a real number?

damemi (commented):
/approve

PR #1188, opened 2023-07-14
safe evict to communicate with volcano

gj199575:
volcano pr:https://github.com/volcano-sh/volcano/pull/2974

this is a safe evict


JaneLiuL:
sorry i can not get the point of this pr, could you share more information for what it do and the benefit?

a7i:
/close 
- no response
- CLA Is not signed
- changes are not clear what the intent is

PR #1189, opened 2023-07-16
feat: Enable open telemetry feature

harshanarayana:
Reworking the changes from #958 as is got too mess to cleanup and rework the changes easily since a lot of things changed during the framework rebuild. 

Thanks and Credits to @Dentrax  for the help and contribution in the original PR under #958


harshanarayana:
@a7i  @damemi  I have ported the changes from #958 into this and addressed most of the comments that were left on that thread in terms of reusing the context and making it simple in terms of invoking the tracers and getting rid of the un-necessary helper functions.  Please take a look when time permits.

harshanarayana:
I am still having issue with `make gen` command like I used to have before.

harshanarayana:
I managed to get `make gen` working inside a container for now.. I've pushed the changes to this branch already.. Please go ahead with review

pkg/descheduler/descheduler.go
```diff
@@ -242,6 +256,7 @@ func Run(ctx context.Context, rs *options.DeschedulerServer) error {
 	}
 
 	if rs.LeaderElection.LeaderElect && rs.DeschedulingInterval.Seconds() == 0 {
+		span.AddEvent("Validation Failue", trace.WithAttributes(attribute.String("err", "leaderElection must be used with deschedulingInterval")))
```
damemi, line 259:
```suggestion
		span.AddEvent("Validation Failure", trace.WithAttributes(attribute.String("err", "leaderElection must be used with deschedulingInterval")))
```

pkg/descheduler/descheduler.go
```diff
@@ -251,6 +266,7 @@ func Run(ctx context.Context, rs *options.DeschedulerServer) error {
 
 	if rs.LeaderElection.LeaderElect && !rs.DryRun {
 		if err := NewLeaderElection(runFn, rsclient, &rs.LeaderElection, ctx); err != nil {
+			span.AddEvent("Leader Election Failue", trace.WithAttributes(attribute.String("err", err.Error())))
```
damemi, line 269:
```suggestion
			span.AddEvent("Leader Election Failure", trace.WithAttributes(attribute.String("err", err.Error())))
```

harshanarayana:
@damemi I am working on getting the tests fixed as we speak.

pkg/descheduler/evictions/evictions.go
```diff
@@ -113,8 +116,13 @@ type EvictOptions struct {
 // EvictPod evicts a pod while exercising eviction limits.
 // Returns true when the pod is evicted on the server side.
 func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod, opts EvictOptions) bool {
+	attributes := trace.WithAttributes(attribute.String("podName", pod.Name), attribute.String("podNamespace", pod.Namespace), attribute.String("reason", opts.Reason))
```
damemi, line 119: this doesn't need to be a variable, you can just pass this `trace.WithAttributes` in the `Tracer().Start(...trace.withAttributes)` below

pkg/descheduler/evictions/evictions.go
```diff
@@ -124,7 +132,9 @@ func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod, opts EvictOptio
 			if pe.metricsEnabled {
 				metrics.PodsEvicted.With(map[string]string{"result": "maximum number of pods per node reached", "strategy": strategy, "namespace": pod.Namespace, "node": pod.Spec.NodeName}).Inc()
 			}
-			klog.ErrorS(fmt.Errorf("Maximum number of evicted pods per node reached"), "limit", *pe.maxPodsToEvictPerNode, "node", pod.Spec.NodeName)
+			err = fmt.Errorf("Maximum number of evicted pods per node reached")
```
damemi, line 135: looks like you're declaring this `err` variable just to call `err.Error()` below. you can just pass the string value directly (and then you can drop the `var err error` declartion on L125)

pkg/descheduler/evictions/evictions.go
```diff
@@ -133,13 +143,16 @@ func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod, opts EvictOptio
 		if pe.metricsEnabled {
 			metrics.PodsEvicted.With(map[string]string{"result": "maximum number of pods per namespace reached", "strategy": strategy, "namespace": pod.Namespace, "node": pod.Spec.NodeName}).Inc()
 		}
-		klog.ErrorS(fmt.Errorf("Maximum number of evicted pods per namespace reached"), "limit", *pe.maxPodsToEvictPerNamespace, "namespace", pod.Namespace)
+		err = fmt.Errorf("Maximum number of evicted pods per namespace reached")
```
damemi, line 146: same `err` note

pkg/tracing/tracing.go
```diff
@@ -0,0 +1,152 @@
...
+	DefaultServiceName = "descheduler"
+
+	// DescheduleOperation is the operation name used for Deschedule() functions.
+	DescheduleOperation = "deschedule"
+
+	// BalanceOperation is the operation name used for Balance() functions.
+	BalanceOperation = "balance"
+
+	// EvictionOption is the operation name used for Evict() functions.
+	EvictionOption = "evict"
```
damemi, line 47: should this be `EvictOperation`?

pkg/tracing/tracing.go
```diff
@@ -0,0 +1,152 @@
...
+	provider trace.TracerProvider
+)
+
+func Tracer() trace.Tracer {
+	return tracer
+}
+
+func Provider() trace.TracerProvider {
+	return provider
+}
```
damemi, line 61: this doesn't seem to be used anywhere (and it's essentially just `otel.GetTracerProvider()`)

pkg/tracing/tracing.go
```diff
@@ -0,0 +1,152 @@
...
+		spanProcessor := sdktrace.NewBatchSpanProcessor(exporter)
+		provider = sdktrace.NewTracerProvider(
+			sdktrace.WithSampler(sdktrace.ParentBased(sdktrace.TraceIDRatioBased(sampleRate))),
+			sdktrace.WithSpanProcessor(spanProcessor),
+			sdktrace.WithResource(resource),
+		)
+	} else {
+		klog.Info("Did not find a trace collector endpoint defined. Switching to NoopTraceProvider")
+		provider = trace.NewNoopTracerProvider()
+	}
```
damemi: you could flip this to the top of the function to fail earlier, like:

```
if endpoint == "" {
  klog.Info("Did not find a trace collector endpoint defined. Switching to NoopTraceProvider")
  provider = trace.NewNoopTracerProvider()
  return
}

// do everything from the if block above
klog.Info("Successfully setup trace provider")
return
```

pkg/framework/plugins/removepodsviolatingtopologyspreadconstraint/main_test.go
```diff
@@ -0,0 +1,12 @@
+package removepodsviolatingtopologyspreadconstraint
+
+import (
+	"context"
+	"testing"
+
+	"sigs.k8s.io/descheduler/pkg/tracing"
+)
+
+func TestMain(m *testing.M) {
```
harshanarayana, line 10: @damemi  @a7i  Should I take this route or add a simple `init` handler on `pkg/tracing` and set the provider to No Op and create a tracer out of it by default that can later be overwritten ?

pkg/tracing/tracing.go
```diff
@@ -0,0 +1,152 @@
...
+		spanProcessor := sdktrace.NewBatchSpanProcessor(exporter)
+		provider = sdktrace.NewTracerProvider(
+			sdktrace.WithSampler(sdktrace.ParentBased(sdktrace.TraceIDRatioBased(sampleRate))),
+			sdktrace.WithSpanProcessor(spanProcessor),
+			sdktrace.WithResource(resource),
+		)
+	} else {
+		klog.V(2).Info("Did not find a trace collector endpoint defined. Switching to NoopTraceProvider")
+		provider = trace.NewNoopTracerProvider()
+	}
```
damemi, line 135: you could flip this function to fail earlier by moving this above (L82) like:

```
if endpoint == "" {
  klog.V(2).Info("Did not find a trace collector endpoint defined. Switching to NoopTraceProvider")
  provider = trace.NewNoopTracerProvider()
  return
}

// do everything from the if block (don't need if/else)
klog.V(2).Info("Successfully setup trace provider")
return
```

damemi (changes requested):
just minor nits at this point but looking good

pkg/descheduler/evictions/evictions.go
harshanarayana: Turned this into a variable to avoid having to keep the lengthy lines having multiple `WithAttributes`, Will move that inline

pkg/framework/plugins/removepodsviolatingtopologyspreadconstraint/main_test.go
harshanarayana: @damemi  That might get a bit tricky. 

```golang
func SetTracerProvider(tp trace.TracerProvider) {
	current := TracerProvider()

	if _, cOk := current.(*tracerProvider); cOk {
		if _, tpOk := tp.(*tracerProvider); tpOk && current == tp {
			// Do not assign the default delegating TracerProvider to delegate
			// to itself.
			Error(
				errors.New("no delegate configured in tracer provider"),
				"Setting tracer provider to it's current value. No delegate will be configured",
			)
			return
		}
	}

	delegateTraceOnce.Do(func() {
		if def, ok := current.(*tracerProvider); ok {
			def.setDelegate(tp)
		}
	})
	globalTracer.Store(tracerProviderHolder{tp: tp})
}
```

So, on the actual runtime the delegate settings might not be configured for the actual trace. Is there a better way to handle this ?

harshanarayana: Only way I can think of is using an env lookup in the `init` and do this conditionally and set that ENV during the tests

```golang
func init() {
   if v, ok := os.LookupEnv("INIT_NOOP_TRACER"); ok && v == "true" {
      // init provider and tracer
   }
}
```
But I am not sure if this is the right thing to do or not.

damemi: something like:

```
var tracer

init() {
  tracer = otel.NoOpTracer()
}

SetTracerProvider() {
  opts := parseOptions()
  tracer = otel.SetTracerProvider()...
}
```

harshanarayana: @damemi  changed the `init` handler to avoid calling `otel.SetTraceProvider` entirely. 

```golang
func init() {
	provider = trace.NewNoopTracerProvider()
	tracer = provider.Tracer(TracerName)
}
```
When imported, it will have the tracer and provider set to NoOp version of it. But it will get overwritten by the right provider and tracer at the runtime and then we call the `otel.SetTraceprovider` so it will take care of the required runtime config at the OTEL level.

pkg/tracing/tracing.go
```diff
@@ -0,0 +1,156 @@
...
+	if err := tp.Shutdown(ctx); err != nil {
+		otel.Handle(err)
+		klog.ErrorS(err, "failed to shutdown the trace exporter")
+		return err
+	}
+	return nil
+}
+
+func init() {
```
damemi, line 153: please move above the other functions

damemi (commented):
one nit, otherwise lgtm. Thanks @harshanarayana, this is really cool!
/approve

harshanarayana:
@damemi  @a7i  PTAL. I have pushed the updated changes based on Mike's suggestion

damemi (approved):

pkg/tracing/tracing.go
```diff
@@ -0,0 +1,156 @@
...
+	DefaultServiceName = "descheduler"
+
+	// DescheduleOperation is the operation name used for Deschedule() functions.
+	DescheduleOperation = "deschedule"
+
+	// BalanceOperation is the operation name used for Balance() functions.
+	BalanceOperation = "balance"
+
+	// EvictionOption is the operation name used for Evict() functions.
+	EvictOption = "evict"
```
a7i, line 47: doesn't seem that these are being referenced anywhere, correct?
```suggestion
```

a7i (commented):
one last nit, and ready to sign LGTM

pkg/tracing/tracing.go
harshanarayana: My bad. I should have used this instead of hardcoding that in a few places. Fixed.

harshanarayana:
@a7i  Commit pushed with requested changes. PTAL when you can

Issue #1204, opened 2023-07-24
TopologySpreadConstraint calculations do not fully respect node selectors, node affinity policy, and/or node taints policy


MarcPow:
descheduler version: 0.27.1
k8s version: v1.26.3
policy: RemovePodsViolatingTopologySpreadConstraint

Simple example:  When running in a cluster with both Windows and Linux nodes, if a pod has a single-os node selector, but topologyBalanceNodeFit is false, the introduction of nodes of the other OS will throw off balance computations and introduce additional churn.

a7i:
Addressed in #1218

Issue #1149, opened 2023-05-16
NodeFit Specification


a7i:
**Is your feature request related to a problem? Please describe.**

Currently nodeFit is a boolean flag and the user has no control over which nodeFit checks should run or not.

**Describe the solution you'd like**

Full control over what nodeFit check is enabled or not

```
nodeFit:
  requests: false
  tolerateTaints: true
  affinity: true
  nodeSelector: true
  schedulable: true
```
  

**Describe alternatives you've considered**

Overwriting the Default Evictor

**What version of descheduler are you using?**

descheduler version: v0.27.0x

**Additional context**

damemi:
Originally nodefit was planned to just eventually merge into the default state, we just made it an option because it was a change in how descheduler currently worked.

But seeing the options enumerated like this is really interesting. What if we just merged nodeFit into the default evictor or PreEvictionFilter? I think it has grown beyond what we originally envisioned and honestly I think it makes sense to expose each of these toggles

a7i:
Adding @ingvagabund's proposal in this Issue:

At the same time someone might ask for disabling only specific checks. So instead of defining a bool argument I would like to suggest an enum.

```
topologyBalanceNodeFit:
- MatchSelector
- TaintsAndTolerations
- InsufficientResources
- SchedulableNodes
```

```
type NodeFitConditions string

var (
	MatchSelectorNodeFitConditions NodeFitConditions = "MatchSelector"
	TaintsAndTolerationsNodeFitConditions NodeFitConditions = "TaintsAndTolerations"
	InsufficientResourcesNodeFitConditions NodeFitConditions = "InsufficientResources"
	SchedulableNodesNodeFitConditions NodeFitConditions = "SchedulableNodes"
)

type RemovePodsViolatingTopologySpreadConstraintArgs struct {
	...
	TopologyBalanceNodeFit *[]NodeFitConditions `json:"topologyBalanceNodeFit,omitempty"`
}
```

Also on DefaultEvictor

ingvagabund:
If there's a way I would like to avoid using pointers in the API definitions. Atm, `TopologyBalanceNodeFit == nil` means to use any reasonable defaults (that can change over time) vs. `*TopologyBalanceNodeFit == []NodeFitConditions` means do not use any node fit conditions, i.e. nodeFit feature disabled.

a7i:
Is there a way to make this "future proof" in case we decide to pull in Scheduler's  Filter extension point?

[...]

PR #1205, opened 2023-07-24
Add handling for node eligibility

MarcPow:
Proposing a potential fix for #1204.

Looking for feedback before adding unit tests, etc.

The current code assumes that a couple of beta feature flags around are currently enabled by default.

### DIFF
--- a/pkg/framework/plugins/removepodsviolatingtopologyspreadconstraint/topologyspreadconstraint.go
+++ b/pkg/framework/plugins/removepodsviolatingtopologyspreadconstraint/topologyspreadconstraint.go
@@ -28,6 +28,8 @@ import (
 	"k8s.io/klog/v2"
 	utilpointer "k8s.io/utils/pointer"
 
+	v1helper "k8s.io/component-helpers/scheduling/corev1"
+	nodeaffinity "k8s.io/component-helpers/scheduling/corev1/nodeaffinity"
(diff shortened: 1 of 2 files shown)


damemi:
/ok-to-test
thanks @MarcPow, I understand wanting to hold off on adding tests but it would be helpful to understand the change if you could add a couple test cases to demonstrate

a7i:
Thank you @MarcPow for this change, looks very thorough.

However, we are planning on introducing a `NodeFitSpec` that will give more flexibility as to what thing to check in order to identify eligible nodes.

Please see [this comment](https://github.com/kubernetes-sigs/descheduler/issues/1149#issuecomment-1576478347) on the Issue. What do you think?

MarcPow:
> Thank you @MarcPow for this change, looks very thorough.
> 
> However, we are planning on introducing a `NodeFitSpec` that will give more flexibility as to what thing to check in order to identify eligible nodes.
> 
> Please see [this comment](https://github.com/kubernetes-sigs/descheduler/issues/1149#issuecomment-1576478347) on the Issue. What do you think?

I think we're all on the same page with respect to the need.   The original proposal was a good one, but I think it was made independent of / prior to the changes that happened in the podTopologySpreadConstraints spec that added more controls here.  At the point that those behaviors were added, I think the call on the descheduler is simply to behave just as the scheduler itself would when placing the pods in the first place.  An independent, separate policy has thus become unnecessary.

Issue #901, opened 2022-08-05
allow namespace filter to use a namespace selector


raffaelespazzoli:
**Problem description**

in a cluster where namespaces are frequently created and destroyed (for example because tenants are onboarded and removed) listing the included/excluded namespaces as an array makes it necessary to frequently patch the Policy object

**Describe the solution you'd like**
a namespace selector (based on label) can be specified in alternative to the include/exclude list. This is the standard that is followed by many other operators.

**Describe alternatives you've considered**
none

**What version of descheduler are you using?**
latest

**Additional context**

JaneLiuL:
yet, good idea~~ 
would you like to have a try on this feature ? @raffaelespazzoli

a7i:
I also like the flexibility it brings but should we do this work as part of `descheduler/v1alpha2` initiative? Given the breaking changes.

cc: @ingvagabund  @damemi

damemi:
Would it be a breaking change to make the namespace include/exclude fields a regex? Current settings with alphanumeric namespace strings would still only match the exact name, so the functionality is the same. As long as no regex special characters are allowed in k8s namespace names.

I am not sure about taking a labelling approach as it decouples descheduler config from the descheduler itself, which I could see causing headaches

a7i:
In my organization, we're discussing use-cases around skipping a namespace via a label. Currently, we have to re-deploy Descheduler to exclude a namespace. [Alternatively, we can label each Deployment as well]

It would be ideal for us to label a namespace and have it be ignored by Descheduler.
example:
```yaml
kind: Namespace
apiVersion: v1
metadata:
  name: example
  labels:
    descheduler.k8s.io/opt-out: "true"
```

Where my Descheduler configuration would define a `namespaceSelector` at a plugin definition:
```yaml
 namespaceSelector:
   matchExpressions:
   - key: descheduler.k8s.io/opt-out
     operator: In
     values: ["true"]
```

cc: @jklaw90

PR #1209, opened 2023-07-30
Allow namespace filter to use regular expressions

superbrothers:
This PR allow namespace filter to use regular expressions.

The current namespace filter does not support cases where namespaces are dynamically created. In #901, namespace selector was proposed, but in my use case, regular expressions are sufficient.

### DIFF
--- a/README.md
+++ b/README.md
@@ -703,6 +703,8 @@ The following strategies accept a `namespaces` parameter which allows to specify
 * `RemovePodsViolatingTopologySpreadConstraint`
 * `RemoveFailedPods`
 
+The list of including, excluding namespaces are treated as string patterns and compiled into regular expressions.
+It adds `^` and `$` at the beginning and end of each pattern to ensure matching the entire string.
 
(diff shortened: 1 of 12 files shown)


a7i:
/ok-to-test

I feel that `namespaceSelector` better aligns with existing Kubernetes resources. For example [Dynamic Admission Controller](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-namespaceselector).

/cc @knelasevero and @ingvagabund thoughts?

superbrothers:
Any updates? Regex namespace filter is good enough for my use case and still maintains backward compatibility, but namespace selector is fine by me.

knelasevero:
> /ok-to-test
> 
> I feel that `namespaceSelector` better aligns with existing Kubernetes resources. For example [Dynamic Admission Controller](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-namespaceselector).
> 
> /cc @knelasevero and @ingvagabund thoughts?

agreed, we shouldn't stray too far in that regard

PR #1210, opened 2023-07-30
feat: Implement preferredDuringSchedulingIgnoredDuringExecution for RemovePodsViolatingNodeAffinity

jordipiqueselles:
Now, the descheduler can detect and evict pods that are not optimally allocated according to the "preferred..." node affinity. It only evicts a pod if it can be scheduled on a node that scores higher in terms of preferred node affinity than the current one.

This can be activated by enabling the RemovePodsViolatingNodeAffinity plugin and passing "preferredDuringSchedulingIgnoredDuringExecution" in the args.

For example, imagine we have a pod that prefers nodes with label "key1: value1" with a weight of 10. If this pod is scheduled on a node that doesn't have "key1: value1" as label but there's another node that has this label and where this pod can potentially run, then the descheduler will evict the pod.

Another effect of this commit is that the
RemovePodsViolatingNodeAffinity plugin will not remove pods that don't fit in the current node but for other reasons than violating the node affinity. Before that, enabling this plugin could cause evictions on pods that were running on tainted nodes without the necessary tolerations.

This commit also fixes the wording of some tests from node_affinity_test.go and some parameters and expectations of these tests, which were wrong.

### DIFF
--- a/pkg/framework/plugins/removepodsviolatingnodeaffinity/node_affinity.go
+++ b/pkg/framework/plugins/removepodsviolatingnodeaffinity/node_affinity.go
@@ -26,6 +26,7 @@ import (
 	nodeutil "sigs.k8s.io/descheduler/pkg/descheduler/node"
 	podutil "sigs.k8s.io/descheduler/pkg/descheduler/pod"
 	frameworktypes "sigs.k8s.io/descheduler/pkg/framework/types"
+	"sigs.k8s.io/descheduler/pkg/utils"
 )
 
 const PluginName = "RemovePodsViolatingNodeAffinity"
@@ -88,7 +89,8 @@ func (d *RemovePodsViolatingNodeAffinity) Deschedule(ctx context.Context, nodes 
 		case "requiredDuringSchedulingIgnoredDuringExecution":
 			// In this specific case, the pod must also violate the nodeSelector to be evicted
 			filterFunc := func(pod *v1.Pod, node *v1.Node, nodes []*v1.Node) bool {
-				return d.handle.Evictor().Filter(pod) &&
+				return utils.PodHasNodeAffinity(pod, utils.RequiredDuringSchedulingIgnoredDuringExecution) &&
+					d.handle.Evictor().Filter(pod) &&
 					nodeutil.PodFitsAnyNode(d.handle.GetPodsAssignedToNodeFunc(), pod, nodes) &&
 					!nodeutil.PodMatchNodeSelector(pod, node)
 			}
@@ -97,9 +99,10 @@ func (d *RemovePodsViolatingNodeAffinity) Deschedule(ctx context.Context, nodes 
 			// In this specific case, the pod must have a better fit on another node than
 			// in the current one based on the preferred node affinity
 			filterFunc := func(pod *v1.Pod, node *v1.Node, nodes []*v1.Node) bool {
-				return d.handle.Evictor().Filter(pod) &&
+				return utils.PodHasNodeAffinity(pod, utils.PreferredDuringSchedulingIgnoredDuringExecution) &&
+					d.handle.Evictor().Filter(pod) &&
 					nodeutil.PodFitsAnyNode(d.handle.GetPodsAssignedToNodeFunc(), pod, nodes) &&
-					(nodeutil.BestPodNodeAffinityWeight(pod, nodes) > nodeutil.PodNodeAffinityWeight(pod, node))
+					(nodeutil.GetBestNodeWeightGivenPodPreferredAffinity(pod, nodes) > nodeutil.GetNodeWeightGivenPodPreferredAffinity(pod, node))
 			}
 			err = d.processNodes(ctx, nodes, filterFunc)
 		default:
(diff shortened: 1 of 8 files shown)


jordipiqueselles:
@a7i @ingvagabund Could you review this PR? Thanks!

pkg/framework/plugins/removepodsviolatingnodeaffinity/node_affinity.go
```diff
@@ -78,40 +78,66 @@ func (d *RemovePodsViolatingNodeAffinity) Name() string {
...
-						klog.V(1).InfoS("Evicting pod", "pod", klog.KObj(pod))
-						d.handle.Evictor().Evict(ctx, pod, evictions.EvictOptions{})
-						if d.handle.Evictor().NodeLimitExceeded(node) {
-							break
-						}
-					}
-				}
+			// In this specific case, the pod must also violate the nodeSelector to be evicted
+			filterFunc := func(pod *v1.Pod, node *v1.Node, nodes []*v1.Node) bool {
+				return d.handle.Evictor().Filter(pod) &&
```
a7i, line 91: Should this be optimized to check that it always has a `RequiredDuringSchedulingIgnoredDuringExecution`?
```suggestion
				return pod.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution != nil && 
					d.handle.Evictor().Filter(pod) &&
```

pkg/framework/plugins/removepodsviolatingnodeaffinity/node_affinity.go
```diff
@@ -78,40 +78,66 @@ func (d *RemovePodsViolatingNodeAffinity) Name() string {
...
+func (d *RemovePodsViolatingNodeAffinity) processNodes(ctx context.Context, nodes []*v1.Node, filterFunc func(*v1.Pod, *v1.Node, []*v1.Node) bool) *frameworktypes.Status {
+	for _, node := range nodes {
+		klog.V(2).InfoS("Processing node", "node", klog.KObj(node))
+
+		// Potentially evictable pods
+		pods, err := podutil.ListPodsOnANode(
+			node.Name,
+			d.handle.GetPodsAssignedToNodeFunc(),
+			podutil.WrapFilterFuncs(d.podFilter, func(pod *v1.Pod) bool {
+				return filterFunc(pod, node, nodes)
```
a7i, line 125:
```suggestion
				return pod.Spec.Affinity != nil && pod.Spec.Affinity.NodeAffinity != nil && filterFunc(pod, node, nodes)
```

pkg/framework/plugins/removepodsviolatingnodeaffinity/node_affinity.go
```diff
@@ -78,40 +78,66 @@ func (d *RemovePodsViolatingNodeAffinity) Name() string {
...
+				return d.handle.Evictor().Filter(pod) &&
+					nodeutil.PodFitsAnyNode(d.handle.GetPodsAssignedToNodeFunc(), pod, nodes) &&
+					!nodeutil.PodMatchNodeSelector(pod, node)
 			}
+			err = d.processNodes(ctx, nodes, filterFunc)
+		case "preferredDuringSchedulingIgnoredDuringExecution":
+			// In this specific case, the pod must have a better fit on another node than
+			// in the current one based on the preferred node affinity
+			filterFunc := func(pod *v1.Pod, node *v1.Node, nodes []*v1.Node) bool {
+				return d.handle.Evictor().Filter(pod) &&
```
a7i, line 100:
```suggestion
				return pod.Spec.Affinity.NodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution != nil &&
				d.handle.Evictor().Filter(pod) &&

File: pkg/utils/predicates.go [hunk 4]
```diff
@@ -275,3 +275,29 @@ func TolerationsEqual(t1, t2 []v1.Toleration) bool {
 	}
 	return true
 }
+
+// Returns the weight that the pod gives to a node by analyzing the
+// soft node affinity of that pod
+// (nodeAffinity.preferredDuringSchedulingIgnoredDuringExecution)
+func PodNodeAffinityWeight(pod *v1.Pod, node *v1.Node) (int32, error) {
```
a7i, line 282: Can we be more explicit in the function name that this only checks preferred?

pkg/framework/plugins/removepodsviolatingnodeaffinity/node_affinity.go
jordipiqueselles: True, that's an easy optimization

pkg/utils/predicates.go
jordipiqueselles: The function has been renamed to `GetNodeWeightGivenPodPreferredAffinity`

pkg/framework/plugins/removepodsviolatingnodeaffinity/node_affinity.go
jordipiqueselles: To apply this optimization, I've created a function called [PodHasNodeAffinity](https://github.com/jordipiqueselles/descheduler/blob/jordi/preferredNodeAffinity-2/pkg/utils/pod.go#L220) that checks if a pod has the node affinity defined (it can be `preferred...` or `required...`). 

This also has helped me to simplify https://github.com/jordipiqueselles/descheduler/blob/jordi/preferredNodeAffinity-2/pkg/utils/predicates.go#L283

a7i:
Thank you @jordipiqueselles  this is a great addition!

/lgtm

knelasevero:
Thanks for this!

/approve

Issue #1213, opened 2023-08-08
Suggestion: Add a Visual Studio Code Dev Container to the repo to allow for a streamlined development experience


MarcPow:
[Dev Containers](https://code.visualstudio.com/docs/devcontainers/create-dev-container) often make it easier for developers to deliver changes quickly, without worrying about how to best set up a development environment on every machine they're using.

Note: I understand if the project would rather not include features that are targeted for particular IDEs.  But I thought I'd propose it - as I find them really useful.

PR #1214, opened 2023-08-08
Add dev container

MarcPow:
Addresses #1213


knelasevero:
> Note: I understand if the project would rather not include features that are targeted for particular IDEs. But I thought I'd propose it - as I find them really useful.

Yeah, this already summarizes my concern. This is very user specific. Maybe we could add to the docs in the getting started guide just as a suggestion if the contributor wants to create this directory and files. But I am generally against adding this directly to this repo. @ingvagabund @a7i WDYT?

MarcPow:
Perhaps your perspective may have changed on these, now that they're moving (slowly) beyond visual studio code - 

https://containers.dev/supporting

PR #1215, opened 2023-08-09
fix: descheduler_loop_duration_seconds has wrong value

Abirdcfly:
see golang/go#60048

I learned this change from https://github.com/golang/tools/blob/9abb02c1b5524bf655a0985bc101496a6b53ff71/internal/gocommand/invoke.go#L265, thanks to author https://github.com/heschi 🙏

### DIFF
--- a/pkg/descheduler/descheduler.go
+++ b/pkg/descheduler/descheduler.go
@@ -112,8 +112,9 @@ func (d *descheduler) runDeschedulerLoop(ctx context.Context, nodes []*v1.Node) 
 	var span trace.Span
 	ctx, span = tracing.Tracer().Start(ctx, "runDeschedulerLoop")
 	defer span.End()
-	loopStartDuration := time.Now()
-	defer metrics.DeschedulerLoopDuration.With(map[string]string{}).Observe(time.Since(loopStartDuration).Seconds())
+	defer func(loopStartDuration time.Time) {
+		metrics.DeschedulerLoopDuration.With(map[string]string{}).Observe(time.Since(loopStartDuration).Seconds())
+	}(time.Now())
 
 	// if len is still <= 1 error out
 	if len(nodes) <= 1 {


a7i:
Nice catch and thanks for the references/links! 🏆 

/ok-to-test
/lgtm

Abirdcfly:
> please assign [ingvagabund](https://github.com/ingvagabund) for approval

/assign @ingvagabund

thanks.

pkg/descheduler/descheduler.go
```diff
@@ -112,8 +112,9 @@ func (d *descheduler) runDeschedulerLoop(ctx context.Context, nodes []*v1.Node)
 	var span trace.Span
 	ctx, span = tracing.Tracer().Start(ctx, "runDeschedulerLoop")
 	defer span.End()
-	loopStartDuration := time.Now()
-	defer metrics.DeschedulerLoopDuration.With(map[string]string{}).Observe(time.Since(loopStartDuration).Seconds())
+	defer func(loopStartDuration time.Time) {
+		metrics.DeschedulerLoopDuration.With(map[string]string{}).Observe(time.Since(loopStartDuration).Seconds())
+	}(time.Now())
```
ingvagabund, line 117: Should not this be more like:
```go
loopStartDuration := time.Now()
defer func(loopStartDuration time.Time) {
		metrics.DeschedulerLoopDuration.With(map[string]string{}).Observe(time.Since(loopStartDuration).Seconds())
	}(loopStartDuration)
```
?

Abirdcfly: I think both of these changes work. 

It's straightforward to use `defer func(...) {...} (time.Now())` because that's the only place the `loopStartDuration` variable is used, and it doesn't seem like it needs an extra variable defined.

ingvagabund: `time.Now()` will have different value for `loopStartDuration := time.Now()` and when `defer` is invoked.  Unless I incorrectly understand how the defer statement invokes the function/code. My assumption is anything after `defer` gets invoke right before a function/method exits.

ingvagabund: https://go.dev/tour/flowcontrol/12 says:
```
The deferred call's arguments are evaluated immediately, but the function call is not executed until the surrounding function returns.
```
Interesting.

Issue #1226, opened 2023-08-24
helmchart install: CronJob in version "v1" cannot be handled as a CronJob


cayla:
**What version of descheduler are you using?**

descheduler version: 0.28

**Does this issue reproduce with the latest release?**

Yes

**Which descheduler CLI options are you using?**

None

**Please provide a copy of your descheduler policy config file**

Stock helm chart.  No options or args

**What k8s version are you using (`kubectl version`)?**

<details><summary><code>kubectl version</code> Output</summary><br><pre>
$ kubectl version
Client Version: v1.28.0
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.27.2
</pre></details>

**What did you do?**

helm install my-release --namespace kube-system descheduler/deschedule

**What did you expect to see?**

```
NAME: test
LAST DEPLOYED: Thu Aug 24 10:00:55 2023
NAMESPACE: kube-system
STATUS: deployed
REVISION: 1
TEST SUITE: None
NOTES:
Descheduler installed as a CronJob.
```

**What did you see instead?**

```
Error: INSTALLATION FAILED: 1 error occurred:
        * CronJob in version "v1" cannot be handled as a CronJob: json: cannot unmarshal object into Go struct field Container.spec.jobTemplate.spec.template.spec.containers.args of type string
```

I believe the issue is the unquoted (or unneeded) `:` in 0.28's `charts/descheduler/templates/cronjob.yaml` 

`- --policy-config-file: "/policy-dir/policy.yaml"`

https://github.com/kubernetes-sigs/descheduler/compare/v0.27.1...descheduler-helm-chart-0.28.0#diff-644d004bc3a84952a3c2a97f5f57a85282a0bade5e1c5f124c7407c7788c217c

Without looking into the proper syntax of the command, if I remove the `:` from `templates/cronjob.yaml` and install from that, it succeeds.

alextricity25:
+1 I'm seeing the same thing. I also created an issue at the same time you did 😆 
https://github.com/kubernetes-sigs/descheduler/issues/1227

cayla:
@Omar007 thanks for pointing that out.  I also verified it against the documentation and it was the same there so I created a PR to resolve this.

PR #1229, opened 2023-08-25
helm: update cronjob args argument to avoid unmarshal error

cayla:
Fixes https://github.com/kubernetes-sigs/descheduler/issues/1226

Make it match both the deployment parallel

https://github.com/kubernetes-sigs/descheduler/blob/33e9a52385394a3f4b2cb55bdf717fe52f1a852f/charts/descheduler/templates/cronjob.yaml#L71

As well as the documented arg

https://github.com/kubernetes-sigs/descheduler/blob/master/docs/user-guide.md#balance-cluster-by-pod-age

### DIFF
--- a/charts/descheduler/templates/cronjob.yaml
+++ b/charts/descheduler/templates/cronjob.yaml
@@ -68,7 +68,7 @@ spec:
               command:
                 {{- toYaml .Values.command | nindent 16 }}
               args:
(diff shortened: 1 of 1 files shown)


a7i:
Thank you @cayla  🏆 

/lgtm
/approve

PR #1233, opened 2023-08-28
topologyspreadconstraint: support `matchLabelKeys`

### LINKED ISSUE #1171: Support MatchLabelKeys in TopologySpreadConstraints Strategy
The topolgoyspreadconstraint strategy doesn't handle [matchLabelKeys](https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/#spread-constraint-definition). We use this for pod-template-hash and it should be considered in the descheduling logic.

The logic should AND the keys with our matchlabels. 

None.

descheduler version: v0.27.1

a7i:
Closes #1171 

In Kubernetes 1.27, this feature flag was graduated to beta and it respects topology spread constraints during rolling upgrades. Descheduler currently does not honor this and apps could get unnecessary evictions during rolling upgrade

Ref: https://github.com/kubernetes/enhancements/issues/3243

### DIFF
--- a/pkg/framework/plugins/removepodsviolatingtopologyspreadconstraint/topologyspreadconstraint.go
+++ b/pkg/framework/plugins/removepodsviolatingtopologyspreadconstraint/topologyspreadconstraint.go
@@ -29,7 +29,7 @@ import (
 	utilpointer "k8s.io/utils/pointer"
 
 	v1helper "k8s.io/component-helpers/scheduling/corev1"
-	nodeaffinity "k8s.io/component-helpers/scheduling/corev1/nodeaffinity"
+	"k8s.io/component-helpers/scheduling/corev1/nodeaffinity"
 	"sigs.k8s.io/descheduler/pkg/descheduler/evictions"
 	"sigs.k8s.io/descheduler/pkg/descheduler/node"
 	podutil "sigs.k8s.io/descheduler/pkg/descheduler/pod"
@@ -48,6 +48,19 @@ type topologyPair struct {
 type topology struct {
 	pair topologyPair
 	pods []*v1.Pod
+}
+
+// topologySpreadConstraint is an internal version for v1.TopologySpreadConstraint
+// and where the selector is parsed.
+// This mirrors scheduler: https://github.com/kubernetes/kubernetes/blob/release-1.28/pkg/scheduler/framework/plugins/podtopologyspread/common.go#L37
+type topologySpreadConstraint struct {
+	maxSkew            int32
+	topologyKey        string
+	selector           labels.Selector
+	nodeAffinityPolicy v1.NodeInclusionPolicy
+	nodeTaintsPolicy   v1.NodeInclusionPolicy
+	podNodeAffinity    nodeaffinity.RequiredNodeAffinity
+	podTolerations     []v1.Toleration
 }
 
 // RemovePodsViolatingTopologySpreadConstraint evicts pods which violate their topology spread constraints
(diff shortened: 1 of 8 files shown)


pkg/framework/plugins/removepodsviolatingtopologyspreadconstraint/topologyspreadconstraint.go
```diff
@@ -496,3 +493,22 @@ func matchNodeInclusionPolicies(tsc *v1.TopologySpreadConstraint, tolerations []
 	}
 	return true
 }
+
+func newTopologyConstraintSet(constraint v1.TopologySpreadConstraint, pod *v1.Pod) topologyConstraintSet {
+	if pod.Labels != nil && len(constraint.MatchLabelKeys) > 0 {
+		if constraint.LabelSelector == nil {
+			constraint.LabelSelector = &metav1.LabelSelector{}
```
ingvagabund, line 500: Safer to make a deep copy of `constraint` before modifying it.

a7i: That's a good idea 👍

a7i: I'll mirror what scheduler is doing https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/framework/plugins/podtopologyspread/common.go#L35

ingvagabund: That sounds good. We should probably keep it in sync in case new features get introduced.

a7i: Added as a separate commit but feel free to squash as you see fit. Look forward to your feedback!

test/e2e/e2e_topologyspreadconstraint_test.go
```diff
@@ -91,29 +98,45 @@ func TestTopologySpreadConstraint(t *testing.T) {
...
 	}
 	for name, tc := range testCases {
 		t.Run(name, func(t *testing.T) {
-			t.Logf("Creating RC %s with %d replicas", name, tc.replicaCount)
-			rc := RcByNameContainer(name, testNamespace.Name, int32(tc.replicaCount), tc.topologySpreadConstraint.LabelSelector.DeepCopy().MatchLabels, nil, "")
-			rc.Spec.Template.Spec.TopologySpreadConstraints = []v1.TopologySpreadConstraint{tc.topologySpreadConstraint}
-			if _, err := clientSet.CoreV1().ReplicationControllers(rc.Namespace).Create(ctx, rc, metav1.CreateOptions{}); err != nil {
-				t.Fatalf("Error creating RC %s %v", name, err)
+			t.Logf("Creating Deployment %s with %d replicas", name, tc.replicaCount)
+			deployment := test.BuildTestDeployment(name, testNamespace.Name, int32(tc.replicaCount), tc.topologySpreadConstraint.LabelSelector.DeepCopy().MatchLabels, func(d *appsv1.Deployment) {
```
a7i, line 120: the e2e tests have to be updated to use Deployment instead of RC. That's because for matchLabelKeys, we want to rely on `pod-template-hash` which is generated by the replicaset

knelasevero:
This 
/lgtm
would it make sense to add a bit about this in our docs?

a7i:
@knelasevero updated the docs.

Would you please take a look? also @ingvagabund need help moving this forward since this was officially supported in 1.27 but not in Descheduler 0.27 and 0.28

knelasevero:
/lgtm

/hold 

Hold for a bit to see if Jan has any comments, we can merge by eow if no comments or when another person has eyes on it

ingvagabund:
@a7i this looks great. Thank you for all the refactoring.

PR #1239, opened 2023-09-07
Add an option to allow opting out of Descheduling

Starblade42:
# DoNotEvict annotation (TL;DR) - (DRAFT) Still need to add tests, and the setting to turn it off. Is this directionally okay?

Up to now, there hasn't been a way to opt out of descheduling unless you use a priority class. There were ways to target or exclude particular namespaces too, which is also useful.

This PR is to add the ability for individual pods to opt out of descheduling via a kubernetes annotation. This gives us flexibility in our large fleet of clusters that are running with many tenants that only have Namespace Admin permissions.

In order to not cause unintended behavior by evaluating the value of the `descheduler.alpha.kubernetes.io/evict` annotation, which only needs to exist to *opt in* to descheduling, I chose to add a new annotation that will be evaluated for existence to opt out.

`descheduler.alpha.kubernetes.io/donotevict: "true"`

If this is present, the pod is filtered out of the evictable pods.

## Be able to turn off DoNotEvict

Being able to say DoNotEvictThis capability is useful, but it should also be possible to turn off the opt-out, because it could be misused by tenants. Opting out too many or the wrong pods can have cost implications.

# More Info, just because

## The deployment 

The application of this DoNotEvict annotation is for a very well behaved and optimized deployment that is punished for being so fault agnostic.

- A deployment with a single pod that has no PDB. 
- The service is tolerant of being rescheduled, and best effort is just fine
- If the service is down for 15 or more minutes we start to be worried about violating our SLAs
- A PDB would make it impossible to deschedule, but also impossible for the Cluster Autoscaler to Evict for node consolidation and scale in. We run hundreds of these deployments in a cluster, so running two is cost prohibitive, and running one with a PDB is also cost prohibitive
 
It has happened many times where the pod is descheduled due to node taints... never quite becomes ready during its startup probe grace period, and then gets descheduled again... and then again, being de/scheduled several times in a 30 minute window. Okay, the service has been not running for 20 minutes or so, and we start violating our SLA.

## Why not use PriorityClass?

In our use case at Adobe, we run relatively large clusters with many namespaces for a variety of internal development teams, so a few things are true:

- Most clusters have many namespaces with very different applications running in them
- Workloads are run by Namespace admins with no Cluster admin access
- Our mandate as the kubernetes devops team is to help development teams of all sizes improve efficiency and decrease spend with as little effort as possible

Having a tenant use a new priority class creates issues for us, because we have admin components that we want to deschedule for cost and availability reasons, but we'd need these tenant pod excluded from descheduling to have a lower priority than our admin components, which are in turn lower priority than `system-node-critical` or `system-cluster-critical`. This would mean we'd need to run multiple deschedulers, one for tenants below a certain priority value, and another for our admin components below the `system-cluster-critical` class and only in our Admin namespaces.

In our use case for example, there are deployments designed to be run as a single pod. We could run more, but that costs more and the service they provide is fault agnostic. Their service is best effort, so a disruption in service is okay if it lasts for a few minutes, even every day. But if the service is down for 30 minutes or more, it could cause issues.

However, there have been many instances where these pods have been descheduled repeatedly and instantly. Adding a PDB would prevent descheduling, but would massively increase cost.

Because they are fairly fault agnostic, adding a PDB would technically solve the problem, but create a huge increase in costs as they could not be evicted.

Using the annotation `descheduler.alpha.kubernetes.io/donotevict: "true"` the descheduler only looks to see if it's set, not the value. This was chosen instead of using the value of descheduler.alpha.kubernetes.io/evict: "true" because that also only checked to see if it was set. Checking the value would be a breaking change.

### DIFF
--- a/pkg/framework/plugins/defaultevictor/defaultevictor.go
+++ b/pkg/framework/plugins/defaultevictor/defaultevictor.go
@@ -33,6 +33,7 @@ import (
 const (
 	PluginName            = "DefaultEvictor"
 	evictPodAnnotationKey = "descheduler.alpha.kubernetes.io/evict"
+	doNotEvictPodAnnotationKey = "descheduler.alpha.kubernetes.io/donotevict"
 )
 
 var _ frameworktypes.EvictorPlugin = &DefaultEvictor{}
@@ -58,6 +59,11 @@ func IsPodEvictableBasedOnPriority(pod *v1.Pod, priority int32) bool {
 // HaveEvictAnnotation checks if the pod have evict annotation
 func HaveEvictAnnotation(pod *v1.Pod) bool {
 	_, found := pod.ObjectMeta.Annotations[evictPodAnnotationKey]
+	return found
+}
+// HaveDoNotEvictAnnotation checks if the pod have donotevict annotation
+func HaveDoNotEvictAnnotation(pod *v1.Pod) bool {
+	_, found := pod.ObjectMeta.Annotations[doNotEvictPodAnnotationKey]
 	return found
 }
 
@@ -170,6 +176,11 @@ func (d *DefaultEvictor) PreEvictionFilter(pod *v1.Pod) bool {
 
 func (d *DefaultEvictor) Filter(pod *v1.Pod) bool {
 	checkErrs := []error{}
+
+	if HaveDoNotEvictAnnotation(pod) {
+        klog.V(4).InfoS("Pod has the donotevict annotation and should be skipped", "pod", klog.KObj(pod))
+		return false
+	}
 
 	if HaveEvictAnnotation(pod) {
 		return true


a7i:
@Starblade42 Thanks for the details in the PR. My organization uses labels to opt-out of Descheduler. All the strategies should support a `labelSelector` so you could do:

```
          args:
            labelSelector:
              matchExpressions:
              - {key: descheduler.example.com/enabled, operator: NotIn, values: ["false"]}
```

Starblade42:
@a7i That is a great tip! I think I should change this PR to adding some documentation about using pod label selectors this way. It ought to be self evident, but clearly I missed it. At least a couple other people are probably as dumb as me!

PR #1242, opened 2023-09-12
fix: golangci config

jklaw90:
timeout is the current variable name and goimports indentation was off.
[timeout ref](https://github.com/golangci/golangci-lint/blob/master/.golangci.reference.yml#L14)


a7i (approved):
/approve

✨

a7i (approved):
/approve

✨

PR #1246, opened 2023-09-20
update image references to point to v0.28.0

a7i:
this was missed as part of v0.28 release

cc: @knelasevero


knelasevero:
Thanks Amir, sorry for the delay last week :(
/approve
/lgtm

knelasevero:
@a7i I think I still don't have write on the repo: https://github.com/kubernetes/org/pull/4162/files

So to force push the tag it needs to be you or Mike or Jan, I think

PR #1058, opened 2023-02-15
avoid descheduling when ReplicaSet=1

jiangxiaobin96:
fix [#786](https://github.com/kubernetes-sigs/descheduler/issues/786)
Compared with setting `PodDisruptionBudget` for every RC/RS, easy to avoid evicting pods with `replicasSet=1`.

PR #1152, opened 2023-05-17
skip eviction when replica count is below evictor threshold setting

jiangxiaobin96:
reconstruct issue [#1058](https://github.com/kubernetes-sigs/descheduler/pull/1058)
```
I0515 16:40:08.689485       1 named_certificates.go:53] "Loaded SNI cert" index=0 certName="self-signed loopback" certDetail="\"apiserver-loopback-client@1684168808\" [serving] validServingFor=[apiserver-loopback-client] issuer=\"apiserver-loopback-client-ca@1684168808\" (2023-05-15 15:40:08 +0000 UTC to 2024-05-14 15:40:08 +0000 UTC (now=2023-05-15 16:40:08.689443092 +0000 UTC))"
I0515 16:40:08.689563       1 secure_serving.go:210] Serving securely on [::]:10258
I0515 16:40:08.689739       1 tlsconfig.go:240] "Starting DynamicServingCertificateController"
I0515 16:40:08.699708       1 reflector.go:287] Starting reflector *v1.Namespace (0s) from k8s.io/client-go/informers/factory.go:150
I0515 16:40:08.699726       1 reflector.go:323] Listing and watching *v1.Namespace from k8s.io/client-go/informers/factory.go:150
I0515 16:40:08.699908       1 reflector.go:287] Starting reflector *v1.Pod (0s) from k8s.io/client-go/informers/factory.go:150
I0515 16:40:08.700081       1 reflector.go:323] Listing and watching *v1.Pod from k8s.io/client-go/informers/factory.go:150
I0515 16:40:08.700517       1 reflector.go:287] Starting reflector *v1.PriorityClass (0s) from k8s.io/client-go/informers/factory.go:150
I0515 16:40:08.700684       1 reflector.go:323] Listing and watching *v1.PriorityClass from k8s.io/client-go/informers/factory.go:150
I0515 16:40:08.699711       1 reflector.go:287] Starting reflector *v1.Node (0s) from k8s.io/client-go/informers/factory.go:150
I0515 16:40:08.700893       1 reflector.go:323] Listing and watching *v1.Node from k8s.io/client-go/informers/factory.go:150
I0515 16:40:08.801649       1 descheduler.go:303] Building a pod evictor
I0515 16:40:08.801674       1 pods.go:263] registry.k8s.io/coredns/coredns:v1.9.3
I0515 16:40:08.801681       1 pods.go:263] docker.io/kindest/kindnetd:v20221004-44d545d1
[...]

PR #1257, opened 2023-10-07
skip eviction when replica count is below evictor `minReplicas` threshold setting

a7i:
alternate implementation to #1152

Closes #1058 

/cc @jiangxiaobin96

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,6 +142,7 @@ The Default Evictor Plugin is used by default for filtering pods before processi
 |`labelSelector`|`metav1.LabelSelector`||(see [label filtering](#label-filtering))|
 |`priorityThreshold`|`priorityThreshold`||(see [priority filtering](#priority-filtering))|
 |`nodeFit`|`bool`|`false`|(see [node fit filtering](#node-fit-filtering))|
+|`minReplicas`|`uint`|`0`| ignore eviction of pods where owner (e.g. `ReplicaSet`) replicas is below this threshold |
 
 ### Example policy
 
@@ -166,6 +167,7 @@ profiles:
         evictFailedBarePods: true
         evictLocalStoragePods: true
         nodeFit: true
+        minReplicas: 2
     plugins:
       # DefaultEvictor is enabled for both `filter` and `preEvictionFilter`
       # filter:
--- a/pkg/descheduler/pod/pods.go
+++ b/pkg/descheduler/pod/pods.go
@@ -142,19 +142,23 @@ func BuildGetPodsAssignedToNodeFunc(podInformer cache.SharedIndexInformer) (GetP
 		if err != nil {
 			return nil, err
 		}
-		pods := make([]*v1.Pod, 0, len(objs))
-		for _, obj := range objs {
-			pod, ok := obj.(*v1.Pod)
-			if !ok {
-				continue
-			}
-			if filter(pod) {
-				pods = append(pods, pod)
-			}
-		}
-		return pods, nil
+		return ConvertToPods(objs, filter), nil
 	}
 	return getPodsAssignedToNode, nil
+}
+
+func ConvertToPods(objs []interface{}, filter FilterFunc) []*v1.Pod {
+	pods := make([]*v1.Pod, 0, len(objs))
+	for _, obj := range objs {
+		pod, ok := obj.(*v1.Pod)
+		if !ok {
+			continue
+		}
+		if filter == nil || filter(pod) {
+			pods = append(pods, pod)
+		}
+	}
+	return pods
 }
 
 // ListPodsOnNodes returns all pods on given nodes.
@@ -213,6 +217,14 @@ func GroupByNamespace(pods []*v1.Pod) map[string][]*v1.Pod {
 // OwnerRef returns the ownerRefList for the pod.
 func OwnerRef(pod *v1.Pod) []metav1.OwnerReference {
 	return pod.ObjectMeta.GetOwnerReferences()
+}
+
+func OwnerRefUIDs(pod *v1.Pod) []string {
+	var ownerRefUIDs []string
+	for _, ownerRef := range OwnerRef(pod) {
+		ownerRefUIDs = append(ownerRefUIDs, string(ownerRef.UID))
+	}
+	return ownerRefUIDs
 }
 
 func IsBestEffortPod(pod *v1.Pod) bool {
--- a/pkg/framework/plugins/defaultevictor/defaultevictor.go
+++ b/pkg/framework/plugins/defaultevictor/defaultevictor.go
@@ -16,13 +16,15 @@ package defaultevictor
 import (
 	// "context"
 	"context"
+	"errors"
 	"fmt"
 
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/util/errors"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog/v2"
 	nodeutil "sigs.k8s.io/descheduler/pkg/descheduler/node"
 	podutil "sigs.k8s.io/descheduler/pkg/descheduler/pod"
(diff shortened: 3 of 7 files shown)


a7i:
@ingvagabund and @knelasevero what are your thoughts on this PR?

knelasevero:
I think this makes sense. However I want to test somethings I just thought about.

knelasevero:
Do we want to backport this and also make this available in v1alpha1 while we still maintain it? Since this is additive it shouldn't be too hard

knelasevero:
> @knelasevero I would prefer to encourage users to migrate to v1alpha2. Adding new functionality is one way to do it. But if you feel that we should support both, I'm happy to add it.

I actually agree. Just asking so the decision is explicit in this PR

pkg/framework/plugins/defaultevictor/defaultevictor.go
```diff
@@ -143,6 +143,21 @@ func New(args runtime.Object, handle frameworktypes.Handle) (frameworktypes.Plug
 		})
 	}
 
+	if defaultEvictorArgs.MinReplicas > 1 {
+		podLister := handle.SharedInformerFactory().Core().V1().Pods().Lister()
```
ingvagabund, line 147: Any chance to list pods by owner UID? E.g. creating a new cache that will get updated based on pod CRUD ops and allow to quickly get a list of pods owned by an owner UID? In huge clusters this might be a bottle neck.

a7i: that's a really good suggestion, let me take a stab at it and see how that works

a7i: I tried this and the indexer only supports a single key to index by, while ownerReferences is an array.

aslafy-z:
@a7i can you please rebase the PR? 🙏

a7i:
> @a7i can you please rebase the PR? 🙏

there hasn't been much traction from other maintainers, but happy to rebase just in case

aslafy-z:
@knelasevero can you please give it a new look? Thank you

pkg/descheduler/pod/pods.go
```diff
@@ -257,3 +258,20 @@ func SortPodsBasedOnAge(pods []*v1.Pod) {
...
 }
+
+// GetOwnerReplicasCount returns the count of pods that match the owner ref of the given pod
+func GetOwnerReplicasCount(pods []*v1.Pod, pod *v1.Pod) uint {
+	var count uint
+	podOwnerRefs := OwnerRef(pod)
+	for _, otherPod := range pods {
+		otherOwnerRefs := OwnerRef(otherPod)
+
+		if slices.EqualFunc(podOwnerRefs, otherOwnerRefs, func(l, r metav1.OwnerReference) bool {
```
ingvagabund, line 269: How often it is to have a pod with two or more owners? Also, if a pod A is owned by owner X and owner Y where X owns 5 replicas and Y owns 6 replicas. Will it mean that when a set of pods is owned by both X and Y the right number of replicas is 5? Or, is it max(5,6)? I'd vote for ignoring any pod that has two or more owners as the first iteration of the implementation. If there's a use case where it make sense to have 2 or more owners we can analyse it per case and alter the implementation. WDYT?

README.md
```diff
@@ -142,6 +142,7 @@ The Default Evictor Plugin is used by default for filtering pods before processi
 |`labelSelector`|`metav1.LabelSelector`||(see [label filtering](#label-filtering))|
 |`priorityThreshold`|`priorityThreshold`||(see [priority filtering](#priority-filtering))|
 |`nodeFit`|`bool`|`false`|(see [node fit filtering](#node-fit-filtering))|
+|`minReplicas`|`uint`|`0`| ignore eviction of pods where owner (e.g. Deployment) replicas is below this threshold |
```
ingvagabund, line 145: Deployment owns ReplicaSet(s) which then own pods. Worth s/Deployment/ReplicaSet to avoid the confusion about what owns what.

pkg/framework/plugins/defaultevictor/defaultevictor.go
ingvagabund: https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/descheduler/pod/pods.go#L131 can return a list of keys. In case the implementation ends up comparing slice with a slice you can always get a list of all pods owned by first owner, second, third and make an intersection of all the lists.

a7i: What do you think is the performance gain here? Loading too much cached objects in memory?

Both approach are already using the informer cache so it's no extra lookup to the apiserver.

ingvagabund: Iterating through too many pods in huge clusters. Though, we can merge the code as it is and improve it once there's a request for it I guess.

pkg/descheduler/pod/pods.go
a7i: `slices.EqualFunc` will return false if the len is not the same. So in the example you gave, it will be 5.

pkg/framework/plugins/defaultevictor/defaultevictor.go
a7i: It's already namespaced. I could limit it further by adding the following labels:
- `pod-template-hash`, if it exists
- `controller-revision-hash`, if it exists

Thoughts? I'll take another look at indexer again

ingvagabund: Right. I missed that namespaced part :) Fine then :)

ingvagabund:
/approve
/hold
feel free to unhold, leaving here in case someone else wants to review.

pkg/descheduler/pod/pods.go
a7i: I just pushed up a commit, let me know if that's what you had in mind

ingvagabund: Yes, that's what I meant :) Looks great. Thank you.

a7i: awesome! thanks for the hint. just squashed commits

pkg/framework/plugins/defaultevictor/types.go
```diff
@@ -33,4 +33,5 @@ type DefaultEvictorArgs struct {
 	LabelSelector           *metav1.LabelSelector  `json:"labelSelector"`
 	PriorityThreshold       *api.PriorityThreshold `json:"priorityThreshold"`
 	NodeFit                 bool                   `json:"nodeFit"`
+	MinReplicas             uint                   `json:"minReplicas"`
```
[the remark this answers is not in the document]
a7i: Please feel free to take over this PR.

At the moment, I have other higher priority items and have low confidence in any movement on this PR (been around since October of last year).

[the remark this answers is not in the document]
ingvagabund: I will put it on my todo list where it will probably stay for some time before I find some free cycle for some more comprehensive though. Thank you for pushing this forward.

pkg/framework/plugins/defaultevictor/defaultevictor.go
```diff
@@ -143,6 +145,38 @@ func New(args runtime.Object, handle frameworktypes.Handle) (frameworktypes.Plug
...
+			}
+
+			ownerRef := pod.OwnerReferences[0]
+			objs, err := indexer.ByIndex(indexName, string(ownerRef.UID))
+			if err != nil {
+				return fmt.Errorf("unable to list pods for minReplicas filter in the policy parameter")
+			}
+
+			pods := podutil.ConvertToPods(objs, nil)
+			ownerReplicas := podutil.GetOwnerReplicasCount(pods, pod)
```
ingvagabund, line 171: `len(objs)` should be already sufficient for the check (assuming the informer already knows about `pod`).

pkg/descheduler/pod/pods.go
```diff
@@ -266,3 +279,20 @@ func GroupByNodeName(pods []*v1.Pod) map[string][]*v1.Pod {
 	}
 	return m
 }
+
+// GetOwnerReplicasCount returns the count of pods that match the owner ref of the given pod
+func GetOwnerReplicasCount(pods []*v1.Pod, pod *v1.Pod) uint {
```
ingvagabund, line 284: `GetOwnerReplicasCount` is no longer needed

pkg/framework/plugins/defaultevictor/validation.go
```diff
@@ -26,5 +28,9 @@ func ValidateDefaultEvictorArgs(obj runtime.Object) error {
 		return fmt.Errorf("priority threshold misconfigured, only one of priorityThreshold fields can be set, got %v", args)
 	}
 
+	if args.MinReplicas == 1 {
```
ingvagabund, line 31: `args.MinReplicas < 2`

pkg/framework/plugins/defaultevictor/types.go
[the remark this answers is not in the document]
ingvagabund: 0 replicas mean a pod does not exist. Which is never a case. Fine to keep it as default = disabled.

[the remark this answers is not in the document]
a7i: I added a warning message in validator to print that it's a no-op when set to 1

Issue #927, opened 2022-08-25
Introduce sort and preEvictionSort extension points


ingvagabund:
Some strategies might required different sorting algorithm to change order of evaluating pods. E.g. `PodLifetime` plugin sorts pods based on their creation time. On the other hand plugins such as `RemovePodsViolatingNodeAffinity` do not have any corresponding sorting mechanism so the natural choice is to sort pods based on their creation time. However, in some cases it might be more preferable to sort pods based on their corresponding priority class. Or sort pods based on a different ordering which is application specific. To provide users ability to change the sorting, a new `sort` extension point is introduced.

The same analogy applies for `preEvictionSort` extension point. Stategies like LowNodeUtilization sort the pods based on their priority class before they are evicted. However, in some cases different ordering might be preferable.

a7i:
@spike-liu  any updates? I can take this if you're not actively working on it.

PR #1259, opened 2023-10-09
RemovePodsViolatingInterPodAntiAffinity: sort to evict newer pods

10hin:
Background: To maintain system stability, when deleting or evicting multiple pods with the same conditions (such as priority), newer pods are generally preferred. For instance, when scaling in a Deployment resource, newer pods are deleted.
I'm not sure the above condition, but this pull request based on it. If this is wrong, please point it.

The "RemovePodsViolatingInterPodAntiAffinity" plugin presently considers only pod priority and QoS tier. This pull request enables the plugin to select the newest pod from pods violating inter pod anti-affinity.

Implementation detail that I want to be reviewed:
- Create a new function "**Stable**SortPodsBasedOnPriorityLowToHigh", that is stable version of "SortPodsBasedOnPriorityLowToHigh", to avoid changing current behavior of other plugins.
- Remove "each node processing" behavior to accurately prioritize pods based on their age. I believe that this modification will have a minimal impact on the plugin.

### DIFF
--- a/pkg/descheduler/pod/pods.go
+++ b/pkg/descheduler/pod/pods.go
@@ -231,7 +231,18 @@ func IsGuaranteedPod(pod *v1.Pod) bool {
 // If pods have same priorities, they will be sorted by QoS in the following order:
 // BestEffort, Burstable, Guaranteed
 func SortPodsBasedOnPriorityLowToHigh(pods []*v1.Pod) {
-	sort.Slice(pods, func(i, j int) bool {
+	sort.Slice(pods, lessFuncByPriority(pods))
+}
+
+// StableSortPodsBasedOnPriorityLowToHigh sorts pods stably based on their priorities from low to high.
+// If pods have same priorities, they will be sorted by QoS in the following order:
+// BestEffort, Burstable, Guaranteed
+func StableSortPodsBasedOnPriorityLowToHigh(pods []*v1.Pod) {
+	sort.SliceStable(pods, lessFuncByPriority(pods))
+}
+
+func lessFuncByPriority(pods []*v1.Pod) func(i, j int) bool {
+	return func(i, j int) bool {
 		if pods[i].Spec.Priority == nil && pods[j].Spec.Priority != nil {
 			return true
 		}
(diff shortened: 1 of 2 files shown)


a7i:
/ok-to-test

/retitle RemovePodsViolatingInterPodAntiAffinity: sort to evict newer pods

a7i:
There have been similar attempts in the past, and the consensus has been to implement Sort extension point. See https://github.com/kubernetes-sigs/descheduler/issues/927

10hin:
@a7i 

> There have been similar attempts in the past, and the consensus has been to implement Sort extension point. See #927

Thank you for your comment. To keep this PR simple, I think current implementation looks not so bad.

10hin:
@a7i @JaneLiuL @knelasevero 
I fixed the formatting issue. Could someone give me a review?

knelasevero:
> @a7i
> 
> > There have been similar attempts in the past, and the consensus has been to implement Sort extension point. See #927
> 
> Thank you for your comment. To keep this PR simple, I think current implementation looks not so bad.

Sorry, I personally don't think we could get this in like this. Now that we have the descheduler framework in place it won't make sense to go back to including custom code in each plugin like that. We need to implement the Sort extension point.

@a7i you mentioned in the past you wanted to take a swing at that?

10hin:
@knelasevero 

> Sorry, I personally don't think we could get this in like this. Now that we have the descheduler framework in place it won't make sense to go back to including custom code in each plugin like that. We need to implement the Sort extension point.

Yes. I checked issue #927 for implementing `the Sort extension point`. Unfortunately, there has been no progress since the comment by @a7i at 2023-08-29, and looks no implementation.

This pull request is focusing on implementing actual priority for deschedule, not changing the descheduler framework. If the descheduler framework already has actual code or rule for `the Sort extension point`, I would be happy to utilize it. However I couldn't find it. Please share any additional information about actual design of `the Sort extension point` design.

a7i:
@10hin thanks for being patient here. This was brought up in the Sig-Scheduling meeting this past week and is prioritized for implementation in the next couple of weeks.

As mentioned in multiple places, this does not align with the future of Descheduler and I ask that we wait until we implement it using the framework.

PR #1281, opened 2023-11-02
fix: display ownerKey in the logs correctly

dimunech:
Logs from `RemoveDuplicates` plugin do not display `ownerKey` correctly. Here's an example:
```
"Average occurrence per node" node="<redacted>" ownerKey={} avg=5"
```

The PR adds an implementation of `Stringer` interface for `podOwner` struct. Example of logs after the change:

```
"Average occurrence per node" node="<redacted>" ownerKey="loki/StatefulSet/loki-write/docker.io/grafana/loki:2.8.2" avg=6
```

### DIFF
--- a/pkg/framework/plugins/removeduplicates/removeduplicates.go
+++ b/pkg/framework/plugins/removeduplicates/removeduplicates.go
@@ -57,6 +57,10 @@ type podOwner struct {
 	imagesHash            string
 }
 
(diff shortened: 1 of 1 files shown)


JaneLiuL (commented):
/lgtm

a7i:
/approve

Thank you for your contributin @dimunech

PR #1289, opened 2023-11-17
feat: Add support for topologySpreadConstraints

UAntBou:
Add support for [topologySpreadConstraints](https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/)

### DIFF
--- a/charts/descheduler/README.md
+++ b/charts/descheduler/README.md
@@ -84,6 +84,7 @@ The following table lists the configurable parameters of the _descheduler_ chart
 | `serviceMonitor.metricRelabelings`  | MetricRelabelConfigs to apply to samples after scraping, but before ingestion                                         | `[]`                                      |
 | `serviceMonitor.relabelings`        | RelabelConfigs to apply to samples before scraping                                                                    | `[]`                                      |
 | `affinity`                          | Node affinity to run the descheduler cronjob/deployment on specific nodes                                             | `nil`                                     |
+| `topologySpreadConstraints`         | Topology Spread Constraints to spread the descheduler cronjob/deployment across the cluster                           | `[]`                                     |
 | `tolerations`                       | tolerations to run the descheduler cronjob/deployment on specific nodes                                               | `nil`                                     |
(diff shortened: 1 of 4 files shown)


charts/descheduler/README.md
```diff
@@ -84,6 +84,7 @@ The following table lists the configurable parameters of the _descheduler_ chart
 | `serviceMonitor.metricRelabelings`  | MetricRelabelConfigs to apply to samples after scraping, but before ingestion                                         | `[]`                                      |
 | `serviceMonitor.relabelings`        | RelabelConfigs to apply to samples before scraping                                                                    | `[]`                                      |
 | `affinity`                          | Node affinity to run the descheduler cronjob/deployment on specific nodes                                             | `nil`                                     |
+| `topologySpreadConstraints`         | Topology Spread Constraints to spread the descheduler cronjob/deployment across the cluster                           | `[]`                                     |
```
a7i, line 87: Spread and skew does not apply to a Job as it's single replica (so there's never a skew)
```suggestion
| `topologySpreadConstraints`         | Topology Spread Constraints to spread the descheduler HA Deployment across the cluster                           | `[]`                                     |
```

charts/descheduler/templates/cronjob.yaml
```diff
@@ -51,6 +51,10 @@ spec:
           affinity:
             {{- toYaml . | nindent 12 }}
           {{- end }}
+          {{- with .Values.topologySpreadConstraints }}
+          topologySpreadConstraints:
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
```
a7i, line 57: Not sure of the value of adding this here, thoughts?
```suggestion
```

a7i (commented):
Do we need to make sure that Descheduler doesn't evict itself?

charts/descheduler/templates/cronjob.yaml
UAntBou: I actually copy/pasted and adapted `affinity` so I included this here. I agree it may not be pertinent but I added it for consistency

PR #1290, opened 2023-11-20
[1.26] bump go version to 1.19.13

markdingram:
I note there are a few backports already on the 1.26 branch... could a tag be made to generate a v0.26.2 image? I would follow up with the following PR for Helm / Docs changes.

(also happy to do corresponding bumps for 1.27?)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-FROM golang:1.19.3
+FROM golang:1.19.13
 
 WORKDIR /go/src/sigs.k8s.io/descheduler
 COPY . .


markdingram:
I note there are a few backports already on the 1.26 branch... could a tag be made to generate a v0.26.2 image? I would follow up with the following PR for Helm / Docs changes.

a7i:
Hi @markdingram is there a reason for the bump? Is it related to any CVEs?

markdingram:
> Hi @markdingram is there a reason for the bump? Is it related to any CVEs?

I started to look through the 30 currently reported by a corporate scanner (twistlock). Nothing that appears material in the first 10 I looked through.

Thought to test the water here to see if publishing a new version would help reduce the scope of investigation before continuing through the rest.

JaneLiuL (commented):
/ok-to-test

a7i:
Hi @markdingram , now that we have released 1.29, we only maintain 3 versions [1.27, 1.28, and 1.29]. Additionally releases require some effort so this will not be a trivial change.

Going to close the PR. Thank you for your contribution!

PR #1305, opened 2023-11-28
release v0.28.1: bump helm chart and images

a7i:
similar to #1304 but target is master


a7i:
/cc @knelasevero @jklaw90 

same PR as #1304 , target is master

PR #1312, opened 2023-12-05
feat: Flag for enabling recording of failed evictions

grzesuav:


grzesuav:
@damemi @jklaw90 if I got it right I need to update md ?

jklaw90:
@grzesuav thank you for the pr! by any chance was there any discussion about this feature anywhere i didn't find an issue.
I could see a use case where we want to enable/disable this per profile. @a7i do you have any opinions here?

grzesuav:
https://kubernetes.slack.com/archives/C09TP78DV/p1700578405762779?thread_ts=1700578405.762779&cid=C09TP78DV

grzesuav:
@damemi @jklaw90 I rebased, discussion linked in the previous message, WDYT ?

cmd/descheduler/app/options/options.go
```diff
@@ -102,6 +102,7 @@ func (rs *DeschedulerServer) AddFlags(fs *pflag.FlagSet) {
 	fs.Float64Var(&rs.Tracing.SampleRate, "otel-sample-rate", 1.0, "Sample rate to collect the Traces")
 	fs.BoolVar(&rs.Tracing.FallbackToNoOpProviderOnError, "otel-fallback-no-op-on-error", false, "Fallback to NoOp Tracer in case of error")
 	fs.BoolVar(&rs.EnableHTTP2, "enable-http2", false, "If http/2 should be enabled for the metrics and health check")
+	fs.BoolVar(&rs.RecordEventsForEvictionErrors, "record-events-for-eviction-errors", false, "Set this flag to record events in case of eviction errors")
```
jklaw90, line 105: probably should be true by default to not change functionality for users.

jklaw90:
I think we should make this scoped to a profile so we can be more exact with the option.
cc @a7i

cmd/descheduler/app/options/options.go
grzesuav: actually the `false` is the default to not change any behaviour

googs1025:
This similar feature has been implemented in this pr https://github.com/kubernetes-sigs/descheduler/pull/1536 :)

PR #1313, opened 2023-12-06
Add Support to ppc64le Architecture

diogomurta:
This request aims to have support for the ppc64le Architecture used in OpenShift installed on IBM Power, widely used in on-premises environments.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -88,7 +88,7 @@ image.arm64:
 	$(CONTAINER_ENGINE) build --build-arg VERSION="$(VERSION)" --build-arg ARCH="arm64" -t $(IMAGE)-arm64 .
 
 image.ppc64le:
-        docker build --build-arg VERSION="$(VERSION)" --build-arg ARCH="ppc64le" -t $(IMAGE)-ppc64le .
+	$(CONTAINER_ENGINE) build --build-arg VERSION="$(VERSION)" --build-arg ARCH="ppc64le" -t $(IMAGE)-ppc64le .
 
 push: image
 	gcloud auth configure-docker
--- a/Makefile
+++ b/Makefile
@@ -71,7 +71,6 @@ build.arm64:
 build.ppc64le:
         CGO_ENABLED=0 GOOS=linux GOARCH=ppc64le go build ${LDFLAGS} -o _output/bin/descheduler sigs.k8s.io/descheduler/cmd/descheduler
 
-
 dev-image: build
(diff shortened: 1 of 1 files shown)


a7i:
I would be supportive for this but we need to make the corresponding changes in [here](https://github.com/kubernetes/k8s.io/blob/main/registry.k8s.io/images/k8s-staging-descheduler/images.yaml) as well. I'm happy to help with that if others think that we should support additional archs.

cc: @knelasevero  and @jklaw90  thoughts?

@diogomurta would you please squash your commits?

Makefile
```diff
@@ -83,6 +87,9 @@ image.arm:
 image.arm64:
 	$(CONTAINER_ENGINE) build --build-arg VERSION="$(VERSION)" --build-arg ARCH="arm64" -t $(IMAGE)-arm64 .
 
+image.ppc64le:
+        docker build --build-arg VERSION="$(VERSION)" --build-arg ARCH="ppc64le" -t $(IMAGE)-ppc64le .
```
a7i, line 91: some of the maintainers use podman so let's continue to support the existing pattern:
```suggestion
	$(CONTAINER_ENGINE) build --build-arg VERSION="$(VERSION)" --build-arg ARCH="ppc64le" -t $(IMAGE)-ppc64le .
```

Makefile
```diff
@@ -68,6 +68,10 @@ build.arm:
 build.arm64:
 	CGO_ENABLED=0 GOOS=linux GOARCH=arm64 go build ${LDFLAGS} -o _output/bin/descheduler sigs.k8s.io/descheduler/cmd/descheduler
 
+build.ppc64le:
+        CGO_ENABLED=0 GOOS=linux GOARCH=ppc64le go build ${LDFLAGS} -o _output/bin/descheduler sigs.k8s.io/descheduler/cmd/descheduler
+
+
```
a7i, line 74:
```suggestion
```

nit: let's be consistent with the rest of the Makefile and use a single new line

jklaw90:
@a7i sounds good, ibm 390x might be worth adding too.

a7i:
/ok-to-test

I can try to get this in 0.30

knelasevero:
cc: @knelasevero and @jklaw90 thoughts?

Also supportive, I would just be mindful of giving it or not giving it official support to those different architectures, and we should decide if we would prioritise architecture specific issues if they arrise (had that issue and discussion in another repo, that's why I'm bringing this up) 

Anyways I don't think this will be a problem.

PR #1314, opened 2023-12-07
helm: allow configuration of ipFamilyPolicy

dongjiang1989:
This pull request adds allow configuration of ipFamilyPolicy for the descheduler service.

Since different versions of k8s have different support for ipv6/ipv4 dual stack, it is not enabled by default.

### DIFF
--- a/charts/descheduler/templates/service.yaml
+++ b/charts/descheduler/templates/service.yaml
@@ -10,6 +10,10 @@ metadata:
 spec:
   clusterIP: None
   ports:
+{{- if .Values.service.ipDualStack.enabled }}
+  ipFamilies: {{ toYaml .Values.service.ipDualStack.ipFamilies | nindent 4 }}
+  ipFamilyPolicy: {{ .Values.service.ipDualStack.ipFamilyPolicy }}
+{{- end }}
   - name: http-metrics
     port: 10258
     protocol: TCP
--- a/charts/descheduler/values.yaml
+++ b/charts/descheduler/values.yaml
@@ -212,3 +212,7 @@ serviceMonitor:
     #   targetLabel: nodename
     #   replacement: $1
     #   action: replace
+  ipDualStack:
+    enabled: false
+    ipFamilies: ["IPv6", "IPv4"]
+    ipFamilyPolicy: "PreferDualStack"


dongjiang1989:
cc @a7i  PTAL

charts/descheduler/values.yaml
```diff
@@ -189,6 +189,10 @@ livenessProbe:
 
 service:
   enabled: false
+  ipDualStack:
+    enabled: false
+    ipFamilies: ["IPv6", "IPv4"]
+    ipFamilyPolicy: "PreferDualStack"
```
a7i, line 195:
```suggestion
  ## @param service.ipFamilyPolicy [string], support SingleStack, PreferDualStack and RequireDualStack
  ##
  ipFamilyPolicy: ""
  ## @param service.ipFamilies [array] List of IP families (e.g. IPv4, IPv6) assigned to the service.
  ## Ref: https://kubernetes.io/docs/concepts/services-networking/dual-stack/
  ## E.g.
  ## ipFamilies:
  ##   - IPv6
  ##
  ipFamilies: []
```

charts/descheduler/templates/service.yaml
```diff
@@ -10,6 +10,10 @@ metadata:
 spec:
   clusterIP: None
   ports:
+{{- if .Values.service.ipDualStack.enabled }}
+  ipFamilies: {{ toYaml .Values.service.ipDualStack.ipFamilies | nindent 4 }}
+  ipFamilyPolicy: {{ .Values.service.ipDualStack.ipFamilyPolicy }}
+{{- end }}
```
a7i, line 16: your implementation breaks `ports`. suggesting below:

```suggestion
  {{- if .Values.service.ipFamilyPolicy }}
  ipFamilyPolicy: {{ .Values.service.ipFamilyPolicy }}
  {{- end }}
  {{- if .Values.service.ipFamilies }}
  ipFamilies: {{ toYaml .Values.service.ipFamilies | nindent 4 }}
  {{- end }}
  ports:
```

charts/descheduler/values.yaml
dongjiang1989: Thanks for your review. Done.

charts/descheduler/templates/service.yaml
dongjiang1989: Thanks for your review. Fixed

dongjiang1989:
> Thank you for the pr @dongjiang1989 In addition to requested changes, would you please squash and sign your commits?

Thanks for your review. 
Done. PTAL @a7i

a7i:
@dongjiang1989 just curious, why do you use a Service for Descheduler?

a7i:
> > @dongjiang1989 just curious, why do you use a Service for Descheduler?
> 
> The [Open Policy Agent](https://github.com/open-policy-agent/gatekeeper) is installed in the online kubernetes cluster. All services must have Dual Stack setting. It does not use this service itself.

Then you should be able to omit creating the `Service`, correct?

dongjiang1989:
> > > @dongjiang1989 just curious, why do you use a Service for Descheduler?
> > 
> > 
> > The [Open Policy Agent](https://github.com/open-policy-agent/gatekeeper) is installed in the online kubernetes cluster. All services must have Dual Stack setting. It does not use this service itself.
> 
> Then you should be able to omit creating the `Service`, correct?

Hmmm... maybe not. OPA must support Deployment and Service appearing in pairs. 🤔️

Issue #1187, opened 2023-07-11
Descheduling framework wrap up


ingvagabund:
Remaining bits:
- create https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/cmd/scheduler/main.go#L48-L63 like interface for users to easily register custom plugins
- simplify `pluginregistry.Register` in https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/descheduler/setupplugins.go#L39C2-L49 to pass less arguments
- ...

knelasevero:
As we discussed in the community meeting, wanted to list a few efforts that we can start looking at again with the current state of the framework (now that we unblocked a bunch of efforts that were paused before):

- We have a bunch of nodeFit discussions and feature requests:
    - https://github.com/kubernetes-sigs/descheduler/issues/640
    - https://github.com/kubernetes-sigs/descheduler/issues/845
    - https://github.com/kubernetes-sigs/descheduler/issues/1067
    - I think we started to draft something initial with https://github.com/kubernetes-sigs/descheduler/issues/1149 but it would come on the next API alpha version: https://github.com/kubernetes-sigs/descheduler/issues/1179
    
- OpenTelemetry tracing
    - We have a PR open with a lot of reviews, but outdated with new codebase: https://github.com/kubernetes-sigs/descheduler/pull/958
    
-  Custom tags for eviction metrics
    -  https://github.com/kubernetes-sigs/descheduler/issues/905
    
- Well, v1aplha3 itself is not exactly blocked and people can work on it https://github.com/kubernetes-sigs/descheduler/issues/1179

- Allowing event-triggered strategy runs 
    - https://github.com/kubernetes-sigs/descheduler/pull/488
    - https://github.com/kubernetes-sigs/descheduler/issues/696
    
- Out of tree plugins:
    - https://github.com/kubernetes-sigs/descheduler/issues/1089
    - https://github.com/kubernetes-sigs/descheduler/issues/679
    - https://github.com/kubernetes-sigs/descheduler/issues/557
    
Let me know if I am forgetting something. We probably should give weights to those with regards to complexity and importance and the prioritize what we want as must haves first.

knelasevero:
@damemi @ingvagabund @a7i

damemi:
Like we talked about on the sig call, we have 2 categories of things to work on now for what's next: tech debt we shelved in the interest of moving along a framework release, and feature development that was blocked on the framework refactoring too much of the code.

[...]

PR #1316, opened 2023-12-08
helm: Update default `descheduler/v1alpha1` to `descheduler/v1alpha2`

dongjiang1989:
This PR setting default deschedulerPolicyAPIVersion from `descheduler/v1alpha1` to `descheduler/v1alpha2`

### DIFF
--- a/charts/descheduler/values.yaml
+++ b/charts/descheduler/values.yaml
@@ -95,7 +95,7 @@ deschedulerPolicy:
   #   sampleRate: 1.0
   #   fallbackToNoOpProviderOnError: true
   profiles:
-    - name: ProfileName
+    - name: DefaultPolicyProfile
       pluginConfig:
       - name: "RemoveDuplicates"
       - name: "RemovePodsHavingTooManyRestarts"
--- a/charts/descheduler/values.yaml
+++ b/charts/descheduler/values.yaml
@@ -79,7 +79,7 @@ cmdOptions:
   v: 3
 
 # Recommended to use the latest Policy API version supported by the Descheduler app version
-deschedulerPolicyAPIVersion: "descheduler/v1alpha1"
+deschedulerPolicyAPIVersion: "descheduler/v1alpha2"
 
 deschedulerPolicy:
   # nodeSelector: "key1=value1,key2=value2"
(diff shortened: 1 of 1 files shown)


charts/descheduler/values.yaml
```diff
@@ -94,40 +94,47 @@ deschedulerPolicy:
...
+          - ScheduleAnyway
+      - name: "LowNodeUtilization"
+        args:
           thresholds:
-            cpu: 20
-            memory: 20
-            pods: 20
+            "cpu" : 20
+            "memory": 20
+            "pods": 20
```
dongjiang1989, line 121:
```suggestion
            cpu: 20
            memory: 20
            pods: 20
```

charts/descheduler/values.yaml
```diff
@@ -94,40 +94,47 @@ deschedulerPolicy:
...
           targetThresholds:
-            cpu: 50
-            memory: 50
-            pods: 50
+            "cpu" : 50
+            "memory": 50
+            "pods": 50
```
dongjiang1989, line 125:
```suggestion
            cpu: 50
            memory: 50
            pods: 50
```

knelasevero:
the other way around, getting rid of v1alpha1 will make it easier to wrap the framework up, and get rid of some of the complex conversions

I say we go with v1alph2 as default already. Quite a few releases with that up

damemi:
#1187 seems pretty open ended, looks like there is some architectural changes in the first comment. But other things like nodeFit updates or other features can be added later. Could we get an update on that issue of what is really in scope of "wrap up"? Having a list of what's left to be done would be helpful

If most of the wrap up work that's left is there to support v1alpha1, then I agree that switching to default here will help nudge users so we can drop v1alpha1.

knelasevero:
> If most of the wrap up work that's left is there to support v1alpha1, then I agree that switching to default here will help nudge users so we can drop v1alpha1.

It is not that most of the wrap up work that's left is there to support v1alpha1. We have to go through that list from Jan and that other list from me anyways. However, some of those items are way harder to work on with the v1alpha1 still in our repo (and the conversions we have here). `Simplifying the register` or even `having less arguments on the register function` all get easier to do when we dont have the conversion logic being carried into the registry (which I had to do since using the universal decoder directly was hard with versions so different [and all that thing about having plugin types indepedently in each plugin and not in the general API]). New versions that we might have (v1alpha2 or v1beta1 or w/e) will all be easier to handle since we probably will be able to just use the universal decoder since their structure will probably be way more similar (I think).

> Having a list of what's left to be done would be helpful

Isn't it already listed there?

damemi:
@knelasevero that issue lists a lot of tasks, and it's not clear which are done or in progress, or the priority that they need to be done in. So, that's why I was asking for an update on it.

If the goal is to make v1alpha2 the default so we can drop v1alpha1, then I agree with @a7i that this seems like something that would be done at the end of the wrap up. That doesn't mean it can't be done now, but we need to clarify what the scope of the wrap up really is.

If there are goals for v1alpha2 that aren't reasonable to support with v1alpha1 (such as simplifying the register), then we can identify those and break off v1alpha1 here. But I would like to see a final status update so we know where everything stands before making that call.

That said, I don't see a reason to block this PR if it's just updating the default value in the Helm chart. I actually think we should merge this, just not with the goal of immediately following up with dropping v1alpha1 (if that makes sense?)

knelasevero:
> @knelasevero that issue lists a lot of tasks, and it's not clear which are done or in progress, or the priority that they need to be done in. So, that's why I was asking for an update on it.

Sorry, I misread it :sweat_smile:  Having an update on each item and if they are finished or how far we are on each makes a lot of sense.

> If the goal is to make v1alpha2 the default so we can drop v1alpha1, then I agree with @a7i that this seems like something that would be done at the end of the wrap up. That doesn't mean it can't be done now, but we need to clarify what the scope of the wrap up really is.

I agree, I think there are conflicting understandings of what the wrap up would be. I would have to think a little bit about that one, and I guess we can go over this in the community call or the async call we wanted to have eventually.

> If there are goals for v1alpha2 that aren't reasonable to support with v1alpha1 (such as simplifying the register), then we can identify those and break off v1alpha1 here. But I would like to see a final status update so we know where everything stands before making that call.

What do you mean by break off? (do we want to move this conversation to somewhere else? I am ok keeping it here, just asking since this is going a bit offtopic for the PR)

> That said, I don't see a reason to block this PR if it's just updating the default value in the Helm chart. I actually think we should merge this, just not with the goal of immediately following up with dropping v1alpha1 (if that makes sense?)

Agree, I should have been more clear in my previous comment.

> the other way around, getting rid of v1alpha1 will make it easier to wrap the framework up, and get rid of some of the complex conversions

> I say we go with v1alph2 as default already. Quite a few releases with that up

When I said this I meant not in this PR, and not necessarily in something that would be an immediate follow up, more like we should start making more efforts in that direction, and this PR is on of the steps.

damemi:
@knelasevero sounds good, sorry to derail this PR. I'm +1 to unblocking it then. By "break off" I meant make it clear where our support/feature set for v1alpha1 is ending (specifically in the context of the v1alpha2 work and how it affects that). But yes, we should move the discussion to a different place too :)

dongjiang1989:
> @dongjiang1989 thoughts on making it match what's in https://github.com/kubernetes-sigs/descheduler/pull/1139/files? That matches closer to what we have for kustomize w.r.t naming

Thanks for your review. I got it. This PR follow it

ingvagabund:
v1alpha2 has been around for some time. +1 making it "officially closed for any other change" and focusing on v1alpha3 instead.

a7i:
Already addressed in #1139 
Thank you for your contribution 🏆 
/close

PR #1317, opened 2023-12-12
fix: topologySpreadConstraint fields in structured logs

SataQiu:
There is a problem with the log display due to the use of non-exported fields：
https://github.com/kubernetes-sigs/descheduler/blob/f0f7ebbe9af680b338709391cf1c0da7c2270ce5/pkg/framework/plugins/removepodsviolatingtopologyspreadconstraint/topologyspreadconstraint.go#L220

Without this patch the `tsc` is printed as empty `{}`:
```sh
I1212 08:34:36.666387       1 topologyspreadconstraint.go:230] "Skipping topology constraint because it is already balanced" constraint={}
```

With this patch:
```sh
I1212 08:41:10.178744       1 topologyspreadconstraint.go:221] "Skipping topology constraint because it is already balanced" constraint={"MaxSkew":1,"TopologyKey":"topology.kubernetes.io/zone","Selector":[{}],"NodeAffinityPolicy":"Honor","NodeTaintsPolicy":"Ignore","PodNodeAffinity":{},"PodTolerations":[{"key":"node.kubernetes.io/not-ready","operator":"Exists","effect":"NoExecute","tolerationSeconds":300},{"key":"node.kubernetes.io/unreachable","operator":"Exists","effect":"NoExecute","tolerationSeconds":300}]}
```


a7i:
Thank you for your contribution @SataQiu 🏆

PR #1318, opened 2023-12-18
`PodLifeTime`: consider pods with container status `ImagePullBackOff`

a7i:
`PodLifeTime` plugin to support pods with container status `ImagePullBackOff`

```yaml
  containerStatuses:
  - image: test-podlifetime:latest
    imageID: ""
    lastState: {}
    name: test-podlifetime
    ready: false
    restartCount: 0
    started: false
    state:
      waiting:
        message: Back-off pulling image "test-podlifetime:latest"
        reason: ImagePullBackOff
```

### DIFF
--- a/pkg/framework/plugins/podlifetime/pod_lifetime_test.go
+++ b/pkg/framework/plugins/podlifetime/pod_lifetime_test.go
@@ -28,6 +28,7 @@ import (
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes/fake"
 	"k8s.io/client-go/tools/events"
+	utilptr "k8s.io/utils/ptr"
 	"sigs.k8s.io/descheduler/pkg/descheduler/evictions"
 	podutil "sigs.k8s.io/descheduler/pkg/descheduler/pod"
 	frameworkfake "sigs.k8s.io/descheduler/pkg/framework/fake"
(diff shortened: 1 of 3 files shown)


pkg/framework/plugins/podlifetime/pod_lifetime_test.go
```diff
@@ -28,6 +28,7 @@ import (
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes/fake"
 	"k8s.io/client-go/tools/events"
+	utilpointer "k8s.io/utils/pointer"
```
kannon92, line 31: This util was deprecated actually and should be replaced by k8s.io/utils/ptr.

a7i: Done, there are a lot of references to utilpointer still, but will track as a separate PR

kannon92: Thanks, that LGTM to me.

jklaw90:
should we mention this in the readme example? https://github.com/kubernetes-sigs/descheduler?tab=readme-ov-file#podlifetime

a7i:
good call @jklaw90 - updated ✔️

PR #1334, opened 2023-12-28
fix(podlifetime): fix failed unittest

xujihui1985:
correct desired pod creation time

"Two pods in the `dev` Namespace, 1 created 605 seconds ago. 1 should be evicted.", the create time stamp should be 605 s before current time

### DIFF
--- a/pkg/framework/plugins/podlifetime/pod_lifetime.go
+++ b/pkg/framework/plugins/podlifetime/pod_lifetime.go
@@ -25,6 +25,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/klog/v2"
+
 	frameworktypes "sigs.k8s.io/descheduler/pkg/framework/types"
 
 	"sigs.k8s.io/descheduler/pkg/descheduler/evictions"
@@ -67,8 +68,8 @@ func New(args runtime.Object, handle frameworktypes.Handle) (frameworktypes.Plug
 	}
 
 	podFilter = podutil.WrapFilterFuncs(podFilter, func(pod *v1.Pod) bool {
-		podAgeSeconds := uint(metav1.Now().Sub(pod.GetCreationTimestamp().Local()).Seconds())
-		return podAgeSeconds > *podLifeTimeArgs.MaxPodLifeTimeSeconds
+		podAgeSeconds := int(metav1.Now().Sub(pod.GetCreationTimestamp().Local()).Seconds())
+		return podAgeSeconds > int(*podLifeTimeArgs.MaxPodLifeTimeSeconds)
 	})
 
 	if len(podLifeTimeArgs.States) > 0 {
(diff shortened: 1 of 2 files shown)


damemi (commented):
/ok-to-test

ingvagabund:
@xujihui1985 nice catch!!!

I wonder how "Two pods in the `dev` Namespace, 1 created 605 seconds ago. 1 should be evicted." test has been passing until now when it should not.

xujihui1985:
> @xujihui1985 nice catch!!!
> 
> I wonder how "Two pods in the `dev` Namespace, 1 created 605 seconds ago. 1 should be evicted." test has been passing until now when it should not.
> 
> Also, "Two pods in the `dev` Namespace, 1 is new and 1 very is old. 1 should be evicted." and "Two pods in the `dev` Namespace, 1 created 605 seconds ago. 1 should be evicted." tests are functionally equivalent wrt. `maxLifeTime`. The same holds for "Two pods in the `dev` Namespace, 2 are new and 0 are old. 0 should be evicted." and "Two pods in the `dev` Namespace, 1 created 595 seconds ago. 0 should be evicted.".


I happened to run the test in my ci system, and can't pass this test, so I look into the test and find out the test condition might be wrong, but I haven't check why this test used to be success.

ingvagabund:
Running the test from the current master:HEAD it passes. Though only because the [computation overflows](https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/framework/plugins/podlifetime/pod_lifetime.go#L70):
```
	podFilter = podutil.WrapFilterFuncs(podFilter, func(pod *v1.Pod) bool {
		podAgeSeconds := uint(metav1.Now().Sub(pod.GetCreationTimestamp().Local()).Seconds())
		return podAgeSeconds > *podLifeTimeArgs.MaxPodLifeTimeSeconds
	})
```
The code does not assume a pod gets created in the future :) Dropping uint the case fails.

ingvagabund:
@xujihui1985 would you please also update the filter code as well? E.g.
```
	maxPodLifeTimeSeconds := int(*podLifeTimeArgs.MaxPodLifeTimeSeconds)
	podFilter = podutil.WrapFilterFuncs(podFilter, func(pod *v1.Pod) bool {
		podAgeSeconds := int(metav1.Now().Sub(pod.GetCreationTimestamp().Local()).Seconds())
		return podAgeSeconds > maxPodLifeTimeSeconds
	})
```
@a7i we should probably ignore any pod created in the future :)

xujihui1985:
sure, glad to do that

xujihui1985:
> @xujihui1985 would you please also update the filter code as well? E.g.
> 
> ```
> 	maxPodLifeTimeSeconds := int(*podLifeTimeArgs.MaxPodLifeTimeSeconds)
> 	podFilter = podutil.WrapFilterFuncs(podFilter, func(pod *v1.Pod) bool {
> 		podAgeSeconds := int(metav1.Now().Sub(pod.GetCreationTimestamp().Local()).Seconds())
> 		return podAgeSeconds > maxPodLifeTimeSeconds
> 	})
> ```
> 
> @a7i we should probably ignore any pod created in the future :)

I see, covert to uint make the podAgeSeconds a large number because of the nagtive sign, I have update the MR according to your suggestion,  but actually this future pod doesn't make sense in real life, so either we compare the creation seconds with convert it to int or if podAgeSeconds is < 0, we just return false

PR #1342, opened 2024-01-25
Allow eviction of DaemonSet pods

tioxy:
#### What type of PR is this?

/kind feature

#### What this PR does / why we need it:

Enables eviction of DaemonSet pods with the key `evictDaemonSetPods: true` for evictors

#### Which issue(s) this PR fixes:

NONE

#### Special notes for your reviewer:

By default, allowing eviction of pods from daemonsets may be dangerous if misconfigured.

With the release of `v1alpha2` profiles, we got the ability to tune granular specs for eviction profiles.  This means now that it is possible to select specific daemonset to allow pod eviction automatically. This is interesting because reduces toil of pod recreation for specific scenarios.

#### Does this PR introduce a user-facing change?

```release-note
Allow eviction of daemonset pods by `evictDaemonSetPods` key
```

#### Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:

```docs
NONE
```

### DIFF
--- a/pkg/api/v1alpha1/conversion.go
+++ b/pkg/api/v1alpha1/conversion.go
@@ -139,6 +139,14 @@ func V1alpha1ToInternal(
 		}
 	}
 
+	evictDaemonSetPods := false
+	if deschedulerPolicy.EvictDaemonSetPods != nil {
+		evictDaemonSetPods = *deschedulerPolicy.EvictDaemonSetPods
+		if evictDaemonSetPods {
+			klog.V(1).Info("Warning: EvictDaemonSetPods is set to True. This could cause eviction of Kubernetes DaemonSet pods.")
+		}
+	}
+
 	ignorePvcPods := false
 	if deschedulerPolicy.IgnorePVCPods != nil {
 		ignorePvcPods = *deschedulerPolicy.IgnorePVCPods
@@ -193,6 +201,7 @@ func V1alpha1ToInternal(
 							Name: defaultevictor.PluginName,
 							Args: &defaultevictor.DefaultEvictorArgs{
 								EvictLocalStoragePods:   evictLocalStoragePods,
+								EvictDaemonSetPods:      evictDaemonSetPods,
 								EvictSystemCriticalPods: evictSystemCriticalPods,
 								IgnorePvcPods:           ignorePvcPods,
 								EvictFailedBarePods:     evictBarePods,
--- a/pkg/api/v1alpha1/types.go
+++ b/pkg/api/v1alpha1/types.go
@@ -41,6 +41,9 @@ type DeschedulerPolicy struct {
 	// EvictSystemCriticalPods allows eviction of pods of any priority (including Kubernetes system pods)
 	EvictSystemCriticalPods *bool `json:"evictSystemCriticalPods,omitempty"`
 
+	// EvictDaemonSetPods allows pods owned by a DaemonSet resource to be evicted.
+	EvictDaemonSetPods *bool `json:"evictDaemonSetPods,omitempty"`
+
 	// IgnorePVCPods prevents pods with PVCs from being evicted.
 	IgnorePVCPods *bool `json:"ignorePvcPods,omitempty"`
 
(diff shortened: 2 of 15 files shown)


tioxy:
One thing that I was thinking: should this key really be supported on `v1alpha1`?  I ended up doing it for the sake of consistency between both versions, but I see some problems:

- It may be more dangerous if misconfigured on the previous version
- Looks like `v1alpha1` is being deprecated already

a7i:
> One thing that I was thinking: should this key really be supported on `v1alpha1`? I ended up doing it for the sake of consistency between both versions, but I see some problems:


As it's currently implemented, setting the value to `true` does not enable the feature since it's not implemented [here](https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/api/v1alpha1/conversion.go#L121) in the conversion code.

a7i:
@tioxy would you please add a unit and/or e2e test for this to verify that it's working as expected? Thank you!

tioxy:
Thank you for the clarification @a7i, I will be trying to code these tests soon.

pkg/api/v1alpha1/types.go
```diff
@@ -38,6 +38,9 @@ type DeschedulerPolicy struct {
 	// EvictLocalStoragePods allows pods using local storage to be evicted.
 	EvictLocalStoragePods *bool `json:"evictLocalStoragePods,omitempty"`
 
+	// EvictDaemonSetPods allows pods owned by a DaemonSet resource to be evicted.
+	EvictDaemonSetPods *bool `json:"evictDaemonSetPods,omitempty"`
```
a7i, line 42: If we want this to work for v1alpha1, then the conversion has to be updated to include it [ref](https://github.com/kubernetes-sigs/descheduler/blob/bb5930eb21605879104e02d9da62bd96180d9492/pkg/api/v1alpha1/conversion.go#L115)

tioxy: To be fair I believe this can be a dangerous behavior, but making it by default disabled and allowing enabling it optionally seems to be a good common ground.

tioxy:
I just added the missing E2E test. To be honest, I got no clue if this test is implemented properly 😅.

I based myself on the `testEvictSystemCritical` structure and followed it.

pkg/api/v1alpha1/conversion.go
```diff
@@ -139,6 +139,14 @@ func V1alpha1ToInternal(
 		}
 	}
 
+	evictDaemonSetPods := false
```
ingvagabund, line 142: `evictDaemonSetPods` needs to be wired to `DefaultEvictorArgs`

ingvagabund: Would you also create a unit test for the conversion? To make sure the conversion produces the expected outcome.

tioxy: Hi @ingvagabund, thank you for the review.

Added the missing reference in the last commit for `DefaultEvictorArgs`.

About the unit test for conversion, where should be the ideal place to write it? I'm little lost if it should be within `strategymigration_test.go` or other file that I may be missing.

ingvagabund: https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/descheduler/policyconfig_test.go#L43 under `TestV1alpha1ToV1alpha2`

tioxy: Added the missing tests

ingvagabund:
@tioxy the PR looks good in overall. Would you please reduce the amount of commits in the PR? So it's easier to follow the changes. E.g. one commit for a new functionality, one for tests, one for documentation. Or, any reasonable distribution of changes.

Note: we are going to remove v1alpha1 type. So, any changes related to v1alpha1 can be safely dropped. Users are strongly recommended to migrate all their v1alpha1 policies to v1alpha2.

tioxy:
Reorganized the commits 👍 

Great review @ingvagabund , thanks

tioxy:
Hi @a7i @ingvagabund 👋 

Do I need to make any extra changes to the PR to be merged?

ingvagabund:
Apologies for the delay. I will put it on my todo list for tomorrow.

ingvagabund:
@tioxy this is on me. I did not realize v1alpha1 is already deprecated when I performed the first review by suggesting to provide testing for v1alpha1 -> v1alpha2 conversion. In the last descheduling community meeting we decided to remove v1alpha1 completely in the next release: https://groups.google.com/g/kubernetes-sig-scheduling/c/OyFjkDntuhg/m/GVjD4gpaBAAJ?utm_medium=email&utm_source=footer. Given v1alpha1 is already deprecated there's no point of extending the existing v1alpha1 type. Would you please drop any code related to v1alpha1 from your PR? Apologies for the unnecessary complications.

a7i:
@ingvagabund your call, but happy to merge this as-is. Given that we're removing the code, this new addition will also be removed.

/approve
/hold for Jan's final review

PR #1346, opened 2024-02-01
trim managedFields in shared informer factory

a7i:
Similar to https://github.com/kubernetes/autoscaler/pull/6159

We recently upgraded to autoscaler 1.29 and noticed a huge memory utilization drop and after further investigation, it appears to be related to omitting managedFields.

For the cluster below, we use server-side-apply (SSA) which stores a big chunk in managedFields.
Screenshot 2024-02-01 at 10 02 50 AM


knelasevero:
nice catch!

Issue #1347, opened 2024-02-01
Single node clusters support


pablo-suazo:
I would like to use descheduler in a single node kubernetes cluster.

It is useful to me to automatically delete pods with too many restart or failed pods.

It is possible to overrride this check in any way?

`if len(nodes) <= 1 {
		klog.V(1).InfoS("The cluster size is 0 or 1 meaning eviction causes service disruption or degradation. So aborting..")
		return fmt.Errorf("the cluster size is 0 or 1")
	}
`

larryrensing:
any movement on this? it would be great to have!

erikgb:
This issue becomes even more annoying when switching from cronjob to deployment workload. Since the rescheduled aborts, this makes the deployment pod crash-loop.

I suggest adding an option to enable support for single-node clusters. The default can still be to abort in single-node clusters to give users more awareness of the additional disturbance running descheduler with just a single cluster node.

panteparak:
At work, I scale down replicas down to 1/2 on many of our apps during night times, which reduced down the number of nodes needed to 1 and cause the same error as above

alvistar:
Hi, I forked the code https://github.com/alvistar/descheduler/tree/singlenode

Here I am bypassing the check provided an env variable is set.

This works for me. I am not opening a pull request as I have no time to proper make tests. Today it seems all tests are e2e on a cluster of 3 nodes. So to do things nicely it would be created a single cluster test just for this feat.

If somebody would take the task from here it would be great.

erikgb:
Still highly desired!

/remove-lifecycle rotten

ingvagabund:
https://github.com/kubernetes-sigs/descheduler/pull/1350#discussion_r1710988841 is still valid. This will need some survey/design considerations. Yet each plugin could decide it can run over a single node cluster. Resp. provide the minimal number of nodes required for running. So if there's only a single (or N nodes) and at least one plugin that accepts this number of nodes the descheduling cycle could run normally.

Issue #1298, opened 2023-11-24
Cluster Size check shouldn't exit process when running as Deployment


markdingram:
**Is your feature request related to a problem? Please describe.**

When there are 0 or 1 nodes the descheduler loop returns error `the cluster size is 0 or 1` & the process exits.

This doesn't play nicely when the descheduler is running as a Deployment - the pod goes into `CrashLoopBackOff` due to the repeated early exits.

**Describe the solution you'd like**

When the descheduler is running as a Deployment `the cluster size is 0 or 1` check shouldn't exit the process. The process should remain running until the next iteration.

Something like (in `runDeschedulerLoop`):

```
	// if len is still <= 1 error out
	if len(nodes) <= 1 {
		klog.V(1).InfoS("The cluster size is 0 or 1 meaning eviction causes service disruption or degradation. So aborting..")
		if d.rs.DeschedulingInterval.Seconds() == 0 {
			return fmt.Errorf("the cluster size is 0 or 1")
		} else {
			return nil
		}
	}
```

**Describe alternatives you've considered**

**What version of descheduler are you using?**

descheduler version:

0.28.0

**Additional context**

Example logs

```

I1124 16:00:06.155189       1 node.go:50] "Node lister returned empty list, now fetch directly"
I1124 16:00:06.160342       1 descheduler.go:121] "The cluster size is 0 or 1 meaning eviction causes service disruption or degradation. So aborting.."
E1124 16:00:06.160453       1 descheduler.go:431] the cluster size is 0 or 1
I1124 16:00:06.160876       1 reflector.go:295] Stopping reflector *v1.Pod (0s) from k8s.io/client-go/informers/factory.go:150
I1124 16:00:06.160913       1 reflector.go:295] Stopping reflector *v1.PriorityClass (0s) from k8s.io/client-go/informers/factory.go:150
I1124 16:00:06.161268       1 secure_serving.go:255] Stopped listening on [::]:10258
I1124 16:00:06.161288       1 tlsconfig.go:255] "Shutting down DynamicServingCertificateController"
I1124 16:00:06.161487       1 reflector.go:295] Stopping reflector *v1.Node (0s) from k8s.io/client-go/informers/factory.go:150
I1124 16:00:06.161529       1 reflector.go:295] Stopping reflector *v1.Namespace (0s) from k8s.io/client-go/informers/factory.go:150
```

```
Events:
  Type     Reason     Age                   From               Message
  ----     ------     ----                  ----               -------
...
[...]

PR #1350, opened 2024-02-11
add `ignoreMinNodesCheck` to allow disabling minimum node size check

a7i:
Closes https://github.com/kubernetes-sigs/descheduler/issues/1347
and https://github.com/kubernetes-sigs/descheduler/issues/1298

### DIFF
--- a/pkg/descheduler/descheduler.go
+++ b/pkg/descheduler/descheduler.go
@@ -25,25 +25,25 @@ import (
 
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
-	"k8s.io/apimachinery/pkg/api/meta"
-	"k8s.io/client-go/discovery"
-	"k8s.io/client-go/informers"
-	"k8s.io/client-go/tools/events"
-	componentbaseconfig "k8s.io/component-base/config"
-	"k8s.io/klog/v2"
-
 	v1 "k8s.io/api/core/v1"
 	policy "k8s.io/api/policy/v1"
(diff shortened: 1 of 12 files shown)


pablo-suazo:
@jklaw90 @lixiang233 Could you check this? Thank you

pkg/descheduler/descheduler.go
```diff
@@ -116,8 +116,9 @@ func (d *descheduler) runDeschedulerLoop(ctx context.Context, nodes []*v1.Node)
 		metrics.DeschedulerLoopDuration.With(map[string]string{}).Observe(time.Since(loopStartDuration).Seconds())
 	}(time.Now())
 
+	shouldCheckMinNodes := !ptr.Deref(d.deschedulerPolicy.IgnoreMinNodesCheck, false)
 	// if len is still <= 1 error out
-	if len(nodes) <= 1 {
+	if shouldCheckMinNodes && len(nodes) <= 1 {
```
ingvagabund, line 121: I wonder whether we could limit this check only for plugins with Balance extension point. Normally, any Descheduler EP should be allowed to operator over any node independent of how many there are.

From https://github.com/kubernetes-sigs/descheduler/issues/1347:
```
It is useful to me to automatically delete pods with too many restart or failed pods.
```

a7i: Good feedback. That makes sense to me

I can take a stab at this soon

ingvagabund: Thx. Lemme know once you need a review. I'll take a look.

Issue #1253, opened 2023-10-03
Add support for pod deletion (vs. eviction)


ichekrygin:
There are times when "evict" is not sufficient to remove pods.

Consider a use-case when the pod is stuck in the "Terminating" state, which could happen for various reasons, but mainly due to the "kubelet" related issues (not running, network, etc).

Currently, we typically run `kubectl delete --force --grace-period=0` to overcome this issue. 

Granted, such a workaround is not always applicable to all workloads (for example, stateful, etc.).
It is still could be very useful to craft a policy that would address the use case of pods stuck in a "Terminating" state.

a7i:
Duplicate of #1170

PR #1354, opened 2024-02-26
KEP-1397: descheduler integration with evacuation API as an alternative to eviction API

ingvagabund:

The descheduler eviction policy is built on top of the eviction API. The API currently does not support eviction requests that are not completed right away. Instead, any eviction needs to either succeed or be rejected in response. Nevertheless, there are cases where an eviction request is expected to only initiate eviction. While getting confirmation or rejection of the eviction initiation (or its promise).

Utilize [evacuation API](https://github.com/kubernetes/enhancements/pull/4565) as an alternative to eviction API. As an interim solution (until the evacuation API is available) allow to interpret pods with a special annotation as a request for eviction initiation instead of expecting an immediate eviction.

Resolves: https://github.com/kubernetes-sigs/descheduler/issues/1397

### DIFF
--- a/keps/1397-evictions-in-background/README.md
+++ b/keps/1397-evictions-in-background/README.md
@@ -1,0 +1,1122 @@
+<!--
+**Note:** When your KEP is complete, all of these comment blocks should be removed.
+
+To get started with this template:
+
+- [ ] **Pick a hosting SIG.**
+  Make sure that the problem space is something the SIG is interested in taking
+  up. KEPs should not be checked in without a sponsoring SIG.
+- [ ] **Create an issue in kubernetes/enhancements**
+  When filing an enhancement tracking issue, please make sure to complete all
+  fields in that template. One of the fields asks for a link to the KEP. You
+  can leave that blank until this KEP is filed, and then go back to the
+  enhancement and add the link.
+- [ ] **Make a copy of this template directory.**
+  Copy this template into the owning SIG's directory and name it
+  `NNNN-short-descriptive-title`, where `NNNN` is the issue number (with no
+  leading-zero padding) assigned to your enhancement above.
+- [ ] **Fill out as much of the kep.yaml file as you can.**
+  At minimum, you should fill in the "Title", "Authors", "Owning-sig",
+  "Status", and date-related fields.
+- [ ] **Fill out this file as best you can.**
+  At minimum, you should fill in the "Summary" and "Motivation" sections.
+  These should be easy if you've preflighted the idea of the KEP with the
+  appropriate SIG(s).
+- [ ] **Create a PR for this KEP.**
+  Assign it to people in the SIG who are sponsoring this process.
+- [ ] **Merge early and iterate.**
+  Avoid getting hung up on specific details and instead aim to get the goals of
+  the KEP clarified and merged quickly. The best way to do this is to just
+  start with the high-level sections and fill out details incrementally in
+  subsequent PRs.
+
+Just because a KEP is merged does not mean it is complete or approved. Any KEP
+marked as `provisional` is a working document and subject to change. You can
+denote sections that are under active debate as follows:
+
+```
+<<[UNRESOLVED optional short context or usernames ]>>
+Stuff that is being argued.
+<<[/UNRESOLVED]>>
+```
+
+When editing KEPS, aim for tightly-scoped, single-topic PRs to keep discussions
+focused. If you disagree with what is already in a document, open a new PR
+with suggested changes.
+
+One KEP corresponds to one "feature" or "enhancement" for its whole lifecycle.
+You do not need a new KEP to move from beta to GA, for example. If
+new details emerge that belong in the KEP, edit the KEP. Once a feature has become
+"implemented", major changes should get new KEPs.
+
+The canonical place for the latest set of instructions (and the likely source
+of this file) is [here](/keps/NNNN-kep-template/README.md).
+
+**Note:** Any PRs to move a KEP to `implementable`, or significant changes once
+it is marked `implementable`, must be approved by each of the KEP approvers.
+If none of those approvers are still appropriate, then changes to that list
+should be approved by the remaining approvers and/or the owning SIG (or
+SIG Architecture for cross-cutting KEPs).
+-->
+# KEP-1397: descheduler integration with evacuation API as an alternative to eviction API
+
+<!--
+This is the title of your KEP. Keep it short, simple, and descriptive. A good
+title can help communicate what the KEP is and should be considered as part of
+any review.
+-->
+
+<!--
+A table of contents is helpful for quickly jumping to sections of a KEP and for
+highlighting any additional information provided beyond the standard KEP
+template.
+
+Ensure the TOC is wrapped with
+  <code>&lt;!-- toc --&rt;&lt;!-- /toc --&rt;</code>
+tags, and then generate with `hack/update-toc.sh`.
+-->
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [Expected Outcomes](#expected-outcomes)
+  - [User Stories (Optional)](#user-stories-optional)
+  - [Annotation vs. evacuation API based eviction](#annotation-vs-evacuation-api-based-eviction)
+  - [Workflow Description](#workflow-description)
+    - [Actors](#actors)
+    - [Workflow Steps](#workflow-steps)
+  - [Notes/Constraints/Caveats (Optional)](#notesconstraintscaveats-optional)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Implementation Strategies](#implementation-strategies)
+    - [API changes](#api-changes)
+    - [v1alpha1](#v1alpha1)
+    - [v1alphaN or beta](#v1alphan-or-beta)
+    - [Code changes](#code-changes)
+    - [Metrics](#metrics)
+  - [Open Questions [optional]](#open-questions-optional)
+  - [Test Plan](#test-plan)
+      - [Prerequisite testing updates](#prerequisite-testing-updates)
+      - [Unit tests](#unit-tests)
+      - [Integration tests](#integration-tests)
+      - [e2e tests](#e2e-tests)
+  - [Graduation Criteria](#graduation-criteria)
+    - [Alpha](#alpha)
+    - [Beta](#beta)
+    - [GA](#ga)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Infrastructure Needed (Optional)](#infrastructure-needed-optional)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+<!--
+**ACTION REQUIRED:** In order to merge code into a release, there must be an
+issue in [kubernetes/enhancements] referencing this KEP and targeting a release
+milestone **before the [Enhancement Freeze](https://git.k8s.io/sig-release/releases)
+of the targeted release**.
+
+For enhancements that make changes to code or processes/procedures in core
+Kubernetes—i.e., [kubernetes/kubernetes], we require the following Release
+Signoff checklist to be completed.
+
+Check these off as they are completed for the Release Team to track. These
+checklist items _must_ be updated for the enhancement to be released.
+-->
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [ ] (R) KEP approvers have approved the KEP status as `implementable`
+- [ ] (R) Design details are appropriately documented
+- [ ] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input (including test refactors)
+  - [ ] e2e Tests for all Beta API Operations (endpoints)
+  - [ ] (R) Ensure GA e2e tests meet requirements for [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
+  - [ ] (R) Minimum Two Week Window for GA e2e tests to prove flake free
+- [ ] (R) Graduation criteria is in place
+  - [ ] (R) [all GA Endpoints](https://github.com/kubernetes/community/pull/1806) must be hit by [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
+- [ ] (R) Production readiness review completed
+- [ ] (R) Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+<!--
+**Note:** This checklist is iterative and should be reviewed and updated every time this enhancement is being considered for a milestone.
+-->
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+The descheduler eviction policy is built on top of the eviction API.
+The API currently does not support eviction requests that are not completed right away.
+Instead, any eviction needs to either succeed or be rejected in response.
+Nevertheless, there are cases where an eviction request is expected to only initiate eviction.
+While getting confirmation or rejection of the eviction initiation (or its promise).
+
+The descheduler treats all pods as cattle.
+Each descheduling loop consists of  pods nomination followed by their eviction.
+When eviction of one pod does not succeed another pod is selected until a limit
+on the number of evictions is reached or all nominated pods are evicted.
+The descheduler does not keep track of pods that failed to be evicted.
+The next descheduling loop repeats the same routine.
+The order of eviction is not guaranteed.
+Thus, any subsequent loop can either evict pods that failed to be evicted
+in the previous loop or completely ignore them if a limit is reached.
+Thus, if an administrator installs a webhook that intercepts eviction requests
+while changing the underlying functionality (e.g. eviction in background
+initiation but returning an error code instead of success),
+the descheduler can do more harm than good.
+
+## Motivation
+
+Kubernetes ecosystem is wide.
+There are various types of workloads in various environments that may require
+different eviction policies to be applied. Some pods can be evicted right away,
+some pods may need more time for clean up. Other pods may need more time
+than a graceful termination period gives or be able to retry eviction
+in case a live migration did not succeed.
+
+### Goals
+
+- **Enhance Descheduler eviction policy**: Allow pods that require eviction
+  without getting evicted right away to get a promise from third part
+  component on handling the eviction without directly invoking the upstream eviction API.
+  Eventually replacing the eviction API with a new [evacuation API](https://github.com/kubernetes/enhancements/pull/4565).
+- **No more evictions, only requests for evacuation**: the descheduler will
+  no longer guarantee eviction. All the configurable limits per namespace/node/run will
+  change their semantics to a maximal number of evacuation requests per namespace/node/run.
+- **New sorting algorithm to prioritize in-progress evictions**: Sort the nominated pods
+  to prefer those with eviction already in progress (or its promise)
+  to minimize workload disruptions.
+- **New prefilter algorithm to exclude in-progress evictions**: Filter out all pods
+  that have an already existing evacuation request present.
+- **Measure pending evictions**: New metric to keep count of evictions in progress/evacuation requests
+
+### Non-Goals
+
+- Implementation of any customized eviction policy: The Descheduler will only
+  acknowledge existence of pods that require different eviction handling.
+  The actual eviction (e.g. live migration) will be implemented by third parties.
+
+## Proposal
+
+This enhancement proposes the implementation of a more informed
+eviction policy for the Descheduler.
+Allowing plugins (constructing lists of pods nominated for eviction)
+to be aware of pods with eviction in progress to avoid evicting
+more than a configured limit allows.
+Reducing unnecessary evictions can improve overall workload
+distribution and save cost of bringing up new pods.
+
+### Expected Outcomes
+
+- Live migration implemented through external admission webhooks
+  are no longer perceived as eviction errors.
+  Currently, no pod with a live migration in background is evicted properly.
+  Thus resulting in eviction of all such pods from all possible nodes
+  until a limit is reached. Which is an undesirable behaviour.
+- Descheduling plugins are more aware of pods with live migration in background.
+  Thus, improving the decision making behind which pods get nominated for eviction.
+- The descheduler will no longer be responsible for direct eviction.
+  Instead, only requesting eviction and letting other components (more
+  informed and following various policies) to perform the actual eviction.
+
+### User Stories (Optional)
+
+- As a cluster admin running a K8s cluster with a descheduler,
+  I want to evict KubeVirt pods that may require live migration of VMs
+  before a pod can be  evicted. A VM live migration can take minutes
+  to complete or may require retries.
+- As an end user deploying my application on a K8s cluster with a descheduler,
+  I want to evict pods while performing pod live migration to persist
+  pod's state without storing any necessary data into pod-independent persistent storage.
+- As a developer I want to be able to implement a custom eviction policy to
+  address various company use cases that are not supported
+  by default when running a descheduler instance.
+- As a descheduler plugin developer I want to be aware of evictions
+  in progress to improve pod nomination and thus avoid unnecessary disruptions.
+- As a security professional I want to make sure any live migration
+  adhers to security policies and protects sensitive data.
+
+### Annotation vs. evacuation API based eviction
+
+The [evacuation API](https://github.com/kubernetes/enhancements/pull/4565) is expected
+to replace the [eviction API](https://kubernetes.io/docs/concepts/scheduling-eviction/api-eviction/).
+Instead of checking whether an eviction was successful or failed, an evacuation request is created.
+In case the request fails to be created a pod eviction is skipped as previously.
+In addition, each cycle will keep a number of evacution requests created
+instead of a number of pods evicted.
+
+The annotation based approach is a special case of invoking the eviction API:
+
+|  | Eviction API | Evacuation API |
+|--|--|--|
+|**Eviction handling**|Eviction of a pod with `descheduler.alpha.kubernetes.io/request-evict-only` annotation. Checking the eviction error, return code and response text. Interpreting a specific error code and a presence of a known suffix in the response text. |Creation of an evacuation CR for the given pod. Checking the evacuation CR creation error.|
+|**Pod nomination and sorting**|Preferring pods with both `descheduler.alpha.kubernetes.io/request-evict-only` and `descheduler.alpha.kubernetes.io/eviction-in-progress` annotations.|Preferring pods that have a corresponding evacuation CR present.|
+|**Cache reconstruction**|When a descheduler gets restarted and a new internal caches (with tracked annotations) cleared the descheduler lists all pods and populate the cache with any pod that has both annotation (`descheduler.alpha.kubernetes.io/request-evict-only` and `descheduler.alpha.kubernetes.io/eviction-in-progress`) present. In case only the first annotation is present but the eviction request was already created, the update event handler will catch the second annotation addition and the cache gets synced. In the worst case the limit of max number of pods evicted gets exceeded.|The descheduler waits until all evacution CRs are synced.|
+
+Given the evacuation API is still a work in progress without any
+existing implementation, the annotation based eviction can be seen
+as an v1alpha1 implemenation of this proposal.
+Integration with the evacuation API as an v1alphaN or beta implementation.
+
+### Workflow Description
+
+This section outlines the workflow for enabling the proposed evictions
+in background that prevent workloads from being evicted right away
+without following additional eviction policies.
+
+#### Actors
+
+- Cluster Administrator: Responsible for configuring descheduling
+  policies and overseeing cluster operations.
+- User: Individuals or services running workloads within the cluster.
+- Descheduler: Component responsible for requesting pod evictions
+
+#### Workflow Steps
+
+1. Policy configuration
+   1. The cluster administrator configures the descheduler to enable the new functionality.
+2. A user deploys a workload through various nodes
+   1. A scheduler distributes the pods based on configured scheduling policies
+3. Eviction nomination
+   1. At the beginning of each descheduling cycle the descheduler increases
+      all the internal counters to take into account all pods that are subjects
+      to background eviction. Either by checking `descheduler.alpha.kubernetes.io/eviction-in-progress` annotation,
+      list of evacuation requests or confronting the internal caches.
+   2. Each descheduling plugin nominates a set of pods to be evicted.
+   3. All the nominated pods are sorted based on the eviction-in-progress first priority.
+4. Pod eviction: the descheduler starts evicting nominated pods until a limit is reached
+   1. If an annotation based eviction is enabled:
+      1. If a pod is not annotated with `descheduler.alpha.kubernetes.io/request-evict-only` evict
+         the pod normally without additional handling.
+      2. If a pod is annotated with `descheduler.alpha.kubernetes.io/request-evict-only` evict the pod.
+         1. If the eviction fails intepret the error as a request for eviction in background.
+         2. Error code, resp. response text is checked to distinguish between
+            a genuine error and a confirmation of an eviction in background.
+   2. If the evacuation API is utilized create an evacution request
+
+### Notes/Constraints/Caveats (Optional)
+
+- The upstream eviction API does not currently support evictions in background
+  implemented by external components. For that, there's no community
+  dedicated error code nor a response text for this type of eviction.
+- The eviction in background error code will be temporary mapped to 429 (TooManyRequests).
+- The response text will be searched for strings
+  containing "Eviction triggered" prefix or similar.
+- The descheduler will still treat all the pods with eviction in background as cattle.
+  It will only prioritize these pods to be processed sooner.
+  There may still be unnecessary/sub-optimal evictions.
+
+### Risks and Mitigations
+
+- If an external component responsible for eviction in background
+  does not work properly the descheduler might evict less pods than expected.
+  Leaving pods that require eviction (e.g. long running or violating third
+  party policies) running longer than expected.
+  The descheduler will need to implement a TTL mechanism to abandon
+  eviction requests that take longer than expected.
+- Most of the heavy lifting happens in the external components.
+  The descheduler only acknowledges pods that are evicted in background.
+  All the internal counters take these pods into account.
+  When these pods are not properly evicted the internal counters might limit
+  eviction of other pods that (when evicted) might improve overall cluster
+  resource utilization.
+- This new functionality needs to stay feature gated until
+  the evacuation API is available and stable.
+
+## Design Details
+
+### Implementation Strategies
+
+#### API changes
+
+- **New category of pods for eviction**: Two new annotations are introduced:
+  - `descheduler.alpha.kubernetes.io/request-evict-only`: to signal pods whose
+    eviction is not expected to be completed right away.
+    Instead, an eviction request is expected to be intercepted by an external
+    component which will initiate the eviction process for the pod.
+    E.g. by performing live migration or scheduling the eviction to a more suitable time.
+  - `descheduler.alpha.kubernetes.io/eviction-in-progress`: to mark pods whose
+    eviction was initiated by an external component.
+- **New eviction limits**: introduction of new `MaxNoOfPodsEvacuationsPerNode` and
+  `MaxNoOfPodsEvacuationsPerNamespace` feature gated v1alpha1 fields.
+- **New configurable list of eviction message prefixes (introduced later when needed)**:
+  each third party validating admission webhook can provide different response message.
+  It is not feasible to ask every party to change their messages for a temporary solution
+  until the evacuation API is in place. Thus, temporary introducing a new feature
+  gated `evictionMessagePrefixes` field (`[]string`) for a list of known prefixes
+  signaling an eviction in background.
+  Normally, "Eviction triggered" prefix will be acknowledged by default.
+
+#### v1alpha1
+
+- Keep track of evictions in progress: A new cache will be introduced
+  to keep track of pods annotated with `descheduler.alpha.kubernetes.io/request-evict-only`
+  that were requested to be evicted through the upstream eviction API.
+  In case a descheduler gets unexpectedly restarted any pod annotated
+  with `descheduler.alpha.kubernetes.io/eviction-in-progress` will be added
+  to the cache if not already present.
+  The cache will help with sorting and avoiding a duplicated eviction
+  request of pods that were not annotated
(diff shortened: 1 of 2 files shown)


songtao98:
Hi, @ingvagabund . I have a question about how does this KEP works in Descheduling Framework. 
1. Do you expect to use this new eviction handling mechanism to replace existing `PodEvictor` （who call Eviciton API directly）, so that all users evict their pods through it? 
2. Do you consider to take one more step to make it as a extension point, e.g. `Evict`, so that users can decide which way to evict their pods, and existing `PodEvictor` or your extension by this KEP can both be implemented as a `EvictPlugin`(different with existing `EvictorPlugin`)?

ingvagabund:
@songtao98 thank you for asking the right questions.

> 1. Do you expect to use this new eviction handling mechanism to replace existing PodEvictor （who call Eviciton API directly）, so that all users evict their pods through it?

Eventually, that's the ultimate goal. However, the first implementations will provide both eviction API and evacuation API. So we have some soft time for early adopters to share their experience and corner case for which evacuation API does not work/scale well. Followed by discussions with evacuation API designers about possible resolutions before the new API goes GA.

> 2. Do you consider to take one more step to make it as a extension point, e.g. Evict, so that users can decide which way to evict their pods, and existing PodEvictor or your extension by this KEP can both be implemented as a EvictPlugin(different with existing EvictorPlugin)?

Not right now. Ultimately, the new evacuation API should be a special case of the eviction API. The details are yet to be discussed.

songtao98:
> thank you for asking the right questions.

@ingvagabund Thank you for your reply! For question 1, the strategy sounds great. For question 2, I'm not sure if someone has ever proposed a custom Evictor mechanism(to customize the real evictor like `PodEvictor`, not currently `EvictorPlugin`), but we think this capability will eventually be meaningful and already implemented one. We are also looking forward to more discussion about details, and if the community think it's the right way, we will be more than happy to contribute : )

ingvagabund:
> We think this capability will eventually be meaningful and already implemented one. We are also looking forward to more discussion about details, and if the community think it's the right way, we will be more than happy to contribute : )

Few sig calls back we discussed the possibility of writing mini-keps. Turning the low level eviction mechanism into a plugin is a good example for it :) If there are good use cases where a user can benefit from a custom eviction mechanism. Would you be willing to write a proposal for this? The current proposal focuses mainly on integration with the new evacuation API (and annotation as an interim solution). Pluginazing the evictor itself is related, yet it deserves its own attention.

songtao98:
> Few sig calls back we discussed the possibility of writing mini-keps. Turning the low level eviction mechanism into a plugin is a good example for it :) If there are good use cases where a user can benefit from a custom eviction mechanism. Would you be willing to write a proposal for this? The current proposal focuses mainly on integration with the new evacuation API (and annotation as an interim solution). Pluginazing the evictor itself is related, yet it deserves its own attention.

For sure! We are certainly willing to draft a proposal for customizing the eviction mechanism. However, it might take some time as I need to organize some details and perform the necessary abstractions. BTW, regarding the mini-keps you mentioned, are there any specific format requirements? If not, I will follow the current KEP template.

keps/1397-evictions-in-background/README.md
```diff
@@ -0,0 +1,895 @@
...
+|**Pod nomination and sorting**|Preferring pods with both `descheduler.alpha.kubernetes.io/request-evict-only` and `descheduler.alpha.kubernetes.io/eviction-in-progress` annotations.|Preferring pods that have a corresponding evacuation CR present.|
+|**Cache reconstruction**|When a descheduler gets restarted and a new internal caches (with tracked annotations) cleared the descheduler lists all pods and populate the cache with any pod that has both annotation (`descheduler.alpha.kubernetes.io/request-evict-only` and `descheduler.alpha.kubernetes.io/eviction-in-progress`) present. In case only the first annotation is present but the eviction request was already created, the update event handler will catch the second annotation addition and the cache gets synced. In the worst case the limit of max number of pods evicted gets exceeded.|The descheduler waits until all evacution CRs are synced.|
+
+Given the evucation API is still a work in progress without any existing implementation, the annotation based eviction can be seen as an v1alpha1 implemenation of this proposal. Integration with the evacuation API as an v1alphaN or beta implementation.
```
songtao98, line 217: > Given the evucation API is still a work in progress without any existing implementation,

Typo here "evucation"

ingvagabund:
> BTW, regarding the mini-keps you mentioned, are there any specific format requirements? If not, I will follow the current KEP template.

We don't have any dedicated template for the mini-keps atm. I have utilized the upstream k8s template (which might put the bar too high). Nevertheless, I find it quite useful. Even filling in the parts that seem to be too far away from the descheduler project. Yet, they help me to ask myself questions that are important to answer. Let's use the upstream template for now until we find a better one.

gcezaralmeida:
Hey, this is something that I'm interested in also. The idea of Descheduler evicting Kubevirt VM pods to Live Migrate then will help a lot the community.

ingvagabund:
Hey @gcezaralmeida, I'd be interested in hearing more about your use cases. Happy to extend the proposal with other use cases if you have any in mind.

keps/1397-evictions-in-background/README.md
```diff
@@ -0,0 +1,1122 @@
...
+3. Eviction nomination
+   1. At the beginning of each descheduling cycle the descheduler increases
+      all the internal counters to take into account all pods that are subjects
+      to background eviction. Either by checking `descheduler.alpha.kubernetes.io/eviction-in-progress` annotation,
+      list of evacuation requests or confronting the internal caches.
+   2. Each descheduling plugin nominates a set of pods to be evicted.
+   3. All the nominated pods are sorted based on the eviction-in-progress first priority.
+4. Pod eviction: the descheduler starts evicting nominated pods until a limit is reached
+   1. If an annotation based eviction is enabled:
+      1. If a pod is not annotated with `descheduler.alpha.kubernetes.io/request-evict-only` evict
```
atiratree, line 311: FYI, this annotation will have to be changed when moving to the Evacuation API by the workloads owners 

ref: https://github.com/kubernetes/enhancements/blob/2d7dbab6c95575d68395c30fa789f87d9a2c3c01/keps/sig-apps/4563-evacuation-api/README.md#evacuator

basically to the following annotation:

```
evacuation.coordination.k8s.io/priority_${EVACUATOR_CLASS}: ${PRIORITY}/${ROLE}
```

The contract will be different though, so I suppose it doesn't make sense to introduce it before the Evacuation API is merged.

But it would be good to mention it here so the users are aware of it. Maybe in the `v1alphaN or beta` section?

keps/1397-evictions-in-background/README.md
```diff
@@ -0,0 +1,1122 @@
...
+         2. Error code, resp. response text is checked to distinguish between
+            a genuine error and a confirmation of an eviction in background.
+   2. If the evacuation API is utilized create an evacution request
+
+### Notes/Constraints/Caveats (Optional)
+
+- The upstream eviction API does not currently support evictions in background
+  implemented by external components. For that, there's no community
+  dedicated error code nor a response text for this type of eviction.
+- The eviction in background error code will be temporary mapped to 429 (TooManyRequests).
```
atiratree, line 324: is it a good choice to use the error code that is already used by the eviction API?

keps/1397-evictions-in-background/README.md
```diff
@@ -0,0 +1,1122 @@
...
+  descheduling loop all internal counters will take into account pods
+  that are either annotated with `descheduler.alpha.kubernetes.io/eviction-in-progress`
+  or are present in the cache of to-be-evicted pods.
+
+#### v1alphaN or beta
+- Replace the current eviction API with the new
+  [evacuation API](https://github.com/kubernetes/enhancements/pull/4565) by default.
+  Both `descheduler.alpha.kubernetes.io/request-evict-only` and
+  `descheduler.alpha.kubernetes.io/eviction-in-progress` will be deprecated.
+  Existing deployments will need to remove `descheduler.alpha.kubernetes.io/request-evict-only`
```
atiratree, line 406: does that mean that delayed eviction/evacuation will not work for the workloads that forget to remove the annotation?

keps/1397-evictions-in-background/README.md
```diff
@@ -0,0 +1,1122 @@
...
+- The evacuation API expects the evacuation request to be reconciled during each descheduling cycle.
+  Given each cycle starts from scratch the descheduler needs to first list
+  all evacuation requests and sort the pods accordingly in each plugin.
+  Other components may create an evacuation request as well.
+  The descheduler needs to take this into account.
+  E.g. by creating a snapshot (per cycle, per strategy) to avoid inconsistencies.
+  An evacuation request may exist without a corresponding pod
+  before the evacuation controller garbage collects the request.
+- The descheduler needs to keep a cache of evacution API requests (CRs)
+  and remove itself from finalizers if the request is no longer needed.
```
atiratree, line 418: We are thinking of removing the cancellation feature. If we remove it, then no more work is expected of the descheduler.

Does descheduler have a use case for canceling evictions/evacuations?

keps/1397-evictions-in-background/README.md
```diff
@@ -0,0 +1,1122 @@
...
+   1. At the beginning of each descheduling cycle the descheduler increases
+      all the internal counters to take into account all pods that are subjects
+      to background eviction. Either by checking `descheduler.alpha.kubernetes.io/eviction-in-progress` annotation,
+      list of evacuation requests or confronting the internal caches.
+   2. Each descheduling plugin nominates a set of pods to be evicted.
+   3. All the nominated pods are sorted based on the eviction-in-progress first priority.
+4. Pod eviction: the descheduler starts evicting nominated pods until a limit is reached
+   1. If an annotation based eviction is enabled:
+      1. If a pod is not annotated with `descheduler.alpha.kubernetes.io/request-evict-only` evict
+         the pod normally without additional handling.
```
atiratree, line 312: Can you please write the workflow twice? One for the `v1alpha1` and second for `v1alphaN or beta`?

atiratree: For example in the Evacuation case, Evacuation object can be created for any pod. Regardless if it supports evacuation or not. The evacuation controller and optional evacuator will ensure that the pod gets deleted eventually.

atiratree: Does it make sense for the descheduler to distinguish/prioritize pods that support evacuation?

keps/1397-evictions-in-background/README.md
```diff
@@ -0,0 +1,1122 @@
...
+    eviction was initiated by an external component.
+- **New eviction limits**: introduction of new `MaxNoOfPodsEvacuationsPerNode` and
+  `MaxNoOfPodsEvacuationsPerNamespace` feature gated v1alpha1 fields.
+- **New configurable list of eviction message prefixes (introduced later when needed)**:
+  each third party validating admission webhook can provide different response message.
+  It is not feasible to ask every party to change their messages for a temporary solution
+  until the evacuation API is in place. Thus, temporary introducing a new feature
+  gated `evictionMessagePrefixes` field (`[]string`) for a list of known prefixes
+  signaling an eviction in background.
+  Normally, "Eviction triggered" prefix will be acknowledged by default.
```
atiratree, line 370: how is this going to change with the Evacuation API?

keps/1397-evictions-in-background/README.md
```diff
@@ -0,0 +1,1122 @@
...
+```
+
+to provide the [descheduling framework](https://github.com/kubernetes-sigs/descheduler/pull/1372)
+a generic way of evicting/evacuating pods.
+The new interface can be used for implementing various testing scenarios.
+Or turned into a plugin if needed.
+
+#### Metrics
+
+A new metric `pod_evacuations` for counting the number of evacuated pods is introduced.
```
atiratree, line 441: how are the metrics going to be changed in beta?

keps/1397-evictions-in-background/README.md
```diff
@@ -0,0 +1,1122 @@
...
+- The eviction in background error code will be temporary mapped to 429 (TooManyRequests).
+- The response text will be searched for strings
+  containing "Eviction triggered" prefix or similar.
+- The descheduler will still treat all the pods with eviction in background as cattle.
+  It will only prioritize these pods to be processed sooner.
+  There may still be unnecessary/sub-optimal evictions.
+
+### Risks and Mitigations
+
+- If an external component responsible for eviction in background
```
atiratree, line 333: does this mean that we will retry the eviction request? E.g. for the case that the eviction information was lost by the evacuator? How is this and the TTL going to look in beta?

keps/1397-evictions-in-background/README.md
```diff
@@ -0,0 +1,1122 @@
...
+Only provide a summary about how many pods were requested to be evacuated.
+If needed, another metric `pod_evacuations_in_progress` for counting
+the number of evacutions that are reconciled can be introduced.
+
+### Open Questions [optional]
+
+This is where to call out areas of the design that require closure before deciding
+to implement the design.
+
+* The evacution API proposal mentions the evacuation instigator should remove
```
atiratree, line 452: maybe cancellation will not be needed? https://github.com/kubernetes-sigs/descheduler/pull/1354#discussion_r1611588675

keps/1397-evictions-in-background/README.md
```diff
@@ -0,0 +1,1122 @@
...
+  its intent when the evacuation is no longer necessary.
+  Should each descheduling cycle reset all the eviction requests
+  or wait until some of the eviction requests were completed
+  to free a "bucket" for new evictions?
+  **Answer**: The first implementation will not reset any evacuation request.
+  The descheduler will account for existing requests and update the internal counters accordingally.
+  In the future a mechanism for deleting too old evacuation requests can be introduced.
+  I.e. based on a new `--max-evacuation-request-ttl` option.
+* The evacution API proposal mentions more than a one entity can request an eviction.
```
atiratree, line 461: does the descheduler have a use for a custom value of `.spec.progressDeadlineSeconds`? We are thinking of using a unified value (`1800 (30m)`) in all of the components. 

Might be good to clarify deschedulers position on this.

As it currently stands, the entity who creates the Evacuation has a control of this value. The following instigators who do not create the object, do not have the controll.

keps/1397-evictions-in-background/README.md
```diff
@@ -0,0 +1,1122 @@
...
+  Should the descheduler take these into account as well?
+  What if another entity decides to evict a pod that is of a low priority
+  from the descheduler's point of view?
+  **Answer**: The first implementation will consider all pods with an existing
+  evacuation request as "already getting evacuated". Independent of the pod priorities.
+* With evacution requests a plugin does not evict a pod right away.
```
atiratree, line 467: Is it possible to move to just using the Evacuations? Would it be simpler to reason about by the plugins?

keps/1397-evictions-in-background/README.md
ingvagabund: There's not many codes recognized by the eviction API. Plus, the expectation here is to use combination of both error code and response text. The response text makes the difference.

keps/1397-evictions-in-background/README.md
ingvagabund: The current `v1alphaN or beta` mentions:
```
Both `descheduler.alpha.kubernetes.io/request-evict-only` and
  `descheduler.alpha.kubernetes.io/eviction-in-progress` will be deprecated.
  Existing deployments will need to remove `descheduler.alpha.kubernetes.io/request-evict-only`
  annotation to utilize the new evacuation API.
  Otherwise, the eviction API will be used as a fallback.
```
What about to extend it with something like:
```
Workload owners utilizing the new evacuation API are expected to replace both annotations
with [evacuation.coordination.k8s.io/priority_${EVACUATOR_CLASS}: ${PRIORITY}/${ROLE}](https://github.com/kubernetes/enhancements/blob/2d7dbab6c95575d68395c30fa789f87d9a2c3c01/keps/sig-apps/4563-evacuation-api/README.md#evacuator")
```
?

keps/1397-evictions-in-background/README.md
ingvagabund: > Does it make sense for the descheduler to distinguish/prioritize pods that support evacuation?

Yes. It's preferable to evacuate pods which have higher chance of getting evicted at the end. Would presence of `evacuation.coordination.k8s.io/priority_${EVACUATOR_CLASS}: ${PRIORITY}/${ROLE}` signal such a support?

keps/1397-evictions-in-background/README.md
ingvagabund: If the descheduler is the only evacuation requester, the evacuation request gets deleted after TTL. So other pods have a chance to be evacuated. Depending on how pods are proceed in the next descheduling cycle the same evacuation request can be re-created.

keps/1397-evictions-in-background/README.md
ingvagabund: Deprecated and removed eventually.

keps/1397-evictions-in-background/README.md
ingvagabund: Yes, this can happen until the deprecated annotations get removed (2 releases minimum after the deprecation is announced). This could be a blocking reason for a broader adoption.

keps/1397-evictions-in-background/README.md
ingvagabund: Only when an evacuation takes too long. Nevertheless, that should not be the case as long as the evacuation is properly reconciled. If the case happens a cluster administrator needs to investigate right away why too many evacuations are pending or not progressing fast enough. As a temporary solution the administrator can increase the descheduler limits to mitigate the situation. Also, TTL functionality will be optional and configurable.

keps/1397-evictions-in-background/README.md
ingvagabund: The way the numbers are figured will stay unchanged. Instead of the number of pods "evicted in background" the metrics will count the number of evacuation requests created.

keps/1397-evictions-in-background/README.md
ingvagabund: Answered in the comment.

keps/1397-evictions-in-background/README.md
ingvagabund: > We are thinking of using a unified value (1800 (30m)) in all of the components.

We can make it the default and configurable if needed. Can an evacuation request be created without `.spec.progressDeadlineSeconds`? I.e. can we expected the default value will get set to a reasonable value/30m?

keps/1397-evictions-in-background/README.md
ingvagabund: Can you more elaborate on what you mean? In my mind evacuations = evacuation requests.

keps/1397-evictions-in-background/README.md
atiratree: Hm, basically yes, but it would be good to highlight that the annotation's meaning will be different.

keps/1397-evictions-in-background/README.md
atiratree: If we were implemented this from ground up, having a different code would help to better distinguish the failures. But okay, I suppose it makes sense to use 429. An Eviction API sends this [when disruptions are not allowed](https://github.com/kubernetes/kubernetes/blob/master/pkg/registry/core/pod/storage/eviction.go#L430). And also returns `RetryAfterSeconds`. Which is almost this case. And clients might handle it well.

Do we have any recommendations and use case for the `RetryAfterSeconds`?

keps/1397-evictions-in-background/README.md
atiratree: That is a good idea, I will add a default of 30m to the API.

keps/1397-evictions-in-background/README.md
atiratree: I forgot what I meant originally.

Can we pass the Evacuation objects through the plugins then and have them decide on their state?

keps/1397-evictions-in-background/README.md
ingvagabund: The current descheduler implementation relies on `--descheduling-interval`. There's no retry logic implemented. The next eviction/evacuation attempt is expected to be performed in the next descheduling cycle.

keps/1397-evictions-in-background/README.md
ingvagabund: Sounds good. This can be implemented as an improvement. Though, every pod should have the same chance of getting evacuated/evicted.

keps/1397-evictions-in-background/README.md
ingvagabund: Good point. We already have metrics that count the number of evicted pods. We can either deprecate the metrics or replace them. Nevertheless, it will take a while before the evacuation API is implemented and the new metrics produced. Meantime, we need to publish it. I am fine deprecating the new metrics we are about to created once they have equivalent representatives produced by the evacuation stack.

keps/1397-evictions-in-background/README.md
ingvagabund: The eviction/evacuation is exposed/expected to be exposed through a generic interface. All the knowledge about how eviction/evacuation processes are implemented is currently expected to be exposed through filters/sorters. So no plugin implementation depends on a particular eviction/evacuation implementation. Some community adopters would like to turn the eviction/evacuation handle into a plugin as well.

keps/1397-evictions-in-background/README.md
ingvagabund: Updated. Mentioning the semantic is different.

keps/1397-evictions-in-background/README.md
ingvagabund: Done

keps/1397-evictions-in-background/README.md
ingvagabund: Mentioned in the proposal

knelasevero:
Do we consider the alpha implementation here to be an workaround until the until the evacuation API is fully integrated (in the future, in beta)?

I understand trying to deliver value earlier and then already start to get feedback, but have you checked with Feature Requesters and stakeholders if that is really what they want? Maybe it makes sense to start with the evacuation API integration from the start.

knelasevero:
In Filip's proposal (I need to read it in more detail, but I skimmed it already), there is this notion for Evacuator Prioritization and multiple evacuators (and a lot of other details not considered here). To me it sounds like when going from alpha implementation (here) to beta, and trying to assimilate the other complexities from the real evacuator API too much would change to call the graduation here a graduation from alpha to beta? Maybe it would be another alpha for a while before becoming beta? (or, again, as above comment, it would be worth to consider all those things from the start...)

keps/1397-evictions-in-background/README.md
```diff
@@ -0,0 +1,1147 @@
...
+- Workloads do not set `descheduler.alpha.kubernetes.io/request-evict-only` annotation
+  when deprecated
+
+#### Beta
+
+- Gather feedback from developers and surveys
+- Evacuation API based evictions
+- Identifying corner cases that are not addressed by the evacuation API
+- `descheduler.alpha.kubernetes.io/request-evict-only` annotation is deprecated
+- Workloads do not set `descheduler.alpha.kubernetes.io/request-evict-only` annotation
```
knelasevero, line 705: @atiratree then also implementation around prioritization, detailed status reporting, and cancellation policies as per the Evacuation API? Then, not a beta graduation, right?

keps/1397-evictions-in-background/README.md
```diff
@@ -0,0 +1,1147 @@
...
+   3. All the nominated pods are sorted based on the eviction-in-progress first priority.
+4. Pod eviction: the descheduler starts evicting nominated pods until a limit is reached
+   1. If a pod is not annotated with `descheduler.alpha.kubernetes.io/request-evict-only` evict
+      the pod normally without additional handling.
+   2. If a pod is annotated with `descheduler.alpha.kubernetes.io/request-evict-only` evict the pod.
+      1. If the eviction fails intepret the error as a request for eviction in background.
+      2. Error code, resp. response text is checked to distinguish between
+         a genuine error and a confirmation of an eviction in background.
+
+##### Evacuation API (v1alphaN or beta)
```
knelasevero, line 319: Oh, i see v1alphaN is considered, I missed this

knelasevero: Maybe more than considered it should be a must given the expected changes

knelasevero:
This looks great and is very detailed. I need to read both proposals a bit more deeply to give more insights, but wanted to give my initial thoughts so the KEP can go forward

keps/1397-evictions-in-background/README.md
```diff
@@ -0,0 +1,1147 @@
...
+- Address feedback on usage/changed behavior, provided on GitHub issues
+- Deprecate the flag
+-->
+
+#### Alpha
+
+- Feature implemented behind a feature flag
+- Initial e2e tests completed and enabled
+- Annotation based evictions
+- Workloads do not set `descheduler.alpha.kubernetes.io/request-evict-only` annotation
```
atiratree, line 696: a bit confusing to see the deprecation in alpha

keps/1397-evictions-in-background/README.md
```diff
@@ -0,0 +1,1147 @@
...
+
+#### Code changes
+
+The code responsible for invoking the eviction API is located
+under `sigs.k8s.io/descheduler/pkg/descheduler/evictions` under private `evictPod` function.
+The function can be generalized into a generic interface:
+
+```go
+type PodEvacuator interface {
+    EvacutePod(ctx context.Context, pod *v1.Pod) error
```
atiratree, line 451: what guarantee we have each plugin will call this interface? Isn't it better to override/change the Evictor interface?

keps/1397-evictions-in-background/README.md
```diff
@@ -0,0 +1,1147 @@
...
+  **Answer**: The first implementation will not reset any evacuation request.
+  The descheduler will account for existing requests and update the internal counters accordingally.
+  In the future a mechanism for deleting too old evacuation requests can be introduced.
+  I.e. based on a new `--max-evacuation-request-ttl` option.
+* The evacuation API proposal mentions more than a one entity can request an eviction.
+  Should the descheduler take these into account as well?
+  What if another entity decides to evict a pod that is of a low priority
+  from the descheduler's point of view?
+  **Answer**: The first implementation will consider all pods with an existing
+  evacuation request as "already getting evacuated". Independent of the pod priorities.
```
atiratree, line 487: The important part to decide for the descheduler is, if it wants to subscribe to evacuating these lower evacuations as well. This is done by setting the finalizer and prevents the evacuation cancellation.

keps/1397-evictions-in-background/README.md
```diff
@@ -0,0 +1,1147 @@
...
+  3. Observe the pod is not removed right away and wait for the configured
+     delay T before checking a pod is removed.
+  4. Check a pod was removed after T, not before.
+- `EvacuationLimitReached` for testing only a limited amount of pods is evacuated
+  with already existing evacuation requests:
+  1. Create evacuation requests (e.g. by annotating pods
+     with `descheduler.alpha.kubernetes.io/eviction-in-progress` or creating evacuation requests).
+  2. Run the descheduler and observe only an expected amount of pods
+     is requested to be evicted/evacuated.
+- `EvacuationTTL` for testing an evacuation request (with a single owner
```
atiratree, line 616: I am a bit confused, do we introduce the TTL for the first version or not?

atiratree:
> Do we consider the alpha implementation here to be an workaround until the until the evacuation API is fully integrated (in the future, in beta)?
>
> I understand trying to deliver value earlier and then already start to get feedback, but have you checked with Feature Requesters and stakeholders if that is really what they want? Maybe it makes sense to start with the evacuation API integration from the start.

FIY, we are still discussing the Evacuation API feature and it was deferred from 1.31

ingvagabund:
> In Filip's proposal (I need to read it in more detail, but I skimmed it already), there is this notion for Evacuator Prioritization and multiple evacuators (and a lot of other details not considered here). To me it sounds like when going from alpha implementation (here) to beta, and trying to assimilate the other complexities from the real evacuator API too much would change to call the graduation here a graduation from alpha to beta? Maybe it would be another alpha for a while before becoming beta? (or, again, as above comment, it would be worth to consider all those things from the start...)

Currently, the promotion to beta requires implementation of the evacuation API. Thus, we need to stay aligned with it and can not promote to beta until the evacuation API is beta.

keps/1397-evictions-in-background/README.md
ingvagabund: The implementation will not get promoted to beta without the evacuation API integration. We will be rolling at least two v1alpha versions.

knelasevero:
> > Do we consider the alpha implementation here to be an workaround until the until the evacuation API is fully integrated (in the future, in beta)?
> 
> Yes
> 
> > I understand trying to deliver value earlier and then already start to get feedback, but have you checked with Feature Requesters and stakeholders if that is really what they want? Maybe it makes sense to start with the evacuation API integration from the start.
> 
> I did. This is an important feature for KubeVirt. There's a strong interest to have this feature merged sooner than later. Waiting until the Evacuation API is implemented/alpha would be the best course of actions. Yet, it's not a blocker for implementing the first alpha of this feature.

I know it is not a blocker. I see it more from the frame of: 

Is this delivering enough value when implemented as a workaround to justify not waiting for the actual thing that you want to use? Is the feedback and initial users you will get now be worth implementing this, even if the thing will be quite different when the Evacuation API is here (in beta or when it is safe to use it or w/e)?

keps/1397-evictions-in-background/README.md
ingvagabund: > `Workloads do not set descheduler.alpha.kubernetes.io/request-evict-only annotation` is not actionable. Maybe s/do/should/ and document this?

What about `Workloads are not expected to set descheduler.alpha.kubernetes.io/request-evict-only annotation`?

keps/1397-evictions-in-background/README.md
ingvagabund: Change to "v1alphaN and beta"

keps/1397-evictions-in-background/README.md
```diff
@@ -0,0 +1,1147 @@
...
+2. A user deploys a workload through various nodes
+   1. A scheduler distributes the pods based on configured scheduling policies
+3. Eviction nomination
+   1. At the beginning of each descheduling cycle the descheduler increases
+      all the internal counters to take into account all pods that are subjects
+      to background eviction. By checking `descheduler.alpha.kubernetes.io/eviction-in-progress`
+      annotation or confronting the internal caches.
+   2. Each descheduling plugin nominates a set of pods to be evicted.
+   3. All the nominated pods are sorted based on the eviction-in-progress first priority.
+4. Pod eviction: the descheduler starts evicting nominated pods until a limit is reached
```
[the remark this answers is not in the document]
ingvagabund: > Is the eviction going to be handled in each plugin separately as it is currently?

Yes. The current mechanism will be persisted.

> What are the guarantees that a plugin will use the Evacuator/Evictor correctly? E.g. calling pre-checks?

None. Each plugin is given a handler for evicting/evacuating a pod. No pod will be aware whether an eviction or evacuation API is invoked under the hood.

By pre-checks you mean excluding pods with already existing evacuation requests?

keps/1397-evictions-in-background/README.md
```diff
@@ -0,0 +1,1147 @@
...
+      the pod normally without additional handling.
+   2. If a pod is annotated with `descheduler.alpha.kubernetes.io/request-evict-only` evict the pod.
+      1. If the eviction fails intepret the error as a request for eviction in background.
+      2. Error code, resp. response text is checked to distinguish between
+         a genuine error and a confirmation of an eviction in background.
+
+##### Evacuation API (v1alphaN or beta)
+
+1. Policy configuration
+   1. The cluster administrator configures the descheduler to enable the new functionality.
```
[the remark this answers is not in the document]
ingvagabund: We run a check for whether an eviction API is supported in https://github.com/kubernetes-sigs/descheduler/blob/c7f8670b11e22b8b7be539e755e57cc6a26748bc/pkg/descheduler/descheduler.go#L250. I expect a similar check for the evacuation API.

keps/1397-evictions-in-background/README.md
```diff
@@ -0,0 +1,1147 @@
...
+  eviction requests that take longer than expected.
+- Most of the heavy lifting happens in the external components.
+  The descheduler only acknowledges pods that are evicted in background.
+  All the internal counters take these pods into account.
+  When these pods are not properly evicted the internal counters might limit
+  eviction of other pods that (when evicted) might improve overall cluster
+  resource utilization.
+- This new functionality needs to stay feature gated until
+  the evacuation API is available and stable.
+- When `descheduler.alpha.kubernetes.io/request-evict-only` annotation is deprecated
```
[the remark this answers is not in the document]
ingvagabund: > So do we skip these workloads? 

These workloads will still be evicted through the eviction API as implemented in v1alpha1. Users will need to explicitly remove the annotation to utilize the evacuation API until the annotation is removed from the descheduler.

> When do we remove this support?

The deprecation will follow the standard timelines. I.e. at least one year from the deprecation.

keps/1397-evictions-in-background/README.md
ingvagabund: That depends on how many alphas we will have. The annotation gets deprecated the moment the evacuation API is implemented as alpha. While we will need to wait until the evacuation API gets promoted to beta before we can promote the eviction-in-background to beta.

keps/1397-evictions-in-background/README.md
ingvagabund: The Evictor interface will get replaced by `PodEvacuator` interface under the hood. Currently, all plugins are expected to invoke `d.handle.Evictor().Evict` method. The method will invoke the evacuation API once available and get renamed to a more suitable name.

keps/1397-evictions-in-background/README.md
ingvagabund: Good point. Yes. We want to set the finalizer even for this pods.

keps/1397-evictions-in-background/README.md
ingvagabund: I see I have mentioned the TTL will not be implemented in the first version under Open Questions. Yet, v1alpha1 section mentions it will. I removed that part from v1alpha1.

v1alpha1 will not introduce the TTL mechanism. Yet, it might get implemented in later v1alpha versions.

ingvagabund:
> Is this delivering enough value when implemented as a workaround to justify not waiting for the actual thing that you want to use? Is the feedback and initial users you will get now be worth implementing this, even if the thing will be quite different when the Evacuation API is here (in beta or when it is safe to use it or w/e)?

Disclaimer: this is my personal comment and it does not have to reflect the current reality.

We don't have any participating initial users besides the discussions I have had with some KubeVirt representatives. So currently there's only one user. The feedback is currently limited and influenced by the evacuation API proposal. I don't have any solid data to support anything I have proposed in the current proposal. The design was crafted in a best-effort manner in my best conscience. Taking into account many what-if scenarios. We still don't have many participating contributors in the descheduler community to extend the discussions/reviews and hand craft a better design. Practically, any contribution from my side is done either in my free time of when I prioritize my work in my current company.

I am aligned with your perspective and acknowledge you asked the right and important questions. I see two options. To wait (hard to say for how long) and collect more feedback with an assumption the design will improve. Or, pin the current design, produce v1alpha1 implementation, collect the feedback and see how well/bad the design proved the be.

knelasevero:
> We still don't have many participating contributors in the descheduler community to extend the discussions/reviews and hand craft a better design

You focus on the design reviews not being comprehensive for lack of contributors, but that is actually an argument to wait, no? Since if you consider the repo to have less support than it needs it would be harder to maintain a workaround than it would be to maintain what you envision to be the final implementation (that would also start at alpha, but would not have big changes of core internals).

> To wait (hard to say for how long) and collect more feedback with an assumption the design will improve. Or, pin the current design, produce v1alpha1 implementation, collect the feedback and see how well/bad the design proves the be.

I also see these two options, and I see pros and cons coming from both of them. I think the workaround design is great, I wouldn't consider it to be bad. I think it was detailed and a lot has been considered. Just raised this discussion to be the devil's advocate and check other paths. 

> with an assumption the design will improve

But you also assume that, or the goal of migrating to the intended final implementation using the actual evacuation API (and ditching the previous one) would not be written down here in you EP.

fossedihelm:
Reading the great discussion here! To me, the proposal looks good, and thank you for this effort!

> `This is an important feature for KubeVirt` -> but how urgent, I think that's the question. Also, do they know of the plan to migrate later to the evacuation API approach? Would they explicitly prefer not to wait, or did they just say they need it asap?

Answering this question: Yes, this is an essential feature for KubeVirt.
Let's start from the beginning: live migration is a key feature of KubeVirt; it allows the workload not to be compromised. For this reason, it is chosen as the default behavior during an eviction, drain, etc...
KubeVirt is also shipped by default with Openshift, and currently, Descheduler and KubeVirt behavior are not compatible with each other. Without this workaround, a cluster admin would have to choose between having an active KubeVirt core feature and the Descheduler, and this would put them in a difficult situation that would be best avoided.
As expressed above this is a workaround that will be overridden with the evacuation API and KubeVirt will move to the evacuation API accordingly.
IOW, yes, KubeVirt side it would definitely be better to have this ASAP.
Thank you!

knelasevero:
> Reading the great discussion here! To me, the proposal looks good, and thank you for this effort!
> 
> > `This is an important feature for KubeVirt` -> but how urgent, I think that's the question. Also, do they know of the plan to migrate later to the evacuation API approach? Would they explicitly prefer not to wait, or did they just say they need it asap?
> 
> Answering this question: Yes, this is an essential feature for KubeVirt. Let's start from the beginning: live migration is a key feature of KubeVirt; it allows the workload not to be compromised. For this reason, it is chosen as the default behavior during an eviction, drain, etc... KubeVirt is also shipped by default with Openshift, and currently, Descheduler and KubeVirt behavior are not compatible with each other. Without this workaround, a cluster admin would have to choose between having an active KubeVirt core feature and the Descheduler, and this would put them in a difficult situation that would be best avoided. As expressed above this is a workaround that will be overridden with the evacuation API and KubeVirt will move to the evacuation API accordingly. IOW, yes, KubeVirt side it would definitely be better to have this ASAP. Thank you!

Thanks! Having your direct input here is important. Thanks for chiming in!

ingvagabund:
I am currently refactoring the current code base so at the end the changes required to implement v1alpha1 of this KEP are minimal. The current PoC is under https://github.com/ingvagabund/descheduler/tree/eviction-in-background-code-2024-07-02.

keps/1397-evictions-in-background/README.md
```diff
@@ -0,0 +1,1147 @@
...
+  by default when running a descheduler instance.
+- As a descheduler plugin developer I want to be aware of evictions
+  in progress to improve pod nomination and thus avoid unnecessary disruptions.
+- As a security professional I want to make sure any live migration
+  adhers to security policies and protects sensitive data.
+
+### Annotation vs. evacuation API based eviction
+
+The [evacuation API](https://github.com/kubernetes/enhancements/pull/4565) is expected
+to replace the [eviction API](https://kubernetes.io/docs/concepts/scheduling-eviction/api-eviction/).
```
damemi, line 264: is there any timeline on deprecation for the eviction API?

keps/1397-evictions-in-background/README.md
```diff
@@ -0,0 +1,1147 @@
...
+
+* The evacuation API proposal mentions the evacuation instigator should remove
+  its intent when the evacuation is no longer necessary.
+  Should each descheduling cycle reset all the eviction requests
+  or wait until some of the eviction requests were completed
+  to free a "bucket" for new evictions?
+  **Answer**: The first implementation will not reset any evacuation request.
+  The descheduler will account for existing requests and update the internal counters accordingally.
+  In the future a mechanism for deleting too old evacuation requests can be introduced.
+  I.e. based on a new `--max-evacuation-request-ttl` option.
```
damemi, line 481: should this be a policy config field?

damemi (commented):
looks good to me, and the evacuation API does seem like a more graceful evolution of the eviction API.

One question though, we have had requests for more immediate pod deletion (like https://github.com/kubernetes-sigs/descheduler/issues/1253), and this seems to go a step further away from that. Should we have any continued support for direct eviction requests? Or is eviction going away entirely?

ingvagabund:
Thank you Mike for going through the proposal.

> Should we have any continued support for direct eviction requests? Or is eviction going away entirely?

@atiratree will know better though based on my understand the eviction API will not go away. Instead, users will still have option to choose between the evacuation (delayed eviction) and the eviction APIs. Yet, with strong recommendation to prefer the evacuation API as it will serve as a cluster scoped coordinator for any eviction.

Wrt. to https://github.com/kubernetes-sigs/descheduler/issues/1253 we will need to decide whether the descheduler will eventually switch to the evacuation API by default or not. Yet, we will still keep the direct eviction API and any potential immediate deletion configurable. It will take some time before the evacuation API becomes GA.

keps/1397-evictions-in-background/README.md
ingvagabund: Not yet. It will strongly depends on when the evacuation API graduates.

keps/1397-evictions-in-background/README.md
ingvagabund: Yeah. A policy config field is versioned. With a feature gate mechanism alpha feature fields can be ignored.

atiratree:
> @atiratree will know better though based on my understand the eviction API will not go away. Instead, users will still have option to choose between the evacuation (delayed eviction) and the eviction APIs. Yet, with strong recommendation to prefer the evacuation API as it will serve as a cluster scoped coordinator for any eviction.

@ingvagabund yes the eviction API is not going away. A better version of it (e.g. Evacuation API, etc.) is still being discussed upstream in sig-apps and should eventually be used by all the major components like the descheduler, cluster autoscaler, etc. Normal users do not really use eviction much.

atiratree:
in accordance with the above consensus
/lgtm

PR #1355, opened 2024-02-26
Clarify node selector

BigValen:
I didn't grasp the subtlety of this option until after a conversation with some folks on #kubernetes-users on the Kubernetes slack. They recommended I submit a PR.

- Move definition down, under the nodeFit option, so it's more obvious they are dependent
- Add examples of nodeSelector
- Mention that pods on nodes that don't match the nodeSelector can still be evicted

### DIFF
--- a/pkg/framework/plugins/removepodshavingtoomanyrestarts/toomanyrestarts.go
+++ b/pkg/framework/plugins/removepodshavingtoomanyrestarts/toomanyrestarts.go
@@ -82,7 +82,13 @@ func New(args runtime.Object, handle frameworktypes.Handle) (frameworktypes.Plug
 				return true
 			}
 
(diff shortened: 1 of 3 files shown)


pkg/framework/plugins/removepodshavingtoomanyrestarts/toomanyrestarts.go
```diff
@@ -82,7 +82,13 @@ func New(args runtime.Object, handle frameworktypes.Handle) (frameworktypes.Plug
 				return true
 			}
 
-			for _, containerStatus := range pod.Status.ContainerStatuses {
+			containerStatuses := pod.Status.ContainerStatuses
```
a7i, line 85: would you please rebase and drop the extra commit?

Issue #1279, opened 2023-10-30
Consider implementing pod (anti)affinity checks in NodeFit 


logyball:
**Is your feature request related to a problem? Please describe.**
 Hello, we have been running into a situation which was somewhat surprising until I [dug into the code](https://github.com/kubernetes-sigs/descheduler/blob/a01322a5e5a74b787912a1e70a5d48f7fea14275/pkg/descheduler/node/node.go#L105) and learned that `NodeFit` does not take into account _all_ scheduler predicates, notably pod (anti)Affinity.  At our organization, we've had multiple occasions where using descheduler `HighNodeUtilization` profile + GKE's `optimize-utilization` [autoscaling profile](https://cloud.google.com/kubernetes-engine/docs/concepts/cluster-autoscaler#autoscaling_profiles) has bit us _because_ we assumed that NodeFit and the kube-scheduler would agree.  

The basic situation is:

- Node A contains Pod X
- Node B contains Pod Y
- Pod X and Pod Y have a pod anti-affinity rule that prevents them from being scheduled on the same node
- Node A is considered "underutilized" by the descheduler
- Node B is _not_ considered "underutilized"
- NodeFit algorithm is run and determines that Pod X can schedule on Node B
- descheduler evicts Pod X
- kube-scheduler determines Pod X cannot schedule on Node B because of Pod X/Y's anti-affinity
- kube-scheduler schedules Pod X on Node A
- The cycle continues with each loop of the descheduler

**Describe the solution you'd like**
I suggest that we look into implementing Pod (anti)affinity checks into the NodeFit calculation, and would be happy to look into contributing it.  It probably adds a decent amount of complexity, but would likely make the descheduler's decisions more closely resemble that of the kube-scheduler, what do you think?

**Describe alternatives you've considered**
- Do nothing: perhaps call out in the documentation that the NodeFit arg does not take into account pod (anti)affinity

**What version of descheduler are you using?**

descheduler version: v0.27.x

logyball:
Also, in a bit of follow up, after digging into a bit of how the scheduler implements these checks, it's not easy to parse for me.  I'd love some guidance here if this is an avenue we'd like to pursue.  It looks like in the past (last seen in v1.9.11 of k8s), there was a [Predicate checker](https://github.com/kubernetes/kubernetes/blob/v1.9.11/plugin/pkg/scheduler/scheduler.go) that we could probably have used to good effect.  However, now the ecosystem looks a bit more complex.

[...]

PR #1356, opened 2024-02-27
Check whether pod matches the inter-pod anti-affinity of another Pod in a given Node in `NodeFit()`

nikimanoledaki:
Closes https://github.com/kubernetes-sigs/descheduler/issues/1279
More context: https://github.com/kubernetes-sigs/descheduler/issues/1149

This PR:
* Adds a check so that NodeFit() assesses whether it is possible for a given Pod to be scheduled on a given Node before descheduling it.
* Uses the logic from the plugin [RemovePodsViolatingInterPodAntiAffinity](https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/framework/plugins/removepodsviolatinginterpodantiaffinity/pod_antiaffinity.go) which already contains logic for fetching and comparing the pod anti-affinity rule of pods in [checkPodsWithAntiAffinityExist](https://github.com/kubernetes-sigs/descheduler/blob/6c865fdf32ac8fa9b20fd9a971b2631eabab0b1d/pkg/framework/plugins/removepodsviolatinginterpodantiaffinity/pod_antiaffinity.go#L152). The logic is extracted from `pkg/framework/plugins/removepodsviolatinginterpodantiaffinity` -> `pkg/utils/predicates.go` (for the most part).
* Updates user-facing docs about this change
* Adds lots of unit tests!

### DIFF
--- a/pkg/descheduler/node/node.go
+++ b/pkg/descheduler/node/node.go
@@ -130,8 +130,10 @@ func NodeFit(nodeIndexer podutil.GetPodsAssignedToNodeFunc, pod *v1.Pod, node *v
 		errors = append(errors, fmt.Errorf("node is not schedulable"))
 	}
 
-	// Check if pod matches pod anti-affinity rule of other pod on node
-	if matchesInterPodAntiAffinityRule(pod, node) {
+	// Check if pod matches inter-pod anti-affinity rule of pod on node
+	if match, err := podMatchesInterPodAntiAffinity(nodeIndexer, pod, node); err != nil {
+		errors = append(errors, err)
+	} else if match {
 		errors = append(errors, fmt.Errorf("pod matches inter-pod anti-affinity rule of other pod on node"))
 	}
 
@@ -329,10 +331,17 @@ func PodMatchNodeSelector(pod *v1.Pod, node *v1.Node) bool {
 	return matches
 }
 
-// matchesInterPodAntiAffinityRule checks if a pod anti-affinity rule matches that of any other pod(s)
-// that are already on the node.
-func matchesInterPodAntiAffinityRule(pod *v1.Pod, node *v1.Node) bool {
-	// if it matches pod anti affinity, return true
-	// if it does not match pod anti affinity, return false
-	return len(node.ObjectMeta.Labels) > 0
+// podMatchesInterPodAntiAffinity checks if the pod matches the anti-affinity rule
+// of another pod that is already on the given node.
+// If a match is found, it returns true.
+func podMatchesInterPodAntiAffinity(nodeIndexer podutil.GetPodsAssignedToNodeFunc, pod *v1.Pod, node *v1.Node) (bool, error) {
+	podsOnNode, err := podutil.ListPodsOnANode(node.Name, nodeIndexer, nil)
+	if err != nil {
+		return false, fmt.Errorf("error listing all pods: %v", err)
+	}
+
+	podsInANamespace := podutil.GroupByNamespace(podsOnNode)
+	nodeMap := utils.CreateNodeMap([]*v1.Node{node})
+
+	return utils.CheckPodsWithAntiAffinityExist(pod, podsInANamespace, nodeMap), nil
 }
--- a/pkg/descheduler/node/node.go
+++ b/pkg/descheduler/node/node.go
@@ -335,6 +335,10 @@ func PodMatchNodeSelector(pod *v1.Pod, node *v1.Node) bool {
 // of another pod that is already on the given node.
 // If a match is found, it returns true.
 func podMatchesInterPodAntiAffinity(nodeIndexer podutil.GetPodsAssignedToNodeFunc, pod *v1.Pod, node *v1.Node) (bool, error) {
+	if pod.Spec.Affinity == nil || pod.Spec.Affinity.PodAntiAffinity == nil {
+		return false, nil
+	}
+
 	podsOnNode, err := podutil.ListPodsOnANode(node.Name, nodeIndexer, nil)
 	if err != nil {
 		return false, fmt.Errorf("error listing all pods: %v", err)
--- a/pkg/descheduler/node/node.go
+++ b/pkg/descheduler/node/node.go
@@ -130,6 +130,11 @@ func NodeFit(nodeIndexer podutil.GetPodsAssignedToNodeFunc, pod *v1.Pod, node *v
 		errors = append(errors, fmt.Errorf("node is not schedulable"))
 	}
 
+	// Check if pod matches pod anti-affinity rule of other pod on node
+	if matchesInterPodAntiAffinityRule(pod, node) {
+		errors = append(errors, fmt.Errorf("pod matches inter-pod anti-affinity rule of other pod on node"))
+	}
+
 	return errors
 }
 
@@ -322,4 +327,12 @@ func PodMatchNodeSelector(pod *v1.Pod, node *v1.Node) bool {
 		return false
 	}
 	return matches
+}
+
+// matchesInterPodAntiAffinityRule checks if a pod anti-affinity rule matches that of any other pod(s)
+// that are already on the node.
+func matchesInterPodAntiAffinityRule(pod *v1.Pod, node *v1.Node) bool {
+	// if it matches pod anti affinity, return true
+	// if it does not match pod anti affinity, return false
+	return len(node.ObjectMeta.Labels) > 0
 }
--- a/pkg/utils/priority.go
+++ b/pkg/utils/priority.go
@@ -14,10 +14,10 @@ import (
 
 const SystemCriticalPriority = 2 * int32(1000000000)
 
-// GetNamespacesFromPodAffinityTerm returns a set of names
+// getNamespacesFromPodAffinityTerm returns a set of names
 // according to the namespaces indicated in podAffinityTerm.
 // If namespaces is empty it considers the given pod's namespace.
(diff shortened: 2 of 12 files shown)


damemi (commented):
/ok-to-test

pkg/descheduler/node/node.go
```diff
@@ -323,3 +330,18 @@ func PodMatchNodeSelector(pod *v1.Pod, node *v1.Node) bool {
 	}
 	return matches
 }
+
+// podMatchesInterPodAntiAffinity checks if the pod matches the anti-affinity rule
+// of another pod that is already on the given node.
+// If a match is found, it returns true.
+func podMatchesInterPodAntiAffinity(nodeIndexer podutil.GetPodsAssignedToNodeFunc, pod *v1.Pod, node *v1.Node) (bool, error) {
```
logyball, line 337: [NIT] - I think the descheduler terminology here is `violates` anti-affinity, e.g. [the RemovePodsViolatingInterPodAntiAffinity plugin](https://github.com/kubernetes-sigs/descheduler?tab=readme-ov-file#removepodsviolatinginterpodantiaffinity).

I suppose it's not necesarily violating anti-affinity, since it hasn't actually happened yet, so probably not a big deal.  `podWillViolateInterPodAntiAffinity` sounds dumb too.  Maybe leave it how it is now that I think through it.

pkg/utils/priority.go
```diff
@@ -17,7 +17,7 @@ const SystemCriticalPriority = 2 * int32(1000000000)
 // GetNamespacesFromPodAffinityTerm returns a set of names
 // according to the namespaces indicated in podAffinityTerm.
 // If namespaces is empty it considers the given pod's namespace.
-func GetNamespacesFromPodAffinityTerm(pod *v1.Pod, podAffinityTerm *v1.PodAffinityTerm) sets.Set[string] {
+func getNamespacesFromPodAffinityTerm(pod *v1.Pod, podAffinityTerm *v1.PodAffinityTerm) sets.Set[string] {
```
logyball, line 20: Im a little concerned about this function. I realize this PR does not modify it heavily (modifying public/private), but I'm doing some research on the [namespaceSelector](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#namespace-selector) functionality within pod affinity/anti-affinity and am wondering if we're missing a condition here.

The definition says that:

> You can also select matching namespaces using namespaceSelector, which is a label query over the set of namespaces. The affinity term is applied to namespaces selected by both namespaceSelector and the namespaces field. Note that an empty namespaceSelector ({}) matches all namespaces, while a null or empty namespaces list and null namespaceSelector matches the namespace of the Pod where the rule is defined.

It looks to me like if we have _no_ namespaceSelector defined (i.e. namespaceSelector is `nil`), the current comparison behavior is correct.  However, if we have namespaceSelector as `{}` or an actually defined label selector, we could be missing out on namespaces here.

I think that the way this works is that we could have several namespaces here (not real examples)

 ```sh
❯ k get ns namespace-a -o yaml
apiVersion: v1
kind: Namespace
metadata:
  labels:
    kubernetes.io/metadata.name: namespace-a
    owner: team-a
  name: namespace-a
  
❯ k get ns namespace-b -o yaml
apiVersion: v1
kind: Namespace
metadata:
  labels:
    kubernetes.io/metadata.name: namespace-b
    owner: team-b
  name: namespace-b

❯ k get ns dev-o yaml
apiVersion: v1
kind: Namespace
metadata:
  labels:
    kubernetes.io/metadata.name: dev
    owner: team-b
  name: dev
```

And we could have a pod with a pod antiAffinity:

```yaml
spec:
      affinity:
        podAntiAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
          - namespaceSelector:
              matchLabels:
                team: team-b
            labelSelector:
              matchExpressions:
              - key: name
                operator: In
                values:
                - tmp
            topologyKey: "kubernetes.io/hostname"
```

I realize this is a somewhat contrived example, but in this case let's assume that this was deployed to `dev`.  It would totally miss out that there may be some pods to check in `namespace-b`, as that would be matched by the namespaceSelector but not by the fact that the `namespace` list was empty.

I'm not sure if this is a blocker or if it just something that we should address at _some point_, I'll leave that up to the maintainers.  I'm not sure how widespread the usage of namespace selectors is (we don't use them), but this may result in surprising behavior.

logyball (changes requested):
Hey @nikimanoledaki , great work here! I've got the tests all passing, and look forward to testing this out with you.  I have one question about the namespaces that we'll be comparing in order to figure out _which_ pods need to checked for anti-affinity, let me know what you think.

pkg/descheduler/node/node.go
nikimanoledaki: The different naming was intentional! In my interpretation, I see a different angle/perspective/approach between what `NodeFit()` and the plugin each do.

The plugin looks at _all_ nodes, namespaces, and pods for pairs of pods that violate an inter-pod anti-affinity rule. If the Pods violate this rule, one of them is evicted. Whereas `NodeFit()` checks to see if there is a match so that it does not schedule it in the first place. My interpretation of "violating the rule" was that the two Pods with anti-affinity should already be present on the Node, otherwise it's just a match - but it's just an interpretation and maybe it could be worded differently or more clearly !

knelasevero:
I think this can get in before we redesign nodeFit, as @a7i  pointed out. 

gonna approve and let lgtm's for other maintainers and @logyball 

/approve

pkg/utils/priority.go
logyball: > If we have an actually defined label namespace selector, it would indeed not evaluate namespaces that aren't included in the selector, but this seems to me like the expected behaviour, no? 🤔 I may be misunderstanding something.

What I'm saying here is that if we had a `namespaceSelector` defined, as is allowed by the pod (anti)affinity spec, we're ignoring it in our list of namespaces to check for descheduler, which would _not_ be expected behavior.

The easier case in my mind is imagining we have _no_ `namespaces` defined in the pod affinity term, but we do have a label selector defined in `namespacesSelector`.  In this case, the function `GetNamespacesFromPodAffinityTerm` will return only the pod itself's Namespace, and not any namespaces that match the `namespaceSelector` `labelSelector`.  The kube-scheduler, however, _will_ match the namespaces in the `namespaceSelector`, and will cause differing behavior between the two.

My question is: does this matter enough to fix now?  This PR isn't _introducing_ this behavior, it's pre-existing.  Perhaps this could be just follow-up work for the `nodeFit` redesign.

nikimanoledaki: Gotcha, I see what you mean - `getNamespacesFromPodAffinityTerm` does not take into account the `namespaceSelector`. Good catch! This does seem like a missing feature. The PR is already rather large and patches one thing but this could definitely cause issues for folks. It would be worth opening a separate issue so that it's tracked and can be tackled in a follow-up PR.

pkg/descheduler/node/node.go
logyball: sound reasonable to me!

pkg/utils/priority.go
logyball: I think that you're right.  I've created this issue:

https://github.com/kubernetes-sigs/descheduler/issues/1361

and resolved my comments, I'll approve and we can follow up.  @knelasevero @a7i - ok to slide this issue into consideration for the nodeFit re-design?

nikimanoledaki: Perfect, thank you @logyball! 😄 

Resolving this for now - we can continue the discussion about `namespaceSelector` in the follow-up issue that you created 👍

logyball (approved):
LGTM! nice work

nikimanoledaki:
Hi @a7i 👋 I would really appreciate your review since we discussed this feature on Slack. Thank you!

nikimanoledaki:
@a7i thanks for taking a look. Is the PR ok to merge? Would appreciate your help with this / next steps, thank you!

pkg/descheduler/node/node.go
```diff
@@ -323,3 +330,18 @@ func PodMatchNodeSelector(pod *v1.Pod, node *v1.Node) bool {
...
+	podsOnNode, err := podutil.ListPodsOnANode(node.Name, nodeIndexer, nil)
```
a7i, line 338: Can we get out early. In a cluster with 20k pods and only a handful of pods with PodAntiAffinity this can get expensive, right?


```suggestion
        if pod.Spec.Affinity == nil || pod.Spec.Affinity. PodAntiAffinity == nil {
          return false, nil
        }
	podsOnNode, err := podutil.ListPodsOnANode(node.Name, nodeIndexer, nil)
```

ingvagabund:
The changes look reasonable in general. Thank you for your time and energy. Would you please reduce the number of commits in the PR to make the code changes easier to follow? E.g. making functions/methods private right away, moving code forth and back in a single commit, building the new functions right way in a single commit, etc.

pkg/descheduler/node/node.go
nikimanoledaki: Absolutely - just added this check in the last commit! Thanks @a7i 👍

nikimanoledaki:
Thank you @ingvagabund! I rewrote the history a bit to remove in-between draft steps & duplications :)

/test all

a7i:
@nikimanoledaki I'm working on getting this merged today. We have a CI job that is stuck

a7i:
seems to be stuck on linting: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_descheduler/1356/pull-descheduler-verify-master/1767621656668278784

a7i:
thank you for your contribution @nikimanoledaki  🏆

PR #1360, opened 2024-03-05
Add argument to remove only pods violating allowlisted taints

### LINKED ISSUE #1353: Add option to apply RemovePodsViolatingNodeTaints only for explicitly included taints

We want to control the taints for which the RemovePodsViolatingNodeTaints strategy gets applied. 

While the `excludedTaints` option offers some control, it only allows configuration of well-known and expected cases. 
It doesn't work for unforeseen situations where unpredictable taints may appear. 

For example during an incident, an engineer might want to manually cordon the node while keeping all the workloads on it in order to not cause further disruption. This engineer might be missing the institutional knowledge of which taints are excluded to achieve this and thus might unknowingly cause the pods to get evicted by the RemovePodsViolatingNodeTaints strategy. 

On the other hand it's impossible for the infrastructure owners to anticipate all possible taints which they would need to exclude. 

We would like to have the option to apply the RemovePodsViolatingNodeTaints strategy only for an explicitly allowlisted set of taints. 

If and only if a taint is in a list of `includedTaints` will the pods on the node be considered candidates for descheduling. 

If `includedTaints` is empty, all taints will be included by default, compatible with the current behavior.

`excludedTaints` still applies to all taints that are `includedTaints`. Meaning a taint that is in both will effectively be excluded.

1. One alternative would be to limit the scope of evictions through node labels. But this would require setting/removing node labels together with taints. This is not supported by most tools designed for node lifecycle management (like [node-problem-detector](https://github.com/kubernetes/node-problem-detector)) 

2. Limit the applicable taints through policy enforcement with e.g. Kyverno. But this would effectively only implement the proposed solution with another tool, which makes maintaining the whole setup a lot more difficult. 

descheduler version: [v0.29.0](https://github.com/kubernetes-sigs/descheduler/releases/tag/v0.29.0)

**Additional context**

etoster:
Fixes #1353. 

Adds the option to apply the RemovePodsViolatingNodeTaints strategy only for an explicitly allowlisted set of taints.

When a list of includedTaints is provided, a taint will be considered if and only if it matches an included key **or** key=value from the list. Otherwise it will be ignored. Leaving includedTaints unset will include any taint by default.

### DIFF
--- a/pkg/framework/plugins/removepodsviolatingnodetaints/node_taint_test.go
+++ b/pkg/framework/plugins/removepodsviolatingnodetaints/node_taint_test.go
@@ -100,6 +100,10 @@ func TestDeletePodsViolatingNodeTaints(t *testing.T) {
 		createPreferNoScheduleTaint("testTaint", "test", 1),
 	}
 
+	node7 := test.BuildTestNode("n7", 2000, 3000, 10, nil)
+	node7 = addTaintsToNode(node7, "testTaint", "test", []int{1})
+	node7 = addTaintsToNode(node7, "testingTaint", "testing", []int{1})
+
 	p1 := test.BuildTestPod("p1", 100, 0, node1.Name, nil)
 	p2 := test.BuildTestPod("p2", 100, 0, node1.Name, nil)
 	p3 := test.BuildTestPod("p3", 100, 0, node1.Name, nil)
@@ -161,6 +165,15 @@ func TestDeletePodsViolatingNodeTaints(t *testing.T) {
 	// PreferNoSchedule:testTaint0=test0 so the pod is not tolarated
 	p13 = addTolerationToPod(p13, "testTaint", "test", 0, v1.TaintEffectPreferNoSchedule)
 
+	p14 := test.BuildTestPod("p14", 100, 0, node7.Name, nil)
+	p14.ObjectMeta.OwnerReferences = test.GetNormalPodOwnerRefList()
+	p14 = addTolerationToPod(p14, "testTaint", "test", 1, v1.TaintEffectNoSchedule)
+
+	p15 := test.BuildTestPod("p15", 100, 0, node7.Name, nil)
+	p15.ObjectMeta.OwnerReferences = test.GetNormalPodOwnerRefList()
+	p15 = addTolerationToPod(p15, "testTaint", "test", 1, v1.TaintEffectNoSchedule)
+	p15 = addTolerationToPod(p15, "testingTaint", "testing", 1, v1.TaintEffectNoSchedule)
+
 	var uint1 uint = 1
 
 	tests := []struct {
@@ -175,6 +188,7 @@ func TestDeletePodsViolatingNodeTaints(t *testing.T) {
 		nodeFit                        bool
 		includePreferNoSchedule        bool
 		excludedTaints                 []string
+		includedTaints                 []string
 	}{
 		{
 			description:             "Pods not tolerating node taint should be evicted",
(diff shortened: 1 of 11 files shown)


knelasevero:
should we let users set both included and excluded taints? When we look at namespace inclusion/exclusion we make inclusion be something like: "this and nothing else". Shouldn't we keep consistency here?

@a7i @damemi @ingvagabund

damemi:
> should we let users set both included and excluded taints? When we look at namespace inclusion/exclusion we make inclusion be something like: "this and nothing else". Shouldn't we keep consistency here?
> 
> @a7i @damemi @ingvagabund

I'm for using one or the other until there's a use case for both. I had mentioned in the original issue that it didn't seem harmful to do both, but I agree that it's safer and more consistent to make them exclusive for now.

etoster:
> should we let users set both included and excluded taints? When we look at namespace inclusion/exclusion we make inclusion be something like: "this and nothing else". Shouldn't we keep consistency here?

@damemi also pointed this out in the [issue](https://github.com/kubernetes-sigs/descheduler/issues/1353#issuecomment-1959451405). 

I started looking into how it's done for namespaces, but I wasn't sure where would be the right place to check in this case. If it would be within the strategy plugin or if there is some linter parsing the deschedulerPolicy resource before (which I didn't find). So I proposed the change without it, but I'm happy to add it, if you could give me some guidance on what you consider the best practice for it!

ingvagabund:
+1 for allowing to set only one of include/exclude lists, not both.

ingvagabund:
> If it would be within the strategy plugin

Yeap. Each strategy/plugin is responsible for its own arg validation. E.g. https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/framework/plugins/removeduplicates/validation.go

etoster:
Thanks @ingvagabund! I added the validation to allow only one of include/exclude taints. Please have a look. 
cc @knelasevero @damemi

pkg/framework/plugins/removepodsviolatingnodetaints/node_taint_test.go
```diff
@@ -322,6 +323,42 @@ func TestDeletePodsViolatingNodeTaints(t *testing.T) {
...
+			includedTaints:          []string{"testTaint1=test1"},
+			expectedEvictedPodCount: 1, // p1 and p3 tolerate the included taint, p2 gets evicted
+		},
+		{
+			description:             "Pods not tolerating not included taints should not get evicted",
+			pods:                    []*v1.Pod{p1, p2, p4},
+			nodes:                   []*v1.Node{node1},
+			evictLocalStoragePods:   false,
+			evictSystemCriticalPods: false,
+			includedTaints:          []string{"testTaint1=test2"},
```
ingvagabund, line 359: This case tests the same case as "Pods not tolerating not included taints should not get evicted". I.e. a node is not tainted with `includedTaints` taints. So no toleration is needed. Worth adding a test where `includeTaints` is empty, a node is tainted with two taints, pods has a list of pods where some tolerate both taints, some don't.

etoster: You're right. I merged both "Pods not tolerating not included taints should not get evicted" tests and created a new one "Pods not tolerating all taints are evicted when includedTaints is empty".

pkg/framework/plugins/removepodsviolatingnodetaints/node_taint_test.go
```diff
@@ -322,6 +336,41 @@ func TestDeletePodsViolatingNodeTaints(t *testing.T) {
 			expectedEvictedPodCount: 0, // p2 and p7 can't be evicted
 			nodeFit:                 true,
 		},
+		{
+			description:             "Pods tolerating included taints should not get evicted",
+			pods:                    []*v1.Pod{p1},
+			nodes:                   []*v1.Node{node1},
```
ingvagabund, line 342: s/node1/node7
This way the test can validate `testingTaint1=testing1` taint has no effect on the eviction.

ingvagabund (changes requested):
One last nit.

The changes are not that significant so they can be part of a single commit (will leave this up to you). Multiple commits are ok as well. In which case would you please integrate the last commit into the first one?

knelasevero:
After Jan's comments are addressed this lgtm

pkg/framework/plugins/removepodsviolatingnodetaints/node_taint_test.go
etoster: Yes, this is a nice double check. Changed it.

etoster:
Thank you all for your quick responses and helpful feedback! This was a great experience as a first time contributor. :)

PR #1362, opened 2024-03-06
Disambiguate node fit log messages

jackfrancis:
This PR adds some additional language across the three flavors of "does this pod fit on a node?" funcs that are used to inform eviction filters. This can aid debugging so that our logs can tell us at a glance where in the eviction flow we are at.


damemi (approved):
/approve

PR #1368, opened 2024-03-28
feat: expose node reached metric

fanhaouu:
Now every strategy incorporates the 'Evictor().NodeLimitExceeded' logic, which leads to an eternal failure to meet the logic in the 'EvictPod' method, specifically 'pe.nodepodCount[pod.Spec.NodeName]+1 > *pe.maxPodsToEvictPerNode'. Consequently, this prevents the exposure of node-reached metrics and the evaluation of reasons for not evicting subsequent pods.

This pr can resolve that.

### DIFF
--- a/pkg/descheduler/evictions/evictions.go
+++ b/pkg/descheduler/evictions/evictions.go
@@ -125,13 +125,20 @@ func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod, opts EvictOptio
 	defer span.End()
 
 	if pod.Spec.NodeName != "" {
-		if pe.maxPodsToEvictPerNode != nil && pe.nodepodCount[pod.Spec.NodeName]+1 > *pe.maxPodsToEvictPerNode {
-			if pe.metricsEnabled {
-				metrics.PodsEvicted.With(map[string]string{"result": "maximum number of pods per node reached", "strategy": opts.StrategyName, "namespace": pod.Namespace, "node": pod.Spec.NodeName, "profile": opts.ProfileName}).Inc()
+		if pe.maxPodsToEvictPerNode != nil {
+			if pe.nodepodCount[pod.Spec.NodeName]+1 == *pe.maxPodsToEvictPerNode {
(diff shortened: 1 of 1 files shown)


ingvagabund:
@fanhaouu can you provide an example of the "eternal" failure? Is there a reproducer for this failure? Is there an issue that can help us to see how the failure affects the descheduler, resp. users?

fanhaouu:
The current various descheduling strategies all use the `NodeLimitExceeded` method in `PodEvictor` to determine in advance whether the next eviction operation is necessary. This ensures that the condition 

> pe.namespacePodCount[pod.Namespace]+1 > *pe.maxPodsToEvictPerNamespace

 will never occur.

For example:
https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/framework/plugins/podlifetime/pod_lifetime.go#L135
https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/framework/plugins/removeduplicates/removeduplicates.go#L214
https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/framework/plugins/removefailedpods/failedpods.go#L107

fanhaouu:
> @fanhaouu can you provide an example of the "eternal" failure? Is there a reproducer for this failure? Is there an issue that can help us to see how the failure affects the descheduler, resp. users?

The current various descheduling strategies all use the `NodeLimitExceeded` method in `PodEvictor` to determine in advance whether the next eviction operation is necessary. This ensures that the condition 

> pe.namespacePodCount[pod.Namespace]+1 > *pe.maxPodsToEvictPerNamespace

 will never occur.

For example:
https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/framework/plugins/podlifetime/pod_lifetime.go#L135
https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/framework/plugins/removeduplicates/removeduplicates.go#L214
https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/framework/plugins/removefailedpods/failedpods.go#L107

PR #1373, opened 2024-04-08
bump(golang.org/x/net)=v0.23.0

ingvagabund:
For https://pkg.go.dev/vuln/GO-2024-2687. Even though http/2 is disabled by default, let's be proactive here just in case.

Alongside https://github.com/kubernetes/test-infra/pull/32403.


damemi (approved):
/lgtm
/approve

PR #1374, opened 2024-04-09
bump to go 1.22.2

jklaw90:
CVE-2023-45288 was handled in Go 1.22.2.


jklaw90:
seems to be our ci is running 1.21.9, this also contains the fix i can use that version if we prefer?

ingvagabund:
> seems like our ci is running 1.21.9? that version also contains the fix for the CVE, i can set our mod to 1.21.9 if it's easier.

Bumped to 1.21.9 through https://github.com/kubernetes/test-infra/pull/32403

ingvagabund:
1.22 works better given k8s 1.30 pins 1.22: https://github.com/kubernetes/kubernetes/blob/77aa9c21c295614817d5d49ea81f20783480691b/go.mod#L9

ingvagabund:
Bumping the go version in CI for the master branch: https://github.com/kubernetes/test-infra/pull/32415

Dockerfile
```diff
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-FROM golang:1.21.5
+FROM golang:1.22.2
```
a7i, line 14: can we use 1.21.9 to be in sync with https://github.com/kubernetes/test-infra/pull/32403

a7i: Just saw @ingvagabund comments above

a7i:
@jklaw90 would you please update [here](https://github.com/kubernetes-sigs/descheduler/blob/master/hack/update-gofmt.sh#L26) and [here](https://github.com/kubernetes-sigs/descheduler/blob/master/hack/verify-gofmt.sh#L26)?

PR #1376, opened 2024-04-09
helm: Allow the use of existing policy configMap

agmimidi:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it:**
Enables the use of an existing policy configMap. This can be useful in complex scenarios where the policy file needs to be constructed via other means (e.g. An existing gitOps/IaC setup)

**Which issue(s) this PR fixes:**
NONE

**Special notes for your reviewer:**
Default behaviour is kept as-is, change only takes effect when  `existingPolicyConfigMap` is set to true.

**Does this PR introduce a user-facing change?**
See above

**Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:**
This PR also modifies the README file


damemi (commented):
/ok-to-test

charts/descheduler/templates/configmap.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if not  .Values.existingPolicyConfigMap }}
```
a7i, line 1: couldn't we just use the existing `deschedulerPolicy`?

```suggestion
{{- if .Values.deschedulerPolicy }}
```

agmimidi: Hi Amir,
Nice catch. I have implemented as recommended.  I also added a few comments on values.yaml to explain the new functionality

charts/descheduler/templates/configmap.yaml
```diff
@@ -10,3 +11,4 @@ data:
     apiVersion: "{{ .Values.deschedulerPolicyAPIVersion }}"
     kind: "DeschedulerPolicy"
 {{ toYaml .Values.deschedulerPolicy | trim | indent 4 }}
+{{- end }}
```
a7i, line 14: would you please modify your editor setting to add a new line at the EOF?

charts/descheduler/values.yaml
```diff
@@ -81,6 +81,10 @@ cmdOptions:
 # Recommended to use the latest Policy API version supported by the Descheduler app version
 deschedulerPolicyAPIVersion: "descheduler/v1alpha2"
 
+# deschedulerPolicy contains the policies the descheduler will execute. 
+# To use policies stored in an existing configMap use:
+# NOTE: The name of the cm should comply to {{ template "descheduler.fullname" . }}
+# deschedulerPolicy: []
```
a7i, line 87:
```suggestion
# deschedulerPolicy: {}
```

charts/descheduler/values.yaml
```diff
@@ -81,6 +81,10 @@ cmdOptions:
 # Recommended to use the latest Policy API version supported by the Descheduler app version
 deschedulerPolicyAPIVersion: "descheduler/v1alpha2"
 
+# deschedulerPolicy contains the policies the descheduler will execute.
```
a7i, line 84:
```suggestion
# deschedulerPolicy contains the policies the descheduler will execute.
```

a7i:
/retitle helm: Allow the use of existing policy configMap
/lgtm

Thank you for your contribution @agmimidi

Issue #1370, opened 2024-03-29
Inter-pod anti-affinity check for `NodeFit` should not check if Pods are on the same Node already


nikimanoledaki:
**What version of descheduler are you using?**

descheduler version: `master` - not released yet, but would be introduced in `0.30.0`

**What did you do?** **What did you expect to see?** **What did you see instead?**

Folow-up to https://github.com/kubernetes-sigs/descheduler/pull/1356

We've been testing the recent addition to `NodeFit` on our environment to check for inter-pod anti-affinity. I found a bug, which means the feature is not currently doing what it's supposed to be doing. It's not creating any major issues or consequences - just not doing what it should be doing! 

It should be assessing whether a Pod fits on another Node based on whether another Pod on this other Node has matching antiaffinities, but it returns early, due to the following issue.

[Here](https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/utils/predicates.go#L326), we evaluate whether both Pods are present on the Node. It only makes sense to check this for the plugin where this function originated from, RemovePodsViolatingInterPodAntiAffinity. That plugin checks if both Pods are on the same Node and whether to evict one of them, if they have matching antiaffinities. It does not make sense to do this when we’re evaluating if a Pod could fit on another Node as part of NodeFit. 

I can open a PR to fix this as soon as possible on Tuesday next week unless someone else gets to it first. Hopefully we can add this to the descheduler `0.30.0` before it's released.

nikimanoledaki:
I can confirm that #1395 addresses the remaining bug with the descheduler's evaluation of inter-pod anti-affinity policies 👍 🎉  Thank you @fanhaouu @ingvagabund for making Kubernetes a little bit (even) more awesome! 😄

PR #1395, opened 2024-05-11
fix the issue that the pod anti-filtering rules are not taking effect

fanhaouu:
Currently, the "podMatchesInterPodAntiAffinity" method has an issue. It is unable to check if the pod matches the anti-affinity rule of another pod that is already on the given node.

The root cause is that the "CheckPodsWithAntiAffinityExist" method in "predicate.go" can only check if pods with AntiAffinity issues exist on the same node

This pr can resolve that.

### DIFF
--- a/pkg/descheduler/node/node_test.go
+++ b/pkg/descheduler/node/node_test.go
@@ -759,6 +759,9 @@ func TestNodeFit(t *testing.T) {
 			"region": "main-region",
 		}
 	})
+
+	nodeNolabel := test.BuildTestNode("node", 64000, 128*1000*1000*1000, 2, nil)
+
 	tests := []struct {
 		description string
 		pod         *v1.Pod
@@ -767,7 +770,7 @@ func TestNodeFit(t *testing.T) {
 		err         error
 	}{
 		{
-			description: "insufficient cpu",
+			description: "Insufficient cpu",
 			pod:         test.BuildTestPod("p1", 10000, 2*1000*1000*1000, "", nil),
 			node:        node,
 			podsOnNode: []*v1.Pod{
@@ -776,7 +779,7 @@ func TestNodeFit(t *testing.T) {
 			err: errors.New("insufficient cpu"),
 		},
 		{
-			description: "insufficient pod num",
+			description: "Insufficient pod num",
 			pod:         test.BuildTestPod("p1", 1000, 2*1000*1000*1000, "", nil),
 			node:        node,
 			podsOnNode: []*v1.Pod{
@@ -786,7 +789,7 @@ func TestNodeFit(t *testing.T) {
 			err: errors.New("insufficient pods"),
 		},
 		{
-			description: "matches inter-pod anti-affinity rule of pod on node",
+			description: "Pod matches inter-pod anti-affinity rule of other pod on node",
 			pod:         test.PodWithPodAntiAffinity(test.BuildTestPod("p1", 1000, 1000, node.Name, nil), "foo", "bar"),
 			node:        node,
 			podsOnNode: []*v1.Pod{
@@ -795,11 +798,36 @@ func TestNodeFit(t *testing.T) {
 			err: errors.New("pod matches inter-pod anti-affinity rule of other pod on node"),
 		},
 		{
-			description: "pod fits on node",
+			description: "Pod doesn't match inter-pod anti-affinity rule of other pod on node, because pod and other pod is not same namespace",
+			pod:         test.PodWithPodAntiAffinity(test.BuildTestPod("p1", 1000, 1000, node.Name, nil), "foo", "bar"),
+			node:        node,
+			podsOnNode: []*v1.Pod{
+				test.PodWithPodAntiAffinity(test.BuildTestPod("p2", 1000, 1000, node.Name, func(pod *v1.Pod) {
+					pod.Namespace = "test"
+				}), "foo", "bar"),
+			},
+		},
+		{
+			description: "Pod doesn't match inter-pod anti-affinity rule of other pod on node, because other pod not match labels of pod",
+			pod:         test.PodWithPodAntiAffinity(test.BuildTestPod("p1", 1000, 1000, node.Name, nil), "foo", "bar"),
+			node:        node,
+			podsOnNode: []*v1.Pod{
+				test.PodWithPodAntiAffinity(test.BuildTestPod("p2", 1000, 1000, node.Name, nil), "foo1", "bar1"),
+			},
+		},
+		{
+			description: "Pod doesn't match inter-pod anti-affinity rule of other pod on node, because node have no topologyKey",
+			pod:         test.PodWithPodAntiAffinity(test.BuildTestPod("p1", 1000, 1000, "node1", nil), "foo", "bar"),
+			node:        nodeNolabel,
+			podsOnNode: []*v1.Pod{
+				test.PodWithPodAntiAffinity(test.BuildTestPod("p2", 1000, 1000, node.Name, nil), "foo", "bar"),
+			},
+		},
+		{
+			description: "Pod fits on node",
 			pod:         test.BuildTestPod("p1", 1000, 1000, "", func(pod *v1.Pod) {}),
 			node:        node,
 			podsOnNode:  []*v1.Pod{},
-			err:         nil,
 		},
 	}
 
--- a/pkg/utils/predicates.go
+++ b/pkg/utils/predicates.go
@@ -24,9 +24,36 @@ import (
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/component-helpers/scheduling/corev1"
 	"k8s.io/klog/v2"
 )
+
+// GetNamespacesFromPodAffinityTerm returns a set of names
+// according to the namespaces indicated in podAffinityTerm.
+// If namespaces is empty it considers the given pod's namespace.
+func GetNamespacesFromPodAffinityTerm(pod *v1.Pod, podAffinityTerm *v1.PodAffinityTerm) sets.Set[string] {
+	names := sets.New[string]()
+	if len(podAffinityTerm.Namespaces) == 0 {
+		names.Insert(pod.Namespace)
+	} else {
+		names.Insert(podAffinityTerm.Namespaces...)
+	}
+	return names
+}
+
+// PodMatchesTermsNamespaceAndSelector returns true if the given <pod>
+// matches the namespace and selector defined by <affinityPod>`s <term>.
+func PodMatchesTermsNamespaceAndSelector(pod *v1.Pod, namespaces sets.Set[string], selector labels.Selector) bool {
+	if !namespaces.Has(pod.Namespace) {
+		return false
+	}
+
+	if !selector.Matches(labels.Set(pod.Labels)) {
+		return false
+	}
+	return true
+}
 
 // The following code has been copied from predicates package to avoid the
 // huge vendoring issues, mostly copied from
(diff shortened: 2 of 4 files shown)


ingvagabund:
@fanhaouu would you please share an example of this case? With `nodeFit=true` the pod anti-affinity check is completely skipped. 

> Currently, the "podMatchesInterPodAntiAffinity" method has an issue. It is unable to check if the pod matches the anti-affinity rule of another pod that is already on the given node.
> The root cause is that the "CheckPodsWithAntiAffinityExist" method in "predicate.go" can only check if pods with AntiAffinity issues exist on the same node

Can you rephrase the description in terms of "a pod P1 on a node N1", "a pod P1 wrt. P2", etc.? The description is still  ambiguous. It would help to start the description with "There are two nodes N1 and N2. Both pods P1 and P2 are scheduled. P1 to node N1, P2 to node N2. There's another pod P3 scheduled to node N? that violates anti-affinity of P?. ...".

fanhaouu:
@ingvagabund I don't know if the example I provided is clear enough?

pkg/utils/predicates.go
```diff
@@ -322,11 +322,21 @@ func CheckPodsWithAntiAffinityExist(pod *v1.Pod, pods map[string][]*v1.Pod, node
...
+						continue
+					}
+
+					node, ok := nodeMap[pod.Spec.NodeName]
+					if !ok {
+						continue
+					}
+
+					if nodeFit {
+						if _, ok := node.Labels[term.TopologyKey]; ok {
```
ingvagabund, line 335: Here we still do not know whether `pod` and `existingPod` are violating the anti-affinity rule. Just that both pods are in anti-affinity relation. Checking whether `pod` is in a given topology domain is not sufficient here.

pkg/utils/predicates_test.go
```diff
@@ -1099,12 +1101,61 @@ func TestCheckPodsWithAntiAffinityExist(t *testing.T) {
 			}), "foo", "bar"),
 			podsInNamespace: map[string][]*v1.Pod{},
 			nodeMap:         map[string]*v1.Node{},
+			nodeFit:         false,
 			expMatch:        false,
 		},
+		{
+			name: "found pod matching pod anti-affinity when node fitting",
+			pod:  test.PodWithPodAntiAffinity(test.BuildTestPod("p1", 1000, 1000, "node", nil), "foo", "bar"),
+			podsInNamespace: map[string][]*v1.Pod{
```
ingvagabund, line 1110: Unless a pod is scheduled to a node that no longer has TopologyKey label key, pod anti-affinity needs presence of at least two pods to violate the rule. The same holds for other newly introduced test cases.

ingvagabund:
> If according to the current logic of podMatchesInterPodAntiAffinity, to determine if pod1 can be scheduled to node2, this method returns match as false and error as nil, but in reality, the match value should be true, meaning that pod1 cannot be scheduled to node2

So it's the nodefit part here that's in question? I.e. when `pod1` is to be evicted you'd expect the descheduler to check whether there's a node where `pod1` (resp. it's clone) can be scheduled to?

ingvagabund:
Is this PR expected to resolve https://github.com/kubernetes-sigs/descheduler/issues/1370?

ingvagabund:
I see now what's going on:

> [Here](https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/utils/predicates.go#L326), we evaluate whether both Pods are present on the Node. It only makes sense to check this for the plugin where this function originated from, RemovePodsViolatingInterPodAntiAffinity. That plugin checks if both Pods are on the same Node and whether to evict one of them, if they have matching antiaffinities. It does not make sense to do this when we’re evaluating if a Pod could fit on another Node as part of NodeFit.

ingvagabund:
> It should be assessing whether a Pod fits on another Node based on whether another Pod on this other Node has matching antiaffinities, but it returns early, due to the following issue.

Let's have P1 scheduled to N1 and P2 scheduled to N2. Pods P1 and P2 have matching anti-affinities. The check here is `NodeFit(P1,N2)` and the claim here is the check "returns early". Because of https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/utils/predicates.go#L358:
```
func hasSameLabelValue(node1, node2 *v1.Node, key string) bool {
	if node1.Name == node2.Name { // returns early
		return true
	}
...
}
```

So the goal here is to ignore node N1 in the check. N1 is retrieved due to accessing `pod.Spec.NodeName` in https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/utils/predicates.go#L326. Instead, just check whether P1 can be scheduled to N2.

@fanhaouu is this the current issue in hand? Just rephrasing so it is clear what this issue and fix is all about.

@nikimanoledaki for awareness.

fanhaouu:
> > It should be assessing whether a Pod fits on another Node based on whether another Pod on this other Node has matching antiaffinities, but it returns early, due to the following issue.
> 
> Let's have P1 scheduled to N1 and P2 scheduled to N2. Pods P1 and P2 have matching anti-affinities. The check here is `NodeFit(P1,N2)` and the claim here is the check "returns early". Because of https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/utils/predicates.go#L358:
> 
> ```
> func hasSameLabelValue(node1, node2 *v1.Node, key string) bool {
> 	if node1.Name == node2.Name { // returns early
> 		return true
> 	}
> ...
> }
> ```
> 
> So the goal here is to ignore node N1 in the check. N1 is retrieved due to accessing `pod.Spec.NodeName` in https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/utils/predicates.go#L326. Instead, just check whether P1 does/does not violate the anti-affinity rule (produced by P2) on node N2 by checking node N2's labels (through `TopologyKey` key).
> 
> @fanhaouu is this the current issue in hand? Just rephrasing so it is clear what this issue and fix is all about.
> 
> @nikimanoledaki for awareness.

yes, master, you are right

pkg/utils/predicates.go
ingvagabund: In that case `nodeHavingExistingPod.Label` instead of `node.Labels`.

ingvagabund:
> It should be assessing whether a Pod fits on another Node based on whether another Pod on this other Node has matching antiaffinities, but it returns early, due to the following issue.

More on this. Again with let's have P1 scheduled to N1 and P2 scheduled to N2. Pods P1 and P2 have matching anti-affinities. The check here is NodeFit(P1,N2) and the claim here is the check "returns early".

Given N1 != N2 then `hasSameLabelValue` is true only when both N1 and N2 have the same `Labels[TopologyKey]` value. Meaning `hasSameLabelValue` returns at the end. @nikimanoledaki would you please elaborate more on what "returns early" means in context of https://github.com/kubernetes-sigs/descheduler/issues/1370?

ingvagabund:
Wrt. "returns early": In case `CheckPodsWithAntiAffinityExist` wants to detect that P1 is in violation with P2 on N2 while P1 and P2 are in "pod anti-affinity" relation and N2 has `Labels[TopologyKey]` (the actual value is irrelevant), `CheckPodsWithAntiAffinityExist` is expected to return `true`. Which can happen only when `hasSameLabelValue` returns true. Which is not the case when N1 (which is different from N2) leads to `hasSameLabelValue` returning false. Causing  `CheckPodsWithAntiAffinityExist` returns false, which turns into `NodeFit` not erroring on `podMatchesInterPodAntiAffinity`.

So the following case leads to `NodeFit(P1, N2)` behaving incorrectly:
1. P1 on N1, P2 on N2, P1 in pod anti-affinity with P2, N1 != N2,
1. N2 has `Labels[TopologyKey]`, N1 does not have `Labels[TopologyKey]`
1. `NodeFit(P1, N2)` returns no error instead of "pod matches inter-pod anti-affinity rule of other pod on node" since N1 (missing `Labels[TopologyKey]` causes `hasSameLabelValue` to return `false`).

nikimanoledaki:
> Given N1 != N2 then hasSameLabelValue is true only when both N1 and N2 have the same Labels[TopologyKey] value. 

This is correct - I came to the same conclusion that when there are multiple nodes, given that both pods have an anti-affinity, a valid check to validate this is if they have the same `TopologyKey`.

I wrote a draft PR a while ago that I didn't have time to finish, here, which has more logging that I was using to debug in our environment. It approached this logic in a slightly different way: https://github.com/kubernetes-sigs/descheduler/pull/1415/files However I do prefer the `bool` approach of this current PR (the one we're reviewing right now).

I also need to go over the logic and then will leave more feedback.

nikimanoledaki:
> would you please elaborate more on what "returns early" means in context of https://github.com/kubernetes-sigs/descheduler/issues/1370?

It returned early [here](https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/utils/predicates.go#L326):

```
// assuming that we have two pods, P1 and P2,
// and assuming that we have two nodes, N1 and N2, 
// where P1 is on N1 and P2 is on N2,
// the following will return early because it evaluates if P1 is on N2, but it isn't.
node, ok := nodeMap[pod.Spec.NodeName] 
```

ingvagabund:
> node, ok := nodeMap[pod.Spec.NodeName] 

Normally, the descheduler is expected to check only a pod that has been scheduled. Which implies `nodeMap[pod.Spec.NodeName] ` will always exist. Though, I can see how `CheckPodsWithAntiAffinityExist` can be functionality tested over a pod that has not been scheduled yet.

nikimanoledaki:
> the descheduler is expected to check only a pod that has been scheduled. Which implies nodeMap[pod.Spec.NodeName]  will always exist. 

Hm not sure we are seeing the same thing. Taking a step back, we know that `NodeFit` should be able to check whether P1, which is on N1, can be scheduled on another node, N2. With this anti-affinity check, it should check if there is an existing Pod on N2, P2, which has anti-affinity with P1.

Currently, the code will check whether P1 and P2 are on N2. Of course, we know that P1 is not yet on N2, so `node, ok := nodeMap[pod.Spec.NodeName]` always return `!ok` and `continue`. We don't want to check this if we are coming from `NodeFit`.

I think the issue is that this was a leftover/oversight from my part when I was refactoring this in https://github.com/kubernetes-sigs/descheduler/pull/1356 where I tried to refactor this anti-affinity check from [RemovePodsViolatingInterPodAntiAffinit](https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/framework/plugins/removepodsviolatinginterpodantiaffinity/pod_antiaffinity.go). However, that plugin was comparing P1 and P2 when they are already both on the same Node. On the other hand, this check coming from `NodeFit` should be able to check when P1 is on N1 and P2 is on N2. I hope this makes some sense 🙈

ingvagabund:
> Currently, the code will check whether P1 and P2 are on N2. Of course, we know that P1 is not yet on N2, so node, ok := nodeMap[pod.Spec.NodeName] will always return !ok and then return early (continue). Therefore, we don't want to check this if we are coming from NodeFit where P1 is not on N2 (yet).

P1 is scheduled on N1 so I wonder why you get !ok when accessing nodeMap[P1.Spec.NodeName]

nikimanoledaki:
> P1 is scheduled on N1 so I wonder why you get !ok when accessing nodeMap[P1.Spec.NodeName]

Because the Node is in fact N2!

nikimanoledaki:
Screenshot 2024-05-23 at 15 55 50

pkg/utils/predicates.go
```diff
@@ -322,11 +322,21 @@ func CheckPodsWithAntiAffinityExist(pod *v1.Pod, pods map[string][]*v1.Pod, node
...
 				for _, existingPod := range pods[namespace] {
-					if existingPod.Name != pod.Name && podMatchesTermsNamespaceAndSelector(existingPod, namespaces, selector) {
-						node, ok := nodeMap[pod.Spec.NodeName]
-						if !ok {
-							continue
+					if existingPod.Name == pod.Name || !podMatchesTermsNamespaceAndSelector(existingPod, namespaces, selector) {
+						continue
+					}
+
+					node, ok := nodeMap[pod.Spec.NodeName]
```
nikimanoledaki, line 329: There is [only one Node being passed](https://github.com/kubernetes-sigs/descheduler/pull/1395/files#diff-fc4033754cd376d78b09ff84a934296f26e774702505885309268f816ba3817cR348) when coming from NodeFit and that is the destination Node (N2). P1 is not on N2, so this will be `!ok`. We should only evaluate this if `nodeFit == false`.

fanhaouu: thank you, should using existingPod here, i have fixed it.

fanhaouu: Perhaps my pr misled ingvagabund,😂

nikimanoledaki: > should using existingPod here

Yes! I agree about reversing to make sure that existingPod is on N2 👍

nikimanoledaki (approved):
LGTM! This should solve #1395 👍

ingvagabund:
@fanhaouu can you do https://github.com/ingvagabund/descheduler/commit/c1b82d15f2de78c487dca5f347230602747fa351 instead?

ingvagabund:
@fanhaouu apologies for pushing the other commit. I mis-clicked.

fanhaouu:
> @fanhaouu apologies for pushing the other commit. I mis-clicked.

haha, let me resubmit it.

pkg/utils/predicates_test.go
```diff
@@ -1099,12 +1101,67 @@ func TestCheckPodsWithAntiAffinityExist(t *testing.T) {
...
 		},
+		{
+			name: "found pod matching pod anti-affinity when node fitting",
+			pod:  test.PodWithPodAntiAffinity(test.BuildTestPod("p1", 1000, 1000, "node1", nil), "foo", "bar"),
+			podsInNamespace: map[string][]*v1.Pod{
+				"default": {
+					test.PodWithPodAntiAffinity(test.BuildTestPod("p2", 1000, 1000, "node2", nil), "foo", "bar"),
+				},
+			},
+			nodeMap: map[string]*v1.Node{
```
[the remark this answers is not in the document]
fanhaouu: this is also not necessary. pod1 is on node1, nodeMap contains the node where pod2 is located.Here we are writing test cases for when nodeFit is true

pkg/utils/predicates.go
```diff
@@ -310,7 +310,7 @@ func CreateNodeMap(nodes []*v1.Node) map[string]*v1.Node {
 }
 
 // CheckPodsWithAntiAffinityExist checks if there are other pods on the node that the current pod cannot tolerate.
-func CheckPodsWithAntiAffinityExist(pod *v1.Pod, pods map[string][]*v1.Pod, nodeMap map[string]*v1.Node) bool {
+func CheckPodsWithAntiAffinityExist(pod *v1.Pod, pods map[string][]*v1.Pod, nodeMap map[string]*v1.Node, nodeFit bool) bool {
```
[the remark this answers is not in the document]
fanhaouu: no problem

pkg/utils/predicates.go
```diff
@@ -322,15 +322,27 @@ func CheckPodsWithAntiAffinityExist(pod *v1.Pod, pods map[string][]*v1.Pod, node
...
+						klog.V(4).InfoS("no matching PodAntiAffinity found for pod on node", "pod with anti-affinity", klog.KObj(pod)))
+						continue
+					}
+
+					nodeHavingExistingPod, ok := nodeMap[existingPod.Spec.NodeName]
+					if !ok {
+						continue
+					}
+
+					if nodeFit {
```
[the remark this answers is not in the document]
fanhaouu: If CheckPodsWithAntiAffinityExist is a common method, it will still be prone to misuse, thus triggering the same bug, without resolving any issues.

fanhaouu:
I made adjustments, to no longer pursue reuse CheckPodsWithAntiAffinityExist method, but realized podMatchesInterPodAntiAffinity method alone, at the same time by adjusting part of the method of parameter names and internal logic. @ingvagabund @nikimanoledaki 
Looking forward to a re-review and merging as soon as possible to avoid impacting users.

pkg/descheduler/node/node_test.go
```diff
@@ -795,11 +798,36 @@ func TestNodeFit(t *testing.T) {
...
+		},
+		{
+			description: "Pod doesn't match inter-pod anti-affinity rule of other pod on node, because node have no topologyKey",
+			pod:         test.PodWithPodAntiAffinity(test.BuildTestPod("p1", 1000, 1000, "node1", nil), "foo", "bar"),
+			node:        nodeNolabel,
+			podsOnNode: []*v1.Pod{
+				test.PodWithPodAntiAffinity(test.BuildTestPod("p2", 1000, 1000, node.Name, nil), "foo", "bar"),
+			},
+		},
+		{
```
ingvagabund, line 826: For completeness can you extend the tests with "Pod matches inter-pod anti-affinity rule" to have `NodeFit` produce "pod matches inter-pod anti-affinity rule of other pod on node" error?

ingvagabund:
Very nice!!! @fanhaouu thank you for your patience and improvements.
/approve
Leaving the final lgtm for @nikimanoledaki

nikimanoledaki (approved):
LGTM too! Thank you @fanhaouu for implementing this change and @ingvagabund for your suggestions and review! 🚀

PR #1396, opened 2024-05-11
KEP-753: Descheduling framework: update diagram

ingvagabund:
Update the profiles to reflect only Deschedule and Balance plugins are run and the order of first Deschedule of all profiles then Balance of all profiles.


damemi (commented):
/approve

PR #1405, opened 2024-05-19
[release 1.30] update kustomize image tags and helm versions

a7i:
last step before cutting release-1.30


ingvagabund:
Should we also update https://github.com/kubernetes-sigs/descheduler/blob/master/.github/workflows/manifests.yaml#L11?

PR #1412, opened 2024-05-21
fix helm's default deschedulerPolicy

duplabe:


a7i:
@duplabe  thank you for your contribution. LGTM but first you need to sign the CLA

PR #1416, opened 2024-05-28
use cmd context instead of using context.Background()

googs1025:
- use cmd context instead of using context.Background().
cmd *cobra.Command provided the context ctx, we should use this one provided by cmd instead of using context.Background(). We should pass the upstream ctx to the downstream.


a7i (approved):
/ok-to-test

a7i:
cc: @jklaw90

Issue #1421, opened 2024-05-29
KEP-NNNN: 


ingvagabund:
**Is your feature request related to a problem? Please describe.**

The `NodeFit` predicate was introduced to allow the descheduler to make better decisions about evictions to avoid cases where there's no feasible node for re-scheduling after a pod gets evicted. To enable the predicate, `DefaultEvictor` plugin provides an optional `nodeFit` option that each plugin can utilize. The list of existing checks has been extended over time in good faith to improve the eviction decisions. The `NodeFit` predicate currently consists of the following checks:
- a pod matches a node selector
- a pod tolerates taints
- a pod fits resource requests
- a node is unschedulable
- a pod matches inter pod anti-affinity

Some plugins adopted the `NodeFit` predicate natively through invocation of additional `PodFitsAnyOtherNode`, `PodFitsAnyNode` and `PodFitsCurrentNode` predicates built on top of `NodeFit`. Nevertheless, there are cases where it's more preferable to check only a subset of existing checks or disable the checks completely. Which is problematic for such plugins where it's impossible to fully disable the checks.

**User stories**

* Plugins like `RemovePodsViolatingNodeAffinity` or `RemovePodsViolatingNodeTaints`
  have subset of `NodeFit` checks enabled natively. These checks can not be disabled
  without disabling the corresponding plugin. Instead, as an administrator
  I'd like to disable specific checks like "a pod fits resource requests" to get
  as close as possible to disabling all `NodeFit` checks. So I can evict and detect
  pending pods and allow cluster autoscalers or other tools to reconcile the situation.
* As an administrator I'd like to configure `PodLifetime` plugin to check
  there are nodes with sufficient resources that can accept any evicted pod
  even though pod node selector does not match any node. So when there are
  too many pending pods due to node label mismatch my automation can label
  existing nodes and allocate more resources or have the multi-cluster scheduler
  reschedule my workload to a different cluster.
* As an administrator I'd like to make sure `RemovePodsViolatingInterPodAntiAffinity`
  plugin evicts pods even though there's currently no node with sufficient resources
  while respecting node affinities and taints. So the cluster autoscaler
  can scale up new nodes when too many Pending pods are observed.
* As an administrator I'd like to run a different scheduler than the default one.
  For that I might need to disable some of the existing `NodeFit` checks that
  are no longer valid or might collide with how the non-default scheduler works.
* As an AI/ML infrastructure administrator I'd like to extend available `NodeFit`
[...]

PR #1423, opened 2024-06-02
add gomaxprocs limit, return node fit error and pod QoS in advance

fanhaouu:
This PR aims to address the following three issues：
1. The default value of GOMAXPROCS is the number of CPU cores of the machine. After GOMAXPROCS is higher than the number of truly usable cores, the Go scheduler will keep switching OS threads, which will affect the performance of descheduler program;
2. There are many filter policies in the current node fit, similar to the filter plugin in the scheduler. The correct logic should be: if any plugin does not meet the conditions, it should be terminated in advance. The current check logic is incomprehensible, and it actually performs all filter check logic, which is not only time-consuming, but also very unnecessary.(https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/descheduler/node/node.go#L107)
3. If the pod's Qos has a value, return the QoS value early. For pods evicted within the descheduler, they are typically already scheduled onto nodes, and in most cases, the QoS value is already present on the pod's status. Therefore, it is unnecessary to start the determination process from scratch.

### DIFF
--- a/cmd/descheduler/descheduler.go
+++ b/cmd/descheduler/descheduler.go
@@ -19,6 +19,7 @@ package main
 import (
 	"os"
 
+	_ "go.uber.org/automaxprocs"
 	"k8s.io/component-base/cli"
 	"sigs.k8s.io/descheduler/cmd/descheduler/app"
 )
--- a/go.mod
+++ b/go.mod
@@ -12,6 +12,7 @@ require (
 	go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc v1.24.0
 	go.opentelemetry.io/otel/sdk v1.24.0
 	go.opentelemetry.io/otel/trace v1.24.0
+	go.uber.org/automaxprocs v1.5.3
 	google.golang.org/grpc v1.62.0
 	k8s.io/api v0.30.0
 	k8s.io/apimachinery v0.30.0
--- a/go.sum
+++ b/go.sum
@@ -141,6 +141,8 @@ github.com/pkg/errors v0.9.1 h1:FEBLx1zS214owpjy7qsBeixbURkuhQAwrK5UwLGTwt4=
 github.com/pkg/errors v0.9.1/go.mod h1:bwawxfHBFNV+L2hUp1rHADufV3IMtnDRdf1r5NINEl0=
 github.com/pmezard/go-difflib v1.0.0 h1:4DBwDE0NGyQoBHbLQYPwSUPoCMWR5BEzIk/f1lZbAQM=
 github.com/pmezard/go-difflib v1.0.0/go.mod h1:iKH77koFhYxTK1pcRnkKkqfTogsbg7gZNVY4sRDYZ/4=
+github.com/prashantv/gostub v1.1.0 h1:BTyx3RfQjRHnUWaGF9oQos79AlQ5k8WNktv7VGvVH4g=
+github.com/prashantv/gostub v1.1.0/go.mod h1:A5zLQHz7ieHGG7is6LLXLz7I8+3LZzsrV0P1IAHhP5U=
 github.com/prometheus/client_golang v1.16.0 h1:yk/hx9hDbrGHovbci4BY+pRMfSuuat626eFsHb7tmT8=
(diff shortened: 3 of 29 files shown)


fanhaouu:
@a7i ,master, can you help me review this pr?

a7i:
Hi @fanhaouu great contribution. Going to copy some of the maintainers for feedback as well:
/cc @jklaw90 @ingvagabund  @damemi

My feedback:
- _regarding point 1_: For `GOMAXPROCS`, as far as I know, we don't run any goroutines in Descheduler. How does this help?

- _regarding point 2_: The idea is to present all NodeFit predicates. The predicate check is not in any particular order, so returning the first one may not present the whole picture to the cluster operator. In a cluster of 35k pods / 15k deployments, a Descheduler run takes a few (single digit) seconds. Given that the shorter run interval can be a minute, I'm not convinced this optimization is worth it. What do you think?

- _regarding point 3_: I like it, that's a great change!

- overall: if you could split this into 3 PRs, I think it would make it easier to provide feedback and request changes.

fanhaouu:
> Hi @fanhaouu great contribution. Going to copy some of the maintainers for feedback as well: /cc @jklaw90 @ingvagabund @damemi
> 
> My feedback:
> 
> * _regarding point 1_: For `GOMAXPROCS`, as far as I know, we don't run any goroutines in Descheduler. How does this help?
> * _regarding point 2_: The idea is to present all NodeFit predicates. The predicate check is not in any particular order, so returning the first one may not present the whole picture to the cluster operator. In a cluster of 35k pods / 15k deployments, a Descheduler run takes a few (single digit) seconds. Given that the shortest frequency can be a minute, I'm not convinced this optimization is worth it. What do you think?
> * _regarding point 3_: I like it, that's a great change!
> * overall: if you could split this into 3 PRs, I think it would make it easier to provide feedback and request changes.



> Hi @fanhaouu great contribution. Going to copy some of the maintainers for feedback as well: /cc @jklaw90 @ingvagabund @damemi
> 
> My feedback:
> 
> * _regarding point 1_: For `GOMAXPROCS`, as far as I know, we don't run any goroutines in Descheduler. How does this help?
> * _regarding point 2_: The idea is to present all NodeFit predicates. The predicate check is not in any particular order, so returning the first one may not present the whole picture to the cluster operator. In a cluster of 35k pods / 15k deployments, a Descheduler run takes a few (single digit) seconds. Given that the shortest frequency can be a minute, I'm not convinced this optimization is worth it. What do you think?
> * _regarding point 3_: I like it, that's a great change!
> * overall: if you could split this into 3 PRs, I think it would make it easier to provide feedback and request changes.

master, thank you for your reply.

point1. At present, descheduler has some goroutine, but the number is not very large, so gomaxproc can be added or not, but the current go runtime cannot recognize the container environment, I think it is better to add it. For example, jvm runtime has long supported container environments; (As descheduler in our company has a lot of goroutine, the performance can be improved significantly after adding this restriction, so I didn't remove this submission from the pr)

point2. Because the default loop time is 5min, this time can be adjusted. The larger the cluster, the longer the nodeFit time, so the full loop time can easily exceed 5min. What if the user sets the loop time to 1 minute? Comparing one by one I think is very, very unnecessary, too time-consuming, we should be the same as the filter policy in the scheduler.

In short, if we can make the program better, faster, without affecting any functionality, I think it makes a lot of sense.

We are also developing a descheduler cache mechanism internally. Currently, descheduler pulls weight data for filtering at policy runtime, which can actually cache a good part of the data in advance, just like the cache in the scheduler. If our company is stable online, I will also contribute to our descheduler community. We can review it together then.

ingvagabund:
As Amir mentioned would you please break the PR into three separate PRs? Some of the suggested changes deserves a dedicated discussion. Wrt. NodeFit I am in the process of composing a KEP: https://github.com/kubernetes-sigs/descheduler/issues/1421. This sounds like a good use case to include in the proposal.

fanhaouu:
hi, masters, i have split this into 3 PRs, looking forward to your feedback:

1. https://github.com/kubernetes-sigs/descheduler/pull/1434
2. https://github.com/kubernetes-sigs/descheduler/pull/1435
3. https://github.com/kubernetes-sigs/descheduler/pull/1436

/cc @a7i @jklaw90 @ingvagabund @damemi

a7i:
/close

it's already split into 3 PRs

Issue #1419, opened 2024-05-29
`failed to convert Descheduler minor version to float` on start


der-eismann:
**What version of descheduler are you using?**

descheduler version: 0.30.0

**Does this issue reproduce with the latest release?**

yes

**Which descheduler CLI options are you using?**

`--kubeconfig container/config-staging --policy-config-file container/policy.yaml`

**Please provide a copy of your descheduler policy config file**

```yaml
apiVersion: "descheduler/v1alpha1"
kind: "DeschedulerPolicy"
nodeSelector: node.kubernetes.io/role=worker
evictLocalStoragePods: true
evictSystemCriticalPods: false
ignorePvcPods: false
strategies:
  "RemovePodsViolatingTopologySpreadConstraint":
    enabled: true
    params:
      includeSoftConstraints: true # soft = 'ScheduleAnyway' constraint
      nodeFit: true
      thresholdPriority: 200100 # only evict pods with 'default' pod priority
```

**What k8s version are you using (`kubectl version`)?**

<details><summary><code>kubectl version</code> Output</summary><br><pre>
$ kubectl version
Client Version: v1.30.1
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.30.0-eks-036c24b

</pre></details>

**What did you do?**

Start the descheduler

**What did you expect to see?**

No warning

**What did you see instead?**

This on every start:
```
W0529 16:12:11.137539  715219 descheduler.go:248] failed to convert Descheduler minor version to float
```

I don't know why the `GitVersion` is used to parse [here](https://github.com/kubernetes-sigs/descheduler/blob/release-1.30/pkg/descheduler/descheduler.go#L290-L293) - it has a format of `v20240519-v0.30.0`. According to the docs of `utilversion.ParseGeneric`:
> ParseGeneric parses a "generic" version string. The version string must consist of two or more dot-separated numeric fields (the first of which can't have leading zeroes), *followed* by arbitrary uninterpreted data (which need not be separated from the final numeric field by punctuation). For convenience, leading and trailing whitespace is ignored, and the version can be preceded by the letter "v". See also ParseSemantic.

The version it's trying to parse however is this (obtained from `descheduler version` - also see the double `v`):
```
v20240519-v0.30.0
```

[...]

PR #1430, opened 2024-06-07
fix: minor version parsing in version compatibility check

a7i:
fixes in #1419 

/cc @damemi

### DIFF
--- a/pkg/descheduler/descheduler.go
+++ b/pkg/descheduler/descheduler.go
@@ -18,9 +18,9 @@ package descheduler
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"math"
+	"strconv"
 	"time"
 
 	"go.opentelemetry.io/otel/attribute"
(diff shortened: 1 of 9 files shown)


pkg/version/version_test.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"testing"
+)
+
+func TestGet(t *testing.T) {
+	tests := []struct {
+		name      string
+		version   string
+		wantMajor string
+		wantMinor string
```
damemi, line 12: could you just have one version object and check that it's equal to the expected? The diff in the output of that check will show whether it's the major or minor that's wrong

a7i: a lot of the fields need to be set to do a reflectEqual, let me push a separate commit to get your feedback

a7i: let me know what you think

a7i: any feedback?

pkg/descheduler/descheduler.go
```diff
@@ -276,30 +277,30 @@ func Run(ctx context.Context, rs *options.DeschedulerServer) error {
...
-	serverVersion, err := utilversion.ParseSemantic(serverVersionInfo.String())
+	kubeServerVersion, err := utilversion.ParseSemantic(kubeServerVersionInfo.String())
 	if err != nil {
 		return errors.New("failed to parse Kubernetes server version")
 	}
 
-	deschedulerVersion, err := utilversion.ParseGeneric(versionInfo.GitVersion)
+	deschedulerMinor, err := strconv.ParseFloat(deschedulerVersionInfo.Minor, 64)
 	if err != nil {
 		return errors.New("failed to convert Descheduler minor version to float")
```
ingvagabund, line 293: Could we also print `deschedulerVersionInfo.Minor` in the error message? So users can see the invalid version string.

pkg/descheduler/descheduler.go
```diff
@@ -276,30 +277,30 @@ func Run(ctx context.Context, rs *options.DeschedulerServer) error {
...
+func validateVersionCompatibility(discovery discovery.DiscoveryInterface, deschedulerVersionInfo version.Info) error {
+	kubeServerVersionInfo, err := discovery.ServerVersion()
 	if err != nil {
 		return errors.New("failed to discover Kubernetes server version")
 	}
 
-	serverVersion, err := utilversion.ParseSemantic(serverVersionInfo.String())
+	kubeServerVersion, err := utilversion.ParseSemantic(kubeServerVersionInfo.String())
 	if err != nil {
 		return errors.New("failed to parse Kubernetes server version")
```
ingvagabund, line 288: Could we also print `kubeServerVersionInfo.String()` in the error message? So users can see the invalid version string.

PR #1432, opened 2024-06-08
chore: reduce repetition of go versions

a7i:
when we cut a new release of descheduler, we have to update the go version in multiple places dquote> which presents an opportunity to miss updating one.

### DIFF
--- a/.github/workflows/helm.yaml
+++ b/.github/workflows/helm.yaml
@@ -35,7 +35,7 @@ jobs:
 
       - uses: actions/setup-go@v3
         with:
-          go-version: '1.22.3'
+          go-version-file: 'go.mod'
 
       - name: Set up chart-testing
         uses: helm/chart-testing-action@v2.2.1
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,7 @@ HAS_GOLANGCI := $(shell ls _output/bin/golangci-lint 2> /dev/null)
 GOFUMPT_VERSION := v0.4.0
 HAS_GOFUMPT := $(shell command -v gofumpt 2> /dev/null)
 
-GO_VERSION := $(shell sed -En 's/^go (.*)$$/\1/p' go.mod)
+GO_VERSION := $(shell (command -v jq > /dev/null && (go mod edit -json | jq -r .Go)) || (sed -En 's/^go (.*)$$/\1/p' go.mod))
 
 # REGISTRY is the container registry to push
 # into. The default is to push to the staging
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -134,7 +136,7 @@ gen:
 	./hack/update-docs.sh
 
 gen-docker:
-	$(CONTAINER_ENGINE) run --entrypoint make -it -v $(CURRENT_DIR):/go/src/sigs.k8s.io/descheduler -w /go/src/sigs.k8s.io/descheduler golang:1.22.3 gen
```
ingvagabund, line 137: What is `gen-docker` used for? Do we use it? Also, the golang image is missing a registry. Should `ocker.io/library/golang` be used as the default?

.github/workflows/helm.yaml
```diff
@@ -35,7 +35,7 @@ jobs:
 
       - uses: actions/setup-go@v3
         with:
-          go-version: '1.22.3'
+          go-version-file: 'go.mod'
```
ingvagabund, line 38: Nice: https://github.com/actions/setup-go?tab=readme-ov-file#getting-go-version-from-the-gomod-file

Makefile
a7i: some users reported having issues running gen, so this was a contribution to run it in docker.

it was added here https://github.com/kubernetes-sigs/descheduler/pull/1189/files

a7i: I'm not sure if it is still used by anyone

Makefile
```diff
@@ -32,6 +32,8 @@ HAS_GOLANGCI := $(shell ls _output/bin/golangci-lint 2> /dev/null)
 GOFUMPT_VERSION := v0.4.0
 HAS_GOFUMPT := $(shell command -v gofumpt 2> /dev/null)
 
+GO_VERSION := $(shell sed -En 's/^go (.*)$$/\1/p' go.mod)
```
ingvagabund, line 35: I like `go mod edit -json | jq -r .Go`. Though, it introduces a dependency on jq.

ingvagabund: From https://go.dev/ref/mod#go-mod-file-go:
```
GoDirective = "go" GoVersion newline .
GoVersion = string | ident .  /* valid release version; see above */
```

Hopefully the syntax does not change in the future. In the worst case users will fail to pull the golang image.

Can you test of `jq` existence and run `go mod edit -json | jq -r .Go`. Otherwise, fall back to the current line?

a7i: seems like unnecessary complexity for a local make command, no?

let me see what other sig projects do

a7i: no real standard. pushed up a commit to apply your suggestion. let me know what you think

PR #1434, opened 2024-06-09
add GOMAXPROCS limit

fanhaouu:
This PR aims to address the following issue：

The default value of GOMAXPROCS is the number of CPU cores of the machine. After GOMAXPROCS is higher than the number of truly usable cores, the Go scheduler will keep switching OS threads, which will affect the performance of descheduler program.

At present, descheduler has some goroutine, but the number is not very large, so gomaxproc can be added or not, but the current go runtime cannot recognize the container environment, I think it is better to add it. For example, jvm runtime has long supported container environments.

### DIFF
--- a/cmd/descheduler/descheduler.go
+++ b/cmd/descheduler/descheduler.go
@@ -19,6 +19,7 @@ package main
 import (
 	"os"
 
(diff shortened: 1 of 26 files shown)


fanhaouu:
image

PR #1435, opened 2024-06-09
return pod qos in advance

fanhaouu:
If the pod's Qos has a value, return the QoS value early. For pods evicted within the descheduler, they are typically already scheduled onto nodes, and in most cases, the QoS value is already present on the pod's status. Therefore, it is unnecessary to start the determination process from scratch.

### DIFF
--- a/pkg/utils/qos.go
+++ b/pkg/utils/qos.go
@@ -20,6 +20,10 @@ func isSupportedQoSComputeResource(name v1.ResourceName) bool {
 // A pod is guaranteed only when requests and limits are specified for all the containers and they are equal.
 // A pod is burstable if limits and requests do not match across all containers.
 func GetPodQOS(pod *v1.Pod) v1.PodQOSClass {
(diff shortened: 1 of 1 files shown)


pkg/utils/qos.go
```diff
@@ -20,6 +20,10 @@ func isSupportedQoSComputeResource(name v1.ResourceName) bool {
 // A pod is guaranteed only when requests and limits are specified for all the containers and they are equal.
 // A pod is burstable if limits and requests do not match across all containers.
 func GetPodQOS(pod *v1.Pod) v1.PodQOSClass {
+	if len(pod.Status.QOSClass) != 0 {
```
a7i, line 23: slightly different variation than kubernetes/kubernete, but close enough: https://github.com/kubernetes/kubernetes/blob/9d63e575f8482b857ef328325d2107df840b9deb/pkg/apis/core/helper/qos/qos.go#L36-L38

a7i (approved):
/lgtm

PR #1436, opened 2024-06-09
return node fit error in advance

fanhaouu:
There are many filter policies in the current node fit, similar to the filter plugin in the scheduler. The correct logic should be: if any plugin does not meet the conditions, it should be terminated in advance. The current check logic is incomprehensible, and it actually performs all filter check logic, which is not only time-consuming, but also very unnecessary.(https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/descheduler/node/node.go#L107)

Because the default loop time is 5min, this time can be adjusted. The larger the cluster, the longer the nodeFit time, so the full loop time can easily exceed 5min. What if the user sets the loop time to 1 minute? Comparing one by one I think is very, very unnecessary, too time-consuming, we should be the same as the filter policy in the scheduler.

When checking the pod at https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/framework/plugins/defaultevictor/defaultevictor.go#L67, it also returns early

### DIFF
--- a/pkg/descheduler/node/node.go
+++ b/pkg/descheduler/node/node.go
@@ -18,19 +18,23 @@ package node
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"sync/atomic"
 
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
-	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	clientset "k8s.io/client-go/kubernetes"
 	listersv1 "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog/v2"
 	podutil "sigs.k8s.io/descheduler/pkg/descheduler/pod"
 	"sigs.k8s.io/descheduler/pkg/utils"
 )
+
+const workersCount = 100
 
 // ReadyNodes returns ready nodes irrespective of whether they are
 // schedulable or not.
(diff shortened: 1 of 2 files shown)


ingvagabund:
Or, we could run the checks in parallel. The question is whether it's useful to notify about all individual checks errors. If the goal here is to reduce the computation time, then parallelization works better.

fanhaouu:
> Or, we could run the checks in parallel. The question is whether it's useful to notify about all individual checks errors. If the goal here is to reduce the computation time, then parallelization works better.

thank you, master, I just added the parallel logic. I forgot to submit the code for parallel processing earlier.

ingvagabund:
>  I just added the parallel logic.

This is another parallelization. Originally, I mean to parallelize internal of `NodeFit` itself. E.g. the individual checks. If we parallelize on the level of PodFitsAnyNode/PodFitsAnyOtherNode we need a different discussion.

ingvagabund:
@fanhaouu in my proposal I'd like to allow users to extend available nodeFit checks through custom plugins. For which we need to make sure none of the individual checks modifies pod/node arguments. Though, I can make a note in the proposal to revisit the parallelization later so I don't block this PR.

pkg/descheduler/node/node.go
```diff
@@ -104,90 +108,109 @@ func IsReady(node *v1.Node) bool {
...
+		err := NodeFit(nodeIndexer, pod, node)
+		if err == nil {
 			klog.V(4).InfoS("Pod fits on node", "pod", klog.KObj(pod), "node", klog.KObj(node))
-			return true
+			atomic.AddInt32(&filteredLen, 1)
+			cancel()
+		} else {
+			klog.V(4).InfoS("Pod does not fit on node", "pod", klog.KObj(pod), "node", klog.KObj(node), "err", err.Error())
 		}
-		klog.V(4).InfoS("Pod does not fit on any node",
```
ingvagabund, line 173: This error message is actually misleading. Thx for fixing it.

ingvagabund (commented):
With the new parallelization would you please create new test cases for both `PodFitsAnyOtherNode` and  `PodFitsAnyNode` to make sure each node is processed properly? E.g. run `PodFitsAnyOtherNode` with a list of e.g. 4 nodes:
- 1st and 3rd nodes do not fit a pod
-  2nd and 4th nodes do not fit a pod

To make sure the parallelization works as expected.

fanhaouu:
> With the new parallelization would you please create new test cases for both `PodFitsAnyOtherNode` and `PodFitsAnyNode` to make sure each node is processed properly? E.g. run `PodFitsAnyOtherNode` with a list of e.g. 4 nodes:
> 
> * 1st and 3rd nodes do not fit a pod
> * 2nd and 4th nodes do not fit a pod
> 
> To make sure the parallelization works as expected.

ok, master

pkg/descheduler/node/node.go
```diff
@@ -104,90 +108,109 @@ func IsReady(node *v1.Node) bool {
...
+		node := nodes[i]
 		// Skip node pod is already on
 		if node.Name == pod.Spec.NodeName {
-			continue
+			return
 		}
-
-		errors := NodeFit(nodeIndexer, pod, node)
-		if len(errors) == 0 {
+		err := NodeFit(nodeIndexer, pod, node)
```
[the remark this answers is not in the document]
fanhaouu: There should be no methods to modify nodes or pods in NodeFit. Currently, there is no modification logic, and even if there were, it should use deepCopy. If we allow users to customize and extend methods in NodeFit in the future, we should document this clearly.

pkg/descheduler/node/node.go
```diff
@@ -104,90 +108,96 @@ func IsReady(node *v1.Node) bool {
...
 
-// PodFitsAnyOtherNode checks if the given pod will fit any of the given nodes, besides the node
-// the pod is already running on. The predicates used to determine if the pod will fit can be found in the NodeFit function.
-func PodFitsAnyOtherNode(nodeIndexer podutil.GetPodsAssignedToNodeFunc, pod *v1.Pod, nodes []*v1.Node) bool {
-	for _, node := range nodes {
-		// Skip node pod is already on
-		if node.Name == pod.Spec.NodeName {
-			continue
-		}
+func podFitsNodes(nodeIndexer podutil.GetPodsAssignedToNodeFunc, pod *v1.Pod, nodes []*v1.Node, filter func(pod *v1.Pod, node *v1.Node) bool) bool {
```
ingvagabund, line 150: s/filter/excludeFilter/

pkg/descheduler/node/node_test.go
```diff
@@ -716,6 +716,152 @@ func TestPodFitsAnyOtherNode(t *testing.T) {
...
+				test.BuildTestNode(nodeNames[3], 8000, 8*1000*1000*1000, 12, func(node *v1.Node) {
+					node.ObjectMeta.Labels = map[string]string{
+						nodeLabelKey: nodeLabelValue,
+					}
+
+					node.Status.Allocatable[v1.ResourceEphemeralStorage] = *resource.NewQuantity(1000*1000*1000*1000, resource.DecimalSI)
+				}),
+			},
+			podsOnNodes: []*v1.Pod{
+				test.BuildTestPod("3-core-pod", 2000, 4*1000*1000*1000, nodeNames[1], func(pod *v1.Pod) {
```
ingvagabund, line 852: > fourth node is the one where the pod is located

Based on the test description this needs to be `nodeNames[3]`

ingvagabund (commented):
Given there's `podFitsNodes` function would you please create a new test specifically for testing the parallelization? Something like this:
```go
nodesTraversed := make(map[string]struct{})
...
podFitsNodes(
	nodeIndexer,
	pod,
	nodes,
	func(pod *v1.Pod, node *v1.Node) bool {
		nodesTraversed[node.Name] = struct{}{}
		return false
	},
)
...
// checking all four nodes have been processed/read.
for _, nodeName := range nodeNames {
	if _, exists := nodesTraversed[nodeName]; !exists {
		t.Failf("Node %v was not proccesed", nodeName)
	}
}
```

To make sure every node is checked. With this the new tests are no longer needed.

fanhaouu:
> Given there's `podFitsNodes` function would you please create a new test specifically for testing the parallelization? Something like this:
> 
> ```go
> nodesTraversed := make(map[string]struct{})
> ...
> podFitsNodes(
> 	nodeIndexer,
> 	pod,
> 	nodes,
> 	func(pod *v1.Pod, node *v1.Node) bool {
> 		nodesTraversed[node.Name] = struct{}{}
> 		return false
> 	},
> )
> ...
> // checking all four nodes have been processed/read.
> for _, nodeName := range nodeNames {
> 	if _, exists := nodesTraversed[nodeName]; !exists {
> 		t.Failf("Node %v was not proccesed", nodeName)
> 	}
> }
> ```
> 
> To make sure every node is checked. With this the new tests are no longer needed.

ok, thank you, master

PR #1441, opened 2024-06-18
refactor some methods in e2e test

googs1025:
/kind cleanup

- Use outer ctx instead of `context.TODO()`
- use `wait.PollUntilContextTimeout` instead of `wait.PollImmediate`


a7i (approved):
thank you 🏆 

/approve

PR #1443, opened 2024-06-19
fix: helm leader-election typo to `resourceNamespace`

a7i:
reported issue: https://kubernetes.slack.com/archives/C09TP78DV/p1718633429179519

Test using `leaderElection.resourceNamescape`
```
helm template deschduler ./descheduler -f ./descheduler/values.yaml --set kind=Deployment --set=leaderElection.enabled=true --set=leaderElection.resourceNamescape=foo

Error: execution error at (descheduler/templates/deployment.yaml:62:16): 'leaderElection.resourceNamescape' is deprecated. instead use 'leaderElection.resourceNamespace'
```

Test using `leaderElection.resourceNamespace`
```
helm template deschduler ./descheduler -f ./descheduler/values.yaml --set kind=Deployment --set=leaderElection.enabled=true --set=leaderElection.resourceNamespace=foo | grep leader-elect-resource-namespace

            - --leader-elect-resource-namespace=foo

### DIFF
--- a/charts/descheduler/templates/_helpers.tpl
+++ b/charts/descheduler/templates/_helpers.tpl
@@ -87,8 +87,10 @@ Leader Election
 {{- if .Values.leaderElection.resourceName }}
 - --leader-elect-resource-name={{ .Values.leaderElection.resourceName }}
 {{- end }}
-{{- if .Values.leaderElection.resourceNamescape }}
-- --leader-elect-resource-namespace={{ .Values.leaderElection.resourceNamescape }}
+{{/* resource namespace value starts with a typo so keeping resourceNamescape for backwards compatibility */}}
+{{- $resourceNamespace := default .Values.leaderElection.resourceNamespace .Values.leaderElection.resourceNamescape -}}
+{{- if $resourceNamespace -}}
+- --leader-elect-resource-namespace={{ $resourceNamespace }}
 {{- end -}}
 {{- end }}
 {{- end }}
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: charts/descheduler/templates/_helpers.tpl [hunk 1]
```diff
@@ -87,8 +87,10 @@ Leader Election
 {{- if .Values.leaderElection.resourceName }}
 - --leader-elect-resource-name={{ .Values.leaderElection.resourceName }}
 {{- end }}
-{{- if .Values.leaderElection.resourceNamescape }}
-- --leader-elect-resource-namespace={{ .Values.leaderElection.resourceNamescape }}
+{{- if .Values.leaderElection.resourceNamespace }}
+- --leader-elect-resource-namespace={{ .Values.leaderElection.resourceNamespace }}
+{{- else if .Values.leaderElection.resourceNamescape }}
+{{- fail "'leaderElection.resourceNamescape' is deprecated. instead use 'leaderElection.resourceNamespace'" }}
```
ingvagabund, line 93: We should not fail when a param is deprecated. Would it make more sense to print a warning but still take `resourceNamescape into account?

a7i: I don't know of a way to print warning in helm.

I think this is a trivial change to prompt users to correct. we could continue to support both indefinitely

ingvagabund: `resourceNamespace` was introduced in Feb 2022. No way to tell how many users rely on that parameter. I am for deprecating `resourceNamespace` in the release notes, supporting both for 2 release and removing the deprecated param then.

a7i: supporting both now, let me know what you think

PR #1446, opened 2024-06-22
fix: return the unmatched cases first

googs1025:
We should return the unmatched cases first,and leave the filter method for last, because it is computationally expensive.


a7i (approved):
/lgtm

/assign @seanmalloy

PR #1452, opened 2024-06-28
fix: indexer cache error when default evictor is re-initialized

### LINKED ISSUE #1440: Unable to create a profile" err="unable to build DefaultEvictor plugin
Hi!

I'm facing an error in Descheduler pod logs:

```
"unable to create a profile" err="unable to build DefaultEvictor plugin: unable to initialize \"DefaultEvictor\" plugin: indexer conflict: map[metadata.ownerReferences:{}]" profile="Descheduler"
```

I'm installing Descheduler via Helm Chart v0.30.1 with Deployment mode.

My policy config looks like this:
```
deschedulerPolicy:
  profiles:
    - name: "Descheduler"
      pluginConfig:
        - name: "DefaultEvictor"
          args:
            evictLocalStoragePods: false
            evictSystemCriticalPods: false
            ignorePvcPods: false
            evictFailedBarePods: false
            nodeFit: true
            minReplicas: 2
        - name: "RemoveDuplicates"
        - name: "LowNodeUtilization"
          args:
            thresholds:
              cpu : 30
              memory: 30
              pods: 20
            targetThresholds:
              cpu : 80
              memory: 80
              pods: 50
        - name: "HighNodeUtilization"
          args:
            thresholds:
              cpu : 80
              memory: 80
              pods: 50
        - name: "RemovePodsViolatingInterPodAntiAffinity"
        - name: "RemovePodsViolatingNodeTaints"
        - name: "RemovePodsViolatingNodeAffinity"
          args:
            nodeAffinityType:
              - "requiredDuringSchedulingIgnoredDuringExecution"
        - name: "RemovePodsViolatingTopologySpreadConstraint"
          args:
            topologyBalanceNodeFit: true
            constraints:
              - "DoNotSchedule"
              - "ScheduleAnyway"
      plugins:
        balance:
          enabled:
            - "RemoveDuplicates"
            - "LowNodeUtilization"
            - "HighNodeUtilization"
            - "RemovePodsViolatingTopologySpreadConstraint"
        deschedule:
          enabled:
            - "RemovePodsViolatingInterPodAntiAffinity"
            - "RemovePodsViolatingNodeAffinity"
            - "RemovePodsViolatingNodeTaints"
```

I also see that when pod starts, everything works fine until CronJob finishes... Then I'm starting to get these errors...

Maybe I'm configuring something wrong? 🤔

a7i:
It appears that DefaultEvictor is re-initialized, which causes a conflict issue when adding the same indexer again.

closes #1440

### DIFF
--- a/pkg/framework/plugins/defaultevictor/defaultevictor.go
+++ b/pkg/framework/plugins/defaultevictor/defaultevictor.go
@@ -244,6 +244,15 @@ func (d *DefaultEvictor) Filter(pod *v1.Pod) bool {
 
 func getPodIndexerByOwnerRefs(indexName string, handle frameworktypes.Handle) (cache.Indexer, error) {
 	podInformer := handle.SharedInformerFactory().Core().V1().Pods().Informer()
+	indexer := podInformer.GetIndexer()
+
+	// do not reinitialize the indexer, if it's been defined already
+	for name := range indexer.GetIndexers() {
+		if name == indexName {
+			return indexer, nil
+		}
+	}
+
 	if err := podInformer.AddIndexers(cache.Indexers{
 		indexName: func(obj interface{}) ([]string, error) {
 			pod, ok := obj.(*v1.Pod)
@@ -257,6 +266,5 @@ func getPodIndexerByOwnerRefs(indexName string, handle frameworktypes.Handle) (c
 		return nil, err
 	}
 
-	indexer := podInformer.GetIndexer()
 	return indexer, nil
 }
--- a/pkg/framework/plugins/defaultevictor/defaultevictor_test.go
+++ b/pkg/framework/plugins/defaultevictor/defaultevictor_test.go
@@ -15,6 +15,7 @@ package defaultevictor
 
 import (
 	"context"
+	"fmt"
 	"testing"
 
 	v1 "k8s.io/api/core/v1"
@@ -31,6 +32,19 @@ import (
 	"sigs.k8s.io/descheduler/test"
 )
 
+type testCase struct {
+	description             string
+	pods                    []*v1.Pod
+	nodes                   []*v1.Node
+	evictFailedBarePods     bool
+	evictLocalStoragePods   bool
+	evictSystemCriticalPods bool
+	priorityThreshold       *int32
+	nodeFit                 bool
+	minReplicas             uint
+	result                  bool
+}
+
 func TestDefaultEvictorPreEvictionFilter(t *testing.T) {
 	n1 := test.BuildTestNode("node1", 1000, 2000, 13, nil)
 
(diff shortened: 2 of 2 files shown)


a7i:
/cc @ingvagabund


any ideas on how to make this better? to be honest, I'm not sure why DefaultEvictor is getting re-initialized. But perhaps this is possible in case of multiple profiles

ingvagabund:
> I'm not sure why DefaultEvictor is getting re-initialized. But perhaps this is possible in case of multiple profiles

Yeah. Each profile gets initialized separately. The same shared informer, a new instance of the default evictor.

ingvagabund:
Some thoughts:
- what if two plugins register the same indexer? We should prefix each indexer with a plugin name (some time in the future)
- should we register generic indexer so the default plugins do not need to register it on their own? This will not resolve the current issue. Yet, it will get rid of the error
- should we create a separate shared informer for each profile? We don't limit the number of profiles. We probably should, e.g. to 100 or 1000 and extend later when needed?

ingvagabund:
The current implementation creates a new profile in each descheduling cycle. Yet, the created indexer never gets removed from the shared informer. Making every other descheduling cycle to error.

ingvagabund:
Worth checking if the code can be cleanly refactored to initialize each plugin only once per the whole descheduler run. The primary motivation for re-initialized each plugin was to start clean. If we can guarantee no plugin will keep a state then this will be a nice improvement to the code base.

a7i:
>what if two plugins register the same indexer? We should prefix each indexer with a plugin name (some time in the future)

Why is that necessarily? This is no plugin logic inside the shared informed. And ideally we _share_ the shared informer.


>should we register generic indexer so the default plugins do not need to register it on their own? This will not resolve the current issue. Yet, it will get rid of the error

we certainly could, but we then need to compare it against the current implementation. I personally prefer this PR but open to feedback from others. For example, no need to initialize the shared informer for policies that do not need minReplicas check.

>should we create a separate shared informer for each profile? We don't limit the number of profiles. We probably should, e.g. to 100 or 1000 and extend later when needed?

is there a reason for this? ideally we share it to reduce memory footprint.

a7i:
>
Worth checking if the code can be cleanly refactored to initialize each plugin only once per the whole descheduler run. The primary motivation for re-initializing each plugin was to start clean. If we can guarantee no plugin will keep a state then this will be a nice improvement to the code base.


I think the issue is that the params/config for each can be different. Correct?

ingvagabund:
We could inject the args as "data". Assuming the args do not interfere with a plugin initialization. Though, that will change the plugin API.

ingvagabund:
> I think the issue is that the params/config for each can be different. Correct?

You are right. That was another motivation for it.

ingvagabund:
Note: We share a single shared informer through all the plugins. As long as the plugins are part of this repository we have a full control over how each indexer gets created. We don't have such control over custom plugins. Thus, we need to rely on good citizenship principles and assume anyone assembling their own deschedulers from the descheduling framework will take the extra step of making sure all the extra plugins create their indexes properly. Plugin owners can already define as many indexers as they like without following any good practices. The same holds for updating items in either of the informer caches without making a deep copy.

PR #1454, opened 2024-07-02
test: generate uid when building a pod

ingvagabund:
UID is an integral part of every pod.
Even though we don't test for UID in any of the existing test cases, future test cases might rely on it.


a7i (approved):
/approve

PR #1455, opened 2024-07-06
feat: pod evictor options

a7i:
when we add a new option to PodEvictor, it typically ends up touch dozens of files because it requires adding a new arg to the initializer. this is the first step in reducing the number of initializer args

### DIFF
--- a/pkg/descheduler/evictions/options.go
+++ b/pkg/descheduler/evictions/options.go
@@ -1,0 +1,45 @@
+package evictions
+
+import (
+	policy "k8s.io/api/policy/v1"
+)
+
+type Options struct {
+	policyGroupVersion         string
+	dryRun                     bool
+	maxPodsToEvictPerNode      *uint
+	maxPodsToEvictPerNamespace *uint
+	metricsEnabled             bool
+}
+
+// NewOptions returns an Options with default values.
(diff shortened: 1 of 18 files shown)


pkg/descheduler/evictions/options.go
```diff
@@ -0,0 +1,45 @@
...
+	dryRun                     bool
+	maxPodsToEvictPerNode      *uint
+	maxPodsToEvictPerNamespace *uint
+	metricsEnabled             bool
+}
+
+// NewOptions returns an Options with default values.
+func NewOptions() *Options {
+	return &Options{
+		policyGroupVersion: policy.SchemeGroupVersion.String(),
```
ingvagabund, line 18: Just a though: I wonder whether there's ever going to be any other version than `policy/v1`. We detect the version at start up. We might as well drop it and assume `policy/v1` is always enabled. Re-introduce the ability to configure the version later.

ingvagabund: https://kubernetes.slack.com/archives/C18NZM5K9/p1720259403385159

a7i: That sounds good, thanks for starting that!

PR #1451, opened 2024-06-28
The descheduler supports limiting the total number of pods evicted per rescheduling cycle

zhifei92:
The MaxNoOfPodsToEvictTotal serves to prevent scenarios where an unintended issue, such as a software bug or incorrect usage of Pod Disruption Budgets (PDBs), could result in a substantial number of pods being evicted simultaneously, potentially compromising the availability of services.

Fixes #1449

PR #1456, opened 2024-07-06
PodEvictor: turn an exceeded limit into an error

ingvagabund:
When checking for node limit getting exceeded the pod eviction never fails. Thus, ignoring the metric reporting when a pod fails to be evicted due to node limit constrains.

The error also allows plugin to react on other limits getting exceeded. E.g. the limit on the number of pods evicted per namespace.

To allow other limits (e.g. as https://github.com/kubernetes-sigs/descheduler/pull/1451) to be easily introduced.
Also, resolves https://github.com/kubernetes-sigs/descheduler/pull/1368.


a7i (approved):
/approve

nice refactoring and changes 🏆

PR #1459, opened 2024-07-08
descheduler_test.go refactoring

ingvagabund:
- initPluginRegistry as a single way to register all plugins in testing
- Use v1alpha2 descheduling policy
- Set OwnerReferences through GetReplicaSetOwnerRefList
- Define initDescheduler for further use: Can be used by other tests executing individual descheduling cycle explicitly.

### DIFF
--- a/pkg/descheduler/descheduler_test.go
+++ b/pkg/descheduler/descheduler_test.go
@@ -106,9 +106,7 @@ func TestTaintsUpdated(t *testing.T) {
 	n2 := test.BuildTestNode("n2", 2000, 3000, 10, nil)
 
 	p1 := test.BuildTestPod(fmt.Sprintf("pod_1_%s", n1.Name), 200, 0, n1.Name, nil)
-	p1.ObjectMeta.OwnerReferences = []metav1.OwnerReference{
-		{},
-	}
+	p1.ObjectMeta.OwnerReferences = test.GetReplicaSetOwnerRefList()
 
 	client := fakeclientset.NewSimpleClientset(n1, n2, p1)
 	eventClient := fakeclientset.NewSimpleClientset(n1, n2, p1)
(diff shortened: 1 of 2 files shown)


pkg/descheduler/descheduler_test.go
```diff
@@ -29,45 +27,120 @@ import (
...
+	sharedInformerFactory := informers.NewSharedInformerFactoryWithOptions(rs.Client, 0, informers.WithTransform(trimManagedFields))
+	eventBroadcaster, eventRecorder := utils.GetRecorderAndBroadcaster(ctx, client)
+
+	descheduler, err := newDescheduler(rs, internalDeschedulerPolicy, "v1", eventRecorder, sharedInformerFactory)
+	if err != nil {
+		eventBroadcaster.Shutdown()
+		t.Fatalf("Unable to create a descheduler instance: %v", err)
+	}
+
+	ctx, cancel := context.WithCancel(ctx)
```
a7i, line 121: returning cancel is an odd signature, perhaps pass in the cancel context?

a7i (approved):
/approve

pkg/descheduler/descheduler_test.go
ingvagabund: The function still needs to return the event broadcaster shutdown. Are you suggesting to find a different way of canceling the context and calling the shutdown?

a7i: ah true. we can leave it as-is then

jklaw90: could create the ctx with cancel at a higher level and pass it in here? 
the broadcaster should shutdown on ctx cancelation?
```StartRecordingToSink(stopCh <-chan struct{})```

ingvagabund: That works, thx :) https://github.com/kubernetes-sigs/descheduler/pull/1460

PR #1458, opened 2024-07-08
pod evictor: make it thread safe

ingvagabund:
Currently, all the plugins are run in a sequence.
No plugin executes evictions in parallel within.
Yet, there's no guarantee a future plugin (e.g. a custom one) will not attemp to evict pods in parallel.

PR #1466, opened 2024-07-14
Introduce RequestEviction feature for evicting pods in background (KEP-1397)

ingvagabund:
Implementation of https://github.com/kubernetes-sigs/descheduler/pull/1354.

When the feature is enabled each pod with `descheduler.alpha.kubernetes.io/request-evict-only` annotation will have the eviction API error examined for a specific error code/reason and message. If matched eviction of such a pod will be interpreted as initiation of an eviction in background.

The first version of the feature does not provide any way to configure a timeout for pending evictions in background. Neither a timeout for how often a pending evictions in background are checked for timeouts. Both set to 10 minutes by default. The default values can be adjusted based on provided feedback.

TODO:
- [x] evictions in background are respested
- [x] unit tests
- [x] feature gating implemented (`EvictionsInBackground` as a new feature gate)
- [x] e2e integration with kubevirt
- [x] feature configuration (currently none)

Refactorings before this PR can be merged:
- https://github.com/kubernetes-sigs/descheduler/pull/1474
- https://github.com/kubernetes-sigs/descheduler/pull/1469
- https://github.com/kubernetes-sigs/descheduler/pull/1472
- https://github.com/kubernetes-sigs/descheduler/pull/1460
- https://github.com/kubernetes-sigs/descheduler/pull/1459
- https://github.com/kubernetes-sigs/descheduler/pull/1458
- https://github.com/kubernetes-sigs/descheduler/pull/1457
- https://github.com/kubernetes-sigs/descheduler/pull/1456
- https://github.com/kubernetes-sigs/descheduler/pull/1454
- https://github.com/kubernetes-sigs/descheduler/pull/1447
- https://github.com/kubernetes-sigs/descheduler/pull/1483
- https://github.com/kubernetes-sigs/descheduler/pull/1484
- https://github.com/kubernetes-sigs/descheduler/pull/1485


ingvagabund:
Compared to virt-handler produced by my local kind these are the logs produced by the upstream CI:
```
{"component":"virt-handler","level":"info","msg":"re-enqueuing VirtualMachineInstance default/vmi-masquerade-3","pos":"vm.go:1712","reason":"server error. command SyncVMI failed: \"LibvirtError(Code=38, Domain=0, Message='Unable to create tap device tap0: Permission denied')\"","timestamp":"2024-07-17T15:01:56.211181Z"}
```

Could be kubevirt does not like combination of running kind in pod with dind enabled.

ingvagabund:
Currently blocked by v1.31.z k8s rebase in https://github.com/kubevirt/client-go.

ingvagabund:
Simplifying the kubevirt client creation and dropping dependency on k8s version.

damemi:
@ingvagabund can you put the vendor changes into their own commit for review?

ingvagabund:
@damemi done

pkg/descheduler/descheduler_test.go
```diff
@@ -28,6 +34,26 @@ import (
 	"sigs.k8s.io/descheduler/test"
 )
 
+var (
+	podEvictionError     = fmt.Errorf("PodEvictionError")
```
fanhaouu, line 38: This is not necessary to use fmt.Errorf

pkg/descheduler/evictions/evictions.go
```diff
@@ -18,24 +18,192 @@ package evictions
...
+type evictionRequestsCache struct {
+	mu                    sync.Mutex
+	requests              map[string]evictionRequestItem
+	rPerNode              map[string]uint
+	rPerNamespace         map[string]uint
+	rTotal                uint
+	assumedRequestTimeout uint
+}
+
+func newEvictionRequestsCache(assumedRequestTimeout uint) *evictionRequestsCache {
```
fanhaouu, line 66: same above

pkg/descheduler/evictions/evictions.go
```diff
@@ -77,7 +250,126 @@ func NewPodEvictor(
 		metricsEnabled:             options.metricsEnabled,
 		nodePodCount:               make(nodePodEvictedCount),
 		namespacePodCount:          make(namespacePodEvictCount),
+		featureGates:               featureGates,
+	}
+
+	if featureGates.Enabled(features.EvictionsInBackground) {
+		erCache := newEvictionRequestsCache(assumedEvictionRequestTimeoutSeconds)
+
+		podInformer.AddEventHandler(
```
fanhaouu, line 259: You should use syncTracker here, https://github.com/kubernetes/kubernetes/pull/113985.  Like scheduler: https://github.com/kubernetes/kubernetes/blob/release-1.31/pkg/scheduler/eventhandlers.go#L364

This evictor is shared, and it's essential to ensure that the cache is fully synchronized before other rescheduling policies can call the evictor to evict pods; otherwise, some constraints may become invalid.

pkg/descheduler/evictions/evictions.go
```diff
@@ -210,13 +569,36 @@ func evictPod(ctx context.Context, client clientset.Interface, pod *v1.Pod, poli
 		},
 		DeleteOptions: deleteOptions,
 	}
-	err := client.PolicyV1().Evictions(eviction.Namespace).Evict(ctx, eviction)
+	err := pe.client.PolicyV1().Evictions(eviction.Namespace).Evict(ctx, eviction)
```
fanhaouu, line 572: Have you considered adding a similar retry mechanism?
	
retriable := func(err error) bool {
		return apierrors.IsTooManyRequests(err) || apierrors.IsInternalError(err) || apierrors.IsServerTimeout(err) || apierrors.IsUnexpectedServerError(err)
	}
	err = retry.OnError(retry.DefaultRetry, retriable, func() error {
		return evictPod(ctx, pe.client, pod, pe.directEvict)
	})

fanhaouu:
No obvious bugs were found; everything except the commented code looks good to me.

pkg/descheduler/descheduler_test.go
ingvagabund: `errors.New("PodEvictionError")` is better :)

pkg/descheduler/evictions/evictions.go
ingvagabund: We wait for all the caches to sync at https://github.com/kubernetes-sigs/descheduler/blob/a3146a17057f7e151c314e85bee585ae740df942/pkg/descheduler/descheduler.go#L425. Are you convinced there's more to wait for?

fanhaouu: No, WaitForCacheSync only ensures that the local indexer has completed synchronization. However, due to the existence of the buffer queue—pendingNotifications, events handler may not have processed all events, leading to incomplete cached views that depend on the event handler.

pkg/descheduler/evictions/evictions.go
ingvagabund: I believe we discussed this in the past. Though, I have not seen anyone reporting this as annoying. We can always update the code. Nevertheless, if the returned errors are returned again for some period of time, retrying will not help. If only few evictions return the error codes the descheduler will try to evict another pod. All the pods are still treated as cattle. So the impact is usually negligible.

fanhaouu: Yes, the impact is not significant. For example, the rescheduling policy uses the method `podutil.SortPodsBasedOnPriorityLowToHigh(removablePods)` to sort the pods. Lower-priority pods at the front may fail to be evicted due to network issues, resulting in the eviction of higher-priority pods later, even though all these pods are evictable.

ingvagabund: You are right. This could be a suboptional case. Too many higher priority pods getting evicted instead of lower priority ones. Nevertheless, this deserves a separate PR/change. Retries are outside the scope of this.

pkg/descheduler/evictions/evictions.go
ingvagabund: The code is collecting the handler's `HasSynced`.

pkg/descheduler/evictions/evictions.go
```diff
@@ -108,6 +423,46 @@ func (pe *PodEvictor) SetClient(client clientset.Interface) {
 	pe.client = client
 }
 
+func (pe *PodEvictor) evictionRequestsTotal() uint {
```
atiratree, line 426: I forgot what was the consensus last time, but do we want to introduce new metrics for this feature? https://github.com/kubernetes-sigs/descheduler/blob/master/keps/1397-evictions-in-background/README.md#metrics

test/e2e/e2e_evictioninbackground_test.go
```diff
@@ -0,0 +1,542 @@
...
+const (
+	vmiCount = 3
+)
+
+func virtualMachineInstance(idx int) *kvcorev1.VirtualMachineInstance {
+	return &kvcorev1.VirtualMachineInstance{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: fmt.Sprintf("kubevirtvmi-%v", idx),
+			Annotations: map[string]string{
+				"descheduler.alpha.kubernetes.io/request-evict-only": "",
```
atiratree, line 45: just to confirm, the feature depends only on the key, and not on the value? Would be good to add a little a bit of the documentation above `EvictionRequestAnnotationKey` and `EvictionInProgressAnnotationKey`

pkg/descheduler/descheduler.go
```diff
@@ -140,9 +141,12 @@ func newDescheduler(rs *options.DeschedulerServer, deschedulerPolicy *api.Desche
...
 	}
 
-	podEvictor := evictions.NewPodEvictor(
-		nil,
+	podEvictor, err := evictions.NewPodEvictor(
+		ctx,
+		rs.Client,
 		eventRecorder,
+		podInformer,
+		rs.DefaultFeatureGates,
```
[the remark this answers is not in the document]
ingvagabund: I was keeping in mind the case where the descheduling framework is run as a simulator. E.g. in case two or more framework instances are created in the same run/binary.

pkg/descheduler/evictions/evictions.go
ingvagabund: We do. Yet, not as part of this PR. I will have a follow up for all other changes.

test/e2e/e2e_evictioninbackground_test.go
ingvagabund: That is correct. +1 for a documentation. I still have a blog post to write.

pkg/descheduler/descheduler.go
```diff
@@ -410,6 +417,7 @@ func RunDeschedulerStrategies(ctx context.Context, rs *options.DeschedulerServer
 
 	sharedInformerFactory.Start(ctx.Done())
 	sharedInformerFactory.WaitForCacheSync(ctx.Done())
+	descheduler.podEvictor.WaitForHandlersSync(ctx)
```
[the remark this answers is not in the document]
ingvagabund: Done

pkg/descheduler/evictions/evictions.go
```diff
@@ -18,24 +18,194 @@ package evictions
...
+	erc.mu.Lock()
+	defer erc.mu.Unlock()
+	uid, err := getPodKey(pod)
+	if err != nil {
+		return fmt.Errorf("unable to get pod key: %v", err)
+	}
+	if _, exists := erc.requests[uid]; exists {
+		return nil
+	}
+	erc.requests[uid] = evictionRequestItem{pod: pod}
```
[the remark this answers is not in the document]
ingvagabund: Done

pkg/descheduler/evictions/evictions.go
```diff
@@ -18,24 +18,194 @@ package evictions
...
+type evictionRequestItem struct {
+	pod               *v1.Pod
+	assumed           bool
+	creationTimestamp metav1.Time
+}
+
+type evictionRequestsCache struct {
+	mu                           sync.Mutex
+	requests                     map[string]evictionRequestItem
+	rPerNode                     map[string]uint
```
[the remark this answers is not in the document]
ingvagabund: Done

pkg/descheduler/evictions/evictions.go
```diff
@@ -18,24 +18,194 @@ package evictions
...
+	assumedEvictionRequestTimeoutSeconds uint          = 10 * 60 // 10 minutes
+	evictionRequestsCacheResyncPeriod    time.Duration = 10 * time.Minute
+	// syncedPollPeriod controls how often you look at the status of your sync funcs
+	syncedPollPeriod = 100 * time.Millisecond
+)
+
+type evictionRequestItem struct {
+	pod               *v1.Pod
+	assumed           bool
+	creationTimestamp metav1.Time
```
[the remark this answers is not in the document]
ingvagabund: Renamed to assumedTimestamp

pkg/descheduler/evictions/evictions.go
```diff
@@ -18,24 +18,194 @@ package evictions
...
+			requestAgeSeconds := uint(metav1.Now().Sub(item.creationTimestamp.Local()).Seconds())
+			if requestAgeSeconds > erc.assumedRequestTimeoutSeconds {
+				klog.V(4).Infof("Assumed eviction request in background timed out after %vs for %v, deleting", erc.assumedRequestTimeoutSeconds, klog.KObj(item.pod))
+				erc.deleteItem(uid)
+			}
+		}
+	}
+}
+
+func (erc *evictionRequestsCache) evictionRequestsPerNode(nodeName string) uint {
```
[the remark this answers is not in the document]
ingvagabund: Done

pkg/descheduler/evictions/evictions.go
```diff
@@ -210,13 +592,36 @@ func evictPod(ctx context.Context, client clientset.Interface, pod *v1.Pod, poli
 		},
 		DeleteOptions: deleteOptions,
 	}
-	err := client.PolicyV1().Evictions(eviction.Namespace).Evict(ctx, eviction)
+	err := pe.client.PolicyV1().Evictions(eviction.Namespace).Evict(ctx, eviction)
+	if pe.featureGates.Enabled(features.EvictionsInBackground) {
```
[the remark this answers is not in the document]
ingvagabund: Done. Just existing faster when err == nil.

cmd/descheduler/app/options/options.go
```diff
@@ -45,6 +50,10 @@ type DeschedulerServer struct {
 	SecureServing  *apiserveroptions.SecureServingOptionsWithLoopback
 	DisableMetrics bool
 	EnableHTTP2    bool
+	// FeatureGates enabled by the user
```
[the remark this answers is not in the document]
ingvagabund: Not applicable here as one may create multiple descheduler instances for simulation purposes.

cmd/descheduler/app/server.go
```diff
@@ -115,7 +116,13 @@ func NewDeschedulerCommand(out io.Writer) *cobra.Command {
 }
 
 func Run(ctx context.Context, rs *options.DeschedulerServer) error {
-	err := tracing.NewTracerProvider(ctx, rs.Tracing.CollectorEndpoint, rs.Tracing.TransportCert, rs.Tracing.ServiceName, rs.Tracing.ServiceNamespace, rs.Tracing.SampleRate, rs.Tracing.FallbackToNoOpProviderOnError)
+	err := features.DefaultMutableFeatureGate.SetFromMap(rs.FeatureGates)
```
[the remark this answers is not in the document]
ingvagabund: Done

pkg/framework/profile/profile.go
```diff
@@ -314,7 +315,7 @@ func (d profileImpl) RunDeschedulePlugins(ctx context.Context, nodes []*v1.Node)
 			span.AddEvent("Plugin Execution Failed", trace.WithAttributes(attribute.String("err", status.Err.Error())))
 			errs = append(errs, fmt.Errorf("plugin %q finished with error: %v", pl.Name(), status.Err))
 		}
-		klog.V(1).InfoS("Total number of pods evicted", "extension point", "Deschedule", "evictedPods", d.podEvictor.TotalEvicted()-evicted)
+		klog.V(1).InfoS("Total number of evictions/requests", "extension point", "Deschedule", "evictedPods", d.podEvictor.TotalEvicted()-evicted, "evictionRequests", d.podEvictor.TotalEvictionRequests()-evictionRequests)
```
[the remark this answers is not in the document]
ingvagabund: Even though there are two places that looks almost similar unifying this through a common function will not help that much. Are you suggesting to do something like:
```go
keysAndValues := []interface{}{
	"extension point", "Deschedule",
	"evictedPods", d.podEvictor.TotalEvicted()-evicted,
	"evictionRequests", d.podEvictor.TotalEvictionRequests()-evictionRequests,
}
```
?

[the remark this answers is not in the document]
atiratree: no, I mean just the format: `Sprintf` vs `keysAndValues` 

for example in evictions.go we have:

```
				klog.V(4).Infof("Assumed eviction request in background timed out after %vs for %v/%v, deleting", erc.assumedRequestTimeoutSeconds, item.podNamespace, item.podName)
```

test/e2e/e2e_evictioninbackground_test.go
atiratree: Ok, sounds good. Let's not forget about the constants and update them after as well.

pkg/descheduler/evictions/evictions.go
```diff
@@ -18,32 +18,204 @@ package evictions
...
+type evictionRequestsCache struct {
+	mu                           sync.RWMutex
+	requests                     map[string]evictionRequestItem
+	requestsPerNode              map[string]uint
+	requestsPerNamespace         map[string]uint
+	requestsTotal                uint
+	assumedRequestTimeoutSeconds uint
+}
+
+func newEvictionRequestsCache(assumedRequestTimeoutSeconds uint) *evictionRequestsCache {
```
atiratree, line 68: It would be nice to introduce unit tests for the cache.

pkg/framework/profile/profile.go
```diff
@@ -306,6 +306,7 @@ func (d profileImpl) RunDeschedulePlugins(ctx context.Context, nodes []*v1.Node)
 		ctx, span = tracing.Tracer().Start(ctx, pl.Name(), trace.WithAttributes(attribute.String("plugin", pl.Name()), attribute.String("profile", d.profileName), attribute.String("operation", tracing.DescheduleOperation)))
 		defer span.End()
 		evicted := d.podEvictor.TotalEvicted()
+		evictionRequests := d.podEvictor.TotalEvictionRequests()
```
atiratree, line 309: It would be easier to track if we prefixed these, for example `evictionRequestsBeforeBalance`?

pkg/descheduler/evictions/evictions.go
```diff
@@ -121,13 +478,35 @@ type EvictOptions struct {
...
+
 	pe.mu.Lock()
 	defer pe.mu.Unlock()
+
 	var span trace.Span
 	ctx, span = tracing.Tracer().Start(ctx, "EvictPod", trace.WithAttributes(attribute.String("podName", pod.Name), attribute.String("podNamespace", pod.Namespace), attribute.String("reason", opts.Reason), attribute.String("operation", tracing.EvictOperation)))
 	defer span.End()
 
-	if pe.maxPodsToEvictTotal != nil && pe.totalPodCount+1 > *pe.maxPodsToEvictTotal {
+	if pe.maxPodsToEvictTotal != nil && pe.totalPodCount+pe.erCache.evictionRequestsTotal()+1 > *pe.maxPodsToEvictTotal {
```
atiratree, line 509: Shouldn't this be gated?

pkg/descheduler/evictions/evictions.go
ingvagabund: There's one in TestEvictionRequestsCacheCleanup

pkg/descheduler/evictions/evictions.go
```diff
@@ -77,7 +255,144 @@ func NewPodEvictor(
...
+					if newPod.Status.Phase == v1.PodSucceeded || newPod.Status.Phase == v1.PodFailed {
+						klog.V(3).Infof("Pod with eviction in background completed. Removing pod %q from the cache.", klog.KObj(newPod))
+						if err := erCache.deletePod(newPod); err != nil {
+							// If the deletion fails the cache may block eviction
+							klog.ErrorS(err, "Unable to delete updated pod from cache", "pod", newPod)
+						}
+						return
+					}
+					// Ignore any pod that does not have eviction in progress
+					if _, exists := newPod.Annotations[EvictionInProgressAnnotationKey]; !exists {
```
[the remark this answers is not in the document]
ingvagabund: Oh. I misunderstood. So it can happen a pod might still have an eviction in background correctly finished. Yet, if `EvictionRequestAnnotationKey` is removed (e.g. by accident or on purpose) the pod is no longer considered for further processing. `EvictionRequestAnnotationKey` is the referential one. We might still check for `EvictionInProgressAnnotationKey` annotation as well just in case there's one. Yet, some implementations might decide to just remove `EvictionRequestAnnotationKey` annotation to signal an eviction in background was aborted/canceled.

pkg/descheduler/evictions/evictions.go
ingvagabund: It probably should to avoid cases where `pe.erCache.evictionRequestsTotal()` misbehaves. In the current implementation `evictionRequestsCache.requestsTotal` will always be 0 when the feature gate is not enabled.

ingvagabund: Done

pkg/framework/profile/profile.go
ingvagabund: Done

pkg/framework/profile/profile.go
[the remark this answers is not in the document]
ingvagabund: Gotcha. Updated.

pkg/descheduler/evictions/evictions.go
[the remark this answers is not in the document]
atiratree: > Yet, if EvictionRequestAnnotationKey is removed (e.g. by accident or on purpose) the pod is no longer considered for further processing.

I am not sure if I see all the loose ends here :D

If the pod was in the cache when the `Request` annotation was removed, then we skip evicting it (thus we process the `InProgress` annotation)

pkg/descheduler/evictions/evictions.go
```diff
@@ -77,20 +257,157 @@ func NewPodEvictor(
...
+				},
+			},
+		)
+		if err != nil {
+			return nil, fmt.Errorf("unable to register event handler for pod evictor: %v", err)
+		}
+
+		podEvictor.registeredHandlers = append(podEvictor.registeredHandlers, handlerRegistration)
+
+		go wait.UntilWithContext(ctx, erCache.cleanCache, evictionRequestsCacheResyncPeriod)
```
atiratree, line 377: wouldn't it be better to create a new method `Run` for this goroutine to make the intentions clearer?

ingvagabund: Done

pkg/descheduler/evictions/evictions.go
```diff
@@ -77,20 +257,157 @@ func NewPodEvictor(
...
+					// Ignore any pod that does not have eviction in progress
+					if _, exists := newPod.Annotations[EvictionInProgressAnnotationKey]; !exists {
+						// In case EvictionInProgressAnnotationKey annotation is not present/removed
+						// it's unclear whether the eviction was restarted or terminated.
+						// If the eviction gets restarted the pod needs to be removed from the cache
+						// to allow re-triggering the eviction.
+						if oldPod.Annotations == nil {
+							return
+						}
+						if _, exists := oldPod.Annotations[EvictionInProgressAnnotationKey]; exists {
```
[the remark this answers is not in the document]
ingvagabund: Gotcha. Updated.

pkg/framework/profile/profile.go
```diff
@@ -334,6 +335,7 @@ func (d profileImpl) RunBalancePlugins(ctx context.Context, nodes []*v1.Node) *f
 		ctx, span = tracing.Tracer().Start(ctx, pl.Name(), trace.WithAttributes(attribute.String("plugin", pl.Name()), attribute.String("profile", d.profileName), attribute.String("operation", tracing.BalanceOperation)))
 		defer span.End()
 		evicted := d.podEvictor.TotalEvicted()
```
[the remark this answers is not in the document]
ingvagabund: Done

pkg/descheduler/evictions/evictions.go
```diff
@@ -121,24 +478,52 @@ type EvictOptions struct {
...
 	defer span.End()
 
-	if pe.maxPodsToEvictTotal != nil && pe.totalPodCount+1 > *pe.maxPodsToEvictTotal {
-		err := NewEvictionTotalLimitError()
-		if pe.metricsEnabled {
-			metrics.PodsEvicted.With(map[string]string{"result": err.Error(), "strategy": opts.StrategyName, "namespace": pod.Namespace, "node": pod.Spec.NodeName, "profile": opts.ProfileName}).Inc()
+	if pe.maxPodsToEvictTotal != nil {
+		totalPodCount := pe.totalPodCount
+		if pe.featureGates.Enabled(features.EvictionsInBackground) {
+			totalPodCount += pe.erCache.evictionRequestsTotal()
```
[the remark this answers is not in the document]
ingvagabund: Done

pkg/descheduler/evictions/evictions.go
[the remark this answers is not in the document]
ingvagabund: There's always some space for errors. What's important is to properly remove entry from the cache eventually so all these "loose ends" are not persistent in the cache forever.

pkg/descheduler/evictions/evictions.go
```diff
@@ -210,13 +596,39 @@ func evictPod(ctx context.Context, client clientset.Interface, pod *v1.Pod, poli
...
 	}
-	err := client.PolicyV1().Evictions(eviction.Namespace).Evict(ctx, eviction)
+	err := pe.client.PolicyV1().Evictions(eviction.Namespace).Evict(ctx, eviction)
+	if err == nil {
+		return false, nil
+	}
+	if pe.featureGates.Enabled(features.EvictionsInBackground) {
+		if pod.Annotations != nil {
+			// eviction in background requested
+			if _, exists := pod.Annotations[EvictionRequestAnnotationKey]; exists {
```
atiratree, line 606: we can simplify the annotations check also here

pkg/descheduler/evictions/evictions.go
```diff
@@ -121,13 +480,35 @@ type EvictOptions struct {
...
 func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod, opts EvictOptions) error {
+	if len(pod.UID) == 0 {
+		klog.InfoS("Ignoring pod eviction due to missing UID", "pod", pod)
+		return fmt.Errorf("Pod %v is missing UID", klog.KObj(pod))
+	}
+
+	if pe.featureGates.Enabled(features.EvictionsInBackground) {
+		if pod.Annotations != nil {
+			// eviction in background requested
+			if _, exists := pod.Annotations[EvictionRequestAnnotationKey]; exists {
```
atiratree, line 491: and here

pkg/descheduler/evictions/evictions.go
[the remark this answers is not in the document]
atiratree: > What's important is to properly remove entry from the cache
which we would not do when `evictionAssumed == false` and `EvictionRequestAnnotationKey` is removed

pkg/descheduler/evictions/evictions.go
```diff
@@ -77,20 +261,155 @@ func NewPodEvictor(
...
+				},
+			},
+		)
+		if err != nil {
+			return nil, fmt.Errorf("unable to register event handler for pod evictor: %v", err)
+		}
+
+		podEvictor.registeredHandlers = append(podEvictor.registeredHandlers, handlerRegistration)
+
+		go erCache.run(ctx)
```
atiratree, line 379: Can create the new goroutine and run this from outside in `RunDeschedulerStrategies`?

pkg/descheduler/evictions/evictions.go
```diff
@@ -77,20 +261,155 @@ func NewPodEvictor(
...
+						klog.ErrorS(nil, "Cannot convert newObj to *v1.Pod", "newObj", newObj)
+						return
+					}
+					if newPod.Annotations == nil {
+						return
+					}
+					// Ignore pod's that are not subject to an eviction in background
+					if _, exists := newPod.Annotations[EvictionRequestAnnotationKey]; !exists {
+						return
+					}
```
atiratree, line 307: +here

pkg/descheduler/evictions/evictions.go
ingvagabund: Done

pkg/descheduler/evictions/evictions.go
ingvagabund: Done

pkg/descheduler/evictions/evictions.go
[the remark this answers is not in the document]
ingvagabund: Good catch. Updating the delete event handler to delete every pod that's in the cache: https://github.com/kubernetes-sigs/descheduler/pull/1466/commits/bb084d6e0f50ae99eafa4567e5f97dd31ab09c38

pkg/descheduler/evictions/evictions.go
```diff
@@ -18,32 +18,208 @@ package evictions
...
+	erc.mu.RLock()
+	defer erc.mu.RUnlock()
+	return uint(len(erc.requests))
+}
+
+// getPodKey returns the string key of a pod.
+func getPodKey(pod *v1.Pod) (string, error) {
+	uid := string(pod.UID)
+	if len(uid) == 0 {
+		return "", errors.New("cannot get cache key for pod with empty UID")
```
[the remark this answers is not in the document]
ingvagabund: Assuming all pods have the UID set this error will never get returned.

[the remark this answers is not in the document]
ingvagabund: +1 for exploring this avenue and filtering out all pods that does not have their uid set from eviction in a separate PR/improvement.

pkg/descheduler/evictions/evictions.go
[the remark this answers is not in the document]
atiratree: That is much better. I still think it is unfortunate that we do not deschedule pods that had both `EvictionRequestAnnotationKey` and `EvictionInProgressAnnotationKey` removed at the same time. If the owner cancels the progress in this way and the pod is not terminating, then the pod will be immune to descheduling.

pkg/descheduler/evictions/evictions.go
ingvagabund: Pod evictor is not mean to be a controller. Thus, no run method there. Currently, it's the most suitable place for placing the cache. Yet, eventually, this needs to go out some place else. Let's leave the run method as it is for now.

pkg/descheduler/evictions/evictions.go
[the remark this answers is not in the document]
ingvagabund: Removing the pod from the cache during update event just in case the pod is in the cache: https://github.com/kubernetes-sigs/descheduler/pull/1466/commits/4425072b93da0b0c43c1fe37e9beec4e12b6ceb5

pkg/descheduler/evictions/evictions.go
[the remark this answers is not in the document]
atiratree: There should be no need for filtering:
- https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#uids
- https://github.com/kubernetes/kubernetes/blob/021dbe9d195165cfcbe756c0b6dd5bb943b64a7c/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L151-L159

pkg/descheduler/evictions/evictions.go
```diff
@@ -77,20 +261,155 @@ func NewPodEvictor(
...
+						// request for eviction again. In case the eviction got restarted requesting
+						// the eviction again is expected to be a no-op. In case the eviction
+						// got terminated with no-retry, requesting a new eviction is a normal
+						// operation.
+						klog.V(3).InfoS("Eviction in background canceled (annotation removed). Removing pod from the cache.", "annotation", EvictionInProgressAnnotationKey, "pod", klog.KObj(newPod))
+						if err := erCache.deletePod(newPod); err != nil {
+							// If the deletion fails the cache may block eviction
+							klog.ErrorS(err, "Unable to delete updated pod from cache", "pod", newPod)
+							return
+						}
```
atiratree, line 340: for example here: the comment and the error is very confusing

pkg/descheduler/evictions/evictions.go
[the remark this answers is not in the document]
atiratree: looks good, thx!

pkg/descheduler/evictions/evictions.go
```diff
@@ -77,20 +248,137 @@ func NewPodEvictor(
...
+						// request for eviction again. In case the eviction got restarted requesting
+						// the eviction again is expected to be a no-op. In case the eviction
+						// got terminated with no-retry, requesting a new eviction is a normal
+						// operation.
+						klog.V(3).InfoS("Eviction in background canceled (annotation removed). Removing pod from the cache.", "annotation", EvictionInProgressAnnotationKey, "pod", klog.KObj(newPod))
+						erCache.deletePod(newPod)
+						return
+					}
+					// Pick up the eviction in progress
+					erCache.addPod(newPod)
```
atiratree, line 319: We should prob log this operations for consistency reasons (we do for the rest of add/delete operations).

Also same concern about the assume operation.

atiratree: and should be guarded by `hasPod`

ingvagabund: Updated

atiratree:
LGTM, thanks!

ingvagabund:
Squashing all the review commits now

ingvagabund:
@atiratree thank you for all your reviews and patience. This really helped.

PR #1468, opened 2024-07-16
fix: Check initContainerStatuses in PodLifeTime states

adammw:
Currently the PodLifeTime check ignores any container statuses from init containers, meaning if any init containers fail with the same errors they are not correctly evicted. 

cc @a7i @jklaw90


pkg/framework/plugins/podlifetime/pod_lifetime.go
```diff
@@ -85,6 +85,13 @@ func New(args runtime.Object, handle frameworktypes.Handle) (frameworktypes.Plug
 				return true
 			}
 
+			// Init Container Status Reason
+			for _, containerStatus := range pod.Status.InitContainerStatuses {
```
a7i, line 89: [RemoveFailedPods](https://github.com/kubernetes-sigs/descheduler/blob/master/README.md#removefailedpods) and [RemovePodsHavingTooManyRestarts](https://github.com/kubernetes-sigs/descheduler/blob/master/README.md#removepodshavingtoomanyrestarts) both offer the flag `includingInitContainers` so to be consistent we should include that flag [ref](https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/framework/plugins/removepodshavingtoomanyrestarts/toomanyrestarts.go#L85-L89). thoughts?

README.md
```diff
@@ -673,12 +673,14 @@ Pods in any state (even `Running`) are considered for eviction.
...
-|`maxPodLifeTimeSeconds`|int||
-|`states`|list(string)|Only supported in v0.25+|
-|`namespaces`|(see [namespace filtering](#namespace-filtering))||
-|`labelSelector`|(see [label filtering](#label-filtering))||
+| Name                           | Type                                              | Notes                    |
+|--------------------------------|---------------------------------------------------|--------------------------|
+| `maxPodLifeTimeSeconds`        | int                                               |                          |
+| `states`                       | list(string)                                      | Only supported in v0.25+ |
+| `includingInitContainers`      | bool                                              | Only supported in v0.31+ |
+| `includingEphemeralContainers` | bool                                              | Only supported in v0.31+ |
```
adammw, line 681: I assume this is the next version - not sure if correct way of noting this?

pkg/framework/plugins/podlifetime/pod_lifetime.go
adammw: updated, however does increase the size of the PR and I would argue that including init containers is likely expected behaviour for erroring statuses

a7i: Understood. If we didn't guard it by a new flag, then it could break existing functionality for users.

PR #1469, opened 2024-07-17
e2e: TestLeaderElection: delete the lease and increase the retry period

ingvagabund:
In some random cases none of the deschedulers acquires a lease. Making the test fail due to already existing lease.

### DIFF
--- a/test/e2e/e2e_leaderelection_test.go
+++ b/test/e2e/e2e_leaderelection_test.go
@@ -27,6 +27,7 @@ import (
 
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	clientset "k8s.io/client-go/kubernetes"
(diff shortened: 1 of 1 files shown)


test/e2e/e2e_leaderelection_test.go
```diff
@@ -141,7 +156,11 @@ func TestLeaderElection(t *testing.T) {
 			t.Logf("Only the pods in %s namespace are evicted. Pods before: %s, Pods after %s", ns2, podListBOrg, podListB)
 		}
 	} else {
-		t.Fatalf("Pods are evicted in both namespaces. For %s namespace Pods before: %s, Pods after %s. And, for %s namespace Pods before: %s, Pods after: %s", ns1, podListAOrg, podListA, ns2, podListBOrg, podListB)
+		if left && right {
+			t.Fatalf("No pods evicted. Probably none of the deschedulers were running.")
```
JaneLiuL, line 160: should it change int t.Lof?

ingvagabund: The test needs to fail here as well. The test expect at least one descheduler running.

JaneLiuL: Oh, ok, then lgtm now~~ thanks for your explain~

JaneLiuL (commented):
/lgtm

test/e2e/e2e_leaderelection_test.go
```diff
@@ -95,9 +99,20 @@ func TestLeaderElection(t *testing.T) {
...
+	}
 	s2.ClientConnection.Kubeconfig = os.Getenv("KUBECONFIG")
 	s2.PolicyConfigFile = "./policy_leaderelection_b.yaml"
 
+	// Delete the descheduler lease
+	err = clientSet.CoordinationV1().Leases("kube-system").Delete(ctx, "descheduler", metav1.DeleteOptions{})
+	if err != nil && !apierrors.IsNotFound(err) {
+		t.Fatalf("Unable to remove kube-system/descheduler lease: %v", err)
+	} else {
+		t.Logf("Removed kube-system/descheduler lease")
```
a7i, line 113: a bit misleading since not found doesn't actually remove

ingvagabund: You are right. Updated. Thx.

PR #1472, opened 2024-07-19
e2e: TestTooManyRestarts: run descheduler as a whole instead of a single plugin

ingvagabund:
Part or replacing direct invocation of plugins with running an entire descheduler while providing a policy with the expected plugins configured.

To goal is to get closer to running a descheduler in production. The current approach still resembles the way the descheduler is exercises in unit tests.


JaneLiuL (commented):
/lgtm

a7i (approved):
/approve

good idea!

PR #1474, opened 2024-07-21
[TestTooManyRestarts] e2e: build a descheduler image and run the descheduler as a pod

ingvagabund:
TODO:
- [x] build a descheduler image before an e2e test is run
- [x] TestTooManyRestarts deploys and runs the descheduler instead of running the actual code directly

Other e2e tests will be updated separately.

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package e2e
 
 import (
+	"bufio"
 	"context"
 	"fmt"
 	"math"
@@ -26,12 +27,15 @@ import (
 	"testing"
 	"time"
 
+	"github.com/ghodss/yaml"
+
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
 	schedulingv1 "k8s.io/api/scheduling/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/informers"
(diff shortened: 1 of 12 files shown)


test/e2e/e2e_test.go
```diff
@@ -60,6 +69,356 @@ import (
...
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+func printPodLogs(t *testing.T, ctx context.Context, kubeClient clientset.Interface, podName string) {
```
jklaw90, line 263: nit: ctx should come first for consistency.

test/e2e/e2e_test.go
```diff
@@ -60,6 +69,356 @@ import (
...
+	request := kubeClient.CoreV1().Pods("kube-system").GetLogs(podName, &v1.PodLogOptions{})
+	readCloser, err := request.Stream(context.TODO())
+	if err != nil {
+		t.Logf("Unable to request stream: %v\n", err)
+		return
+	}
+
+	defer readCloser.Close()
+	r := bufio.NewReader(readCloser)
```
jklaw90, line 272: NewScanner may slightly cleaner here

ingvagabund: Looks better :) Thx

test/e2e/e2e_test.go
```diff
@@ -60,6 +68,348 @@ import (
...
+func deschedulerSA() *v1.ServiceAccount {
+	return &v1.ServiceAccount{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "descheduler-sa",
+			Namespace: "kube-system",
+		},
+	}
+}
+
+func deschedulerClusterRole() *rbacv1.ClusterRole {
```
a7i, line 84: is there a reason that we're duplicating these, as opposed to using [this](https://github.com/kubernetes-sigs/descheduler/blob/master/kubernetes/base/rbac.yaml)?

ingvagabund: Good point. We need to reconcile. I prefer the Go version of all the manifests so we can skip the unmarshaling part. I wonder if it would make sense to create e.g. `hack/manifests2go.sh` that will do the trick and gets run right before e2e.

ingvagabund:
@jklaw90 thanks for the continuous lgtms :) Hopefully this is the last change.

test/e2e/e2e_test.go
ingvagabund: @a7i SA, CR and CRB are now applied from `kubernetes/base/rbac.yaml`

test/e2e/e2e_test.go
```diff
@@ -26,23 +27,30 @@ import (
 	"testing"
 	"time"
 
+	"github.com/ghodss/yaml"
+
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
```
a7i, line 34: looks like this is unused, hence the checks failing

a7i (approved):
/approve
/lgtm

PR #1475, opened 2024-07-23
skip eviction when pod creation time is below minPodAge threshold setting

victorgs:
In the default initialization phase of the descheduler, add a new constraint to not evict pods that creation time is below minPodAge threshold. I have used minutes for the threshold, but if you prefer seconds, I don't mind changing it.

Added value:

- Avoid crazy pod movement when the autoscaler scales up and down.

- Avoid evicting pods when they are warming up.

- Decreases the overall cost of eviction as no pod will be evicted before doing significant amount of work.

- Guard against scheduling. Descheduling loops in situations where the descheduler has a different node fit logic from scheduler, like not considering topology spread constraints.

---

I was testing the change in a local cluster:
`I0716 13:51:27.441401       1 defaultevictor.go:249] "Pod fails the following checks" pod="kube-system/nginx-deployment-8-68c4499688-vjvp8" checks="pod age is not older than MinPodAge: 10000 minutes"`

We think adding this feature upstream is more beneficial than maintaining our own fork, and it may be useful to others.

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,6 +145,7 @@ The Default Evictor Plugin is used by default for filtering pods before processi
 |`priorityThreshold`|`priorityThreshold`||(see [priority filtering](#priority-filtering))|
 |`nodeFit`|`bool`|`false`|(see [node fit filtering](#node-fit-filtering))|
 |`minReplicas`|`uint`|`0`| ignore eviction of pods where owner (e.g. `ReplicaSet`) replicas is below this threshold |
+|`minPodAge`|`uint`|`0`| ignore eviction of pods with a creation time within this threshold (in minutes) |
 
 ### Example policy
 
--- a/README.md
+++ b/README.md
@@ -145,7 +145,7 @@ The Default Evictor Plugin is used by default for filtering pods before processi
 |`priorityThreshold`|`priorityThreshold`||(see [priority filtering](#priority-filtering))|
 |`nodeFit`|`bool`|`false`|(see [node fit filtering](#node-fit-filtering))|
 |`minReplicas`|`uint`|`0`| ignore eviction of pods where owner (e.g. `ReplicaSet`) replicas is below this threshold |
-|`minPodAge`|`metav1.Duration`|`0`| ignore eviction of pods with a creation time within this threshold (in minutes) |
+|`minPodAge`|`metav1.Duration`|`0`| ignore eviction of pods with a creation time within this threshold |
 
 ### Example policy
 
--- a/README.md
+++ b/README.md
@@ -145,7 +145,7 @@ The Default Evictor Plugin is used by default for filtering pods before processi
 |`priorityThreshold`|`priorityThreshold`||(see [priority filtering](#priority-filtering))|
 |`nodeFit`|`bool`|`false`|(see [node fit filtering](#node-fit-filtering))|
 |`minReplicas`|`uint`|`0`| ignore eviction of pods where owner (e.g. `ReplicaSet`) replicas is below this threshold |
-|`minPodAge`|`uint`|`0`| ignore eviction of pods with a creation time within this threshold (in minutes) |
+|`minPodAge`|`metav1.Duration`|`0`| ignore eviction of pods with a creation time within this threshold (in minutes) |
 
 ### Example policy
 
--- a/pkg/api/v1alpha1/conversion.go
+++ b/pkg/api/v1alpha1/conversion.go
@@ -19,6 +19,7 @@ package v1alpha1
 import (
 	"context"
 	"fmt"
+
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/conversion"
 	"k8s.io/apimachinery/pkg/runtime"
--- a/pkg/api/v1alpha1/conversion.go
+++ b/pkg/api/v1alpha1/conversion.go
@@ -19,7 +19,6 @@ package v1alpha1
 import (
 	"context"
 	"fmt"
-
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/conversion"
 	"k8s.io/apimachinery/pkg/runtime"
--- a/pkg/api/v1alpha1/types.go
+++ b/pkg/api/v1alpha1/types.go
@@ -55,9 +55,6 @@ type DeschedulerPolicy struct {
 
 	// MaxNoOfPodsToTotal restricts maximum of pods to be evicted total.
 	MaxNoOfPodsToEvictTotal *uint `json:"maxNoOfPodsToEvictTotal,omitempty"`
-
-	// MinPodAge prevents recently created pods (within the specified minutes) from being evicted.
-	MinPodAge *uint `json:"minPodAge,omitempty"`
 }
 
 type (
--- a/pkg/api/v1alpha1/types.go
+++ b/pkg/api/v1alpha1/types.go
@@ -55,6 +55,9 @@ type DeschedulerPolicy struct {
 
 	// MaxNoOfPodsToTotal restricts maximum of pods to be evicted total.
 	MaxNoOfPodsToEvictTotal *uint `json:"maxNoOfPodsToEvictTotal,omitempty"`
+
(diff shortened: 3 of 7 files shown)


pkg/api/v1alpha1/types.go
```diff
@@ -55,6 +55,9 @@ type DeschedulerPolicy struct {
 
 	// MaxNoOfPodsToTotal restricts maximum of pods to be evicted total.
 	MaxNoOfPodsToEvictTotal *uint `json:"maxNoOfPodsToEvictTotal,omitempty"`
+
+	// MinPodAge prevents recently created pods (within the specified minutes) from being evicted.
+	MinPodAge *uint `json:"minPodAge,omitempty"`
```
ingvagabund, line 60: `v1alpha1` is deprecated, we will be removing this type soon. Better to avoid extending it. Also, a conversion needs to be wired to take the new field into account.

pkg/framework/plugins/defaultevictor/types.go
```diff
@@ -35,4 +35,5 @@ type DefaultEvictorArgs struct {
 	PriorityThreshold       *api.PriorityThreshold `json:"priorityThreshold"`
 	NodeFit                 bool                   `json:"nodeFit"`
 	MinReplicas             uint                   `json:"minReplicas"`
+	MinPodAge               uint                   `json:"minPodAge"`
```
ingvagabund, line 38: What about to use `*time.Duration` instead? Users can then decide which unit they want to choose.

ingvagabund (changes requested):
Thank you for this improvement. This is a useful addition to the project. Much appreciated.

Users will need to keep in mind to avoid combining PodLifeTime's `maxPodLifeTimeSeconds` with `minPodAge` incorrectly. We currently do not validate accross plugin configurations to detect invalid combinations.

pkg/framework/plugins/defaultevictor/types.go
victorgs: Thanks for the comment! I did the change here https://github.com/kubernetes-sigs/descheduler/pull/1475/commits/ec7ad04c51929cb50cb6a946df63212c4c80154c . Let me know if thats what you expect! ty

pkg/api/v1alpha1/types.go
victorgs: Thank you for you comments Jan :)

I did not get this one tbh, I removed `MinPodAge` from `v1alpha1` and move it to `DefaultEvictorArgs` https://github.com/kubernetes-sigs/descheduler/blob/ec7ad04c51929cb50cb6a946df63212c4c80154c/pkg/framework/plugins/defaultevictor/types.go#L25-L39

So if we want to avoid extending `v1alpha1`, I do not know which conversion we need to do.

README.md
```diff
@@ -145,7 +145,7 @@ The Default Evictor Plugin is used by default for filtering pods before processi
 |`priorityThreshold`|`priorityThreshold`||(see [priority filtering](#priority-filtering))|
 |`nodeFit`|`bool`|`false`|(see [node fit filtering](#node-fit-filtering))|
 |`minReplicas`|`uint`|`0`| ignore eviction of pods where owner (e.g. `ReplicaSet`) replicas is below this threshold |
-|`minPodAge`|`uint`|`0`| ignore eviction of pods with a creation time within this threshold (in minutes) |
+|`minPodAge`|`metav1.Duration`|`0`| ignore eviction of pods with a creation time within this threshold (in minutes) |
```
ingvagabund, line 148: "in minutes" can be dropped

pkg/framework/plugins/defaultevictor/defaultevictor.go
```diff
@@ -186,10 +186,10 @@ func New(args runtime.Object, handle frameworktypes.Handle) (frameworktypes.Plug
 		})
 	}
 
-	if defaultEvictorArgs.MinPodAge > 0 {
+	if defaultEvictorArgs.MinPodAge != nil {
 		ev.constraints = append(ev.constraints, func(pod *v1.Pod) error {
-			if pod.Status.StartTime == nil || time.Since(pod.Status.StartTime.Time) < time.Duration(defaultEvictorArgs.MinPodAge)*time.Minute {
-				return fmt.Errorf("pod age is not older than MinPodAge: %d minutes", defaultEvictorArgs.MinPodAge)
+			if pod.Status.StartTime == nil || time.Since(pod.Status.StartTime.Time) < defaultEvictorArgs.MinPodAge.Duration {
+				return fmt.Errorf("pod age is not older than MinPodAge: %d seconds", uint(defaultEvictorArgs.MinPodAge.Duration.Seconds()))
```
ingvagabund, line 192: `Duration.String()` will help to keep the value short and redeable when user chooses high duration. https://pkg.go.dev/time#Duration.String

ingvagabund (changes requested):

pkg/api/v1alpha1/conversion.go
```diff
@@ -19,7 +19,6 @@ package v1alpha1
 import (
 	"context"
 	"fmt"
-
```
ingvagabund, line 22: Can you please undone the new line removal? To keep this file unchanged.

pkg/framework/plugins/defaultevictor/types.go
ingvagabund: This works. Thank you

pkg/api/v1alpha1/types.go
ingvagabund: This is perfect :) I meant any new field under v1alpha1 type needs a conversion from v1alpha1 -> v1alpha2 included as well. Nothing else to be done here. Thank you.

pkg/framework/plugins/defaultevictor/defaultevictor_test.go
```diff
@@ -326,6 +329,8 @@ func TestDefaultEvictorFilter(t *testing.T) {
 	lowPriority := int32(800)
 	highPriority := int32(900)
 
+	minPodAge := metav1.Duration{50 * time.Minute}
```
ingvagabund, line 332:
```
pkg/framework/plugins/defaultevictor/defaultevictor_test.go:332:15: k8s.io/apimachinery/pkg/apis/meta/v1.Duration struct literal uses unkeyed fields
```

`metav1.Duration{Duration: 50 * time.Minute}`

README.md
victorgs: Done in https://github.com/kubernetes-sigs/descheduler/pull/1475/commits/a8502e4e484dc9a647783d7b3f90141350dea680 ty!

pkg/framework/plugins/defaultevictor/defaultevictor.go
victorgs: Done in https://github.com/kubernetes-sigs/descheduler/pull/1475/commits/9adea9fb46c31466989eaef8c93055804581776b

pkg/api/v1alpha1/conversion.go
victorgs: Done in https://github.com/kubernetes-sigs/descheduler/pull/1475/commits/61b7b995b99122065a13a0fa8ff87fbccbc91d52

pkg/framework/plugins/defaultevictor/defaultevictor_test.go
victorgs: Done in https://github.com/kubernetes-sigs/descheduler/pull/1475/commits/c7de3c40c19da27114a6bceab147959c1318eda1

pkg/framework/plugins/defaultevictor/defaultevictor.go
```diff
@@ -64,6 +65,7 @@ func HaveEvictAnnotation(pod *v1.Pod) bool {
 }
 
 // New builds plugin from its arguments while passing a handle
+// nolint: gocyclo
```
victorgs, line 68: @ingvagabund do not want to forget about this. I need to add it to add it to make the linter pass. I saw the codebase has a bunch of this, but if you want to proceed with something else let me know.

victorgs:
@ingvagabund thanks for you help and review 🙇

ingvagabund:
Happy to help. Thank you for your time and contribution.

PR #1480, opened 2024-08-06
Plugin args: tag arguments with omitempty to reduce the marshalled json size

ingvagabund:
```yaml
- args:
    namespaces:
      include:
      - includedNamespace
```
looks better than
```yaml
- args:
    excludedTaints: null
    includePreferNoSchedule: false
    includedTaints: null
    labelSelector: null
    namespaces:
      exclude: []
      include:
      - includedNamespace
```

### DIFF
--- a/pkg/framework/plugins/podlifetime/types.go
+++ b/pkg/framework/plugins/podlifetime/types.go
@@ -25,10 +25,10 @@ import (
 type PodLifeTimeArgs struct {
 	metav1.TypeMeta `json:",inline"`
 
-	Namespaces                   *api.Namespaces       `json:"namespaces"`
-	LabelSelector                *metav1.LabelSelector `json:"labelSelector"`
-	MaxPodLifeTimeSeconds        *uint                 `json:"maxPodLifeTimeSeconds"`
-	States                       []string              `json:"states"`
-	IncludingInitContainers      bool                  `json:"includingInitContainers"`
-	IncludingEphemeralContainers bool                  `json:"includingEphemeralContainers"`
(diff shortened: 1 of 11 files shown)


pkg/framework/plugins/podlifetime/types.go
```diff
@@ -25,10 +25,10 @@ import (
 type PodLifeTimeArgs struct {
 	metav1.TypeMeta `json:",inline"`
 
-	Namespaces                   *api.Namespaces       `json:"namespaces"`
-	LabelSelector                *metav1.LabelSelector `json:"labelSelector"`
-	MaxPodLifeTimeSeconds        *uint                 `json:"maxPodLifeTimeSeconds"`
-	States                       []string              `json:"states"`
-	IncludingInitContainers      bool                  `json:"includingInitContainers"`
-	IncludingEphemeralContainers bool                  `json:"includingEphemeralContainers"`
+	Namespaces                   *api.Namespaces       `json:"namespaces,omitempty"`
```
damemi, line 28: is there anywhere we're specifically making a null check on pointer fields like this?

ingvagabund: Every plugin validation code is doing some validation. E.g. https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/framework/plugins/podlifetime/validation.go#L37.
Also, each plugin is expected to check for the nil pointer. E.g. https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/framework/plugins/podlifetime/pod_lifetime.go#L54-L57.

PR #1483, opened 2024-08-12
tests: de-duplicate framework handle initialization

ingvagabund:
Initialize a plugin handler at once through a single function invocation. To avoid duplicating the same initialization code throughout the codebase.


a7i (approved):
nice cleanup

/lgtm

PR #1496, opened 2024-08-14
bump k8s.io libs to v0.31.0

a7i:
Closes https://github.com/kubernetes-sigs/descheduler/issues/1487

In a follow-up PR we need to decide between our custom `--enable-http2` vs `--disable-http2-serving`


jklaw90:
Fwiw I like `--disable-http2-serving`

a7i:
> Fwiw I like `--disable-http2-serving`

Any suggestions on how we deprecate our custom flag?

My thinking is to make it optional/nil with a deprecated message. And then remove in 2 releases?

ingvagabund:
I will have another review capacity beg of the next week if that's ok.

a7i:
> I will have another review capacity beg of the next week if that's ok.

sounds good, I'll have that PR ready by then. for now, I'll get this merged since it's standard upgrade changes

PR #1497, opened 2024-08-15
[WIP] [TestLeaderElection] e2e: build a descheduler image and run the descheduler as a pod

hsunwenfang:
[TODO] policy function like func tooManyRestartsPolicy()
[TODO] func initPluginRegistry()

### DIFF
--- a/test/e2e/e2e_leaderelection_test.go
+++ b/test/e2e/e2e_leaderelection_test.go
@@ -19,7 +19,6 @@ package e2e
 import (
 	"context"
 	"fmt"
-	"os"
 	"reflect"
 	"strings"
 	"testing"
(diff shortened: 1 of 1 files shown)


hsunwenfang:
[ASK]
Hi @ingvagabund
I'm working on TestLeaderElection and is refering your commit for TestTooManyRestarts
https://github.com/kubernetes-sigs/descheduler/commit/ab467a5dd29925eb3fd794d2d93965f4a58281fc
Where L57 of e2e_toomanyrestarts_test.go the "sigs.k8s.io/descheduler/pkg/framework/plugins/removepodshavingtoomanyrestarts" is imported
While there is no corresponding folder for e2e_leaderelection_test.go
Should "sigs.k8s.io/descheduler/pkg/framework/plugins/defaultevictor" be used instead?
Or should there be another way to do it
Thanks!!

test/e2e/e2e_leaderelection_test.go
```diff
@@ -33,9 +33,19 @@ import (
 	clientset "k8s.io/client-go/kubernetes"
 	utilptr "k8s.io/utils/ptr"
 	"sigs.k8s.io/descheduler/cmd/descheduler/app/options"
+	"sigs.k8s.io/descheduler/pkg/api"
 	"sigs.k8s.io/descheduler/pkg/descheduler"
 )
 
+// Should use something like "sigs.k8s.io/descheduler/pkg/framework/plugins/removepodshavingtoomanyrestarts"
```
ingvagabund, line 40: yes, in this case it would be `"sigs.k8s.io/descheduler/pkg/framework/plugins/podlifetime"`.

ingvagabund:
> While there is no corresponding folder for leaderelection

leaderelection does not have a corresponding plugin. It's a configuration of the whole descheduler. So first you need to create a podlifetime policy (similar to `tooManyRestartsPolicy`). And then create a descheduler condfiguration with the leader election enabled.

ingvagabund:
> func initPluginRegistry()

No need to update `initPluginRegistry` as it already registers all the necessary plugins.

test/e2e/e2e_leaderelection_test.go
```diff
@@ -35,13 +35,46 @@ import (
...
+						Args: &podlifetime.PodLifeTimeArgs{
+							MaxPodLifeTimeSeconds: maxPodLifeTimeSeconds, // [TODO] The values to set
+							Namespaces: &api.Namespaces{
+								Include: []string{targetNamespace},
+							},
+						},
+					},
+					{
+						Name: defaultevictor.PluginName,
+						Args: &defaultevictor.DefaultEvictorArgs{
```
hsunwenfang, line 60: Args in &defaultevictor.DefaultEvictorArgs follows #1472 
Which may not be the proper args here

test/e2e/e2e_leaderelection_test.go
```diff
@@ -124,10 +157,18 @@ func TestLeaderElection(t *testing.T) {
 	}
 	t.Logf("Removed kube-system/descheduler lease")
 
+	tc := struct {
```
hsunwenfang, line 160: use single tc here instead of the tests []struct in #1472 
Not sure if it's better to use []struct, includes the cases of ns1 and ns2 then run
But I suppose ns1 and ns2 should belongs to the same test case

ingvagabund: There's currently only a single test to be run. With two descheduler pods run in the same namespace.

hsunwenfang:
Hi @ingvagabund 
Thanks you for all the instruction 
I added a commit and comments, please help on commenting~
And I did the 'make test-e2e' locally
It seems to be failed at the first e2e_duplicatepods test
Should there be a PR to address that also?
`+ go test sigs.k8s.io/descheduler/test/e2e/ -v
=== RUN   TestRemoveDuplicates
    e2e_duplicatepods_test.go:47: Error during client creation with unable to build in cluster config: unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined
--- FAIL: TestRemoveDuplicates (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0xc8 pc=0x1d0d319]`

ingvagabund:
The goal here is to create two descheduler Deployments: https://github.com/kubernetes-sigs/descheduler/blob/master/test/e2e/e2e_toomanyrestarts_test.go#L213. Each deployment with its policy in its own configmap: https://github.com/kubernetes-sigs/descheduler/blob/master/test/e2e/e2e_toomanyrestarts_test.go#L195. Both deployments running in the same namespace.

hsunwenfang:
Hi @ingvagabund 
Thank you for the explanation, allow me to clarify further.
So the goal here is to replace direct call of `RunDeschedulerStrategies()` with deployment.
Should the namespace for the descheduler deployments be 'kube-system' or `"e2e-" + strings.ToLower(t.Name())`?
And how is the descheduler servers in ns1 and ns2 interact with the deployments then?
thx!

test/e2e/e2e_leaderelection_test.go
```diff
@@ -33,9 +33,52 @@ import (
...
+func PodLifeTimePolicy(targetNamespace string, maxPodLifeTimeSeconds *uint) *api.DeschedulerPolicy {
+	return &api.DeschedulerPolicy{
+		Profiles: []api.DeschedulerProfile{
+			{
+				Name: "PodLifeTimeProfile",
+				PluginConfigs: []api.PluginConfig{
+					{
+						Name: podlifetime.PluginName,
+						Args: &podlifetime.PodLifeTimeArgs{
+							MaxPodLifeTimeSeconds: maxPodLifeTimeSeconds, // [TODO] The values to set
```
ingvagabund, line 52: 5 based on https://github.com/kubernetes-sigs/descheduler/blob/master/test/e2e/policy_leaderelection_a.yaml#L8C9-L8C33

ingvagabund:
> Should the namespace for the descheduler deployments be 'kube-system' or "e2e-" + strings.ToLower(t.Name())?

kube-system

> And how is the descheduler servers in ns1 and ns2 interact with the deployments then?

If you mean a descheduler deployment, then:
- descheduler server for ns1 will get replaced with one descheduler deployment
- descheduler server for ns2 will get replaced with another descheduler deployment

If you mean `leaderelection` deployments created through createDeployment:
- descheduler server for ns1 deployment will target ns1 in its policy to evict pods produced by deployment1
- descheduler server for ns2 deployment will target ns2 in its policy to evict pods produced by deployment1

> https://github.com/kubernetes-sigs/descheduler/blob/master/test/e2e/e2e_toomanyrestarts_test.go#L186
Seems rs is a returned struct not used by codes afterwards. Should L186-191 be removed?

Correct. This is a leftover I forgot to remove.

fanhaouu:
Unfortunately, when I initially submitted the code, I made all the adjustments at once without splitting them up. I merged many duplicate methods.

hsunwenfang:
Hi @ingvagabund 
Thanks for the help always
During testing, my minikube node failed to pull docker.io/library/descheduler 
And My docker account seems not authorized as well
`Trying to pull docker.io/library/descheduler:latest...
Error: initializing source docker://descheduler:latest: reading manifest latest in docker.io/library/descheduler: errors:
denied: requested access to the resource is denied
unauthorized: authentication required`
How should I auth for this image?
thx!

ingvagabund:
You can update [DESCHEDULER_IMAGE](https://github.com/kubernetes-sigs/descheduler/blob/master/test/run-e2e-tests.sh#L30) to point to `localhost/descheduler:latest`.

PR #1507, opened 2024-08-30
feat(helm): make securityContext conditional in Deployment and CronJob

bendikp:
This adds a conditional around `securityContext` in Deployment and CronJob template in order to be able to unset this property. In the same way `podSecurityContext` is handled.


a7i (approved):
/approve

PR #1508, opened 2024-09-03
[e2e] no test timeouts, produce the same image tag as in production

fanhaouu:
1、Avoid e2e test timeout, after migrating all e2e tests to containers, the execution time may increase, and e2e tests might time out；
2、Modify IMAGE_TAG to fix the version parsing issue;

### DIFF
--- a/test/run-e2e-tests.sh
+++ b/test/run-e2e-tests.sh
@@ -54,4 +54,4 @@ fi
 kubectl apply -f kubernetes/base/rbac.yaml
 
 PRJ_PREFIX="sigs.k8s.io/descheduler"
-go test ${PRJ_PREFIX}/test/e2e/ -v
+go test ${PRJ_PREFIX}/test/e2e/ -v -timeout 0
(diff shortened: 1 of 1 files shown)


fanhaouu:
@ingvagabund Hi, master, this is the first of the splited PRs. I will submit the remaining PRs sequentially. Need your approval.

ingvagabund:
Thank you for opening the PR.

> Modify IMAGE_TAG to fix the version parsing issue

Would you please more elaborate on the parsing issue? E.g. providing an example and an error that is produced.

fanhaouu:
> Thank you for opening the PR.
> 
> > Modify IMAGE_TAG to fix the version parsing issue
> 
> Would you please more elaborate on the parsing issue? E.g. providing an example and an error that is produced.
image


1、make test-e2e, VERSION="${IMAGE_TAG}"=v0.30.0-127-g0f1890e5c, 
image
2、The `VERSION` is eventually passed into `version.go`.
image
3、splitVersion return's minorVersion will be ""
https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/version/version.go#L72
4、This will cause `validateVersionCompatibility` to return an error.
https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/descheduler/descheduler.go#L299


We should ensure that the methods for obtaining the version in the e2e tests and in the Makefile are consistent.

PR #1525, opened 2024-09-25
first step in plugins declaring what informers they use, so that dry …

john7doe:
…run has all the expected resource types

Suspect that this is no longer needed:

https://github.com/john7doe/descheduler/blob/8a2b2eb37c479b6269cbf02fe8e5dd324c5695f2/pkg/descheduler/node/node.go#L53-L69

as we are now making sure the same set of listers are registered and we call `WaitForCacheSync` before the plugins run.


john7doe:
opened prematurely, sorry!

PR #1526, opened 2024-09-30
Make sure dry runs sees all the resources a normal run would do

john7doe:
This PR addresses that by implementing a more generic resource handling approach, ensuring that dry runs can access all resource types and objects as expected

I discovered an issue with dry-runs only seeing a subset of resources during execution when I was working on adding an option to [filter pods on pod disruption budgets](https://github.com/john7doe/descheduler/tree/john7doe/addPDBOption) (WIP, will PR later).

The root case was the order of the calls to create listers and the call to `WaitForCacheSync`.

I think this bug is the likely reason why this [special casing ](https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/descheduler/node/node.go#L53-L69)was added.  With the new code thats no longer needed (if reviews agree with that, then I can remove it as part of this PR).

The code also simplifies the number of changes need when a new type of resource is needed by a plugin.

### DIFF
--- a/pkg/descheduler/descheduler.go
+++ b/pkg/descheduler/descheduler.go
@@ -70,7 +70,7 @@ type profileRunner struct {
 
 type descheduler struct {
 	rs                     *options.DeschedulerServer
-	ir                     InformerResources
+	ir                     *informerResources
 	getPodsAssignedToNode  podutil.GetPodsAssignedToNodeFunc
 	sharedInformerFactory  informers.SharedInformerFactory
 	deschedulerPolicy      *api.DeschedulerPolicy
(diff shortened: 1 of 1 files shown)


ingvagabund:
This looks great. I like how cleaner the solution is.
/approve

ingvagabund:
> I think this bug is the likely reason why this [special casing ](https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/descheduler/node/node.go#L53-L69)was added. With the new code thats no longer needed (if reviews agree with that, then I can remove it as part of this PR).

Looks like that code got introduced 7 years ago through https://github.com/kubernetes-sigs/descheduler/commit/7d079813e581a72c90941d0e3c6463e5e38ce9f0: The cached client got implemented few years back: https://github.com/kubernetes-sigs/descheduler/pull/677.

PR #1536, opened 2024-10-26
feature(eviction): add event when EvictPod failed

googs1025:
/kind feature
- add event when EvictPod failed.
- add unit test coverage

When it comes to eviction, what we really care about are the errors that occur during evictions. So I tend to include events when an eviction fails. Besides metrics and application logs, events are also a valuable source of information. Typically, operations personnel respond to events to manage the cluster effectively.

### DIFF
--- a/pkg/descheduler/evictions/evictions.go
+++ b/pkg/descheduler/evictions/evictions.go
@@ -206,20 +206,21 @@ type (
 )
 
 type PodEvictor struct {
-	mu                         sync.RWMutex
-	client                     clientset.Interface
-	policyGroupVersion         string
-	dryRun                     bool
-	maxPodsToEvictPerNode      *uint
-	maxPodsToEvictPerNamespace *uint
-	maxPodsToEvictTotal        *uint
-	nodePodCount               nodePodEvictedCount
-	namespacePodCount          namespacePodEvictCount
-	totalPodCount              uint
-	metricsEnabled             bool
-	eventRecorder              events.EventRecorder
-	erCache                    *evictionRequestsCache
-	featureGates               featuregate.FeatureGate
+	mu                               sync.RWMutex
+	client                           clientset.Interface
+	policyGroupVersion               string
+	dryRun                           bool
+	evictionFailureEventNotification bool
+	maxPodsToEvictPerNode            *uint
+	maxPodsToEvictPerNamespace       *uint
+	maxPodsToEvictTotal              *uint
+	nodePodCount                     nodePodEvictedCount
+	namespacePodCount                namespacePodEvictCount
+	totalPodCount                    uint
+	metricsEnabled                   bool
+	eventRecorder                    events.EventRecorder
+	erCache                          *evictionRequestsCache
+	featureGates                     featuregate.FeatureGate
 
 	// registeredHandlers contains the registrations of all handlers. It's used to check if all handlers have finished syncing before the scheduling cycles start.
 	registeredHandlers []cache.ResourceEventHandlerRegistration
@@ -238,17 +239,18 @@ func NewPodEvictor(
 	}
 
 	podEvictor := &PodEvictor{
-		client:                     client,
-		eventRecorder:              eventRecorder,
-		policyGroupVersion:         options.policyGroupVersion,
-		dryRun:                     options.dryRun,
-		maxPodsToEvictPerNode:      options.maxPodsToEvictPerNode,
-		maxPodsToEvictPerNamespace: options.maxPodsToEvictPerNamespace,
-		maxPodsToEvictTotal:        options.maxPodsToEvictTotal,
-		metricsEnabled:             options.metricsEnabled,
-		nodePodCount:               make(nodePodEvictedCount),
-		namespacePodCount:          make(namespacePodEvictCount),
-		featureGates:               featureGates,
+		client:                           client,
+		eventRecorder:                    eventRecorder,
+		policyGroupVersion:               options.policyGroupVersion,
+		dryRun:                           options.dryRun,
+		evictionFailureEventNotification: options.evictionFailureEventNotification,
+		maxPodsToEvictPerNode:            options.maxPodsToEvictPerNode,
+		maxPodsToEvictPerNamespace:       options.maxPodsToEvictPerNamespace,
+		maxPodsToEvictTotal:              options.maxPodsToEvictTotal,
+		metricsEnabled:                   options.metricsEnabled,
+		nodePodCount:                     make(nodePodEvictedCount),
+		namespacePodCount:                make(namespacePodEvictCount),
+		featureGates:                     featureGates,
 	}
 
 	if featureGates.Enabled(features.EvictionsInBackground) {
@@ -481,6 +483,9 @@ func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod, opts EvictOptio
 		}
 		span.AddEvent("Eviction Failed", trace.WithAttributes(attribute.String("node", pod.Spec.NodeName), attribute.String("err", err.Error())))
 		klog.ErrorS(err, "Error evicting pod", "limit", *pe.maxPodsToEvictTotal)
+		if pe.evictionFailureEventNotification {
+			pe.eventRecorder.Eventf(pod, nil, v1.EventTypeWarning, "EvictionFailed", "Descheduled", "pod eviction from %v node by sigs.k8s.io/descheduler failed: total eviction limit exceeded (%v)", pod.Spec.NodeName, *pe.maxPodsToEvictTotal)
+		}
 		return err
 	}
 
@@ -492,6 +497,9 @@ func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod, opts EvictOptio
 			}
 			span.AddEvent("Eviction Failed", trace.WithAttributes(attribute.String("node", pod.Spec.NodeName), attribute.String("err", err.Error())))
 			klog.ErrorS(err, "Error evicting pod", "limit", *pe.maxPodsToEvictPerNode, "node", pod.Spec.NodeName)
+			if pe.evictionFailureEventNotification {
+				pe.eventRecorder.Eventf(pod, nil, v1.EventTypeWarning, "EvictionFailed", "Descheduled", "pod eviction from %v node by sigs.k8s.io/descheduler failed: node eviction limit exceeded (%v)", pod.Spec.NodeName, *pe.maxPodsToEvictPerNode)
+			}
 			return err
 		}
 	}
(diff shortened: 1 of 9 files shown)


pkg/descheduler/evictions/evictions_test.go
```diff
@@ -40,35 +45,57 @@ func TestEvictPod(t *testing.T) {
...
 		},
 	}
 
 	for _, test := range tests {
-		fakeClient := &fake.Clientset{}
-		fakeClient.Fake.AddReactor("list", "pods", func(action core.Action) (bool, runtime.Object, error) {
-			return true, &v1.PodList{Items: test.pods}, nil
+		t.Run(test.description, func(t *testing.T) {
+			fakeClient := fake.NewClientset(test.pods...)
+			fakeClient.PrependReactor("create", "pods/eviction", func(action core.Action) (handled bool, ret runtime.Object, err error) {
```
googs1025, line 81: The original `fakeClient.Fake.AddReactor("list", "pods", func(action core.Action) (bool, runtime.Object, error)` is not used in the test

pkg/descheduler/evictions/evictions.go
```diff
@@ -167,6 +170,7 @@ func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod, opts EvictOptio
 		if pe.metricsEnabled {
 			metrics.PodsEvicted.With(map[string]string{"result": "error", "strategy": opts.StrategyName, "namespace": pod.Namespace, "node": pod.Spec.NodeName, "profile": opts.ProfileName}).Inc()
 		}
+		pe.eventRecorder.Eventf(pod, nil, v1.EventTypeWarning, "EvictionFailed", "Descheduled", "pod evicted from %v node by sigs.k8s.io/descheduler failed: %v", pod.Spec.NodeName, err.Error())
```
ingvagabund, line 173: Nit: pod evicted -> pod eviction

googs1025: done

pkg/descheduler/evictions/evictions_test.go
```diff
@@ -40,35 +45,57 @@ func TestEvictPod(t *testing.T) {
...
+				if test.wantErrMsg == fmt.Sprintf("pod not found when evicting \"%s\": pods \"%s\" not found", test.evictedPod.Name, test.evictedPod.Name) {
```
ingvagabund, line 82: Worth turning the text into a constant given it's used on multiple places. To avoid unintentionally changing the text. E.g.
```go
const (
	notFoundText = "pod not found when evicting \"%s\": pods \"%s\" not found"
	tooManyRequests = "error when evicting pod (ignoring) \"%s\": Too many requests: too many requests"
)
```

pkg/descheduler/evictions/evictions_test.go
```diff
@@ -40,35 +45,57 @@ func TestEvictPod(t *testing.T) {
...
+			pods:        []runtime.Object{test.BuildTestPod("p2", 400, 0, "node1", nil), test.BuildTestPod("p3", 450, 0, "node1", nil)},
+			wantErr:     true,
+			wantErrMsg:  fmt.Sprintf("pod not found when evicting \"%s\": pods \"%s\" not found", pod1.Name, pod1.Name),
+		},
+		{
+			description: "test pod eviction - pod absent (too many requests error)",
+			node:        node1,
+			evictedPod:  pod1,
+			pods:        []runtime.Object{test.BuildTestPod("p2", 400, 0, "node1", nil), test.BuildTestPod("p3", 450, 0, "node1", nil)},
+			wantErr:     true,
```
ingvagabund, line 73: `wantErr` is currently `true` in both cases. Is the new field really needed?

pkg/descheduler/evictions/evictions_test.go
```diff
@@ -40,35 +45,57 @@ func TestEvictPod(t *testing.T) {
...
+			wantErrMsg:  fmt.Sprintf("error when evicting pod (ignoring) \"%s\": Too many requests: too many requests", pod1.Name),
 		},
 	}
 
 	for _, test := range tests {
-		fakeClient := &fake.Clientset{}
-		fakeClient.Fake.AddReactor("list", "pods", func(action core.Action) (bool, runtime.Object, error) {
-			return true, &v1.PodList{Items: test.pods}, nil
+		t.Run(test.description, func(t *testing.T) {
```
ingvagabund, line 79: +1 for `t.Run`

pkg/descheduler/evictions/evictions_test.go
```diff
@@ -117,51 +144,152 @@ func TestPodTypes(t *testing.T) {
...
+	type podEvictorTest struct {
+		description                string
+		pod                        *v1.Pod
+		dryRun                     bool
+		maxPodsToEvictTotal        *uint
+		maxPodsToEvictPerNode      *uint
+		maxPodsToEvictPerNamespace *uint
+		expectedNodeEvictions      uint
+		expectedTotalEvictions     uint
+		expectedError              error
```
ingvagabund, line 158: s/expectedError/err as expected is used for comparison with actual. s/expectedError/actualErr will work as well.

pkg/descheduler/evictions/evictions_test.go
```diff
@@ -40,35 +45,57 @@ func TestEvictPod(t *testing.T) {
...
 	}{
 		{
 			description: "test pod eviction - pod present",
 			node:        node1,
-			pod:         pod1,
-			pods:        []v1.Pod{*pod1},
-			want:        nil,
+			evictedPod:  pod1,
+			pods:        []runtime.Object{pod1},
+			wantErr:     false,
```
googs1025, line 58: https://github.com/kubernetes-sigs/descheduler/pull/1536/files#r1837802730

we have other case like this

pkg/descheduler/evictions/evictions_test.go
googs1025: replace err -> actualErr

pkg/descheduler/evictions/evictions_test.go
googs1025: https://github.com/kubernetes-sigs/descheduler/pull/1536/files#r1837948302

pkg/descheduler/evictions/evictions_test.go
ingvagabund: Right. I missed this one. Me slowly disappearing ...

pkg/descheduler/evictions/evictions_test.go
googs1025: done

pkg/descheduler/evictions/evictions_test.go
```diff
@@ -33,42 +38,69 @@ import (
...
 	}
 
 	for _, test := range tests {
-		fakeClient := &fake.Clientset{}
-		fakeClient.Fake.AddReactor("list", "pods", func(action core.Action) (bool, runtime.Object, error) {
-			return true, &v1.PodList{Items: test.pods}, nil
+		t.Run(test.description, func(t *testing.T) {
+			fakeClient := fake.NewClientset(test.pods...)
+			fakeClient.PrependReactor("create", "pods/eviction", func(action core.Action) (handled bool, ret runtime.Object, err error) {
+				if test.wantErrMsg == fmt.Sprintf(notFoundText, test.evictedPod.Name, test.evictedPod.Name) {
```
ingvagabund, line 87: What about to pass `k8serrors.XXX` directly? I.e.
```
[]struct {
	description string
	node        *v1.Node
	evictedPod  *v1.Pod
	pods        []runtime.Object
	wantErr     error
}{
	...
	{
		description: "test pod eviction - pod absent (not found error)",
		node:        node1,
		evictedPod:  pod1,
		pods:        []runtime.Object{test.BuildTestPod("p2", 400, 0, "node1", nil), test.BuildTestPod("p3", 450, 0, "node1", nil)},
		wantErr:  k8serrors.NewNotFound(v1.Resource("pods"), pod1.Name),
	},
}
...
fakeClient.PrependReactor("create", "pods/eviction", func(action core.Action) (handled bool, ret runtime.Object, err error) {
	return true, nil, test.wantErr
})
```
The original wantErr bool can be turned into `wantErr != nil`.

pkg/descheduler/evictions/evictions_test.go
```diff
@@ -117,51 +149,152 @@ func TestPodTypes(t *testing.T) {
...
-		t.Errorf("Expected a pod eviction, got an eviction error instead: %v", err)
-	}
+	for _, test := range tests {
+		t.Run(test.description, func(t *testing.T) {
+			fakeClient := fake.NewSimpleClientset(pod1)
+			fakeClient.PrependReactor("create", "pods/eviction", func(action core.Action) (handled bool, ret runtime.Object, err error) {
+				if test.expectedError != nil {
+					return true, nil, test.expectedError
+				}
+				return true, nil, nil
```
ingvagabund, line 245: The whole function body can be reduced to `return true, nil, test.expectedError`.

ingvagabund (changes requested):

pkg/descheduler/evictions/evictions_test.go
googs1025: done

pkg/descheduler/evictions/evictions_test.go
googs1025: Because we return using the `fmt.Error`, we cannot directly use `k8serrors.XXX`, but we can simplify it to directly use `fmt.Error` to return

https://github.com/kubernetes-sigs/descheduler/blob/da52983b2773cfb5d014a6d3f7df0d0a0b3bf6bf/pkg/descheduler/evictions/evictions.go#L213-L221

pkg/descheduler/evictions/evictions.go
```diff
@@ -481,6 +481,7 @@ func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod, opts EvictOptio
 		}
 		span.AddEvent("Eviction Failed", trace.WithAttributes(attribute.String("node", pod.Spec.NodeName), attribute.String("err", err.Error())))
 		klog.ErrorS(err, "Error evicting pod", "limit", *pe.maxPodsToEvictTotal)
+		pe.eventRecorder.Eventf(pod, nil, v1.EventTypeWarning, "EvictionFailed", "Descheduled", "pod eviction from %v node by sigs.k8s.io/descheduler failed: total eviction limit exceeded (%v)", pod.Spec.NodeName, *pe.maxPodsToEvictTotal)
```
ingvagabund, line 484: If these events get produced by default in a cluster with many pods the cluster gets flooded with events. This needs to be disabled by default and enabled if needed.

googs1025: Ok, this sounds more reasonable. I will add a new field in the DeschedulerConfiguration. Because we have similar monitoring cluster-specific failure events (ex: scheduling failure, eviction failure, etc.), this can quickly let cluster administrators know important event information.

googs1025: i will handle this in weekend

googs1025: done

PR #1543, opened 2024-11-13
nodeutilization: NodeUtilization: make pod utilization extraction configurable

ingvagabund:
Extracting requested resource is not the only way to get pod's utilization. Consuming actual usage from metrics is another option. Thus, abstracting `NodeUtilization` so the utilization function can be replaced as needed.

### DIFF
--- a/pkg/descheduler/node/node.go
+++ b/pkg/descheduler/node/node.go
@@ -218,7 +218,12 @@ func fitsRequest(nodeIndexer podutil.GetPodsAssignedToNodeFunc, pod *v1.Pod, nod
 		resourceNames = append(resourceNames, name)
 	}
 
-	availableResources, err := nodeAvailableResources(nodeIndexer, node, resourceNames)
+	availableResources, err := nodeAvailableResources(nodeIndexer, node, resourceNames,
+		func(pod *v1.Pod) (v1.ResourceList, error) {
+			req, _ := utils.PodRequestsAndLimits(pod)
+			return req, nil
+		},
+	)
 	if err != nil {
 		return false, err
 	}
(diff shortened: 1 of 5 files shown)


ingvagabund:
Proceeding with a self-review due to a limited availability of reviewers, in order to maintain momentum and ensure timely progress on this project. While immediate peer review is ideal, continuing development is essential to avoid delays and meet our project goals. Contributors are still encouraged to perform a post-merge review to identify any areas for further improvement or to address potential regressions that may have been missed in the initial review. Post-merge feedback remains valuable, as it helps us enhance code quality, prevent issues from accumulating, and keep our workflow agile.

pkg/descheduler/node/node.go
```diff
@@ -218,7 +218,12 @@ func fitsRequest(nodeIndexer podutil.GetPodsAssignedToNodeFunc, pod *v1.Pod, nod
 		resourceNames = append(resourceNames, name)
 	}
 
-	availableResources, err := nodeAvailableResources(nodeIndexer, node, resourceNames)
+	availableResources, err := nodeAvailableResources(nodeIndexer, node, resourceNames,
+		func(pod *v1.Pod) (v1.ResourceList, error) {
+			req, _ := utils.PodRequestsAndLimits(pod)
```
atiratree, line 223: I suppose that the new future implementations of this function should ensure that there is no overlap between the pod requests and the other sources.

atiratree (commented):
/lgtm

pkg/descheduler/node/node.go
ingvagabund: It's every invoker's responsibility to make sure the returned resource utilization is valid. E.g. metrics might do aggregation of resources to provide a meaningful value.

PR #1544, opened 2024-11-13
nodeutilization: evictPodsFromSourceNodes: iterate through existing resources

ingvagabund:
So `evictPodsFromSourceNodes` can process any list of resources.

Built on top of https://github.com/kubernetes-sigs/descheduler/pull/1543.

### DIFF
--- a/pkg/framework/plugins/nodeutilization/nodeutilization.go
+++ b/pkg/framework/plugins/nodeutilization/nodeutilization.go
@@ -240,10 +240,9 @@ func evictPodsFromSourceNodes(
 	continueEviction continueEvictionCond,
 ) {
 	// upper bound on total number of pods/cpu/memory and optional extended resources to be moved
-	totalAvailableUsage := map[v1.ResourceName]*resource.Quantity{
-		v1.ResourcePods:   {},
-		v1.ResourceCPU:    {},
-		v1.ResourceMemory: {},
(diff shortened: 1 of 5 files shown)


pkg/framework/plugins/nodeutilization/nodeutilization.go
```diff
@@ -349,9 +353,15 @@ func evictPods(
 
 				keysAndValues := []interface{}{
 					"node", nodeInfo.node.Name,
-					"CPU", nodeInfo.usage[v1.ResourceCPU].MilliValue(),
-					"Mem", nodeInfo.usage[v1.ResourceMemory].Value(),
-					"Pods", nodeInfo.usage[v1.ResourcePods].Value(),
+				}
+				if quantity, exists := nodeInfo.usage[v1.ResourceCPU]; exists {
```
atiratree, line 357: Why do we use nodeInfo here? Can we us totalAvailableUsage variable instead? And extract this block for resolving keysAndValues for logging into a function?

ingvagabund: `totalAvailableUsage` includes available usage across all overutilized nodes. `nodeInfo.usage` is node specific.

atiratree: Ack, thanks for the explanation.

atiratree:
lgtm

PR #1545, opened 2024-11-13
nodeutilization: separate code responsible for requested resource extraction into a dedicated usage client

ingvagabund:
Turning a usage client into an interface allows to implement other kinds of usage clients like actual usage or prometheus based resource collection.

Built on top of https://github.com/kubernetes-sigs/descheduler/pull/1544.


pkg/framework/plugins/nodeutilization/lownodeutilization.go
```diff
@@ -102,22 +106,15 @@ func (l *LowNodeUtilization) Name() string {
 
 // Balance extension point implementation for the plugin
 func (l *LowNodeUtilization) Balance(ctx context.Context, nodes []*v1.Node) *frameworktypes.Status {
-	nodeUsage, err := getNodeUsage(nodes, l.resourceNames, l.handle.GetPodsAssignedToNodeFunc())
-	if err != nil {
+	if err := l.usageSnapshot.capture(nodes); err != nil {
```
atiratree, line 109: IMO it would be better not to store the data in the usageSnapshot/usageClient (the naming is a bit confusing). And return a new struct when we do the capture. Then we can pass the the struct around without unnecessarily binding it with the client.

pkg/framework/plugins/nodeutilization/nodeutilization.go
```diff
@@ -238,10 +219,11 @@ func evictPodsFromSourceNodes(
 	podFilter func(pod *v1.Pod) bool,
 	resourceNames []v1.ResourceName,
 	continueEviction continueEvictionCond,
+	usageSnapshot usageClient,
 ) {
 	// upper bound on total number of pods/cpu/memory and optional extended resources to be moved
 	totalAvailableUsage := map[v1.ResourceName]*resource.Quantity{}
-	for _, resourceName := range resourceNames {
+	for _, resourceName := range usageSnapshot.resourceNames() {
```
atiratree, line 226: Wouldn't it be better to use only one source of truth for the resource names. Either from the client or from the plugin struct?

pkg/framework/plugins/nodeutilization/nodeutilization.go
```diff
@@ -250,6 +232,10 @@ func evictPodsFromSourceNodes(
 		taintsOfDestinationNodes[node.node.Name] = node.node.Spec.Taints
 
 		for _, name := range resourceNames {
```
atiratree, line 234: for example we use a different reference here

pkg/framework/plugins/nodeutilization/nodeutilization.go
atiratree: Is it useful to get the resourceNames from the client/snapshot? Can it change over time?

pkg/framework/plugins/nodeutilization/lownodeutilization.go
ingvagabund: The current implementation of the client based on kubernetes metrics does not capture pod usage. It reads the usage as low node utilization plugin goes and asks for it. Which would make the captured snapshot only partial.

pkg/framework/plugins/nodeutilization/nodeutilization.go
atiratree: I see. I think it would be better not channel it through the usageSnapshot/usageClient and use directly. We can still pass it to the client but don't have to expose it through it.

pkg/framework/plugins/nodeutilization/lownodeutilization.go
ingvagabund: I agree using a word snapshot does not fully correspond to the intention. Would `sync` work instead of `capture`?

ingvagabund: I will rename the variable to usageClient. I would like to avoid any splitting atm. I don't think it's worth doing it given the usage client is not expected to be exported elsewhere.

Both low/high node utilization plugins are expected to invoke capture/sync right after `Balance` method is invoked. There's no cache invalidation as each `Balance` is expected to get the latest data.

atiratree: +1, and it would be good to document that in the interface

ingvagabund: Done

pkg/framework/plugins/nodeutilization/usageclients.go
```diff
@@ -0,0 +1,110 @@
...
+)
+
+type usageClient interface {
+	nodeUtilization(node string) map[v1.ResourceName]*resource.Quantity
+	nodes() []*v1.Node
+	pods(node string) []*v1.Pod
+	// Both low/high node utilization plugins are expected to invoke sync right
+	// after Balance method is invoked. There's no cache invalidation so each
+	// Balance is expected to get the latest data by invoking sync.
+	sync(nodes []*v1.Node) error
```
atiratree, line 36: nit: let's put the sync first or last

pkg/framework/plugins/nodeutilization/usageclients.go
```diff
@@ -0,0 +1,110 @@
...
+	podUsage(pod *v1.Pod) (map[v1.ResourceName]*resource.Quantity, error)
+}
+
+type requestedUsageClient struct {
+	_resourceNames        []v1.ResourceName
+	getPodsAssignedToNode podutil.GetPodsAssignedToNodeFunc
+
+	_nodes           []*v1.Node
```
atiratree, line 44: not a big fan of the prefix, what about `cachedNodes`, etc.?

pkg/framework/plugins/nodeutilization/usageclients.go
```diff
@@ -0,0 +1,110 @@
...
+	utilptr "k8s.io/utils/ptr"
+	nodeutil "sigs.k8s.io/descheduler/pkg/descheduler/node"
+	podutil "sigs.k8s.io/descheduler/pkg/descheduler/pod"
+	"sigs.k8s.io/descheduler/pkg/utils"
+)
+
+type usageClient interface {
+	nodeUtilization(node string) map[v1.ResourceName]*resource.Quantity
+	nodes() []*v1.Node
+	pods(node string) []*v1.Pod
```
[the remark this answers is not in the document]
ingvagabund: Skipping some nodes from processing might go unnoticed. Better to error. I will update the usage client so there's no need to return a list of cached nodes.

pkg/framework/plugins/nodeutilization/usageclients.go
```diff
@@ -0,0 +1,110 @@
...
+	// Both low/high node utilization plugins are expected to invoke sync right
+	// after Balance method is invoked. There's no cache invalidation so each
+	// Balance is expected to get the latest data by invoking sync.
+	sync(nodes []*v1.Node) error
+	podUsage(pod *v1.Pod) (map[v1.ResourceName]*resource.Quantity, error)
+}
+
+type requestedUsageClient struct {
+	_resourceNames        []v1.ResourceName
```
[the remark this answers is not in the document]
ingvagabund: Done

pkg/framework/plugins/nodeutilization/usageclients.go
ingvagabund: Done. Put as first.

pkg/framework/plugins/nodeutilization/usageclients.go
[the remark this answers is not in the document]
atiratree: ok, thx for the update

atiratree:
lgtm

PR #1547, opened 2024-11-14
Fix: Critical Vulnerabilities | EOL Increased

sagar-18:
1. Scanned using trivy.
2. updated golang to v1.23.1 
3. EOL Increased

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-FROM golang:1.23.1
+FROM golang:1.23.3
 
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-FROM golang:1.22.7
+FROM golang:1.23.1
```
haosdent, line 14: Should we use the latest minor version like 1.23.3

haosdent:
I guess need to update https://github.com/kubernetes-sigs/descheduler/blob/af8a7445a4ef6d73785150d088d40c0ced56358a/hack/lib/go.sh#L22 as well

PR #1549, opened 2024-11-14
nodeutilization: usage2KeysAndValues for constructing a key:value list for InfoS printing resource usage

ingvagabund:
Code deduplication

### DIFF
--- a/pkg/framework/plugins/nodeutilization/nodeutilization.go
+++ b/pkg/framework/plugins/nodeutilization/nodeutilization.go
@@ -226,6 +226,26 @@ func classifyNodes(
 	return lowNodes, highNodes
 }
 
+func usageToKeysAndValues(usage map[v1.ResourceName]*resource.Quantity) []interface{} {
+	// log message in one line
+	keysAndValues := []interface{}{}
+	if quantity, exists := usage[v1.ResourceCPU]; exists {
+		keysAndValues = append(keysAndValues, "CPU", quantity.MilliValue())
+	}
+	if quantity, exists := usage[v1.ResourceMemory]; exists {
+		keysAndValues = append(keysAndValues, "Mem", quantity.Value())
+	}
+	if quantity, exists := usage[v1.ResourcePods]; exists {
+		keysAndValues = append(keysAndValues, "Pods", quantity.Value())
+	}
+	for name := range usage {
+		if !node.IsBasicResource(name) {
+			keysAndValues = append(keysAndValues, string(name), usage[name].Value())
+		}
+	}
+	return keysAndValues
+}
+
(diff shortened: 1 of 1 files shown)


pkg/framework/plugins/nodeutilization/nodeutilization.go
```diff
@@ -226,6 +226,26 @@ func classifyNodes(
 	return lowNodes, highNodes
 }
 
+func usage2KeysAndValues(usage map[v1.ResourceName]*resource.Quantity) []interface{} {
```
atiratree, line 229:
```suggestion
func usageToKeysAndValues(usage map[v1.ResourceName]*resource.Quantity) []interface{} {
```

better to be consistent and not use number shortcuts in the names

pkg/framework/plugins/nodeutilization/nodeutilization.go
```diff
@@ -351,24 +356,8 @@ func evictPods(
...
-					keysAndValues = append(keysAndValues, "Pods", quantity.Value())
-				}
-				for name := range totalAvailableUsage {
-					if !nodeutil.IsBasicResource(name) {
-						keysAndValues = append(keysAndValues, string(name), totalAvailableUsage[name].Value())
-					}
-				}
-
+				keysAndValues := usage2KeysAndValues(nodeInfo.usage)
+				keysAndValues = append(keysAndValues, "node", nodeInfo.node.Name)
```
atiratree, line 360: might be better to keep the same order

pkg/framework/plugins/nodeutilization/nodeutilization.go
```diff
@@ -351,24 +356,8 @@ func evictPods(
...
-				if quantity, exists := nodeInfo.usage[v1.ResourceCPU]; exists {
-					keysAndValues = append(keysAndValues, "CPU", quantity.MilliValue())
-				}
-				if quantity, exists := nodeInfo.usage[v1.ResourceMemory]; exists {
-					keysAndValues = append(keysAndValues, "Mem", quantity.Value())
-				}
-				if quantity, exists := nodeInfo.usage[v1.ResourcePods]; exists {
-					keysAndValues = append(keysAndValues, "Pods", quantity.Value())
-				}
-				for name := range totalAvailableUsage {
```
atiratree, line 366: so the mixing of node and total stats was incorrect

ingvagabund: Yeah. I probably misunderstood your previous comment.

pkg/framework/plugins/nodeutilization/nodeutilization.go
ingvagabund: Done

atiratree:
lgtm

Issue #225, opened 2020-02-03
Use actual node resource utilization in the strategy "LowNodeUtilization"


zhiyxu:
Currently, pods' request resource requirements are considered for computing node resource utilization in the strategy "LowNodeUtilization", is it more rational to use actual node resource utilization as judgement?

It is common that resource limit of pod is larger than request, and after the scheduling of default scheduler (by resource request), It's probably that cluster is balanced. But the actual resources usage of pod could be much larger than request, which may lead some nodes under pressure.

So is it much more reasonable to combine with metrics server and use actual node resource utilization.

seanmalloy:
@zhiyxu it looks like you are not the first person to request this feature. See the discussions in #123 , #118, and #7. Based on the discussions in those issues it looks like the descheduler `LowNodeUtilization` strategy is still using requests because this aligns with how the k8s scheduler works. Also, this feature is mentioned in the [roadmap](https://github.com/kubernetes-sigs/descheduler#roadmap).

@damemi @aveshagarwal @ravisantoshgudimetla has anything changed recently to enable the k8s scheduler to use real load metrics during scheduling? For example could the new scheduler framework some how enable this feature in the scheduler? Maybe a custom plugin using the scheduler framework could be created to take real load metrics in account?

zhiyxu:
@seanmalloy @ravisantoshgudimetla @damemi @aveshagarwal Any update or plan about this feature?

kangtiann:
+1, We need this feature too.

If we can make a PR for this ?

@seanmalloy @ravisantoshgudimetla @damemi @aveshagarwal

seanmalloy:
@zhiyxu and @kangtiann here are my initial thoughts on what the API spec might look like. Please let me know what you think. I'm pretty confident the v1alpha2 `LowNodeUtilization` strategy will need to be adjusted.

I believe it would be a good idea to write a proposal for this and have SIG scheduling review this.

Create a  new v1alpha1 `LowNodeAllocation` strategy. This strategy will work identically to the request based v1alpha `LowNodeUtilization` strategy. The usage of the work allocation is inspired by the discussion in #7. 
```
apiVersion: "descheduler/v1alpha1"
kind: "DeschedulerPolicy"
strategies:
  "LowNodeAllocation":
     enabled: true
     params:
       nodeResourceAllocatedThresholds:
         thresholds:
           "cpu" : 20
           "memory": 20
           "pods": 20
         targetThresholds:
[...]

PR #1555, opened 2024-11-15
Use actual node resource utilization by consuming kubernetes metrics

ingvagabund:
Extend LowNodeUtilization with getting nodes and pod usage from kubernetes metrics.

Policy configuration:
```yaml
kind: "DeschedulerPolicy"
metricsCollector:
  enabled: true
profiles:
- name: ProfileName
  pluginConfig:
  - name: "LowNodeUtilization"
    args:
      thresholds:
        "cpu" : 20
        "memory": 20
        "pods": 20
      targetThresholds:
        "cpu" : 50
        "memory": 50
        "pods": 50
      metricsUtilization:
        metricsServer: true
  plugins:
    balance:
      enabled:
        - "LowNodeUtilization"
```

Fixes: https://github.com/kubernetes-sigs/descheduler/issues/225


pkg/descheduler/metricscollector/metricscollector.go
```diff
@@ -0,0 +1,132 @@
...
+
+func (mc *MetricsCollector) MetricsClient() metricsclient.Interface {
+	return mc.metricsClientset
+}
+
+func (mc *MetricsCollector) Collect(ctx context.Context) error {
+	if mc == nil {
+		return fmt.Errorf("metrics collector not initialized")
+	}
+	mc.mu.Lock()
```
jklaw90, line 99: seems expensive to lock here when we list nodes and metrics for each node. not sure what the timing is like but our window is only 5 seconds. 

perhaps make a new map  then lock and update all at once?

pkg/framework/plugins/nodeutilization/usageclients.go
```diff
@@ -100,3 +103,84 @@ func (s *requestedUsageClient) sync(nodes []*v1.Node) error {
 
 	return nil
 }
+
+type actualUsageClient struct {
+	resourceNames         []v1.ResourceName
+	getPodsAssignedToNode podutil.GetPodsAssignedToNodeFunc
+	metricsCollector      *metricscollector.MetricsCollector
+
+	_pods            map[string][]*v1.Pod
```
jklaw90, line 112: do we need `_` already not exported anyways?

ingvagabund: `actualUsageClient` already defines `pods` as a method. Prefixing a private variable with `_` is a common practice. Yet, `internalPodList` or similar would do as well.

pkg/descheduler/metricscollector/metricscollector.go
ingvagabund: `sync.RWMutex` might help?

pkg/descheduler/metricscollector/metricscollector.go
```diff
@@ -0,0 +1,132 @@
...
+		clientset:        clientset,
+		metricsClientset: metricsClientset,
+		nodes:            make(map[string]map[v1.ResourceName]*resource.Quantity),
+	}
+}
+
+func (mc *MetricsCollector) Run(ctx context.Context) {
+	wait.NonSlidingUntil(func() {
+		mc.Collect(ctx)
+	}, 5*time.Second, ctx.Done())
```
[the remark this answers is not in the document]
ingvagabund: I did not want to spawn the initial implementation with additional configuration. The API can be extended at any point if 5 seconds interval is not sufficient. We could wait for feedback? Additionally, it might help to wait for e.g. 20 seconds to collect more samples before the descheduling cycle starts to soften the captured cpu/memory utilization in case some nodes are wildly changing their utilization. The next iteration could extend the API with:
```
metricsCollector:
  collectionInterval: 10s # collect a new sample every 10s
  initialDelay: 50s # wait for 50s before evicting to soften the captured cpu/memory utilization
  initialSamples: 5 # collect at least 5 samples before evicting to soften the captured cpu/memory utilization
```

pkg/descheduler/metricscollector/metricscollector.go
```diff
@@ -0,0 +1,132 @@
...
+
+func NewMetricsCollector(clientset kubernetes.Interface, metricsClientset metricsclient.Interface) *MetricsCollector {
+	return &MetricsCollector{
+		clientset:        clientset,
+		metricsClientset: metricsClientset,
+		nodes:            make(map[string]map[v1.ResourceName]*resource.Quantity),
+	}
+}
+
+func (mc *MetricsCollector) Run(ctx context.Context) {
```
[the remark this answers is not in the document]
ingvagabund: I like the idea of making the developer facing API simpler. In general, I tend to invoke a go routine explicitly so I am aware there's a go routine. Rather then hiding it. Which I would like to avoid in this case.

> Could be nice to fail on boot if their metrics server isn't available?

We could replace `PollWithContext` with `wait.PollWithContext` and set the timeout to e.g. 1 minute? To give the metric server a fighting chance to come up. This could be another configurable option in the next iteration.

pkg/descheduler/metricscollector/metricscollector.go
ingvagabund: MetricsCollector is currently listing all nodes. In case a user specifies a node selector the collector will store more data than needed. This needs to be addressed.

ingvagabund: Adding `AllNodesUsage()` to get metrics from all nodes at once to reduce the number of locking/unlocking.

pkg/descheduler/metricscollector/metricscollector.go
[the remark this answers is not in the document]
ingvagabund: Replacing `PollUntilWithContext` with `PollWithContext`.

pkg/descheduler/metricscollector/metricscollector.go
jklaw90: Ahh sorry I was trying to say only lock on after the network calls.

pkg/framework/plugins/nodeutilization/lownodeutilization_test.go
```diff
@@ -847,71 +1000,211 @@ func TestLowNodeUtilization(t *testing.T) {
...
+					metricsClientset.Tracker().Create(podsgvr, podmetrics, podmetrics.Namespace)
+				}
+
+				fakeClient := fake.NewSimpleClientset(objs...)
+
+				collector := metricscollector.NewMetricsCollector(fakeClient, metricsClientset, "")
+				err := collector.Collect(ctx)
+				if err != nil {
+					t.Fatalf("unable to collect metrics: %v", err)
+				}
```
atiratree, line 1153: shouldn't this be gated?

pkg/descheduler/metricscollector/metricscollector.go
```diff
@@ -0,0 +1,142 @@
...
+	nodeSelector     string
+
+	nodes map[string]map[v1.ResourceName]*resource.Quantity
+
+	mu sync.RWMutex
+	// hasSynced signals at least one sync succeeded
+	hasSynced bool
+}
+
+func NewMetricsCollector(clientset kubernetes.Interface, metricsClientset metricsclient.Interface, nodeSelector string) *MetricsCollector {
```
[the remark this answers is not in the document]
ingvagabund: The metrics collector is expected to always return cpu and memory. Do you foresee this might not be the case?

pkg/descheduler/metricscollector/metricscollector.go
```diff
@@ -0,0 +1,142 @@
...
+
+func (mc *MetricsCollector) Run(ctx context.Context) {
+	wait.NonSlidingUntil(func() {
+		mc.Collect(ctx)
+	}, 5*time.Second, ctx.Done())
+}
+
+func weightedAverage(prevValue, value int64) int64 {
+	return int64(math.Floor(beta*float64(prevValue) + (1-beta)*float64(value)))
+}
```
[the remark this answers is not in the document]
ingvagabund: With `math.Round` instead of `math.Floor` the final values stops at 492.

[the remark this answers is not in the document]
ingvagabund: With `math.Random` the value stops at 496.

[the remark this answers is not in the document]
ingvagabund: Even setting the scale from Milli to Micro and going from 1400 to 900 the value stops at 901.

pkg/framework/plugins/nodeutilization/lownodeutilization_test.go
ingvagabund: Done

pkg/descheduler/metricscollector/metricscollector.go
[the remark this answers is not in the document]
ingvagabund: Keeping the scale to Micro to reduce the difference

pkg/framework/plugins/nodeutilization/usageclients.go
```diff
@@ -100,3 +104,89 @@ func (s *requestedUsageClient) sync(nodes []*v1.Node) error {
...
+	}
+
+	return totalUsage, nil
+}
+
+func (client *actualUsageClient) sync(nodes []*v1.Node) error {
+	client._nodeUtilization = make(map[string]map[v1.ResourceName]*resource.Quantity)
+	client._pods = make(map[string][]*v1.Pod)
+
+	nodesUsage, err := client.metricsCollector.AllNodesUsage()
```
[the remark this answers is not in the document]
ingvagabund: Done

pkg/descheduler/metricscollector/metricscollector.go
```diff
@@ -0,0 +1,142 @@
...
+
+	return allNodesUsage, nil
+}
+
+func (mc *MetricsCollector) NodeUsage(node *v1.Node) (map[v1.ResourceName]*resource.Quantity, error) {
+	mc.mu.RLock()
+	defer mc.mu.RUnlock()
+
+	if _, exists := mc.nodes[node.Name]; !exists {
+		klog.V(4).Infof("unable to find node %q in the collected metrics", node.Name)
```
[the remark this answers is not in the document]
ingvagabund: Done

pkg/descheduler/metricscollector/metricscollector.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+func (mc *MetricsCollector) MetricsClient() metricsclient.Interface {
+	return mc.metricsClientset
+}
+
+func (mc *MetricsCollector) Collect(ctx context.Context) error {
+	mc.mu.Lock()
+	defer mc.mu.Unlock()
+	nodes, err := mc.clientset.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{LabelSelector: mc.nodeSelector})
```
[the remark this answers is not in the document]
ingvagabund: Done

pkg/descheduler/metricscollector/metricscollector.go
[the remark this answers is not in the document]
atiratree: I am fine with the Exponentially Weighted Average convergence. But maybe we could detect that we have stopped converging and just use the new value?

pkg/framework/plugins/nodeutilization/usageclients.go
```diff
@@ -100,3 +104,95 @@ func (s *requestedUsageClient) sync(nodes []*v1.Node) error {
...
+		if err != nil {
+			klog.V(2).InfoS("Node will not be processed, error accessing its pods", "node", klog.KObj(node), "err", err)
+			return fmt.Errorf("error accessing %q node's pods: %v", node.Name, err)
+		}
+
+		nodeUsage, ok := nodesUsage[node.Name]
+		if !ok {
+			return fmt.Errorf("unable to find node %q in the collected metrics", node.Name)
+		}
+		if _, exists := nodeUsage[v1.ResourceCPU]; !exists {
```
atiratree, line 184: shouldn't we intersect this with the client.resourceNames similar to what we do in podUsage?

pkg/descheduler/metricscollector/metricscollector.go
```diff
@@ -0,0 +1,144 @@
...
+				v1.ResourceMemory: utilptr.To[resource.Quantity](metrics.Usage.Memory().DeepCopy()),
+			}
+		} else {
+			// get MilliValue to reduce loss of precision
+			mc.nodes[node.Name][v1.ResourceCPU].SetScaled(
+				weightedAverage(mc.nodes[node.Name][v1.ResourceCPU].ScaledValue(resource.Micro), metrics.Usage.Cpu().ScaledValue(resource.Micro)),
+				resource.Micro,
+			)
+			mc.nodes[node.Name][v1.ResourceMemory].SetScaled(
+				weightedAverage(mc.nodes[node.Name][v1.ResourceMemory].ScaledValue(resource.Micro), metrics.Usage.Memory().ScaledValue(resource.Micro)),
```
atiratree, line 136: This seems a bit dangerous for the memory. We should make sure we do not cause overflow. Isn't just calling Value enough here?

pkg/descheduler/metricscollector/metricscollector.go
```diff
@@ -0,0 +1,144 @@
...
+
+		if _, exists := mc.nodes[node.Name]; !exists {
+			mc.nodes[node.Name] = map[v1.ResourceName]*resource.Quantity{
+				v1.ResourceCPU:    utilptr.To[resource.Quantity](metrics.Usage.Cpu().DeepCopy()),
+				v1.ResourceMemory: utilptr.To[resource.Quantity](metrics.Usage.Memory().DeepCopy()),
+			}
+		} else {
+			// get MilliValue to reduce loss of precision
+			mc.nodes[node.Name][v1.ResourceCPU].SetScaled(
+				weightedAverage(mc.nodes[node.Name][v1.ResourceCPU].ScaledValue(resource.Micro), metrics.Usage.Cpu().ScaledValue(resource.Micro)),
```
atiratree, line 132: and mili on the CPU?

pkg/descheduler/metricscollector/metricscollector.go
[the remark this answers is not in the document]
atiratree: No, I just meant to have the correct intersection between the resourceNames passed to the actualUsageClient and the real capability.

pkg/descheduler/metricscollector/metricscollector.go
ingvagabund: Good point.

pkg/descheduler/metricscollector/metricscollector.go
ingvagabund: Going back to milli gives 5 as delta, switching to `Value()` for memory gives 5 as delta too.

pkg/framework/plugins/nodeutilization/lownodeutilization_test.go
```diff
@@ -847,71 +1000,214 @@ func TestLowNodeUtilization(t *testing.T) {
...
-					getAction := action.(core.CreateAction)
-					obj := getAction.GetObject()
-					if eviction, ok := obj.(*policy.Eviction); ok {
-						if _, exists := podsForEviction[eviction.Name]; exists {
-							return true, obj, nil
-						}
-						evictionFailed = true
-						return true, nil, fmt.Errorf("pod %q was unexpectedly evicted", eviction.Name)
+				var collector *metricscollector.MetricsCollector
+				if len(tc.nodemetricses) > 0 || len(tc.podmetricses) > 0 {
```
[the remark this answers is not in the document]
ingvagabund: I see this as equivalent. Each test can either set `metricsEnabled` or specify either of `nodemetricses`/`podmetricses` fields. From the opposite perspective if a test is not expected to test the metrics collector none of `nodemetricses`/`podmetricses` is allowed to be set.

pkg/descheduler/metricscollector/metricscollector.go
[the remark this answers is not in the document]
ingvagabund: Both metrics collector and usage client operate independently. For that, the check needs to live outside of both.

pkg/descheduler/metricscollector/metricscollector.go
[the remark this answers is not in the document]
ingvagabund: The check would need to compute the next step. Which beats the intention. Unless I misunderstood.

pkg/descheduler/metricscollector/metricscollector.go
```diff
@@ -0,0 +1,144 @@
...
+
+	return allNodesUsage, nil
+}
+
+func (mc *MetricsCollector) NodeUsage(node *v1.Node) (map[v1.ResourceName]*resource.Quantity, error) {
+	mc.mu.RLock()
+	defer mc.mu.RUnlock()
+
+	if _, exists := mc.nodes[node.Name]; !exists {
+		klog.V(4).InfoS("unable to find node in the collected metrics", "node", node.Name)
```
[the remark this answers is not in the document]
ingvagabund: Done

pkg/framework/plugins/nodeutilization/usageclients.go
ingvagabund: Done

pkg/descheduler/metricscollector/metricscollector.go
ingvagabund: Done

pkg/descheduler/metricscollector/metricscollector.go
ingvagabund: Done

pkg/framework/plugins/nodeutilization/lownodeutilization_test.go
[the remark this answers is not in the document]
atiratree: I meant the MetricsCollector should not be used when `metricsUtilization.metricsServer == false`

pkg/descheduler/metricscollector/metricscollector.go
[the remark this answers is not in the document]
atiratree: We could check if the computed value by weightedAverage equals to `prevValue`. If yes then we know the difference is small and can use the `value` instead.

This way we would the same deterministic behaviour when the descheduler has just been restarted or has been running for some time.

pkg/framework/plugins/nodeutilization/usageclients.go
```diff
@@ -100,3 +104,95 @@ func (s *requestedUsageClient) sync(nodes []*v1.Node) error {
...
+
+		nodeUsage, ok := nodesUsage[node.Name]
+		if !ok {
+			return fmt.Errorf("unable to find node %q in the collected metrics", node.Name)
+		}
+		nodeUsage[v1.ResourcePods] = resource.NewQuantity(int64(len(pods)), resource.DecimalSI)
+
+		for _, resourceName := range client.resourceNames {
+			if _, exists := nodeUsage[resourceName]; !exists {
+				return fmt.Errorf("unable to find %q resource for collected %q node metric", resourceName, node.Name)
```
atiratree, line 188: +1, so now we do not need to do check the resourceNames in the client.metricsCollector. It can return anything it wants and we just return an error here if it does not comply.

pkg/framework/plugins/nodeutilization/usageclients.go
```diff
@@ -100,3 +104,95 @@ func (s *requestedUsageClient) sync(nodes []*v1.Node) error {
...
+	podMetrics, err := client.metricsCollector.MetricsClient().MetricsV1beta1().PodMetricses(pod.Namespace).Get(context.TODO(), pod.Name, metav1.GetOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("unable to get podmetrics for %q/%q: %v", pod.Namespace, pod.Name, err)
+	}
+
+	totalUsage := make(map[v1.ResourceName]*resource.Quantity)
+	for _, container := range podMetrics.Containers {
+		for _, resourceName := range client.resourceNames {
+			if _, exists := container.Usage[resourceName]; !exists {
+				continue
```
atiratree, line 151: wouldn't it be better to error out here as well?

pkg/descheduler/metricscollector/metricscollector.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+}
+
+func (mc *MetricsCollector) Run(ctx context.Context) {
+	wait.NonSlidingUntil(func() {
+		mc.Collect(ctx)
+	}, 5*time.Second, ctx.Done())
+}
+
+func weightedAverage(prevValue, value int64) int64 {
```
atiratree, line 68: would be good to have some documentation above this function once we resolve https://github.com/kubernetes-sigs/descheduler/pull/1555#discussion_r1846945435 thread

pkg/framework/plugins/nodeutilization/usageclients.go
ingvagabund: You are right. podmetrics should provide both cpu and memory all the time. Done.

pkg/framework/plugins/nodeutilization/lownodeutilization_test.go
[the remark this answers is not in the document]
ingvagabund: Oh, there's `metricsEnabled` local variable actually :) Updated.

atiratree:
Thanks!
/lgtm

ingvagabund:
Squashing commits.

langchain-infra:
any idea on when this might be released @ingvagabund ? I see it's not in any of the release notes yet

ingvagabund:
Currently targeted for v0.32.0

a7i:
the e2e tests fail frequently, can someone take a look? thanks
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_descheduler/1570/pull-descheduler-test-e2e-k8s-master-1-31/1867751211424288768

ingvagabund:
`Process did not finish before 40m0s timeout`

PR #1565, opened 2024-12-03
Bump to Go 1.23.3

seanmalloy:
The k/k repo was bumped to Go 1.23.3. See below PR for reference.

https://github.com/kubernetes/kubernetes/pull/128852

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-FROM golang:1.22.7
+FROM golang:1.23.3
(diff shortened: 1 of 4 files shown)


seanmalloy:
I need to update the CI configuration with the new Go version before merging this.

/hold

a7i:
does this need to be updated as well? 
https://github.com/kubernetes-sigs/descheduler/blob/master/hack/lib/go.sh#L22

a7i:
@seanmalloy the error is:
```
Unknown go version 'go1.23.3', skipping gofmt.
make: *** [Makefile:118: verify-gofmt] Error 1
```

this file needs to be updated to include 1.23: https://github.com/kubernetes-sigs/descheduler/blob/master/hack/lib/go.sh#L22

PR #1582, opened 2024-12-22
Move values.yaml comment about defining your own configmap to a NOTES.txt output instead

tspearconquest:
This PR intends to make the chart easier to use by providing clear instructions visible to the user after an install. I found the comment in values.yaml to be confusing and ultimately made a PR in error trying to fix something which wasn't broken, so I hope that this PR will be found to be useful and more helpful, not to mention visible.

### DIFF
--- a/charts/descheduler/templates/NOTES.txt
+++ b/charts/descheduler/templates/NOTES.txt
@@ -10,3 +10,13 @@ WARNING: You enabled DryRun mode, you can't use Leader Election.
 {{- end}}
 {{- end}}
 {{- end}}
+{{- if .Values.deschedulerPolicy }}
+A default DeschedulerPolicy has been applied for you. You can view the policy with:
+
+kubectl get configmap -n {{ include "descheduler.namespace" . }} {{ template "descheduler.fullname" . }} -o yaml
+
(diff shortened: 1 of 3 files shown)


charts/descheduler/templates/NOTES.txt
```diff
@@ -10,3 +10,13 @@ WARNING: You enabled DryRun mode, you can't use Leader Election.
 {{- end}}
 {{- end}}
 {{- end}}
+{{- if .Values.deschedulerPolicy }}
+A default DeschedulerPolicy has been applied for you. You can view the policy with:
```
a7i, line 14: It's not actually default, because you can control the definition via the values file, right?

tspearconquest:
Just a quick response from my email tied to the PR to confirm that the
other comment by @speeddymon was, in fact, me.

tspearconquest:
I've removed the mention of "default" as requested.

tspearconquest:
Hello, could someone review and merge please?

jklaw90 (commented):
/lgtm

tspearconquest:
Thank you!

Issue #1584, opened 2024-12-25
Perform a scheduling simulation in dry-run mode.


zhifei92:
**Is your feature request related to a problem? Please describe.**

**Scenario:** in dry-run mode.If the descheduler determines that 10 pods in the cluster can be evicted, and each of these pods requires 8 cores and 16 GB of memory (8c/16g), while the total available resources in the cluster are sufficient, but only one node meets the 8c/16g requirement. Under the current logic, all pods are considered evictable. 

In dry-run mode, no resources are deducted because the pods are not actually evicted and rescheduled, nor is scheduling simulated.

loged:
```
klog.V(1).InfoS("Evicted pod in dry run mode", "pod", klog.KObj(pod), "reason", opts.Reason, "strategy", opts.StrategyName, "node", pod.Spec.NodeName, "profile", opts.ProfileName)
```

Using dry-run to preview which pods can be rescheduled may mislead users and cause unexpected issues.

**Describe the solution you'd like**

In dry-run mode, pod scheduling can be simulated to find fitting nodes. This enhances the accuracy of the dry-run results.
**Describe alternatives you've considered**

**What version of descheduler are you using?**
master
descheduler version:

**Additional context**

googs1025:
IIUC: Descheduler dry run mode 
- checks and reports which Pods can be evicted, but does not actually perform these operations. 
- No resource deduction: Because Pods are not actually evicted or rescheduled, resources in the cluster are not actually deducted. 
- No simulation of scheduling: In dry run mode, Descheduler does not simulate the rescheduling process, that is, it does not try to find a new suitable node for the evicted Pod.

So I understand the original mechanism to be reasonable.

zhifei92:
Here is an example: 
- In a cluster with four nodes (A, B, C, D), after a certain rescheduling strategy calculation, it is determined that two Pods with resource requirements of 2 CPU cores and 4 GB of memory need to be evicted from node A. 
- Node B can accommodate one 2c/4g Pod, while nodes C and D can only handle Pods with requirements of 1.5 CPU cores and 3 GB of memory. 

In a non-dry-run scenario, only one Pod on node A can be evicted because the second Pod cannot be successfully rescheduled onto another node. However, in dry-run mode, both Pods on node A are reported as evictable, which is inaccurate since it does not consider the feasibility of rescheduling these Pods onto other nodes.

zhifei92:
> No simulation of scheduling

[...]

PR #1585, opened 2024-12-25
Perform a scheduling simulation in dry-run mode.

zhifei92:
### A bug in dry-run mode.

**Scenario:** If the descheduler determines that 10 pods in the cluster can be evicted, and each of these pods requires 8 cores and 16 GB of memory (8c/16g), while the total available resources in the cluster are sufficient, but only one node meets the 8c/16g requirement. Under the current logic, all pods are considered evictable. 

In dry-run mode, no resources are deducted because the pods are not actually evicted and rescheduled, nor is scheduling simulated.

loged:
```
klog.V(1).InfoS("Evicted pod in dry run mode", "pod", klog.KObj(pod), "reason", opts.Reason, "strategy", opts.StrategyName, "node", pod.Spec.NodeName, "profile", opts.ProfileName)
```

Using dry-run to preview which pods can be rescheduled may mislead users and cause unexpected issues.

Fix: #1584

### DIFF
--- a/pkg/descheduler/evictions/evictions.go
+++ b/pkg/descheduler/evictions/evictions.go
@@ -38,6 +38,7 @@ import (
 
 	"sigs.k8s.io/descheduler/metrics"
 	eutils "sigs.k8s.io/descheduler/pkg/descheduler/evictions/utils"
+	"sigs.k8s.io/descheduler/pkg/descheduler/node"
 	"sigs.k8s.io/descheduler/pkg/features"
 	"sigs.k8s.io/descheduler/pkg/tracing"
 )
@@ -546,6 +547,7 @@ func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod, opts EvictOptio
 	}
 
 	if pe.dryRun {
+		pe.simulateSchedulingInDryRun(pod)
 		klog.V(1).InfoS("Evicted pod in dry run mode", "pod", klog.KObj(pod), "reason", opts.Reason, "strategy", opts.StrategyName, "node", pod.Spec.NodeName, "profile", opts.ProfileName)
 	} else {
 		klog.V(1).InfoS("Evicted pod", "pod", klog.KObj(pod), "reason", opts.Reason, "strategy", opts.StrategyName, "node", pod.Spec.NodeName, "profile", opts.ProfileName)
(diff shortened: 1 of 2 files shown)


pkg/descheduler/node/node.go
```diff
@@ -173,6 +177,14 @@ func podFitsNodes(nodeIndexer podutil.GetPodsAssignedToNodeFunc, pod *v1.Pod, no
 	return filteredLen > 0
 }
 
+func addFitNodeNameToPod(pod *v1.Pod, node *v1.Node) {
```
zhifei92, line 180: Adding a fit node to the pod's annotation is not the most appropriate best practice. It seems more reasonable to add a return value to `type FilterFunc func(*v1.Pod) bool`, changing it to `type FilterFunc func(*v1.Pod) (string, bool)`. However, many filters do not require this string return value. Is there a better implementation approach?

pkg/descheduler/evictions/evictions.go
```diff
@@ -561,6 +563,24 @@ func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod, opts EvictOptio
 	return nil
 }
 
+func (pe *PodEvictor) simulateSchedulingInDryRun(pod *v1.Pod) {
```
googs1025, line 566: I don't quite understand why we need to create a pod in dry run mode. The purpose of dry run mode is to simulate the behavior of Descheduler to preview which pods may be evicted. 🤔

zhifei92: see comments：https://github.com/kubernetes-sigs/descheduler/issues/1584#issuecomment-2564966754

> I don't quite understand why we need to create a pod in dry run mode. 

The purpose is to reserve the resources.

If any part is not described clearly, please feel free to comment.

pkg/descheduler/node/node.go
```diff
@@ -161,6 +164,7 @@ func podFitsNodes(nodeIndexer podutil.GetPodsAssignedToNodeFunc, pod *v1.Pod, no
 		if err == nil {
 			klog.V(4).InfoS("Pod fits on node", "pod", klog.KObj(pod), "node", klog.KObj(node))
 			atomic.AddInt32(&filteredLen, 1)
+			addFitNodeNameToPod(pod, node)
```
googs1025, line 167: Are we going to label all pods, whether they are dry-run or non-dry-run, with this label?

pkg/descheduler/evictions/evictions.go
```diff
@@ -561,6 +563,24 @@ func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod, opts EvictOptio
...
+	if !pe.dryRun {
+		return
+	}
```
googs1025, line 569: It seems to be repeated, and the outside world has already made a judgment

pkg/descheduler/node/node.go
zhifei92: It would be ideal if this behavior only occurred during dry-run and if it could pass the nodename without setting it as an annotation. However, I haven't figured out how to implement this yet. If this approach is accepted, we can give it a try.

pkg/descheduler/evictions/evictions.go
zhifei92: Indeed. Thank you.

pkg/descheduler/evictions/evictions.go
googs1025: got this. thanks

PR #1603, opened 2025-01-07
feature: add PodsWithResourceClaims parameter in DefaultEvictorArgs PodProtections

googs1025:
- add PodsWithResourceClaims parameter in DefaultEvictorArgs PodProtections
now we are gradually supporting DRA features
like `PodsWithResourceClaims ` param,  we sometime also don't want Pods with ResourceClaims to be rescheduled frequently.

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,10 +182,11 @@ The Default Evictor Plugin is used by default for filtering pods before processi
 
 > Setting a value in `extraEnabled` **enables an additional protection rule**. This means the specified type of Pods will be **protected** from eviction.
 
-| Value              | Meaning                                                                |
-|--------------------|------------------------------------------------------------------------|
-| `"PodsWithPVC"`    | Prevents eviction of Pods using Persistent Volume Claims (PVCs). |
-| `"PodsWithoutPDB"` | Prevents eviction of Pods without a PodDisruptionBudget (PDB). |
+| Value                      | Meaning                                                          |
+|----------------------------|------------------------------------------------------------------|
+| `"PodsWithPVC"`            | Prevents eviction of Pods using Persistent Volume Claims (PVCs). |
+| `"PodsWithoutPDB"`         | Prevents eviction of Pods without a PodDisruptionBudget (PDB).   |
+| `"PodsWithResourceClaims"` | Prevents eviction of Pods using ResourceClaims.                  |
 
 ### Example policy
 
--- a/pkg/framework/plugins/defaultevictor/defaultevictor_test.go
+++ b/pkg/framework/plugins/defaultevictor/defaultevictor_test.go
@@ -28,6 +28,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes/fake"
+	"k8s.io/utils/ptr"
 	"sigs.k8s.io/descheduler/pkg/api"
 	evictionutils "sigs.k8s.io/descheduler/pkg/descheduler/evictions/utils"
 	podutil "sigs.k8s.io/descheduler/pkg/descheduler/pod"
@@ -860,6 +861,24 @@ func TestDefaultEvictorFilter(t *testing.T) {
 			},
 			result: false,
 		},
+		{
+			description: "Pod with ResourceClaims is not evicted because 'PodsWithResourceClaims' is in extraPodProtections",
+			pods: []*v1.Pod{
+				test.BuildTestPod("p20", 400, 0, n1.Name, func(pod *v1.Pod) {
+					pod.ObjectMeta.OwnerReferences = test.GetNormalPodOwnerRefList()
+					pod.Spec.ResourceClaims = []v1.PodResourceClaim{
+						{
+							Name:              "test-claim",
+							ResourceClaimName: ptr.To("test-resource-claim"),
+						},
+					}
+				}),
+			},
+			podProtections: PodProtections{
+				ExtraEnabled: []PodProtection{PodsWithResourceClaims},
+			},
+			result: false,
+		},
 	}
 
 	for _, test := range testCases {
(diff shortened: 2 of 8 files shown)


a7i:
```
!!! 'gofmt -s' needs to be run on the following files: 
./pkg/framework/plugins/defaultevictor/defaultevictor_test.go
make: *** [Makefile:118: verify-gofmt] Error 1
```

googs1025:
> @googs1025 can you expand on your use-case a bit more?
> 
> Also, note that README and chart values.yaml will need to be updated

Yes, of course. We now have the DRA feature in kubernetes. We are also testing it in our internal environment. We hope that pods with resourceclaims will not be evicted frequently (because of the device). Therefore, we hope to have an ignorePodsWithResourceClaims parameter like the ignorePvcPods parameter to selectively ignore such pods.

pkg/apis/componentconfig/v1alpha1/types.go
```diff
@@ -57,6 +57,9 @@ type DeschedulerConfiguration struct {
 	// IgnorePVCPods sets whether PVC pods should be allowed to be evicted
 	IgnorePVCPods bool `json:"ignorePvcPods,omitempty"`
 
+	// IgnorePodsWithResourceClaims sets whether pods using resource claims should be allowed to be evicted
+	IgnorePodsWithResourceClaims bool `json:"ignorePodsWithResourceClaims"`
```
a7i, line 61: other fields use `omitempty` is that needed here?

googs1025: whoops... I miss that.
fixed

googs1025:
friendly ping @ingvagabund @ricardomaraschini
can you help this?

ricardomaraschini (commented):
This is nice. I wonder if we shouldn't allow users to select which ResourceClaim they want to avoid the eviction. What do you think @ingvagabund ?

Not that it should be implemented on this PR but if this is something we want to do in the future then it would be better to call the config `IgnorePodsWithAnyResourceClaim`.

README.md
```diff
@@ -133,19 +133,20 @@ These are top level keys in the Descheduler Policy that you can use to configure
...
+| `nodeFit`                      | `bool`                 | `false`                                         | (see [node fit filtering](#node-fit-filtering))                                                                             |
+| `minReplicas`                  | `uint`                 | `0`                                             | ignore eviction of pods where owner (e.g. `ReplicaSet`) replicas is below this threshold                                    |
+| `minPodAge`                    | `metav1.Duration`      | `0`                                             | ignore eviction of pods with a creation time within this threshold                                                          |
+| `ignorePodsWithoutPDB`         | `bool`                 | `false`                                         | set whether pods without PodDisruptionBudget should be evicted or ignored                                                   |
+| `ignorePodsWithResourceClaims` | `bool`                 | `false`                                         | set whether pods with resource claims should be evicted or ignored.                                                         |
```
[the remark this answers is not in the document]
googs1025: done

googs1025:
> This is nice. I wonder if we shouldn't allow users to select which ResourceClaim they want to avoid the eviction. What do you think @ingvagabund ?
> 

We should not be specific to "some ResourceClaim", I think 🤔 
but change to `IgnorePodsWithAnyResourceClaim ` is ok

pkg/framework/plugins/defaultevictor/defaultevictor_test.go
```diff
@@ -802,6 +804,36 @@ func TestDefaultEvictorFilter(t *testing.T) {
...
+							Name:              "test-name",
+							ResourceClaimName: ptr.To("test-resourceclaim"),
+						},
+					}
+				}),
+			},
+			ignorePodsWithResourceClaims: false,
+			result:                       true,
+		}, {
+			description: "ignorePodsWithResourceClaims is set, pod with ResourceClaims, evicts",
```
a7i, line 823:
```suggestion
			description: "ignorePodsWithResourceClaims is set, pod with ResourceClaims, no evuction",
```

a7i:
/approve

pending the minor change requested

pkg/framework/plugins/defaultevictor/defaultevictor_test.go
googs1025: done

pkg/apis/componentconfig/v1alpha1/types.go
```diff
@@ -57,6 +57,9 @@ type DeschedulerConfiguration struct {
 	// IgnorePVCPods sets whether PVC pods should be allowed to be evicted
 	IgnorePVCPods bool `json:"ignorePvcPods,omitempty"`
 
+	// IgnorePodsWithResourceClaims sets whether pods using resource claims should be allowed to be evicted
+	IgnorePodsWithResourceClaims bool `json:"ignorePodsWithResourceClaims,omitempty"`
```
ingvagabund, line 61: Fields from `DeschedulerConfiguration` are no longer used. Only the default evictor options are valid.

pkg/framework/plugins/defaultevictor/types.go
```diff
@@ -25,16 +25,17 @@ import (
...
+	EvictSystemCriticalPods      bool                   `json:"evictSystemCriticalPods,omitempty"`
+	IgnorePvcPods                bool                   `json:"ignorePvcPods,omitempty"`
+	EvictFailedBarePods          bool                   `json:"evictFailedBarePods,omitempty"`
+	LabelSelector                *metav1.LabelSelector  `json:"labelSelector,omitempty"`
+	PriorityThreshold            *api.PriorityThreshold `json:"priorityThreshold,omitempty"`
+	NodeFit                      bool                   `json:"nodeFit,omitempty"`
+	MinReplicas                  uint                   `json:"minReplicas,omitempty"`
+	MinPodAge                    *metav1.Duration       `json:"minPodAge,omitempty"`
+	IgnorePodsWithoutPDB         bool                   `json:"ignorePodsWithoutPDB,omitempty"`
+	IgnorePodsWithResourceClaims bool                   `json:"ignorePodsWithResourceClaims,omitempty"`
```
ingvagabund, line 40: keyword `ignore` is ambiguous. This will case some misunderstanding. Ignore pods with resource claims as "let such pods be and skip them from eviction" or "don't care about resource claims and treat a pod is if there were node". There's already `ignorePvcPods` that creates a precedence for the former case.

As of now with the new option we have:
- `ignorePvcPods`
- `ignorePodsWithoutPDB`
- `ignorePodsWithResourceClaims`

There are also:
- `evictLocalStoragePods`
- `evictSystemCriticalPods`
- `evictFailedBarePods`

All six as a boolean type. Instead of creating new names prefixed with `ignore` and/or `exclude` what about to create a new enum and pass a list of ignored/excluded cases instead? With a single keyword (instead of two). E.g.
```
evictionProtection:
- withPVC
- withoutPDB
- withResourceClaims
- withLocalStorage
- systemCritical
- failedBarePods
```

With `withLocalStorage`, `systemCritical` and `failedBarePods` set as default. Alongside deprecating all 5 options.

googs1025: This sounds like a more reasonable config. I'd like to submit a new PR to refactor the configuration first, and then work on this PR, how about that? 😄

ingvagabund: By all means please :) The new naming/keywords are a suggestion. The naming can be different at the end.

googs1025:
/hold
until https://github.com/kubernetes-sigs/descheduler/issues/1664 been resovled

ingvagabund:
@googs1025 looks good in overall :) Thank you. Can you also update https://github.com/kubernetes-sigs/descheduler?tab=readme-ov-file#supported-values-for-podprotectionsextraenabled?

Issue #1595, opened 2025-01-03
helm chart version 0.32.0 lacks poddisruptionbudgets.policy perms


linuxmaniac:
**What version of descheduler are you using?**

descheduler version: v0.32.0

**Does this issue reproduce with the latest release?**

This is latest release.

**Which descheduler CLI options are you using?**

**Please provide a copy of your descheduler policy config file**
```
---
apiVersion: source.toolkit.fluxcd.io/v1
kind: HelmRepository
metadata:
  name: descheduler
  namespace: flux-system
spec:
  interval: 24h
  url: https://kubernetes-sigs.github.io/descheduler/
---
apiVersion: helm.toolkit.fluxcd.io/v2
kind: HelmRelease
metadata:
  name: descheduler
  namespace: flux-system
spec:
  interval: 30m
  releaseName: descheduler
  targetNamespace: kube-system
  serviceAccountName: helm-controller
  chart:
    spec:
      chart: descheduler
      version: 0.32.x
      sourceRef:
        kind: HelmRepository
        name: descheduler
        namespace: flux-system
      interval: 12h
  values:
    kind: Deployment
    service:
      enabled: true
    serviceMonitor:
      enabled: true
    deschedulerPolicy:
      nodeSelector: node-role.kubernetes.io/worker=true
      profiles:
        - name: ProfileName
          pluginConfig:
            - name: DefaultEvictor
              args:
                nodeFit: true
            - name: RemoveDuplicates
            - name: RemovePodsViolatingInterPodAntiAffinity
            - name: RemovePodsHavingTooManyRestarts
              args:
                podRestartThreshold: 50
                includingInitContainers: true
            - name: LowNodeUtilization
              args:
                thresholds:
                  cpu: 20
                  memory: 20
                  pods: 20
                targetThresholds:
                  cpu: 50
                  memory: 50
                  pods: 50
                evictableNamespaces:
                  exclude:
                    - kube-system
                    - democratic-csi
            - name: HighNodeUtilization
              args:
                thresholds:
                  cpu: 20
                  memory: 20
                  pods: 20
                evictableNamespaces:
                  exclude:
                    - kube-system
                    - democratic-csi
            - name: PodLifeTime
              args:
                maxPodLifeTimeSeconds: 86400
                states:
                  - Pending
                  - PodInitializing
          plugins:
            balance:
              enabled:
                - HighNodeUtilization
                - LowNodeUtilization
                - RemoveDuplicates
            deschedule:
[...]

PR #1599, opened 2025-01-03
charts: add needed perms if metricsCollector is enabled

linuxmaniac:
fixes #1597

```
E0103 12:37:27.584970       1 metricscollector.go:128] "Error fetching metrics" err="nodes.metrics.k8s.io \"k3s-node3\" is forbidden: User \"system:serviceaccount:kube-system:descheduler\" cannot get resource \"nodes\" in API group \"metrics.k8s.io\" at the cluster scope" node="k3s-node3"
```

PR #1604, opened 2025-01-07
chart: add missing poddisruptionbudgets cluster role policy

cayla:
Fixes: https://github.com/kubernetes-sigs/descheduler/issues/1595


cayla:
Nevermind, I see https://github.com/kubernetes-sigs/descheduler/pull/1599 now.

Issue #1605, opened 2025-01-09
flaky test: TestTooManyRestarts/test-one-evictions-use-gracePeriodSeconds


a7i:
flaky test in the recent https://github.com/kubernetes-sigs/descheduler/pull/1538

```
=== RUN   TestTooManyRestarts/test-one-evictions-use-gracePeriodSeconds
    e2e_toomanyrestarts_test.go:167: Creating "descheduler-policy-configmap" policy CM with RemovePodsHavingTooManyRestarts configured...
    e2e_toomanyrestarts_test.go:182: Creating descheduler deployment descheduler
    e2e_toomanyrestarts_test.go:202: Waiting for the descheduler pod running
    e2e_test.go:1768: Waiting for 1 pods to be running, got 0 instead
    e2e_toomanyrestarts_test.go:216: Waiting for grace period of 55 seconds
    e2e_toomanyrestarts_test.go:221: preRunNames: [restart-pod-8c9597889-gbfkw restart-pod-8c9597889-p4dpp restart-pod-8c9597889-p826f restart-pod-8c9597889-r84cp], currentRunNames: [restart-pod-8c9597889-gbfkw restart-pod-8c9597889-p4dpp restart-pod-8c9597889-p826f restart-pod-8c9597889-r84cp], actualEvictedPodCount: 0
    e2e_toomanyrestarts_test.go:237: preRunNames: [restart-pod-8c9597889-gbfkw restart-pod-8c9597889-p4dpp restart-pod-8c9597889-p826f restart-pod-8c9597889-r84cp], currentRunNames: [restart-pod-8c9597889-gbfkw restart-pod-8c9597889-p4dpp restart-pod-8c9597889-p826f restart-pod-8c9597889-r84cp], actualEvictedPodCount: 0
    e2e_toomanyrestarts_test.go:239: Expecting 4 number of pods evicted, got 0 instead
    e2e_toomanyrestarts_test.go:237: preRunNames: [restart-pod-8c9597889-gbfkw restart-pod-8c9597889-p4dpp restart-pod-8c9597889-p826f restart-pod-8c9597889-r84cp], currentRunNames: [restart-pod-8c9597889-gbfkw restart-pod-8c9597889-p4dpp restart-pod-8c9597889-p826f restart-pod-8c9597889-r84cp], actualEvictedPodCount: 0
    e2e_toomanyrestarts_test.go:239: Expecting 4 number of pods evicted, got 0 instead
[...]

PR #1606, opened 2025-01-09
flake: rollback gracePeriod for evict pod in e2e test

googs1025:
part of issue: https://github.com/kubernetes-sigs/descheduler/issues/1605

I have observed many flaky tests. My current idea is to roll back the e2e test first. Because gracePeriod is indeed effective, but it seems that the e2e test is not stable. I don’t have much time to conduct a deeper investigation at the moment. I will fix this when I have time.


a7i:
/lgtm

I'll re-open the PR if it gets closed

a7i:
/approve

this is blocking other PRs from being merged

PR #1610, opened 2025-01-18
refactor(utils): refactor GetPodQOS to make it more readable and maintainable

googs1025:
- refactor ComputePodQOS to make it more readable and maintainable
- add unit test for ComputePodQOS
- remove unuse func

### DIFF
--- a/pkg/utils/qos.go
+++ b/pkg/utils/qos.go
@@ -15,6 +15,8 @@ func isSupportedQoSComputeResource(name v1.ResourceName) bool {
 	return supportedQoSComputeResources.Has(string(name))
 }
 
+var zeroQuantity = resource.MustParse("0")
+
 // GetPodQOS returns the QoS class of a pod.
 // A pod is besteffort if none of its containers have specified any requests or limits.
 // A pod is guaranteed only when requests and limits are specified for all the containers and they are equal.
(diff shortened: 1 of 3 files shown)


googs1025:
friendly ping @ingvagabund @a7i

googs1025:
@ricardomaraschini @ingvagabund can you help this?

pkg/utils/qos.go
```diff
@@ -26,64 +28,82 @@ func GetPodQOS(pod *v1.Pod) v1.PodQOSClass {
...
+	for _, container := range getAllContainers(pod) {
+		// Use a logical AND operation to accumulate the isGuaranteed status,
+		// ensuring that all containers must meet the Guaranteed condition.
+		isGuaranteed = processContainerResources(container, &requests, &limits) && isGuaranteed
+	}
+	if len(requests) == 0 && len(limits) == 0 {
+		return v1.PodQOSBestEffort
+	}
+	// Check is requests match limits for all resources.
+	if isGuaranteed && areRequestsMatchingLimits(requests, limits) {
```
ricardomaraschini, line 41: What is `isGuaranteed` here ? Should this variable be called something else ?

pkg/utils/qos.go
```diff
@@ -26,64 +28,82 @@ func GetPodQOS(pod *v1.Pod) v1.PodQOSClass {
...
+// processContainerResources processes the resources of a single container and updates the provided requests and limits lists.
+func processContainerResources(container v1.Container, requests, limits *v1.ResourceList) bool {
+	isGuaranteed := true
+	processResourceList(*requests, container.Resources.Requests)
+	qosLimitsFound := getQOSResources(container.Resources.Limits)
+	processResourceList(*limits, container.Resources.Limits)
+	if !qosLimitsFound.HasAll(string(v1.ResourceMemory), string(v1.ResourceCPU)) {
+		isGuaranteed = false
+	}
+	return isGuaranteed
```
ricardomaraschini, line 56:
```suggestion
	return qosLimitsFound.HasAll(string(v1.ResourceMemory), string(v1.ResourceCPU))
```

pkg/utils/qos.go
```diff
@@ -26,64 +28,82 @@ func GetPodQOS(pod *v1.Pod) v1.PodQOSClass {
```
ricardomaraschini, line 28: Shouldn't we be calling https://pkg.go.dev/k8s.io/kubernetes/pkg/apis/core/helper/qos#ComputePodQOS and drop the rest of this function ?

googs1025: Back here again, I understand that k/k does not separate the `apis package` as a tool library (like api package). If we want to use `ComputePodQOS func` in descheduler, we may need to introduce it in go.mod, such as: `k8s.io/kubernetes v1.33.0`, which I think is too heavy for go vendor.

```go
require (
	...
	k8s.io/api v0.33.0
	k8s.io/kubernetes v1.33.0     // new add 
```

ricardomaraschini: Makes sense.

pkg/utils/qos.go
```diff
@@ -26,64 +28,82 @@ func GetPodQOS(pod *v1.Pod) v1.PodQOSClass {
 
 	requests := v1.ResourceList{}
 	limits := v1.ResourceList{}
-	zeroQuantity := resource.MustParse("0")
 	isGuaranteed := true
+	for _, container := range getAllContainers(pod) {
+		// Use a logical AND operation to accumulate the isGuaranteed status,
+		// ensuring that all containers must meet the Guaranteed condition.
+		isGuaranteed = processContainerResources(container, &requests, &limits) && isGuaranteed
```
ricardomaraschini, line 35: Isn't this simply overwriting the `isGuaranteed` every cycle ?

ingvagabund:
> refactor ComputePodQOS to make it more readable and maintainable

Can you more elaborate on why it's less readable and maintainable now?

googs1025:
/close 

No longer necessary, just keep the original

PR #1611, opened 2025-01-20
rbac for metricsCollector

jisnardo:
Fix rbac permissions when metricsCollector is enabled.

### DIFF
--- a/charts/descheduler/templates/clusterrole.yaml
+++ b/charts/descheduler/templates/clusterrole.yaml
@@ -29,8 +29,8 @@ rules:
   verbs: ["get", "watch", "list"]
 - apiGroups: ["policy"]
   resources: ["poddisruptionbudgets"]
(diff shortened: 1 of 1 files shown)


charts/descheduler/templates/clusterrole.yaml
```diff
@@ -27,6 +27,17 @@ rules:
 - apiGroups: ["policy"]
   resources: ["poddisruptionbudgets"]
   verbs: ["get", "watch", "list"]
+- apiGroups: ["policy"]
```
googs1025, line 30: this seem to be saved in  this pr https://github.com/kubernetes-sigs/descheduler/pull/1599/files

jisnardo: Great so.

googs1025:
so we close this pr first. free to reopen if we have more discuss 😄 
/close

PR #1614, opened 2025-01-27
[nodeutilization]: allow to set a metrics source as a string so it can be later extended for exclusive configuration

ingvagabund:
Introducing an enum/string will restrict a metrics sources to a single source without extra validation.
While deprecating the current `metricsEnabled` field. Both fields can be used. Only the mention of `metricsEnabled` is removed from the README.md.

A similar approach holds for the top level metrics collector configuration. Deprecating the current API for introducing a list of metrics providers than can be configured in parallel.


ingvagabund:
/hold
for now given https://github.com/kubernetes-sigs/descheduler/releases/tag/v0.32.1 is already out.

ingvagabund:

pkg/framework/plugins/nodeutilization/types.go
```diff
@@ -45,19 +45,31 @@ type LowNodeUtilizationArgs struct {
...
+	// KubernetesMetrics enables metrics from a Kubernetes metrics server.
+	// Please see https://kubernetes-sigs.github.io/metrics-server/ for more.
+	KubernetesMetrics MetricsSource = "KubernetesMetrics"
+)
+
 // MetricsUtilization allow to consume actual resource utilization from metrics
 type MetricsUtilization struct {
 	// metricsServer enables metrics from a kubernetes metrics server.
 	// Please see https://kubernetes-sigs.github.io/metrics-server/ for more.
+	// Deprecated. Use MetricsSource instead.
```
googs1025, line 69: When we deprecate  this field, do we need to keep it in the docs and indicate in which version it will be officially deleted? 🤔

pkg/framework/plugins/nodeutilization/types.go
```diff
@@ -45,19 +45,31 @@ type LowNodeUtilizationArgs struct {
 type HighNodeUtilizationArgs struct {
 	metav1.TypeMeta `json:",inline"`
 
-	Thresholds         api.ResourceThresholds `json:"thresholds"`
-	NumberOfNodes      int                    `json:"numberOfNodes,omitempty"`
-	MetricsUtilization MetricsUtilization     `json:"metricsUtilization,omitempty"`
```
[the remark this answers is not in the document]
ingvagabund: The field has never been used in the plugin

pkg/framework/plugins/nodeutilization/types.go
ingvagabund: You are right. Readme updated to mention `metricsServer` field is deprecated.

README.md
```diff
@@ -288,7 +288,7 @@ A resource consumption above (resp. below) this window is considered as overutil
 This approach is chosen in order to maintain consistency with the kube-scheduler, which follows the same
 design for scheduling pods onto nodes. This means that resource usage as reported by Kubelet (or commands
 like `kubectl top`) may differ from the calculated consumption, due to these components reporting
-actual usage metrics. Metrics-based descheduling can be enabled by setting `metricsUtilization.metricsServer` field.
+actual usage metrics. Metrics-based descheduling can be enabled by setting `metricsUtilization.metricsServer` (deprecated) or `metricsUtilization.source` field to `KubernetesMetrics`.
```
ricardomaraschini, line 291: Would you mind breaking this line ?

ricardomaraschini (approved):

README.md
ingvagabund: Done

atiratree (approved):

atiratree:
@ingvagabund shouldn't we also update https://github.com/kubernetes-sigs/descheduler/blame/17d9b152a2dc058ffd28fc8312fd9db043cbd4ed/pkg/api/types.go#L112?

ingvagabund:
@atiratree `MetricsCollector` field updated to be a pointer to.

I did not fully though through what other source of metrics could be integrated in the policy. The top policy is intended to provide a global configuration for all plugins. These days we can use the Kubernetes metrics or Prometheus client+PromQuery. It's valid to have one plugin to rely on Kubernetes metrics and another on Prometheus. So the sources can be configured in parallel.

ingvagabund:
Each `MetricsCollector` will have a different set of fields in general. Unless we define `MetricsCollector` as `interface{}`. I'd rather wait to see how many other metrics providers will be introduce and bump the policy to v1alpha3 while reshaping the way the metric providers are integrated.

pkg/descheduler/policyconfig.go
```diff
@@ -150,9 +150,15 @@ func validateDeschedulerConfiguration(in api.DeschedulerPolicy, registry pluginr
 				continue
 			}
 			if err := pluginUtilities.PluginArgValidator(pluginConfig.Args); err != nil {
-				errorsInProfiles = append(errorsInProfiles, fmt.Errorf("in profile %s: %s", profile.Name, err.Error()))
+				errorsInPolicy = append(errorsInPolicy, fmt.Errorf("in profile %s: %s", profile.Name, err.Error()))
 			}
 		}
 	}
-	return utilerrors.NewAggregate(errorsInProfiles)
+	if len(in.MetricsProviders) > 1 {
```
atiratree, line 157: What is the reason behind limiting the functionality? Do we want to enforce setting only a single source until the code base is ready for it?

pkg/descheduler/descheduler.go
```diff
@@ -166,7 +166,7 @@ func newDescheduler(ctx context.Context, rs *options.DeschedulerServer, deschedu
 	}
 
 	var metricsCollector *metricscollector.MetricsCollector
-	if deschedulerPolicy.MetricsCollector.Enabled {
+	if (deschedulerPolicy.MetricsCollector != nil && deschedulerPolicy.MetricsCollector.Enabled) || (len(deschedulerPolicy.MetricsProviders) > 0 && deschedulerPolicy.MetricsProviders[0].Source == api.KubernetesMetrics) {
```
atiratree, line 169: Can we make a utility function from this? And even though we allow only a single source now, it would be nicer to go over the whole slice to check the source.

pkg/descheduler/policyconfig.go
ingvagabund: Yes. I will extend the test through https://github.com/kubernetes-sigs/descheduler/pull/1533.

atiratree:
LGTM

Issue #725, opened 2022-02-15
HighNodeUtilization does nothing when all nodes are underutilized


slobo:
**What version of descheduler are you using?**

descheduler version: 0.23

**Does this issue reproduce with the latest release?**
Yes

**Which descheduler CLI options are you using?**
Using helm chart 0.23.1 with these overrides:
```yaml
deschedulerPolicy:
  strategies:
    HighNodeUtilization:
      enabled: true
      params:
        nodeResourceUtilizationThresholds:
          thresholds:
            memory: 20
        numberOfNodes: 0
    LowNodeUtilization:
      enabled: false
schedule: 5 10 * * *
```

**What k8s version are you using (`kubectl version`)?**

<details><summary><code>kubectl version</code> Output</summary><br><pre>
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.2", GitCommit:"092fbfbf53427de67cac1e9fa54aaa09a28371d7", GitTreeState:"clean", BuildDate:"2021-06-16T12:59:11Z", GoVersion:"go1.16.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"21+", GitVersion:"v1.21.5-eks-bc4871b", GitCommit:"5236faf39f1b7a7dabea8df12726f25608131aa9", GitTreeState:"clean", BuildDate:"2021-10-29T23:32:16Z", GoVersion:"go1.16.8", Compiler:"gc", Platform:"linux/amd64"}
</pre></details>

**What did you do?**

Installed descheduler. Ran the cron job manually.

**What did you expect to see?**

Descheduler evict pods from some of the underutilized instances.

**What did you see instead?**

Descheduler didn't evict any pods. Saw this in the log:

```
I0214 21:57:50.046412       1 highnodeutilization.go:90] "Criteria for a node below target utilization" CPU=100 Mem=20 Pods=100
I0214 21:57:50.046417       1 highnodeutilization.go:91] "Number of underutilized nodes" totalNumber=4
I0214 21:57:50.046421       1 highnodeutilization.go:102] "All nodes are underutilized, nothing to do here"
```

[...]

PR #1616, opened 2025-01-29
Make HighNodeUtilization select 1 node if all nodes are underutilized

zoonage:
Eventually reconcile the issue in https://github.com/kubernetes-sigs/descheduler/issues/725 by tainting and removing 1 node at a time on each run.

There's definitely a more efficient way by working out how many nodes could be removed to achieve a certain resource utilisation density on the remaining nodes, however this is a quick fix that eventually reconciles the cluster into the desired state


zoonage:
This is the first time I've tried to use the de-scheduler, how are the *NodeUtilization plugins preventing pods from being re-scheduled onto the node they've just been evicted from?

zoonage:
> This is the first time I've tried to use the de-scheduler, how are the *NodeUtilization plugins preventing pods from being re-scheduled onto the node they've just been evicted from?

I've added PreferredNoSchedule tainting to nodes now to avoid scheduling on nodes we're trying to remove

ricardomaraschini:
Hi @zoonage , do you have any plans on moving this forward ? I think this is a very nice addition to the plugin.

zoonage:
> Hi @zoonage , do you have any plans on moving this forward ? I think this is a very nice addition to the plugin.

Heya, yeah I do, I've just been time constrained a lot this year. Hoping to get some time for it in August to wrap this up

SarCode:
Hey @zoonage, this feature sounds what I was trying with #1766, but your integration is much better

zoonage:
> Hey @zoonage, this feature sounds what I was trying with #1766, but your integration is much better

Skimming your change description it looks like what this needs is a combination of the two to work, my approach was flawed because it assumes you can treat each node as identical, however there are so many cases where that isn't true (e.g. node groups, pods with AZ bound PVs etc)

Currently the place I'm working isn't using k8s so I don't have the time (or infrastructure) to test these changes right now unfortunately, however if you want to take any of my changes please feel free to use them

PR #1618, opened 2025-01-30
tracing: fix context cancellation during shutdown

LukeCarrier:
`tracing.Shutdown()` must be invoked before `done()`, else the context will be cancelled before it's invoked and the following error will be emitted:

```text
E0130 14:31:04.963323       1 tracing.go:156] "failed to shutdown the trace exporter" err="context canceled"
```

~~Interestingly, not observed when running with a `deschedulingInterval`, but reliably reproducible when running as a `CronJob`.~~

Nope, I do see it, I just also see metrics that have already been flushed.

### DIFF
--- a/cmd/descheduler/app/server.go
+++ b/cmd/descheduler/app/server.go
@@ -107,7 +107,6 @@ func Run(rootCtx context.Context, rs *options.DeschedulerServer) error {
 	if err != nil {
 		klog.ErrorS(err, "failed to create tracer provider")
 	}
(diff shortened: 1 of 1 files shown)


ricardomaraschini:
@LukeCarrier Did you have the chance of checking my comment ?

ricardomaraschini:
This one has been replace by https://github.com/kubernetes-sigs/descheduler/pull/1747 and can be closed.

PR #1622, opened 2025-02-02
fix: graceperiodseconds e2e test

googs1025:
- fix graceperiodseconds e2e test


test/e2e/e2e_toomanyrestarts_test.go
```diff
@@ -198,6 +204,43 @@ func TestTooManyRestarts(t *testing.T) {
 			if len(deschedulerPods) != 0 {
 				deschedulerPodName = deschedulerPods[0].Name
 			}
+
+			// Check if grace period is enabled and wait accordingly
+			if tc.enableGracePeriod {
```
googs1025, line 209: After checking the problem, it should be here: we cannot use the `wait.PollUntilContextTimeout`directly. We need to continuously poll within a period of time to ensure that no pod is evicted during this period. `wait.PollUntilContextTimeout` cannot achieve this.
FYI: 
https://github.com/kubernetes-sigs/descheduler/pull/1606/files#diff-40509b8123ffe5c09c142ea31f2175330f82ddc10a17b8ec098b2200bb41678fL210-L230

googs1025:
/cc @a7i @ingvagabund 
I do some testing on CI, seems to be stable

test/e2e/e2e_toomanyrestarts_test.go
ingvagabund: Can you elaborate more on why you are convinced replacing `wait.PollUntilContextTimeout` with the current approach works?

The duration is known so you might create a loop, run it for at most 55 iterations and sleep for one second during each iteration. So the extra context is not needed.

googs1025:
It seems that there are still some issues that have not been resolved. I need to investigate again.  ....

ingvagabund:
@googs1025 would you please extend the PR description with the root cause analysis and how to fix the identified cause? So it's clear what does not work properly and what needs to be fixed to make the review focused.

PR #1625, opened 2025-02-06
WIP: [nodeutilization]: allow to set/remove a soft taint from overutilized nodes

tiraboschi:
Dynamically set/remove a soft taint (effect: PreferNoSchedule) to/from nodes that the descheduler identified as overutilized according to the nodeutilization plugin.
After #1555 the nodeutilization plugin can consume utilization data from actual kubernetes metrics and not only static reservations (requests).
On the other side, the default kube-scheduler is only going to ensure that Pod's specific resource requests can be satisfied.
The soft taint will simply gave an hint to the scheduler to try avoiding nodes that looks overutilized at
descheduler eyes.
Since it's just a soft taint (do it just if possible) there is no need to restrict it only to a certain amount of nodes in the cluster or be strict on error handling.
With a fine-grained `ValidatingAdmissionPolicy` (GA in k8s 1.30) 
we can enforce that the `descheduler-sa` is only allowed
to apply/remove soft-taints (enforcing `PreferNoSchedule` effect)
and with a predefined key-prefix (eventually customizable with a parameter).

TODO: add more unit and e2e tests

Fixes: #1626

### DIFF
--- a/README.md
+++ b/README.md
@@ -291,19 +291,28 @@ like `kubectl top`) may differ from the calculated consumption, due to these com
 actual usage metrics. Metrics-based descheduling can be enabled by setting `metricsUtilization.metricsServer` field.
 In order to have the plugin consume the metrics the metric collector needs to be configured as well.
 See `metricsCollector` field at [Top Level configuration](#top-level-configuration) for available options.
+That gap can only be reduced using the SoftTainter to dinamically apply/remove soft taints (`effect: PreferNoSchedule`) to/from nodes.
+The soft taint will simply gave an hint to the scheduler to try avoiding nodes that looks overutilized at
+the descheduler eyes. Since it's just a soft taint (do it just if possible) there is no guaranteee or results, on the other side even in the worst case it will not introduce any disruption or shortage of workers' nodes.
+The descheduler SoftTainter can be enabled by setting `softTainter.applySoftTaints` field.
+Since it's just an optional sub-feature, the `descheduler-cluster-role` is not granting the right to amend nodes.
+Cluster admins that want to use this feature should create an additional cluster role to grant `patch` and `update` verbs on `nodes` and bind it to the `descheduler-sa` ServiceAccount.
 
 **Parameters:**
 
-|Name|Type|
-|---|---|
-|`useDeviationThresholds`|bool|
-|`thresholds`|map(string:int)|
-|`targetThresholds`|map(string:int)|
-|`numberOfNodes`|int|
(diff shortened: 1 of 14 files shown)


README.md
```diff
@@ -291,19 +291,28 @@ like `kubectl top`) may differ from the calculated consumption, due to these com
 actual usage metrics. Metrics-based descheduling can be enabled by setting `metricsUtilization.metricsServer` field.
 In order to have the plugin consume the metrics the metric collector needs to be configured as well.
 See `metricsCollector` field at [Top Level configuration](#top-level-configuration) for available options.
+That gap can only be reduced using the SoftTainter to dinamically apply/remove soft taints (`effect: PreferNoSchedule`) to/from nodes.
```
googs1025, line 294: I'm not sure if setting taint to node has left the descheduler itself 🤔

pkg/descheduler/node/node.go
```diff
@@ -386,3 +389,112 @@ func podMatchesInterPodAntiAffinity(nodeIndexer podutil.GetPodsAssignedToNodeFun
 
 	return false, nil
 }
+
+// IsNodeSoftTainted checks if the node is already tainted
+// with the provided key and value and PreferNoSchedule effect.
+// If a match is found, it returns true.
+func IsNodeSoftTainted(node *v1.Node, taintKey, taintValue string) bool {
```
googs1025, line 396: `IsNodeSoftTainted`  `AddOrUpdateSoftTaint`  `RemoveSoftTaint` maybe we can remove `Soft`

README.md
googs1025: What I am concerned about is whether this is not a feature of the descheduler. I understand that after rescheduling, new pods may still be scheduled to this node again. But we need to think whether adding taints to the descheduler is a good choice. Maybe we should use a custom component to accomplish this? 🤔  I don't seem to see other plugins that have similar operations for modifying node fields or taints (maybe I'm wrong, I haven't delved into all the plugins).

googs1025: similar issue from : https://github.com/kubernetes/node-problem-detector/pull/565

There are similar discussions in the NDP project. I understand that the two projects are different, but when we introduce similar feature, do we have to think clearly or consider it more general? 🤔

googs1025: What I mean is: Does this deviate from the original intention of the descheduler? But I understand that the feature is reasonable. 🤔

JaneLiuL:
welcome for First-time contributor~~ look forward to see you more and more PR and contribute~~

tiraboschi:
Probably it was out of scope for the core descheduler, let's close it.

PR #1633, opened 2025-02-20
bugfix: honor pod Spec.GracePeriodSeconds firstly when evicting

NoicFank:
if pod already set `.Spec.TerminationGracePeriodSeconds`, then should honor it.  

Currently,  the deletionGracePeriodSeconds of pod will use `pe.gracePeriodSeconds` once it set, no matter `pod.Spec.TerminationGracePeriodSeconds` is set or not.

In this case, if `pod.Spec.TerminationGracePeriodSeconds` is set to 70, and `pe.gracePeriodSeconds` set to `40`, it will shrink the grace existing time of pod, which is unexpected.

### DIFF
--- a/pkg/descheduler/evictions/evictions.go
+++ b/pkg/descheduler/evictions/evictions.go
@@ -565,9 +565,14 @@ func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod, opts EvictOptio
 
 // return (ignore, err)
 func (pe *PodEvictor) evictPod(ctx context.Context, pod *v1.Pod) (bool, error) {
-	deleteOptions := &metav1.DeleteOptions{
-		GracePeriodSeconds: pe.gracePeriodSeconds,
+	deleteOptions := &metav1.DeleteOptions{}
+
+	// Honor the pod's TerminationGracePeriodSeconds firstly.
+	// If the gracePeriodSeconds of pod is not set, then set our value.
+	if pod.Spec.TerminationGracePeriodSeconds == nil {
+		deleteOptions.GracePeriodSeconds = pe.gracePeriodSeconds
 	}
+
 	// GracePeriodSeconds ?
 	eviction := &policy.Eviction{
 		TypeMeta: metav1.TypeMeta{


googs1025:
This makes sense to me. We should meet the TerminationGracePeriodSeconds of each pod first, and then meet the config  in descheduler second.

pkg/descheduler/evictions/evictions.go
```diff
@@ -565,9 +565,14 @@ func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod, opts EvictOptio
 
 // return (ignore, err)
 func (pe *PodEvictor) evictPod(ctx context.Context, pod *v1.Pod) (bool, error) {
-	deleteOptions := &metav1.DeleteOptions{
-		GracePeriodSeconds: pe.gracePeriodSeconds,
+	deleteOptions := &metav1.DeleteOptions{}
+
+	// Honor the pod's TerminationGracePeriodSeconds firstly.
+	// If the gracePeriodSeconds of pod is not set, then set our value.
+	if pod.Spec.TerminationGracePeriodSeconds == nil {
```
googs1025, line 572: But there is a question: do we need to consider the value size? For example, if pod.Spec.TerminationGracePeriodSeconds is less than GracePeriodSeconds, should the descheduler config be respected first?

NoicFank: Thanks quick review~

For me, directly using `.Spec.TerminationGracePeriodSeconds` may more appropriately.

But it's also fine to select maximum of `.Spec.TerminationGracePeriodSeconds`  and `pe.gracePeriodSeconds`

googs1025: After looking deeper, maybe kubelet has already done 🤔 

https://github.com/kubernetes/kubernetes/blob/0b1ef24a9912d1cbbbfe5641bf9cd2cdc91f32bd/pkg/kubelet/pod_workers.go#L976-L1010

a7i, line 572: this will never be nil, since default value is 30

https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/

>The default terminationGracePeriodSeconds setting is 30 seconds.

a7i:
I'm not entirely sure I understand the reasoning behind this PR.

Could you clarify why the gracePeriodSeconds is being set on the policy in the first place?

pkg/descheduler/evictions/evictions.go
NoicFank: Yes, correct, this value will never be nil. This ` pod.Spec.TerminationGracePeriodSeconds == nil ` is meaningless.

@googs1025 could you pls tell the reason why adding `gracePeriodSeconds` args?

NoicFank:
/close

`.Spec.TerminationGracePeriodSeconds` never be nil.

PR #1640, opened 2025-03-07
ReferencedResourceList: alias for map[v1.ResourceName]*resource.Quantity to avoid the type definition duplication

ingvagabund:
Too many places where `map[v1.ResourceName]*resource.Quantity` is getting defined.


ricardomaraschini (approved):
Oh yes @ingvagabund , we need this ! Please.

PR #1641, opened 2025-03-07
feat(token): automountServiceAccountToken set hard to true on deployment

jmederer:
In our Environment all ServiceAccounta are mutatet  to automountServiceAccountToken: false 
And its only allowed to set the automountServiceAccountToken: true in a deployment.

But at the moment descheduler doesnt provide this.

automountServiceAccountToken is necessary for descheduler, for that reason i added it hard coded.

Would be nice if you can go with it.

many thanks

### DIFF
--- a/charts/descheduler/templates/cronjob.yaml
+++ b/charts/descheduler/templates/cronjob.yaml
@@ -70,6 +70,9 @@ spec:
           priorityClassName: {{ .Values.priorityClassName }}
           {{- end }}
           serviceAccountName: {{ template "descheduler.serviceAccountName" . }}
+          {{- if kindIs "bool" .Values.automountServiceAccountToken }}
+          automountServiceAccountToken: {{ .Values.automountServiceAccountToken }}
+          {{- end }}
           restartPolicy: "Never"
           {{- with .Values.imagePullSecrets }}
           imagePullSecrets:
@@ -103,6 +106,9 @@ spec:
               volumeMounts:
                 - mountPath: /policy-dir
                   name: policy-volume
+                {{- if and .Values.extraServiceAccountVolumeMounts (not .Values.automountServiceAccountToken) }}
+                  {{ toYaml .Values.extraServiceAccountVolumeMounts | nindent 16 }}
+                {{- end }}
           {{- if .Values.podSecurityContext }}
           securityContext:
             {{- toYaml .Values.podSecurityContext | nindent 12 }}
@@ -111,4 +117,7 @@ spec:
           - name: policy-volume
             configMap:
               name: {{ template "descheduler.fullname" . }}
+          {{- if and .Values.extraServiceAccountVolumes (not .Values.automountServiceAccountToken) }}
+            {{ toYaml .Values.extraServiceAccountVolumes | nindent 10 }}
+          {{- end }}
 {{- end }}
--- a/charts/descheduler/templates/deployment.yaml
+++ b/charts/descheduler/templates/deployment.yaml
@@ -39,6 +39,9 @@ spec:
       priorityClassName: {{ .Values.priorityClassName }}
       {{- end }}
       serviceAccountName: {{ template "descheduler.serviceAccountName" . }}
+      {{- if kindIs "bool" .Values.automountServiceAccountToken }}
+      automountServiceAccountToken: {{ .Values.automountServiceAccountToken }}
+      {{- end }}
       {{- with .Values.imagePullSecrets }}
       imagePullSecrets:
       {{- toYaml . | nindent 6 }}
@@ -75,6 +78,9 @@ spec:
           volumeMounts:
             - mountPath: /policy-dir
               name: policy-volume
+            {{- if and .Values.extraServiceAccountVolumeMounts (not .Values.automountServiceAccountToken) }}
+              {{ toYaml .Values.extraServiceAccountVolumeMounts | nindent 12 }}
+            {{- end }}
       {{- if .Values.podSecurityContext }}
       securityContext:
         {{- toYaml .Values.podSecurityContext | nindent 8 }}
@@ -83,6 +89,9 @@ spec:
         - name: policy-volume
           configMap:
             name: {{ template "descheduler.fullname" . }}
+        {{- if and .Values.extraServiceAccountVolumes (not .Values.automountServiceAccountToken) }}
+          {{ toYaml .Values.extraServiceAccountVolumes | nindent 8}}
+        {{- end }}
       {{- with .Values.nodeSelector }}
       nodeSelector:
         {{- toYaml . | nindent 8 }}
--- a/charts/descheduler/templates/serviceaccount.yaml
+++ b/charts/descheduler/templates/serviceaccount.yaml
@@ -1,6 +1,9 @@
 {{- if .Values.serviceAccount.create -}}
 apiVersion: v1
 kind: ServiceAccount
+{{- if kindIs "bool" .Values.serviceAccount.automountServiceAccountToken }}
+automountServiceAccountToken: {{ .Values.serviceAccount.automountServiceAccountToken }}
+{{- end }}
 metadata:
   name: {{ template "descheduler.serviceAccountName" . }}
   namespace: {{ include "descheduler.namespace" . }}
--- a/charts/descheduler/values.yaml
+++ b/charts/descheduler/values.yaml
@@ -194,6 +194,15 @@ serviceAccount:
   name:
   # Specifies custom annotations for the serviceAccount
   annotations: {}
+  # Opt out of API credential automounting
+  # 
+  # automountServiceAccountToken Default is not set 
+  # automountServiceAccountToken: true
+  
+
+# Mount the ServiceAccountToken in the Pod of a CronJob or Deployment 
+# Default is not set - but only implied by the ServiceAccount
+# automountServiceAccountToken: true
 
 podAnnotations: {}
 
@@ -245,3 +254,30 @@ serviceMonitor:
     #   targetLabel: nodename
     #   replacement: $1
     #   action: replace
+
+## Additional Volume mounts when automountServiceAccountToken is false
+# extraServiceAccountVolumeMounts: 
+# - mountPath: /var/run/secrets/kubernetes.io/serviceaccount
+#   name: kube-api-access
+#   readOnly: true
+
+## Additional Volumes when automountServiceAccountToken is false
+# extraServiceAccountVolumes:
+# - name: kube-api-access
+#   projected:
+#     defaultMode: 0444
+#     sources:
+#     - configMap:
+#         items:
+#         - key: ca.crt
+#           path: ca.crt
+#         name: kube-root-ca.crt
+#     - downwardAPI:
+#         items:
+#         - fieldRef:
+#             apiVersion: v1
+#             fieldPath: metadata.namespace
+#           path: namespace
+#     - serviceAccountToken:
+#         expirationSeconds: 3600
+#         path: token


jkroepke:
Hi all, we have the same requirement as @jmederer

charts/descheduler/templates/deployment.yaml
```diff
@@ -39,6 +39,7 @@ spec:
       priorityClassName: {{ .Values.priorityClassName }}
       {{- end }}
       serviceAccountName: {{ template "descheduler.serviceAccountName" . }}
+      automountServiceAccountToken: true
```
googs1025, line 42: Can we configure it in helm `value.yaml` instead of hard-coding it? 🤔

jmederer: Hi @googs1025,

of course. I changed the MR.

charts/descheduler/values.yaml
```diff
@@ -199,6 +199,11 @@ serviceAccount:
   name:
   # Specifies custom annotations for the serviceAccount
   annotations: {}
+  # Opt out of API credential automounting
+  automountServiceAccountToken: true
+
+# Mount the ServiceAccountToken in the Pod of a CronJob or Deployment 
+automountServiceAccountToken: true
```
googs1025, line 207: Does the automountServiceAccountToken have to be set to true? If not, maybe we can set it to false by default, and users can set it themselves if necessary.

jmederer: If we set automountServiceAccountToken in Line 206 to false - the result will be no token in the POD of a CJ or Deployment. Its necessary for descheduler so we need this default to true.
I Think false is only for some edge cases a option

jkroepke: @googs1025 - is there is not serviceAccount, descheulder can't talk to Kubernetes API.

googs1025: I'm not sure if this is a cluster-specific issue? If I understand correctly, the default ServiceAccountToken should be mounted, so why is this setting needed?

jmederer: If you set automountServiceAccountToken to false on the service Account. This is what happend in our Environment through Kyverno then we have no Change to use descheduler. Because also the Pods has the Settings automountServiceAccountToken false (implicit if not set in Deployment)

For Descheduler but it absolutly necessary therfore you need automountServiceAccountToken TRUE in your deployment 

https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/

tspearconquest: IMHO - this should default to false. To mitigate the errors and allow descheduler to communicate with Kubernetes API, the volume has to be explicitly defined by the helm chart: https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#bound-service-account-token-volume

Example:
```
apiVersion: v1
kind: Deployment
metadata:
  name: example-deployment
spec:
  template:
    spec:
      containers:
      - name: example-container
        image: example-image:latest
        volumeMounts:
        - mountPath: /var/run/secrets/kubernetes.io/serviceaccount
          name: kube-api-access
          readOnly: true
      volumes:
      - name: kube-api-access
        projected:
          defaultMode: 292
          sources:
          - configMap:
              items:
              - key: ca.crt
                path: ca.crt
              name: kube-root-ca.crt
          - downwardAPI:
              items:
              - fieldRef:
                  apiVersion: v1
                  fieldPath: metadata.namespace
                path: namespace
          - serviceAccountToken:
              expirationSeconds: 3600
              path: token
```

You can explicitly define the service account token volume and volume mount as above, which allows only the containers with the `volumeMount` to talk to the API, and avoids issues with Kyverno and Gatekeeper Policies that block `automountServiceAccountToken: true`

I propose adding the volume and mount to the helm chart and defaulting the automount value to false.

jkroepke: > I propose adding the volume and mount to the helm chart and defaulting the automount value to false.

You can do that, but please mark this as breaking change in the helm chart and consider to bump the major version.

tspearconquest: Hi @jkroepke unfortunately this PR isn't on my fork, so I can't make the change, and was moreso proposing it to the submitter.

charts/descheduler/templates/cronjob.yaml
```diff
@@ -67,6 +67,7 @@ spec:
           priorityClassName: {{ .Values.priorityClassName }}
           {{- end }}
           serviceAccountName: {{ template "descheduler.serviceAccountName" . }}
+          automountServiceAccountToken: {{ .Values.automountServiceAccountToken }}
```
jkroepke, line 70:
```suggestion
          {{- if kindIs "bool" .Values.automountServiceAccountToken  }}
          automountServiceAccountToken: {{ .Values.automountServiceAccountToken }}
          {{- end }}
```

charts/descheduler/templates/deployment.yaml
```diff
@@ -39,6 +39,7 @@ spec:
       priorityClassName: {{ .Values.priorityClassName }}
       {{- end }}
       serviceAccountName: {{ template "descheduler.serviceAccountName" . }}
+      automountServiceAccountToken: {{ .Values.automountServiceAccountToken }}
```
jkroepke, line 42:
```suggestion
      {{- if kindIs "bool" .Values.automountServiceAccountToken  }}
      automountServiceAccountToken: {{ .Values.automountServiceAccountToken }}
      {{- end }}
```

charts/descheduler/templates/serviceaccount.yaml
```diff
@@ -1,6 +1,7 @@
 {{- if .Values.serviceAccount.create -}}
 apiVersion: v1
 kind: ServiceAccount
+automountServiceAccountToken: {{ .Values.serviceAccount.automountServiceAccountToken }}
```
jkroepke, line 4:
```suggestion
{{- if kindIs "bool" .Values.serviceAccount.automountServiceAccountToken}}
automountServiceAccountToken: {{ .Values.serviceAccount.automountServiceAccountToken}}
{{- end }}
```

charts/descheduler/values.yaml
```diff
@@ -199,6 +199,11 @@ serviceAccount:
   name:
   # Specifies custom annotations for the serviceAccount
   annotations: {}
+  # Opt out of API credential automounting
+  automountServiceAccountToken: true
```
jkroepke, line 203:
```suggestion
  # automountServiceAccountToken: false
```

charts/descheduler/values.yaml
```diff
@@ -199,6 +199,11 @@ serviceAccount:
...
+
+# Mount the ServiceAccountToken in the Pod of a CronJob or Deployment 
+automountServiceAccountToken: true
```
jkroepke, line 206:
```suggestion
# automountServiceAccountToken: false
```

charts/descheduler/values.yaml
jkroepke: The change you requested feel unnatural, just speaking as prometheus community helm chart maintainer. Unnatural in terms of compared to how the rest of the world may do it. It doesn't make sense to manually add the volume, if Kubernetes as a property for it.

At Grafana helm charts, the suggested behavior is `automountServiceAccountToken: false` at ServiceAccount and true at `automountServiceAccountToken: true` at Pod Spec.

---

To be honest, I will add a suggestion where the property automountServiceAccountToken is undefined. This should be a safe default for everyone, since it does not modify existing behavior.

jkroepke (commented):
Since we can't decide what could be the best default. I recommend to leave property undefined. This pattern guarantees that existing behavior is not modified, but it can be modified if requested.

This pattern is used at the Prometheus helm chart.

https://github.com/prometheus-community/helm-charts/blob/ed7cf08e755dbdf265d371b3d8fbca7f3a101afc/charts/prometheus/templates/deploy.yaml#L53-L55

jmederer:
Hi,

We've implemented two possible solutions for this part of automountServiceToken – Variant A and Variant B – and would appreciate your input on which direction to go. Both options are working, but they differ in readability.

Variant A:
from @jkroepke  with` if kindIs "bool"`

Variant B:
from @tspearconquest with volumeMount

I'm happy to proceed with either – just let me know which you prefer!

Thanks in advance for your guidance.

Best regards,
Jan

jkroepke:
@tspearconquest @googs1025 @jmederer Can I help here to get this merged soon? Feel free to ping me in Slack.

jmederer:
@tspearconquest @googs1025 - Hello, I wanted to kindly follow up on this PR. It's been inactive for some time, and I would appreciate any feedback or guidance on how to proceed. Thank you!

tspearconquest:
I don't think I should provide guidance on this; because I am not a maintainer for this project I think I would be overstepping if I did. I will leave the decision to them.

jmederer:
Thanks for your reply! Since you're not a maintainer, would you mind mentioning the relevant maintainer(s) here so they can take a look or provide guidance? That would be really helpful. Appreciate your support!

tspearconquest:
Looking at other PR which was merged already, I think it's @googs1025 but I am not certain.

charts/descheduler/values.yaml
```diff
@@ -199,6 +199,38 @@ serviceAccount:
...
+automountServiceAccountToken: false
+
+# Additional Volume mounts
+extraVolumeMounts: 
+- mountPath: /var/run/secrets/kubernetes.io/serviceaccount
+  name: kube-api-access
+  readOnly: true
+
+# Additional Volumes
+extraVolumes:
```
jklaw90: this seems like a lot. do you know of any other helm chart that does this when we set automount to false?

jmederer: No, I'm not aware of any other Helm chart that does this when automountServiceAccountToken is set to false. I would also prefer the approach suggested by @jkroepke ->  https://github.com/kubernetes-sigs/descheduler/pull/1641#pullrequestreview-2764934719

charts/descheduler/values.yaml
jklaw90: @googs1025 i really think we should keep this to ignored or true by default. so the 2nd commit with the value being true in value.yaml was fine to me. OR something like @jkroepke suggested like [this](https://github.com/kubernetes-sigs/descheduler/pull/1641#discussion_r2121186958)

I also question the extra volumes, is anyone actually setting it to false here? 
FWIW we often find helm charts unable to do everything we need as an org and leverage kustomize + helm to get it there.

charts/descheduler/values.yaml
jklaw90: ahh so ignored by default, yeah that would be fine too

jmederer: I'll go ahead and update the PR accordingly.

tspearconquest: It's not in the chart values itself but cert-manager includes it in this document: https://cert-manager.io/docs/installation/best-practice/#restrict-auto-mount-of-service-account-tokens

I agree too that it _probably_ should not be in the values directly; can it be added in `_helpers.tpl`and included in the Deployment and CJ based on the `kindIs`?

tspearconquest: Pseudocode because I'm on my phone: If kind is bool and value is false for auto mount, then include the volume and volume mount from the helpers

jmederer: I've updated the PR accordingly. It's now aligned with the suggested approach. Additionally, it's still possible to configure volumes and volumeMounts as before.

Let me know if anything else needs adjusting — thanks!

charts/descheduler/templates/serviceaccount.yaml
```diff
@@ -1,6 +1,9 @@
 {{- if .Values.serviceAccount.create -}}
 apiVersion: v1
 kind: ServiceAccount
+{{- if kindIs "bool" .Values.serviceAccount.automountServiceAccountToken }}
+automountServiceAccountToken: {{ .Values.serviceAccountautomountServiceAccountToken }}
```
jklaw90:
```suggestion
automountServiceAccountToken: {{ .Values.serviceAccount.automountServiceAccountToken }}
```

charts/descheduler/templates/cronjob.yaml
```diff
@@ -100,6 +103,9 @@ spec:
               volumeMounts:
                 - mountPath: /policy-dir
                   name: policy-volume
+                {{- if .Values.extraVolumeMounts }}
```
jklaw90: probably should add `extraVolumeMounts` and `extraVolumes` commented out to the value file as well.

googs1025:
can you squash commit to one? 😄

googs1025:
> can you squash commit to one? 😄 

/hold

for this

jmederer:
> > can you squash commit to one? 😄
> 
> /hold
> 
> for this
 I've squashed the commits as requested.
Please have another look when you get a chance. 🙂

googs1025:
/lgtm

Final review for @ingvagabund

charts/descheduler/values.yaml
```diff
@@ -244,3 +253,30 @@ serviceMonitor:
     #   targetLabel: nodename
     #   replacement: $1
     #   action: replace
+
+## Additional Volume mounts
```
ingvagabund: s/Additional Volume mounts/Additional Volume mounts when automountServiceAccountToken is false/

charts/descheduler/values.yaml
```diff
@@ -244,3 +253,30 @@ serviceMonitor:
...
+# extraVolumeMounts: 
+# - mountPath: /var/run/secrets/kubernetes.io/serviceaccount
+#   name: kube-api-access
+#   readOnly: true
+
+## Additional Volumes
```
ingvagabund: s/Additional Volumes/Additional Volumes when automountServiceAccountToken is false/

charts/descheduler/templates/cronjob.yaml
ingvagabund: Should apply only when `.Values.automountServiceAccountToken` is false

charts/descheduler/templates/cronjob.yaml
```diff
@@ -108,4 +114,7 @@ spec:
           - name: policy-volume
             configMap:
               name: {{ template "descheduler.fullname" . }}
+          {{- if .Values.extraVolumes }}
```
ingvagabund: Should apply only when `.Values.automountServiceAccountToken` is false

charts/descheduler/values.yaml
```diff
@@ -244,3 +253,30 @@ serviceMonitor:
     #   targetLabel: nodename
     #   replacement: $1
     #   action: replace
+
+## Additional Volume mounts
+# extraVolumeMounts:
```
ingvagabund: s/extraVolumeMounts/extraServiceAccountVolumeMounts/

charts/descheduler/values.yaml
ingvagabund: s/extraVolumes/extraServiceAccountVolumes/

ingvagabund:
+1 for not changing the defaults and keeping `automountServiceAccountToken` commented out.

jmederer:
@ingvagabund  could you check my MR again ?

ingvagabund:
Thank you

Issue #1499, opened 2024-08-16
Replace namespace include/exclude filter with label selector


nabokihms:
**Is your feature request related to a problem? Please describe.**
I have productive and test namespaces labeled accordingly (env: prod, env: test). I'd like to deschedule pods based on these labels. As of today, every time engineers create new namespace, it is required to go and change the rescheduled config as well.

This problem affects the ability to provide self-service for my teams.

**Describe the solution you'd like**
I'd like to see `namespaces.labelSelector` option which can filter namespaces based on labels. This is more flexible in all terms.

Including and excluding namespaces by names (as it works now) can also be implemented using this option (so we no longer need include/exclude at all), e.g., 
```
namespaces:
  include:
  - ns-a
  - ns-b
  - ns-c
```
equals to

```
labelSelector:
  matchExpressions:
    - key: kubernetes.io/metadata.name
      operator: In
      values: [ns-a, ns-b, ns-c]
```

**Describe alternatives you've considered**

--

**What version of descheduler are you using?**

descheduler version: latest

**Additional context**

damemi:
+1 this is honestly how we probably should have done it. @ingvagabund @knelasevero do you know any reason why we did the include/exclude pattern instead of just label selectors?

ingvagabund:
We can not guarantee mutual exclusion of included and excluded namespaces with the whole expressiveness of `labelSelector`. If there's an non-empty intersection of the namespaces should we ignore the shared namespaces? Should we be more restrictive and ignore the shared included namespaces?

nabokihms:
@ingvagabund thanks for your answer, but it is puzzling. Are you saying that the feature will not be accepted because it is not a simple feature? 

What is with the use case then? What can you suggest go solve the issue?

damemi:
@ingvagabund if we get rid of the `include` and `exclude` fields and make it one label selector (ie, where people can set `IN` or `NOT IN` themselves) then I don't think there's any issue with conflicting includes/excludes. It just becomes label selector semantics

ingvagabund:
What you are saying makes sense. Still trying to recall why we decided to use combination of included/excluded namespaces. I can think of performance reasons. If we allow combination of `IN` and `NOT IN` it might be time expensive to compute the final set of matched namespaces. With a list of only included or (exclusive or) only excluded namespaces it's fast to get a list of matched namespaces. If we allow combination of both the computation might get expensive. E.g. 
```
labelSelector:
  matchExpressions:
[...]

PR #1643, opened 2025-03-10
Allow `include` namespaces filtering in node utilization plugin

jjcohere:
The node utilization plugin currently does not allow specifying namespaces with an allowlist via `include`, preferring to filter instead with `exclude`. I could not find a technical reason for this and found other instances where both were allowed. I propose to lift this restriction in the interest of flexibility, at least until #1499/#1501 can be integrated.

### DIFF
--- a/pkg/framework/plugins/nodeutilization/nodeutilization.go
+++ b/pkg/framework/plugins/nodeutilization/nodeutilization.go
@@ -304,7 +304,9 @@ func evictPods(
 	usageClient usageClient,
 ) error {
 	var excludedNamespaces sets.Set[string]
+	var includedNamespaces sets.Set[string]
 	if evictableNamespaces != nil {
+		includedNamespaces = sets.New(evictableNamespaces.Include...)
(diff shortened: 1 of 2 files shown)


ricardomaraschini:
@jjcohere I think this makes sense. Can you rebase this PR and also sign the CLA ? We can try to expedite this.

googs1025:
@jjcohere Is it still active? Maybe I can take it over. 😄

googs1025:
take over by: https://github.com/kubernetes-sigs/descheduler/pull/1733

ingvagabund:
I'd like to avoid allowing inclusion of namespaces for either of the node utilization plugins. Including too few namespaces could degrade the effectiveness of both plugins and make the overall balancing less effective. By requiring users to exclude namespaces instead, they have to think more carefully about the scope and are less likely to exclude too many pods. Using a list of excluded namespaces also makes it harder to unintentionally reduce the set of included namespaces.

PR #1644, opened 2025-03-11
nodeutilization: invoke ValidateLowNodeUtilizationArgs instead of validateLowNodeUtilizationThresholds to make the test more generic

ingvagabund:
So it can be later extended with non threshold related validation tests.


ingvagabund:

googs1025:
/lgtm 
This makes the test case more general

PR #1648, opened 2025-03-13
feat: introduces new node usage classification engine

ricardomaraschini:
> [!NOTE]  
> The code introduced by this PR isn't in use yet. I will have a follow up PR to migrate the `nodeutilization` plugin to start using this new engine. Raising this now so I can have others looking into it. 

Implements a new node usage classifier. This new classifier works with two maps of any data types. Maps are indexed by an arbitrary string.

As an usage example this is how one can filter out for usages using a list of thresholds:

```go
result, err := classifier.Classify(
	// first argument is a list of arbitrary usages.
	map[string]v1.ResourceList{
		"node1": {
			v1.ResourceCPU:    ptr.To(resource.MustParse("2")),
			v1.ResourceMemory: ptr.To(resource.MustParse("2Gi")),
		},
		...
	},
	// second argument is a list of thresholds, also indexed by an
	// arbitrary string. both collected usage and the thresholds
	// must be of the same time (on this case v1.ResourceList).
	map[string][]v1.ResourceList{
		"node1": {
			{
				v1.ResourceCPU:    ptr.To(resource.MustParse("4")),
				v1.ResourceMemory: ptr.To(resource.MustParse("4Gi")),
			},
			{
				v1.ResourceCPU:    ptr.To(resource.MustParse("16")),
				v1.ResourceMemory: ptr.To(resource.MustParse("16Gi")),
			},
		},
		...
	},
	// the third (and subsequent) argument is a list of classifiers.
	// one classifier must be provided for each threshold present in
	// each of the thresholds.
	classifier.ForMap[string, v1.ResourceName, resource.Quantity, v1.ResourceList(
		func(usage, limit resource.Quantity) int {
			return usage.Cmp(limit)
		},
	),
	classifier.ForMap[string, v1.ResourceName, resource.Quantity, v1.ResourceList(
		func(usage, limit resource.Quantity) int {
			return limit.Cmp(usage)
		},
	),
)
```

Classify() returns a slice of maps. Each position in the returned slice correspond to one of the provided Classifiers. i.e. if n classifiers were provided then the returned value will contain n positions.

## Generic Data Types

Everything must start with a list of values we want to classify. `Values` is a map of "values" (e.g. node cpu usages) indexed by an arbitrary key (e.g. node names):

```go
type Values[K comparable, V any] map[K]V
```

With the values at hand we now need some limits to operate upon (e.g. cpu thresholds for a node). We may have different thresholds as well (e.g. a bottom and a top limits for cpu usage). `Limits` is a list of one or more limits (e.g. cpu load thresholds) indexed by an arbitrary key:

```go
type Limits[K comparable, V any] map[K][]V
```

Now with values and limits defined we need to define a function that classify the values. `Classifier` is defined as a function that upon receiving an index and two arguments (the `Value` and the `Limit`) of the same type return a boolean. If the returned value is `true` then the item has been classified by the function and so belongs to that specific group.

```go
type Classifier[K comparable, V any] func(K, V, V) bool
```

> [!IMPORTANT]
> `Values` are compared against `Limits` using the index key. Multiple `Limits` may exist for a given `Value` but for each `Limit` a `Classifier` is expected to be passed in.

## Notes

- This new classification engine supports different thresholds per Node.
- This new classification engine supports more than two thresholds (today we only use `low` and `high`).
- For a good usage example please look at the [TestUsingDeviationThresholds](https://github.com/kubernetes-sigs/descheduler/pull/1648/files#diff-6d7cff398892b6b18ef7e4bd379d7fa0356f73d89a8c74798c8ac5d60c09daacR898) test.

### DIFF
--- a/pkg/framework/plugins/nodeutilization/classifier/classifier.go
+++ b/pkg/framework/plugins/nodeutilization/classifier/classifier.go
@@ -1,0 +1,84 @@
+/*
+Copyright 2025 The Kubernetes Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package classifier
+
+// Classifier is a function that classifies a resource usage based on a limit.
+// The function should return true if the resource usage matches the classifier
+// intent.
+type Classifier[K comparable, V any] func(K, V, V) bool
+
+// Comparer is a function that compares two objects. This function should return
+// -1 if the first object is less than the second, 0 if they are equal, and 1 if
+// the first object is greater than the second. Of course this is a simplification
+// and any value between -1 and 1 can be returned.
+type Comparer[V any] func(V, V) int
+
+// Values is a map of values indexed by a comparable key. An example of this
+// can be a list of resources indexed by a node name.
+type Values[K comparable, V any] map[K]V
+
+// Limits is a map of list of limits indexed by a comparable key. Each limit
+// inside the list requires a classifier to evaluate.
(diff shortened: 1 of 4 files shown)


ricardomaraschini:
I am closing this one. This is too big for a proper review. I will be opening it again in stages as I integrate it with the code.

PR #1649, opened 2025-03-14
feat: refactoring thresholds and usage assessment

ricardomaraschini:
this commit refactors the thresholds and usage assessment for the node utilization plugins. both high and low plugins are affected by this change.

### DIFF
--- a/pkg/framework/plugins/nodeutilization/highnodeutilization.go
+++ b/pkg/framework/plugins/nodeutilization/highnodeutilization.go
@@ -28,6 +28,7 @@ import (
 	nodeutil "sigs.k8s.io/descheduler/pkg/descheduler/node"
 
 	podutil "sigs.k8s.io/descheduler/pkg/descheduler/pod"
+	"sigs.k8s.io/descheduler/pkg/framework/plugins/nodeutilization/normalizer"
 	frameworktypes "sigs.k8s.io/descheduler/pkg/framework/types"
 )
 
@@ -42,6 +43,7 @@ type HighNodeUtilization struct {
 	podFilter                func(pod *v1.Pod) bool
 	underutilizationCriteria []interface{}
 	resourceNames            []v1.ResourceName
+	extendedResourceNames    []v1.ResourceName
 	targetThresholds         api.ResourceThresholds
 	usageClient              usageClient
 }
@@ -57,7 +59,7 @@ func NewHighNodeUtilization(args runtime.Object, handle frameworktypes.Handle) (
 
 	targetThresholds := make(api.ResourceThresholds)
 	setDefaultForThresholds(highNodeUtilizatioArgs.Thresholds, targetThresholds)
-	resourceNames := getResourceNames(targetThresholds)
+	resourceNames := getResourceNames(highNodeUtilizatioArgs.Thresholds)
 
 	underutilizationCriteria := []interface{}{
 		"CPU", highNodeUtilizatioArgs.Thresholds[v1.ResourceCPU],
@@ -77,14 +79,19 @@ func NewHighNodeUtilization(args runtime.Object, handle frameworktypes.Handle) (
 		return nil, fmt.Errorf("error initializing pod filter function: %v", err)
 	}
 
+	extendedResourceNames := uniquifyResourceNames(
+		append(resourceNames, v1.ResourceCPU, v1.ResourceMemory, v1.ResourcePods),
+	)
+
 	return &HighNodeUtilization{
 		handle:                   handle,
 		args:                     highNodeUtilizatioArgs,
 		resourceNames:            resourceNames,
+		extendedResourceNames:    extendedResourceNames,
 		targetThresholds:         targetThresholds,
 		underutilizationCriteria: underutilizationCriteria,
 		podFilter:                podFilter,
-		usageClient:              newRequestedUsageClient(resourceNames, handle.GetPodsAssignedToNodeFunc()),
+		usageClient:              newRequestedUsageClient(extendedResourceNames, handle.GetPodsAssignedToNodeFunc()),
 	}, nil
 }
 
@@ -102,11 +109,18 @@ func (h *HighNodeUtilization) Balance(ctx context.Context, nodes []*v1.Node) *fr
 	}
 
 	nodesMap, nodesUsageMap, podListMap := getNodeUsageSnapshot(nodes, h.usageClient)
-	nodeThresholdsMap := getStaticNodeThresholds(nodes, h.args.Thresholds, h.targetThresholds)
-	nodesUsageAsNodeThresholdsMap := nodeUsageToResourceThresholds(nodesUsageMap, nodesMap)
+	capacities := referencedResourceListForNodesCapacity(nodes)
+
+	usage, thresholds := assessNodesUsagesAndStaticThresholds(
+		nodesUsageMap,
+		capacities,
+		h.args.Thresholds,
+		h.targetThresholds,
+	)
+
 	nodeGroups := classifyNodeUsage(
-		nodesUsageAsNodeThresholdsMap,
-		nodeThresholdsMap,
+		usage,
+		thresholds,
 		[]classifierFnc{
 			// underutilized nodes
 			func(nodeName string, usage, threshold api.ResourceThresholds) bool {
@@ -128,7 +142,12 @@ func (h *HighNodeUtilization) Balance(ctx context.Context, nodes []*v1.Node) *fr
 	category := []string{"underutilized", "overutilized"}
 	for i := range nodeGroups {
 		for nodeName := range nodeGroups[i] {
-			klog.InfoS("Node is "+category[i], "node", klog.KObj(nodesMap[nodeName]), "usage", nodesUsageMap[nodeName], "usagePercentage", resourceUsagePercentages(nodesUsageMap[nodeName], nodesMap[nodeName], true))
+			klog.InfoS(
+				fmt.Sprintf("Node is %s", category[i]),
+				"node", klog.KObj(nodesMap[nodeName]),
+				"usage", nodesUsageMap[nodeName],
+				"usagePercentage", normalizer.Round(usage[nodeName]),
+			)
 			nodeInfos[i] = append(nodeInfos[i], NodeInfo{
 				NodeUsage: NodeUsage{
 					node:    nodesMap[nodeName],
@@ -136,8 +155,8 @@ func (h *HighNodeUtilization) Balance(ctx context.Context, nodes []*v1.Node) *fr
 					allPods: podListMap[nodeName],
 				},
 				thresholds: NodeThresholds{
-					lowResourceThreshold:  resourceThresholdsToNodeUsage(nodeThresholdsMap[nodeName][0], nodesMap[nodeName]),
-					highResourceThreshold: resourceThresholdsToNodeUsage(nodeThresholdsMap[nodeName][1], nodesMap[nodeName]),
+					lowResourceThreshold:  resourceThresholdsToNodeUsage(thresholds[nodeName][0], capacities[nodeName], h.extendedResourceNames),
+					highResourceThreshold: resourceThresholdsToNodeUsage(thresholds[nodeName][1], capacities[nodeName], h.extendedResourceNames),
 				},
 			})
 		}
@@ -189,7 +208,7 @@ func (h *HighNodeUtilization) Balance(ctx context.Context, nodes []*v1.Node) *fr
 		h.handle.Evictor(),
 		evictions.EvictOptions{StrategyName: HighNodeUtilizationPluginName},
 		h.podFilter,
-		h.resourceNames,
+		h.extendedResourceNames,
 		continueEvictionCond,
 		h.usageClient,
 		nil,
@@ -199,21 +218,15 @@ func (h *HighNodeUtilization) Balance(ctx context.Context, nodes []*v1.Node) *fr
 }
 
 func setDefaultForThresholds(thresholds, targetThresholds api.ResourceThresholds) {
-	// check if Pods/CPU/Mem are set, if not, set them to 100
-	if _, ok := thresholds[v1.ResourcePods]; !ok {
-		thresholds[v1.ResourcePods] = MaxResourcePercentage
+	if _, ok := thresholds[v1.ResourcePods]; ok {
+		targetThresholds[v1.ResourcePods] = MaxResourcePercentage
 	}
-	if _, ok := thresholds[v1.ResourceCPU]; !ok {
-		thresholds[v1.ResourceCPU] = MaxResourcePercentage
+	if _, ok := thresholds[v1.ResourceCPU]; ok {
+		targetThresholds[v1.ResourceCPU] = MaxResourcePercentage
 	}
-	if _, ok := thresholds[v1.ResourceMemory]; !ok {
-		thresholds[v1.ResourceMemory] = MaxResourcePercentage
+	if _, ok := thresholds[v1.ResourceMemory]; ok {
+		targetThresholds[v1.ResourceMemory] = MaxResourcePercentage
 	}
-
-	// Default targetThreshold resource values to 100
-	targetThresholds[v1.ResourcePods] = MaxResourcePercentage
-	targetThresholds[v1.ResourceCPU] = MaxResourcePercentage
-	targetThresholds[v1.ResourceMemory] = MaxResourcePercentage
 
 	for name := range thresholds {
 		if !nodeutil.IsBasicResource(name) {
--- a/pkg/framework/plugins/nodeutilization/highnodeutilization.go
+++ b/pkg/framework/plugins/nodeutilization/highnodeutilization.go
@@ -34,73 +34,94 @@ import (
 
 const HighNodeUtilizationPluginName = "HighNodeUtilization"
 
-// HighNodeUtilization evicts pods from under utilized nodes so that scheduler can schedule according to its plugin.
-// Note that CPU/Memory requests are used to calculate nodes' utilization and not the actual resource usage.
-
-type HighNodeUtilization struct {
-	handle                   frameworktypes.Handle
-	args                     *HighNodeUtilizationArgs
-	podFilter                func(pod *v1.Pod) bool
-	underutilizationCriteria []interface{}
-	resourceNames            []v1.ResourceName
-	extendedResourceNames    []v1.ResourceName
-	targetThresholds         api.ResourceThresholds
-	usageClient              usageClient
-}
-
+// this lines makes sure that HighNodeUtilization implements the BalancePlugin
+// interface.
 var _ frameworktypes.BalancePlugin = &HighNodeUtilization{}
 
-// NewHighNodeUtilization builds plugin from its arguments while passing a handle
-func NewHighNodeUtilization(args runtime.Object, handle frameworktypes.Handle) (frameworktypes.Plugin, error) {
-	highNodeUtilizatioArgs, ok := args.(*HighNodeUtilizationArgs)
+// HighNodeUtilization evicts pods from under utilized nodes so that scheduler
+// can schedule according to its plugin. Note that CPU/Memory requests are used
+// to calculate nodes' utilization and not the actual resource usage.
+type HighNodeUtilization struct {
+	handle         frameworktypes.Handle
+	args           *HighNodeUtilizationArgs
+	podFilter      func(pod *v1.Pod) bool
+	criteria       []any
+	resourceNames  []v1.ResourceName
+	highThresholds api.ResourceThresholds
+	usageClient    usageClient
+}
+
+// NewHighNodeUtilization builds plugin from its arguments while passing a handle.
+func NewHighNodeUtilization(
+	genericArgs runtime.Object, handle frameworktypes.Handle,
+) (frameworktypes.Plugin, error) {
+	args, ok := genericArgs.(*HighNodeUtilizationArgs)
 	if !ok {
-		return nil, fmt.Errorf("want args to be of type HighNodeUtilizationArgs, got %T", args)
+		return nil, fmt.Errorf(
+			"want args to be of type HighNodeUtilizationArgs, got %T",
+			genericArgs,
+		)
 	}
 
-	targetThresholds := make(api.ResourceThresholds)
-	setDefaultForThresholds(highNodeUtilizatioArgs.Thresholds, targetThresholds)
-	resourceNames := getResourceNames(highNodeUtilizatioArgs.Thresholds)
-
-	underutilizationCriteria := []interface{}{
-		"CPU", highNodeUtilizatioArgs.Thresholds[v1.ResourceCPU],
-		"Mem", highNodeUtilizatioArgs.Thresholds[v1.ResourceMemory],
-		"Pods", highNodeUtilizatioArgs.Thresholds[v1.ResourcePods],
-	}
-	for name := range highNodeUtilizatioArgs.Thresholds {
-		if !nodeutil.IsBasicResource(name) {
-			underutilizationCriteria = append(underutilizationCriteria, string(name), int64(highNodeUtilizatioArgs.Thresholds[name]))
-		}
+	// this plugins worries only about thresholds but the nodeplugins
+	// package was made to take two thresholds into account, one for low
+	// and another for high usage. here we make sure we set the high
+	// threshold to the maximum value for all resources for which we have a
+	// threshold.
+	highThresholds := make(api.ResourceThresholds)
+	for rname := range args.Thresholds {
+		highThresholds[rname] = MaxResourcePercentage
 	}
 
-	podFilter, err := podutil.NewOptions().
+	// criteria is a list of thresholds that are used to determine if a node
+	// is underutilized. it is used only for logging purposes.
+	criteria := []any{}
+	for rname, rvalue := range args.Thresholds {
+		criteria = append(criteria, rname, rvalue)
+	}
+
+	podFilter, err := podutil.
+		NewOptions().
 		WithFilter(handle.Evictor().Filter).
 		BuildFilterFunc()
 	if err != nil {
 		return nil, fmt.Errorf("error initializing pod filter function: %v", err)
 	}
 
-	extendedResourceNames := uniquifyResourceNames(
-		append(resourceNames, v1.ResourceCPU, v1.ResourceMemory, v1.ResourcePods),
+	// resourceNames is a list of all resource names this plugin cares
+	// about. we care about the resources for which we have a threshold and
+	// all we consider the basic resources (cpu, memory, pods).
+	resourceNames := uniquifyResourceNames(
+		append(
+			getResourceNames(args.Thresholds),
+			v1.ResourceCPU,
+			v1.ResourceMemory,
+			v1.ResourcePods,
+		),
 	)
 
 	return &HighNodeUtilization{
-		handle:                   handle,
-		args:                     highNodeUtilizatioArgs,
-		resourceNames:            resourceNames,
-		extendedResourceNames:    extendedResourceNames,
-		targetThresholds:         targetThresholds,
-		underutilizationCriteria: underutilizationCriteria,
-		podFilter:                podFilter,
-		usageClient:              newRequestedUsageClient(extendedResourceNames, handle.GetPodsAssignedToNodeFunc()),
+		handle:         handle,
+		args:           args,
+		resourceNames:  resourceNames,
+		highThresholds: highThresholds,
+		criteria:       criteria,
+		podFilter:      podFilter,
+		usageClient: newRequestedUsageClient(
+			resourceNames,
+			handle.GetPodsAssignedToNodeFunc(),
+		),
 	}, nil
 }
 
-// Name retrieves the plugin name
+// Name retrieves the plugin name.
 func (h *HighNodeUtilization) Name() string {
 	return HighNodeUtilizationPluginName
 }
 
-// Balance extension point implementation for the plugin
+// Balance holds the main logic of the plugin. It evicts pods from under
+// utilized nodes. The goal here is to concentrate pods in fewer nodes so that
+// less nodes are used.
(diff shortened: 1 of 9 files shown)


pkg/framework/plugins/nodeutilization/nodeutilization.go
```diff
@@ -523,3 +425,75 @@ func classifyPods(pods []*v1.Pod, filter func(pod *v1.Pod) bool) ([]*v1.Pod, []*
...
+
+// ResourceUsageNormalizer is an implementation of a Normalizer that converts a
+// set of resource usages and totals into percentage. This function operates on
+// Quantity Value() for all the resources except CPU, where it uses
+// MilliValue().
+func ResourceUsageNormalizer(usages, totals api.ReferencedResourceList) api.ResourceThresholds {
+	result := api.ResourceThresholds{}
+	for rname, value := range usages {
+		total, ok := totals[rname]
+		if !ok || value == nil || total == nil {
```
ingvagabund, line 485: We should panic when `usages` and `totals` lists do not have identical keys. Anyone calling `ResourceUsageNormalizer` needs to make sure the incoming lists are valid.

pkg/framework/plugins/nodeutilization/nodeutilization.go
```diff
@@ -523,3 +425,75 @@ func classifyPods(pods []*v1.Pod, filter func(pod *v1.Pod) bool) ([]*v1.Pod, []*
...
+	)
+
+	return usage, thresholds
+}
+
+// ResourceUsageNormalizer is an implementation of a Normalizer that converts a
+// set of resource usages and totals into percentage. This function operates on
+// Quantity Value() for all the resources except CPU, where it uses
+// MilliValue().
+func ResourceUsageNormalizer(usages, totals api.ReferencedResourceList) api.ResourceThresholds {
```
ingvagabund, line 481: We can omit the `normalizer` keyword and be explicit. E.g. `ResourceUsage2ResourceThreshold`.

pkg/framework/plugins/nodeutilization/nodeutilization.go
```diff
@@ -523,3 +425,75 @@ func classifyPods(pods []*v1.Pod, filter func(pod *v1.Pod) bool) ([]*v1.Pod, []*
...
+		if !ok || value == nil || total == nil {
+			continue
+		}
+
+		used, avail := value.Value(), total.Value()
+		if rname == v1.ResourceCPU {
+			used, avail = value.MilliValue(), total.MilliValue()
+		}
+
+		// ensures that the percentage is between 0 and 100.
```
ingvagabund, line 494: This should always be the case. If `used < 0` or `used > avail` the data are invalid. This needs to be handled during data validation. Otherwise, the normalizer might hide it and later on the code might produce incorrect results.

pkg/framework/plugins/nodeutilization/normalizer/normalizer.go
```diff
@@ -0,0 +1,119 @@
...
+
+// Min receives a minimum value and a list of values. Returns a list of values
+// where each value is greater than the minimum value.
+func Min[K comparable, N Number, V ~map[K]N](values []V, limit N) []V {
+	result := []V{}
+	for _, item := range values {
+		newItem := V{}
+		for key, value := range item {
+			newItem[key] = N(
+				math.Max(float64(value), float64(limit)),
```
ingvagabund, line 79: If `value < 1` => 1, if `value > 1` => value. What about the case where the average plus the user provided higher threshold are greater than 100%?

pkg/framework/plugins/nodeutilization/normalizer/normalizer.go
```diff
@@ -0,0 +1,119 @@
...
+	result := map[K]V{}
+	for _, key := range keys {
+		result[key] = value
+	}
+	return result
+}
+
+// Min receives a minimum value and a list of values. Returns a list of values
+// where each value is greater than the minimum value.
+func Min[K comparable, N Number, V ~map[K]N](values []V, limit N) []V {
```
ingvagabund, line 73: Better to keep the function as simple as possible. I.e. `(values V, limit N) V` and:
```
items := 		normalizer.Deviate(
				normalizer.Average(usage),
				[]api.ResourceThresholds{
					lowSpan.Negative(),
					highSpan,
				},
			),
[]api.ResourceThresholds{
	normalizer.Min(items[0], 1),
	normalizer.Min(items[1], 1),
},
```
I know I mentioned it will be great to be able to chain the operators. In this case introducing a `map` function would solve it:
```
	map([]api.ResourceThresholds{
			normalizer.Sum(normalizer.Average(usage), lowSpan.Negative()),
			normalizer.Sum(normalizer.Average(usage), highSpan),
		}, 
		func(item api.ResourceThresholds) { return normalizer.Min(item, 1) }
	)
```

pkg/api/types.go
```diff
@@ -75,6 +78,25 @@ type (
 	ResourceThresholds map[v1.ResourceName]Percentage
 )
 
+// Negative returns a new ResourceThresholds with all its values negated.
+func (r ResourceThresholds) Negative() ResourceThresholds {
```
ingvagabund, line 82: `Negative` and `Round` are two examples of methods iterating through the list of floats64 here. If we abstract from the percentages and promote it into a map of float64 number in general, the type can be extend with `Ceil` and `Floor` methods that could round any negative number to 0 and any number higher than 100 (or 1) to 100 (1).

ingvagabund: Also, in general Kubernetes types that are part of a data API are not extended with methods. Rather, functions that take the types as arguments is the way to go.

pkg/api/types.go
```diff
@@ -114,3 +136,31 @@ type MetricsCollector struct {
 
 // ReferencedResourceList is an adaption of v1.ResourceList with resources as references
 type ReferencedResourceList = map[v1.ResourceName]*resource.Quantity
+
+// ReferencedResourceListForNodesCapacity returns a ReferencedResourceList for
+// the capacity of a list of nodes. If allocatable resources are present, they
+// are used instead of capacity.
+func ReferencedResourceListForNodesCapacity(node []*v1.Node) map[string]ReferencedResourceList {
```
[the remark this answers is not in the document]
ricardomaraschini: Understood. I have removed the latter and consolidated the code into a a single function.

pkg/framework/plugins/nodeutilization/nodeutilization.go
```diff
@@ -523,3 +425,75 @@ func classifyPods(pods []*v1.Pod, filter func(pod *v1.Pod) bool) ([]*v1.Pod, []*
...
 	return nonRemovablePods, removablePods
 }
+
+// AssessNodesUsagesAndThresholds converts the raw usage data into percentage.
+// In case we need to calculate deviations from the average this function does
+// so as well. Thresholds are provided in percentage but they slightly differ
+// in behavior: if no deviation is enabled they are interpreted as is but with
+// deviations enabled they are considered deviations from the node's average.
+// Returns the usage (pct) and the thresholds (pct) for each node.
+func AssessNodesUsagesAndThresholds(
```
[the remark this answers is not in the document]
ricardomaraschini: I was trying to make this function usable somewhere else :-) . I have now split it into two functions and made it private. Please take a look.

pkg/framework/plugins/nodeutilization/normalizer/normalizer.go
```diff
@@ -0,0 +1,119 @@
...
+		}
+		result = append(result, newItem)
+	}
+	return result
+}
+
+// Deviate receives an object and a list of "deviations", returns an slice with
+// returns an slice of the same input where the internal value is added to each
+// deviation. Negative deviations are allowed.
+func Deviate[K comparable, N Number, V ~map[K]N](values V, deviations []V) []V {
```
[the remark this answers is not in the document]
ricardomaraschini: Done. For sake of not having to calculate the average twice the call for `Average()` was done earlier in the code flow. Please take a look.

pkg/framework/plugins/nodeutilization/normalizer/normalizer.go
```diff
@@ -0,0 +1,119 @@
...
+// Normalizer is a function that receives two values of the same type and
+// return an object of a different type. An usage case can be a function
+// that converts a memory usage from mb to % (the first argument would be
+// the memory usage in mb and the second argument would be the total memory
+// available in mb).
+type Normalizer[V, N any] func(V, V) N
+
+// Indexer is a function that receives an object and returns a key. This
+// function is used to generate a key for a map based on the object.
+type Indexer[K comparable, O any] func(O) K
```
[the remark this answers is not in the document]
ricardomaraschini: Oh gosh, this is a "left over" from the other code I was working on. Nice catch. Removed.

pkg/framework/plugins/nodeutilization/nodeutilization.go
```diff
@@ -523,3 +425,75 @@ func classifyPods(pods []*v1.Pod, filter func(pod *v1.Pod) bool) ([]*v1.Pod, []*
...
+	// into api.Percentage values.
+	usage := normalizer.Normalize(
+		rawUsages, rawCapacities, ResourceUsageNormalizer,
+	)
+
+	// if we are not taking the average and applying deviations to it we
+	// can simply replicate the same threshold across all nodes and return.
+	if !calculateDeviation {
+		thresholds := normalizer.Replicate(
+			slices.Collect(maps.Keys(rawUsages)),
```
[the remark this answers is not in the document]
ricardomaraschini: Done.

pkg/framework/plugins/nodeutilization/nodeutilization.go
```diff
@@ -523,3 +425,75 @@ func classifyPods(pods []*v1.Pod, filter func(pod *v1.Pod) bool) ([]*v1.Pod, []*
...
+		)
+		return usage, thresholds
+	}
+
+	// calculate the average usage and then deviate it according to the
+	// user provided thresholds. We also ensure that the value after the
+	// deviation is at least 1%. this call also replicates the thresholds
+	// across all nodes.
+	thresholds := normalizer.Replicate(
+		slices.Collect(maps.Keys(rawUsages)),
```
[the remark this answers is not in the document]
ricardomaraschini: Done.

pkg/framework/plugins/nodeutilization/nodeutilization.go
ricardomaraschini: We do have some tests where we expect some nodes to do not have resource capacities set, see [here](https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/framework/plugins/nodeutilization/highnodeutilization_test.go#L263). Should we start panic on these scenarios ? [This](https://github.com/kubernetes-sigs/descheduler/blob/master/pkg/framework/plugins/nodeutilization/highnodeutilization_test.go#L309) test even mentions that the extra resources won't exist in some nodes.

pkg/framework/plugins/nodeutilization/nodeutilization.go
ricardomaraschini: Done.

pkg/framework/plugins/nodeutilization/nodeutilization.go
ricardomaraschini: Done.

pkg/framework/plugins/nodeutilization/normalizer/normalizer.go
ricardomaraschini: Done, please take a look to see if my understanding was correct.

pkg/api/types.go
ricardomaraschini: I think I achieved what you meant. Please take a look.

pkg/framework/plugins/nodeutilization/nodeutilization.go
ingvagabund: If a node does not have the corresponding capacity for a given resourcename set, invoker needs to make sure the capacity is extended with resourcename with a zero quantity.

pkg/framework/plugins/nodeutilization/normalizer/normalizer.go
```diff
@@ -84,17 +84,15 @@ func Min[K comparable, N Number, V ~map[K]N](values []V, limit N) []V {
...
+// Sum receives an object and a "deviation", returns an slice with the internal
+// values "summed" to each deviation.
+func Sum[K comparable, N Number, V ~map[K]N](values, deviation V) V {
+	result := V{}
+	for name, value := range values {
+		if _, ok := deviation[name]; !ok {
+			continue
 		}
-		result = append(result, current)
+		result[name] = value + deviation[name]
```
ingvagabund, line 132: s/values, deviation/mapA, mapB/ or similar naming to make the function even more generic.

pkg/framework/plugins/nodeutilization/nodeutilization.go
ingvagabund: Yes. This can be skipped for now and fixed later.

pkg/framework/plugins/nodeutilization/normalizer/normalizer.go
ricardomaraschini: Done.

ricardomaraschini:
I have the impression that this e2e passed just a few minutes back. Are we dealing with a flake ?

/test pull-descheduler-test-e2e-k8s-master-1-30

pkg/framework/plugins/nodeutilization/nodeutilization.go
```diff
@@ -523,3 +425,112 @@ func classifyPods(pods []*v1.Pod, filter func(pod *v1.Pod) bool) ([]*v1.Pod, []*
...
+) (map[string]api.ResourceThresholds, map[string][]api.ResourceThresholds) {
+	// first we normalize the node usage from the raw data (Mi, Gi, etc)
+	// into api.Percentage values.
+	usage := normalizer.Normalize(
+		rawUsages, rawCapacities, ResourceUsageToResourceThreshold,
+	)
+
+	// calculate the average usage and then deviate it according to the
+	// user provided thresholds.
+	average := normalizer.Average(usage)
```
ingvagabund, line 466: There's also https://github.com/kubernetes-sigs/descheduler/pull/1650 case that's currently does not fit into the generic average function :(.

pkg/framework/plugins/nodeutilization/normalizer/normalizer.go
```diff
@@ -0,0 +1,159 @@
...
+func Average[J, K comparable, N Number, V ~map[J]N](values map[K]V) V {
+	result := V{}
+	for _, imap := range values {
+		for name, value := range imap {
+			result[name] += value
+		}
+	}
+
+	for name := range result {
+		result[name] /= N(len(values))
```
ingvagabund, line 155: `len(values)` currently corresponds to the number of nodes. Though, not every node will have all the resourcenames (here `name`) set. Instead, the function needs to keep a track of how many times each `name` was accessed and divide the total sum at the end by this number.

pkg/framework/plugins/nodeutilization/normalizer/normalizer_test.go
```diff
@@ -0,0 +1,710 @@
...
+					v1.ResourcePods:   resource.MustParse("10"),
+				},
+			},
+			expected: api.ResourceThresholds{
+				v1.ResourceCPU:    50.5,
+				v1.ResourceMemory: 35,
+				v1.ResourcePods:   60,
+			},
+		},
+	} {
```
ingvagabund, line 358: Worth extending the tests with nodes where some nodes have extended resources and some not to have `Average` compute the right sum based on the number of how many nodes have a given resource compared to dividing by all nodes.

pkg/framework/plugins/nodeutilization/nodeutilization.go
```diff
@@ -84,74 +86,6 @@ const (
...
-			}
-		}
-	}
-	lowThreshold, highThreshold := api.ResourceThresholds{}, api.ResourceThresholds{}
-	for resource, value := range total {
-		average[resource] = value / api.Percentage(numberOfNodes)
-		// If either of the spans are 0, ignore the resource. I.e. 0%:5% is invalid.
-		// Any zero span signifies a resource is either not set or is to be ignored.
-		if lowSpan[resource] == MinResourcePercentage || highSpan[resource] == MinResourcePercentage {
-			lowThreshold[resource] = 1
```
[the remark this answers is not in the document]
ingvagabund: 1 here means 100%. Which is the correct value here. So with this removed the new code needs to be able to tell when both low and high thresholds were not set.

pkg/framework/plugins/nodeutilization/nodeutilization.go
```diff
@@ -523,3 +425,112 @@ func classifyPods(pods []*v1.Pod, filter func(pod *v1.Pod) bool) ([]*v1.Pod, []*
...
+	// across all nodes.
+	thresholds := normalizer.Replicate(
+		slices.Collect(maps.Keys(usage)),
+		normalizer.Map(
+			[]api.ResourceThresholds{
+				normalizer.Sum(average, normalizer.Negate(lowSpan)),
+				normalizer.Sum(average, highSpan),
+			},
+			func(thresholds api.ResourceThresholds) api.ResourceThresholds {
+				return normalizer.Clamp(thresholds, 1, 100)
```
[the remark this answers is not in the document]
ingvagabund: So as mentioned in https://github.com/kubernetes-sigs/descheduler/pull/1649/files#r1999511172 if both `lowSpan` and highSpan` for a given resource are `0` then both `[]api.ResourceThresholds` needs to be set to `1`.

pkg/framework/plugins/nodeutilization/normalizer/normalizer.go
ricardomaraschini: Done.

pkg/framework/plugins/nodeutilization/normalizer/normalizer.go
```diff
@@ -0,0 +1,159 @@
...
+	return result
+}
+
+// Sum sums up the values of two maps. Values are expected to be of Number
+// type. Original values are preserved. If a key is present in one map but
+// not in the other, the key is ignored.
+func Sum[K comparable, N Number, V ~map[K]N](mapA, mapB V) V {
+	result := V{}
+	for name, value := range mapA {
+		if _, ok := mapB[name]; !ok {
```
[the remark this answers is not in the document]
ricardomaraschini: Done.

pkg/framework/plugins/nodeutilization/normalizer/normalizer.go
```diff
@@ -0,0 +1,159 @@
...
+		result[key] = N(math.Round(float64(value)))
+	}
+	return result
+}
+
+// Floor receives a minimum value and a map of values. Returns a list of
+// values where each value is greater or equal than the minimum value. In
+// other words, this function rounds up the values to the minimum value
+// provided.
+func Floor[K comparable, N Number, V ~map[K]N](values V, limit N) V {
```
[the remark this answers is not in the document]
ricardomaraschini: Inlined both functions. Floor and Ceil functions removed.

pkg/framework/plugins/nodeutilization/nodeutilization.go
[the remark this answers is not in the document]
ricardomaraschini: I am not sure I completely followed what you mean here. What I heard is: if both thresholds aren't defined by the user then they should both be defined to 100%. This is kinda strange to me as then all nodes could be pottentially classified as underutilized. I craft this [PR](https://github.com/kubernetes-sigs/descheduler/pull/1649/commits/e8ffee2fcda6017766226ab79d780693d0361093) but I am not sure this is right, I mean, this should be validated way earlier when parsing the config IMHO (and abort immediately if the config is wrong). Things like having a low threshold over a high threshold, for example, should trigger an immediate error.

pkg/framework/plugins/nodeutilization/nodeutilization.go
```diff
@@ -455,6 +458,9 @@ func assessNodesUsagesAndRelativeThresholds(
 	rawUsages, rawCapacities map[string]api.ReferencedResourceList,
 	lowSpan, highSpan api.ResourceThresholds,
 ) (map[string]api.ResourceThresholds, map[string][]api.ResourceThresholds) {
+	// we ensure that the user provided thresholds are valid here.
+	lowSpan, highSpan = sanitizeUserProvidedThresholds(lowSpan, highSpan)
```
ingvagabund, line 462: There's already some validation in place:
https://github.com/kubernetes-sigs/descheduler/blob/e283c3103026a3d134cc86a0c0fd67a108674e31/pkg/framework/plugins/nodeutilization/lownodeutilization.go#L287-L316. Which guarantees all the right thresholds are properly set. If not, `setDefaultForLNUThresholds` needs to be updated.

The current logic in place:
- if a threshold for the static case is not set for either of cpu, memory or pods resources, the resource is artificaly added and both low and high thresholds are set to 100%. Since later, once all the nodes are classified, the high threshold is used to compute node's the remaining resources for moving:
- LowNodeUtilization (moving from overutilized to underutilized): actual usage of each underutilized is substracted from the high threshold (how much of a given resource can be moved to an underutilized nodes before it becomes uverutilized)
- HighNodeUtilization (moving from underutilized to overutilized): The high threshold is always set to 100% since non-underutilized has the whole remaining capacity (100%) to be filed.

ingvagabund: The dynamic/deviation case (only for LowNodeUtilization) is the same in how the high threshold is used to get the remaining resources left in moving/landing evicted pods. However, `setDefaultForLNUThresholds` sets missing resources to 0% (as a special case that is not valid when both low and high thresholds are set to 0) to tell `getNodeThresholdsFromAverageNodeUsage` function this is the case. Which is handled by completely ignoring the average computation.

ingvagabund: So right now if both low and high thresholds get set to 100%, `normalizer.Sum(average, normalizer.Negate(lowSpan))` will produce a number <= 0. Which will get truncated to 0 instead of 1.

ingvagabund: Or, you could create a `thresholdNormalizer` function that will set thresholds for all artificially added resources to 100%. Invoking it right after `assessNodesUsagesAndRelativeThresholds` to avoid any extra data manipulation here.

pkg/framework/plugins/nodeutilization/nodeutilization.go
ricardomaraschini: Done.

pkg/framework/plugins/nodeutilization/nodeutilization.go
ricardomaraschini: Done.

pkg/framework/plugins/nodeutilization/nodeutilization.go
ingvagabund: This still remains unresolved.

ingvagabund: `NewLowNodeUtilization` can be extended with `thresholdNormalizer` function. Depending on which resources needs to be added artificially the function can set these to 100% and drop the code in `setDefaultForLNUThresholds`. Resp. `setDefaultForLNUThresholds` can be reshaped into `thresholdNormalizer` and invoked within `Balance` right after `getNodeThresholdsFromAverageNodeUsage`, resp. `assessNodesUsagesAndStaticThresholds`.

pkg/framework/plugins/nodeutilization/lownodeutilization.go
```diff
@@ -202,15 +215,21 @@ func (l *LowNodeUtilization) Balance(ctx context.Context, nodes []*v1.Node) *fra
 					allPods: podListMap[nodeName],
 				},
 				thresholds: NodeThresholds{
-					lowResourceThreshold:  resourceThresholdsToNodeUsage(nodeThresholdsMap[nodeName][0], nodesMap[nodeName]),
-					highResourceThreshold: resourceThresholdsToNodeUsage(nodeThresholdsMap[nodeName][1], nodesMap[nodeName]),
+					lowResourceThreshold:  resourceThresholdsToNodeUsage(thresholds[nodeName][0], capacities[nodeName], append(l.extendedResourceNames, v1.ResourceCPU, v1.ResourceMemory, v1.ResourcePods)),
```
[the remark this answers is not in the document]
ricardomaraschini: You might have commented this in an old version of the PR. This code is no more :-) I checked and this has been addressed with my last commit here.

pkg/framework/plugins/nodeutilization/lownodeutilization.go
```diff
@@ -202,15 +215,21 @@ func (l *LowNodeUtilization) Balance(ctx context.Context, nodes []*v1.Node) *fra
...
+					highResourceThreshold: resourceThresholdsToNodeUsage(thresholds[nodeName][1], capacities[nodeName], append(l.extendedResourceNames, v1.ResourceCPU, v1.ResourceMemory, v1.ResourcePods)),
```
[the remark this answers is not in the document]
ricardomaraschini: You might have commented this in an old version of the PR. This code is no more :-) I checked and this has been addressed with my last commit here.

pkg/framework/plugins/nodeutilization/nodeutilization.go
```diff
@@ -544,3 +441,155 @@ func classifyPods(pods []*v1.Pod, filter func(pod *v1.Pod) bool) ([]*v1.Pod, []*
...
+// assessNodesUsagesAndStaticThresholds converts the raw usage data into
+// percentage. Returns the usage (pct) and the thresholds (pct) for each
+// node.
+func assessNodesUsagesAndStaticThresholds(
+	rawUsages, rawCapacities map[string]api.ReferencedResourceList,
+	lowSpan, highSpan api.ResourceThresholds,
+) (map[string]api.ResourceThresholds, map[string][]api.ResourceThresholds) {
+	// first we normalize the node usage from the raw data (Mi, Gi, etc)
+	// into api.Percentage values.
+	usage := normalizer.Normalize(
```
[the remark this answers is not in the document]
ricardomaraschini: Is this comment in the right place ? I did not understand it. The `normalizer.Normalize` call converts to percent.

pkg/framework/plugins/nodeutilization/lownodeutilization.go
```diff
@@ -152,25 +156,29 @@ func (l *LowNodeUtilization) Balance(ctx context.Context, nodes []*v1.Node) *fra
...
+		usage, thresholds = assessNodesUsagesAndRelativeThresholds(
+			filterResourceNamesFromNodeUsage(nodesUsageMap, l.resourceNames),
+			capacities,
+			l.args.Thresholds,
+			l.args.TargetThresholds,
+		)
 	} else {
-		nodeThresholdsMap = getStaticNodeThresholds(nodes, l.args.Thresholds, l.args.TargetThresholds)
+		usage, thresholds = assessNodesUsagesAndStaticThresholds(
+			filterResourceNamesFromNodeUsage(nodesUsageMap, l.resourceNames),
```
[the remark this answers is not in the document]
ricardomaraschini: Done.

pkg/framework/plugins/nodeutilization/highnodeutilization.go
```diff
@@ -32,205 +32,222 @@ import (
...
 )
 
-const HighNodeUtilizationPluginName = "HighNodeUtilization"
+// this lines makes sure that HighNodeUtilization implements the BalancePlugin
+// interface.
+var _ frameworktypes.BalancePlugin = &HighNodeUtilization{}
 
-// HighNodeUtilization evicts pods from under utilized nodes so that scheduler can schedule according to its plugin.
-// Note that CPU/Memory requests are used to calculate nodes' utilization and not the actual resource usage.
+const HighNodeUtilizationPluginName = "HighNodeUtilization"
```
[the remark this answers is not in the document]
ricardomaraschini: Done.

pkg/framework/plugins/nodeutilization/highnodeutilization.go
```diff
@@ -32,205 +32,222 @@ import (
...
-			// underutilized nodes
+			// underutilized nodes.
 			func(nodeName string, usage, threshold api.ResourceThresholds) bool {
 				return isNodeBelowThreshold(usage, threshold)
 			},
-			// every other node that is schedulable
+			// schedulable nodes.
 			func(nodeName string, usage, threshold api.ResourceThresholds) bool {
-				if nodeutil.IsNodeUnschedulable(nodesMap[nodeName]) {
-					klog.V(2).InfoS("Node is unschedulable", "node", klog.KObj(nodesMap[nodeName]))
```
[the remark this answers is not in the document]
ricardomaraschini: Done.

pkg/framework/plugins/nodeutilization/lownodeutilization.go
```diff
@@ -32,198 +32,218 @@ import (
...
 )
 
-const LowNodeUtilizationPluginName = "LowNodeUtilization"
+// this lines makes sure that HighNodeUtilization implements the BalancePlugin
+// interface.
+var _ frameworktypes.BalancePlugin = &LowNodeUtilization{}
 
-// LowNodeUtilization evicts pods from overutilized nodes to underutilized nodes. Note that CPU/Memory requests are used
-// to calculate nodes' utilization and not the actual resource usage.
+const LowNodeUtilizationPluginName = "LowNodeUtilization"
```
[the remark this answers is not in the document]
ricardomaraschini: Done.

pkg/framework/plugins/nodeutilization/nodeutilization.go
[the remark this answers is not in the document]
ricardomaraschini: But we are returning `usage`. This function returns both usage (in pct) and thresholds. What am I missing.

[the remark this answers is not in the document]
ingvagabund: You are right. Please ignore.

ingvagabund:
/approve
/lgtm

@ricardomaraschini thank you very much

PR #1657, opened 2025-03-27
chore: log average and assessed thresholds

ricardomaraschini:
when calculating the average an applying the deviations it would be nice to also see the assessed values.

this commit makes the descheduler logs these values when using level 3.

### DIFF
--- a/pkg/framework/plugins/nodeutilization/highnodeutilization.go
+++ b/pkg/framework/plugins/nodeutilization/highnodeutilization.go
@@ -74,13 +74,6 @@ func NewHighNodeUtilization(
 		highThresholds[rname] = MaxResourcePercentage
 	}
 
-	// criteria is a list of thresholds that are used to determine if a node
-	// is underutilized. it is used only for logging purposes.
-	criteria := []any{}
-	for rname, rvalue := range args.Thresholds {
-		criteria = append(criteria, rname, rvalue)
-	}
-
 	podFilter, err := podutil.
 		NewOptions().
 		WithFilter(handle.Evictor().Filter).
@@ -106,7 +99,7 @@ func NewHighNodeUtilization(
 		args:           args,
 		resourceNames:  resourceNames,
 		highThresholds: highThresholds,
-		criteria:       criteria,
+		criteria:       thresholdsToKeysAndValues(args.Thresholds),
 		podFilter:      podFilter,
 		usageClient: newRequestedUsageClient(
 			resourceNames,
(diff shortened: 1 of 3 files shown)


pkg/framework/plugins/nodeutilization/highnodeutilization.go
```diff
@@ -197,7 +188,7 @@ func (h *HighNodeUtilization) Balance(ctx context.Context, nodes []*v1.Node) *fr
 
 	lowNodes, schedulableNodes := nodeInfos[0], nodeInfos[1]
 
-	klog.V(1).InfoS("Criteria for a node below target utilization", h.criteria...)
+	klog.V(1).InfoS("Criteria for a node below target utilization", thresholdsToKeysAndValues(h.args.Thresholds)...)
```
ingvagabund, line 191: Originally, the output for InfoS was precomputed to avoid extra computation each time the plugin is run. Let's keep `criteria` field around and just `criteria=thresholdsToKeysAndValues(h.args.Thresholds)`.

pkg/framework/plugins/nodeutilization/nodeutilization.go
```diff
@@ -487,25 +497,37 @@ func assessNodesUsagesAndRelativeThresholds(
...
+	// calculate the average usage.
 	average := normalizer.Average(usage)
+	klog.V(3).InfoS(
+		"Assessed average usage",
+		thresholdsToKeysAndValues(average)...,
+	)
+
+	// decrease the provided threshold from the average to get the low
+	// span. also make sure the resulting values are between 0 and 100.
+	lowSpan = normalizer.Clamp(
```
ingvagabund, line 509: Based on https://dave.cheney.net/2017/04/30/if-a-map-isnt-a-reference-variable-what-is-it a map is not a reference. So this should not change the actual value of `lowSpan` passed by invoker.

pkg/framework/plugins/nodeutilization/highnodeutilization.go
ricardomaraschini: Do we call `Balance()` multiple times for the same plugin during the same deschedule run ?

pkg/framework/plugins/nodeutilization/nodeutilization.go
ricardomaraschini: Done.

pkg/framework/plugins/nodeutilization/highnodeutilization.go
ricardomaraschini: My point here is that between executions we need to read/parse/reload the descheduler policy file and when we do that we may potentially have different thresholds.  This performance gain (populate `criteria` only when the plugin is instantiated) can be accomplished only if the whole pod goes down when the policy config change and while the latter does not change we don't create a new plugin instance. Please let me know where my assumptions are wrong.

ricardomaraschini: Done.

PR #1660, opened 2025-04-03
feat(helm): run descedulerPolicy thru tpl func for more chart control

schahal:
## Why

- Users may run descheduler in many differing environments (e.g., production, development, etc)
- A common practice is to have `values` files per environment with own settings
- Because `deschedulerPolicy` can get large, copying the whole policy for each environment just to update one parameter can get messy due to repetitive blocks and is difficult to synthesize

### Solution

Run the deschedulerPolicy thru the `tpl()` function.

This is a safe no-op by default unless users decide to use it [**see `Testing` section below**].

Users can then define something like `.Values.global.someVal` in each of their environment values files of and reference in the single it in the single `.Values.deschedulerPolicy` object [**see `Testing` section below**].

### No Diffs As-Is

Ran `helm template test .` before and after this change and confirmed **_no changes to the rendered manifests_**.

#### Input

-  Made following changes to local `values.yaml`

    ```diff
    --- a/charts/descheduler/values.yaml
    +++ b/charts/descheduler/values.yaml
    @@ -2,6 +2,9 @@
     [...]
    +global:
    +  cpuTargetThreshold: 75
    +
     [...]
    
    @@ -134,7 +137,7 @@ deschedulerPolicy:
                 [...]
                 targetThresholds:
    -              cpu: 50
    +              cpu: '{{ .Values.global.cpuTargetThreshold }}'
                   memory: 50
                   pods: 50
           plugins:
    ```

#### Output

-  Confirmed the following rendered manifest diff

    ```diff
    @@ -49,7 +49,7 @@
           - name: RemovePodsViolatingTopologySpreadConstraint
           - args:
               targetThresholds:
    -            cpu: 50
    +            cpu: '75'
                 memory: 50
                 pods: 50
               thresholds:
    @@ -145,7 +145,7 @@
             metadata:
               name: test-descheduler
               annotations:
    -            checksum/config: d74ea58f954e6a91c12674f2e9374dd67ead0ceadd26c971fc8aa2cd238c1034
    +            checksum/config: ab0598f0622575878545286b9399397518b413f48820edd1ef09fae9fd169fba
               labels:
                 app.kubernetes.io/name: descheduler
                 app.kubernetes.io/instance: test
        ```

### DIFF
--- a/charts/descheduler/templates/configmap.yaml
+++ b/charts/descheduler/templates/configmap.yaml
@@ -10,5 +10,5 @@ data:
   policy.yaml: |
     apiVersion: "{{ .Values.deschedulerPolicyAPIVersion }}"
     kind: "DeschedulerPolicy"
-{{ toYaml .Values.deschedulerPolicy | trim | indent 4 }}
+{{ tpl (toYaml .Values.deschedulerPolicy) . | trim | indent 4 }}
 {{- end }}


schahal:
☝️ these tests seem to a bit flaky and unrelated

```
E0403 13:45:27.941704       1 evictions.go:515] "Error evicting pod" err="maximum number of evicted pods per namespace reached" limit=1 namespace="default" pod="default/virt-launcher-kubevirtvmi-1-v58vh"
E0403 13:45:27.941733       1 pod_lifetime.go:164] eviction failed: maximum number of evicted pods per namespace reached
```

I'll re-run

LaikaN57:
@ingvagabund Any ideas on the tests there? (Looks like rerun failed.)

ingvagabund:
The tests needed to be adjusted. Passing now. Sorry for the trouble.

PR #1663, opened 2025-04-08
feat: introduce strict eviction policy

ricardomaraschini:
with strict eviction policy the descheduler only evict pods if the pod contains a request for the given threshold. for example, if using a threshold for an extended resource called `example.com/gpu` only pods who request such a resource will be evicted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -405,6 +405,12 @@ strategy evicts pods from `underutilized nodes` (those with usage below `thresho
 so that they can be recreated in appropriately utilized nodes.
 The strategy will abort if any number of `underutilized nodes` or `appropriately utilized nodes` is zero.
 
+To control pod eviction from underutilized nodes, use the `evictionModes`
+array. A lenient policy, which evicts pods regardless of their resource
+requests, is the default. To enable a stricter policy that only evicts pods
+with resource requests defined for the provided threshold resources, add the
+option `OnlyThresholdingResources` to the `evictionModes` configuration.
+
 **NOTE:** Node resource consumption is determined by the requests and limits of pods, not actual usage.
 This approach is chosen in order to maintain consistency with the kube-scheduler, which follows the same
 design for scheduling pods onto nodes. This means that resource usage as reported by Kubelet (or commands
@@ -417,7 +423,14 @@ actual usage metrics. Implementing metrics-based descheduling is currently TODO 
 |---|---|
 |`thresholds`|map(string:int)|
 |`numberOfNodes`|int|
+|`evictionModes`|list(string)|
 |`evictableNamespaces`|(see [namespace filtering](#namespace-filtering))|
+
+**Supported Eviction Modes:**
+
+|Name|Description|
+|---|---|
+|`OnlyThresholdingResources`|Evict only pods that have resource requests defined for the provided threshold resources.|
 
 **Example:**
 
@@ -437,6 +450,8 @@ profiles:
           exclude:
           - "kube-system"
           - "namespace1"
+        evictionModes:
+          - "OnlyThresholdingResources"
     plugins:
       balance:
         enabled:
--- a/pkg/framework/plugins/nodeutilization/nodeutilization.go
+++ b/pkg/framework/plugins/nodeutilization/nodeutilization.go
@@ -32,6 +32,7 @@ import (
 	"k8s.io/utils/ptr"
 	"sigs.k8s.io/descheduler/pkg/descheduler/evictions"
 	nodeutil "sigs.k8s.io/descheduler/pkg/descheduler/node"
+	"sigs.k8s.io/descheduler/pkg/descheduler/pod"
 	podutil "sigs.k8s.io/descheduler/pkg/descheduler/pod"
 	"sigs.k8s.io/descheduler/pkg/framework/plugins/nodeutilization/normalizer"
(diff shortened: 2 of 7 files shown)


pkg/framework/plugins/nodeutilization/types.go
```diff
@@ -48,6 +48,12 @@ type HighNodeUtilizationArgs struct {
 	Thresholds    api.ResourceThresholds `json:"thresholds"`
 	NumberOfNodes int                    `json:"numberOfNodes,omitempty"`
 
+	// RequireResourceRequestForEviction makes the descheduler evict only
+	// pods that have a resource request defined for any of the user
+	// provided thresholds. If the pod does not request the resource, it
+	// will not be evicted.
+	RequireResourceRequestForEviction bool `json:"requireResourceRequestForEviction,omitempty"`
```
ingvagabund, line 55: One way is to limit this new option for the requested resources only. The other is to name it the way that can be easily extended in the future. E.g. when a new usage client (which makes this filter meaningful) gets introduced. What about `exclusivePodEviction` or similar name? Something giving the idea only pods with resources listed in the `thresholds` field are considered.

ingvagabund: Or, instead of creating a bool option what about an enum? E.g.:
```go
type EvictionMode string

var (
	EvictAll EvictionMode = "All"
	EvictOnlyThresholdingResources EvictionMode = "OnlyThresholdingResources"
)
```
. Which could be later extended with other modes as they go without deprecating the boolean option.

ingvagabund: Or, we could also allow to combine multiple modes:
```yaml
kind: "DeschedulerPolicy"
profiles:
  - name: ProfileName
    pluginConfig:
    - name: "HighNodeUtilization"
      args:
        thresholds:
          "cpu" : 20
          "memory": 20
          "pods": 20
        evictableNamespaces:
          exclude:
          - "kube-system"
          - "namespace1"
        evictionModes:
          - OnlyThresholdingResources
          - NoSorting
    plugins:
      balance:
        enabled:
          - "HighNodeUtilization"
```

README.md
```diff
@@ -405,6 +405,13 @@ strategy evicts pods from `underutilized nodes` (those with usage below `thresho
 so that they can be recreated in appropriately utilized nodes.
 The strategy will abort if any number of `underutilized nodes` or `appropriately utilized nodes` is zero.
 
+When evicting pods from underutilized nodes you can choose between a more
+lenient eviction policy (`requireResourceRequestForEviction` set to `false`) or
+a stricter one (`requireResourceRequestForEviction` set to `true`). The lenient
+policy will evict any pod that is not a system critical pod, while the stricter
+one will only evict pods that are not system critical and have resource
```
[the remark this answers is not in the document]
ricardomaraschini: Rephrased, please take a look.

pkg/descheduler/pod/pods.go
```diff
@@ -68,8 +68,8 @@ func NewOptions() *Options {
 
 // WithFilter sets a pod filter.
 // The filter function should return true if the pod should be returned from ListPodsOnANode
-func (o *Options) WithFilter(filter FilterFunc) *Options {
-	o.filter = filter
+func (o *Options) WithFilter(filters ...FilterFunc) *Options {
```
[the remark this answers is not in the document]
ricardomaraschini: Done.

pkg/framework/plugins/nodeutilization/lownodeutilization.go
```diff
@@ -327,8 +329,8 @@ func validatePrometheusMetricsUtilization(args *LowNodeUtilizationArgs) error {
 		return fmt.Errorf("prometheus query is missing")
 	}
 
-	uResourceNames := getResourceNames(args.Thresholds)
-	oResourceNames := getResourceNames(args.TargetThresholds)
+	uResourceNames := slices.Collect(maps.Keys(args.Thresholds))
```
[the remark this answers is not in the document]
ricardomaraschini: Done.

pkg/framework/plugins/nodeutilization/nodeutilization.go
```diff
@@ -752,3 +753,19 @@ func assessAvailableResourceInNodes(
 
 	return available, nil
 }
+
+// WithResourceRequestForAny returns a filter function that checks if a pod
+// has a resource request specified for any of the given resources names.
+func WithResourceRequestForAny(names ...v1.ResourceName) pod.FilterFunc {
```
ingvagabund, line 759: This can stay private as it is used only within the package.

README.md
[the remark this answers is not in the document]
ingvagabund: Looks great :) Thank you

pkg/framework/plugins/nodeutilization/nodeutilization.go
ricardomaraschini: done.

README.md
```diff
@@ -437,6 +450,8 @@ profiles:
           exclude:
           - "kube-system"
           - "namespace1"
+        evictionModes:
```
googs1025, line 453: Do we have this mode parameter enabled by default? It would be reasonable for this plugin to only evict pods that use the configured resources by default?

Issue #1664, opened 2025-04-11
follow-up feature: refactor DefaultEvictorArgs for more clear config


googs1025:
**Is your feature request related to a problem? Please describe.**

base on: https://github.com/kubernetes-sigs/descheduler/pull/1603#discussion_r2039166461

so we create an issue to track. And I will try to discuss the changed config struct in this issue.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**What version of descheduler are you using?**

descheduler version:

**Additional context**

googs1025:
go struct
```go
type EvictionProtectionType string

const (
    withPvc           EvictionProtectionType = "withPvc"
    withoutPdb        EvictionProtectionType = "withoutPdb"
    withLocalStorage   EvictionProtectionType = "withLocalStorage"
    withSystemCritical EvictionProtectionType = "withSystemCritical"
    withFailedBarePods EvictionProtectionType = "withFailedBarePods"
)

type DefaultEvictorArgs struct {
	metav1.TypeMeta `json:",inline"`

	NodeSelector                 string                 `json:"nodeSelector,omitempty"`
	EvictLocalStoragePods        bool                   `json:"evictLocalStoragePods,omitempty"`
	EvictDaemonSetPods           bool                   `json:"evictDaemonSetPods,omitempty"`
	EvictSystemCriticalPods      bool                   `json:"evictSystemCriticalPods,omitempty"`
	IgnorePvcPods                bool                   `json:"ignorePvcPods,omitempty"`
	EvictFailedBarePods          bool                   `json:"evictFailedBarePods,omitempty"`
	LabelSelector                *metav1.LabelSelector  `json:"labelSelector,omitempty"`
	PriorityThreshold            *api.PriorityThreshold `json:"priorityThreshold,omitempty"`
	NodeFit                      bool                   `json:"nodeFit,omitempty"`
	MinReplicas                  uint                   `json:"minReplicas,omitempty"`
	MinPodAge                    *metav1.Duration       `json:"minPodAge,omitempty"`
	IgnorePodsWithoutPDB         bool                   `json:"ignorePodsWithoutPDB,omitempty"`
	
        EvictionProtection []EvictionProtectionType `json:"evictionProtection,omitempty"`
}

```

yaml: 

```yaml
defaultEvictor:
  evictionProtection:
    - withPvc           # Protect Pods with PVC
    - withoutPdb        # Protect Pods without PDB
[...]

PR #1665, opened 2025-04-12
add PodProtections for DefaultEvictorArgs

googs1025:
fix: https://github.com/kubernetes-sigs/descheduler/issues/1664


pkg/framework/plugins/defaultevictor/defaults.go
```diff
@@ -49,6 +49,9 @@ func SetDefaults_DefaultEvictorArgs(obj runtime.Object) {
 	if args.PriorityThreshold == nil {
 		args.PriorityThreshold = nil
 	}
+	if args.EvictionProtection == nil {
+		args.EvictionProtection = nil
+	}
```
ricardomaraschini, line 54: What is the purpose of this function ?

googs1025: oh... This is also where I am confused. I just added it in the same way as before...

README.md
```diff
@@ -192,11 +194,26 @@ profiles:
...
-        evictLocalStoragePods: true
+        # Deprecated: Use `disabledDefaultPodProtections` with "systemCriticalPods" instead.
+        # evictSystemCriticalPods: true
+        # Deprecated: Use `disabledDefaultPodProtections` with "failedBarePods" instead.
+        # evictFailedBarePods: true
+        # Deprecated: Use `disabledDefaultPodProtections` with "withLocalStorage" instead.
+        # evictLocalStoragePods: true
         nodeFit: true
         minReplicas: 2
+        # List of disabled default pod protections. This allows users to selectively disable certain protection rules.
```
googs1025, line 205: We will recommend users to use the new config here

pkg/framework/plugins/defaultevictor/validation.go
```diff
@@ -25,12 +25,23 @@ func ValidateDefaultEvictorArgs(obj runtime.Object) error {
...
-		return fmt.Errorf("priority threshold misconfigured, only one of priorityThreshold fields can be set, got %v", args)
+		return fmt.Errorf("priority threshold misconfigured, only one of priorityThreshold fields can be set, got Value: %v, Name: %q",
+			*args.PriorityThreshold.Value, args.PriorityThreshold.Name)
 	}
 
 	if args.MinReplicas == 1 {
 		klog.V(4).Info("DefaultEvictor minReplicas must be greater than 1 to check for min pods during eviction. This check will be ignored during eviction.")
 	}
 
+	// check if any deprecated fields are set to true
```
googs1025, line 36: validate check here: We do not want users to mix old and new config. If any of the old config is true and there is a value in the new config list (including: DisabledDefaultPodProtections and ExtraPodProtections), it means that they are mixed and the validation fails.

pkg/framework/plugins/defaultevictor/constraints.go
```diff
@@ -0,0 +1,283 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/klog/v2"
+
+	"sigs.k8s.io/descheduler/pkg/api"
+	podutil "sigs.k8s.io/descheduler/pkg/descheduler/pod"
+	frameworktypes "sigs.k8s.io/descheduler/pkg/framework/types"
+	"sigs.k8s.io/descheduler/pkg/utils"
+)
+
```
googs1025, line 31: I think add new file constraints.go to be more clear

googs1025:
I hope we can reach an agreement and merge it into the master before the next release. 😄 
@ingvagabund  @a7i @ricardomaraschini  Can you help with this PR? I know this may take some time....

pkg/framework/plugins/defaultevictor/defaults.go
ingvagabund: Unfortunate product of duplicating the no-op constructs.

pkg/framework/plugins/defaultevictor/types.go
```diff
@@ -25,16 +25,67 @@ import (
...
+	PriorityThreshold   *api.PriorityThreshold `json:"priorityThreshold,omitempty"`
+	NodeFit             bool                   `json:"nodeFit,omitempty"`
+	MinReplicas         uint                   `json:"minReplicas,omitempty"`
+	MinPodAge           *metav1.Duration       `json:"minPodAge,omitempty"`
+	// Deprecated: Use ExtraPodProtection with "withoutPDB" instead.
+	IgnorePodsWithoutPDB bool `json:"ignorePodsWithoutPDB,omitempty"`
+
+	// DisabledDefaultPodProtections is a list that specifies which default Pod protection mechanisms are disabled.
+	// Users can selectively disable certain default protection rules via this field.
+	DisabledDefaultPodProtections []DisabledDefaultPodProtection `json:"disabledDefaultPodProtections,omitempty"`
```
ingvagabund, line 49: What about to put the list of disabled and extra enabled policies under a single field? E.g.
```yaml
# Pod anti-eviction/protection policies.
# The list of policies enabled by default:
# - protect pods with local storage, extend disabled with 'podsWithLocalStorage' to disable the protection
# - protect daemon set pods, extend disabled with `daemonSetPods` to disable the protection
# - ...
protectionPolicies:
  # list of extra enabled policies
  extraEnabled:
  # list of disabled pod policies
  disabled:
```
?

pkg/framework/plugins/defaultevictor/types.go
```diff
@@ -25,16 +25,67 @@ import (
...
+
+	// ExtraPodProtections is a list that specifies additional Pod protection mechanisms that users can enable.
+	// These protection rules are optional and allow users to extend the protection scope based on their needs.
+	// Users can selectively enable certain extra protection rules via this field.
+	ExtraPodProtections []ExtraPodProtection `json:"extraPodProtections,omitempty"`
 }
+
+// DisabledDefaultPodProtection defines the type for default Pod protection mechanisms.
+// Each value represents a specific protection rule that can be disabled.
+type DisabledDefaultPodProtection string
```
ingvagabund, line 59: Both `DisabledDefaultPodProtection` and `ExtraPodProtection` can be consolidated into a single type. E.g.
```
type PodProtectionPolicy string

const (
	PodsWithLocalStorage PodProtectionPolicy = "podsWithLocalStorage"
	DaemonSetPods PodProtectionPolicy = "daemonSetPods"
	SystemCriticalPods PodProtectionPolicy = "systemCriticalPods"
	FailedBarePods PodProtectionPolicy = "failedBarePods"
	PodsWithPVC PodProtectionPolicy = "podsWithPVC"
	PodsWithoutPDB PodProtectionPolicy = "podsWithoutPDB"
)
```

With some of the policies enabled by default. Other used for extra enabling or disabling. Also, would you please have `defaults.go` set the default list of enabled policies too? E.g.
```
args.defaultPodProtectionPolicies = [PodsWithLocalStorage, DaemonSetPods, ...]
```
while extending the new private `defaultPodProtectionPolicies` field with a comment saying the list of default policies is a subject to change?

pkg/framework/plugins/defaultevictor/types.go
googs1025: Here is a good suggestion 
we can config like this: 

```
- name: "DefaultEvictor"
  args:
    # Deprecated: Use `disabledDefaultPodProtections` with "systemCriticalPods" instead.
    # evictSystemCriticalPods: true
    # Deprecated: Use `disabledDefaultPodProtections` with "failedBarePods" instead.
    # evictFailedBarePods: true
    # Deprecated: Use `disabledDefaultPodProtections` with "withLocalStorage" instead.
    # evictLocalStoragePods: true
    nodeFit: true
    minReplicas: 2
    protectionPolicies:
      extraEnabled:
         - withPVC
         - withoutPDB
      # list of disabled pod policies
      disabled:
         - withLocalStorage
         - systemCriticalPods
         - failedBarePods
         - daemonSetPods
```

pkg/framework/plugins/defaultevictor/types.go
googs1025: I thinks this(set default value) will make the behavior incompatible with previous. I think using two types will not have any impact for user. It will also avoid coding errors of different strategies in the delegate (because we distinguish enable and disable as different types).

ingvagabund: `defaultPodProtectionPolicies` is going to be a private variable only, not exposed to customers. Only for internal use.

googs1025: got this

pkg/framework/plugins/defaultevictor/types.go
```diff
@@ -25,16 +25,65 @@ import (
...
+
+var BuiltInPodProtectionPolicies = []PodProtectionPolicy{
+	PodsWithLocalStorage,
+	DaemonSetPods,
+	SystemCriticalPods,
+	FailedBarePods,
+}
+
+// PodProtectionPolicy defines the protection policy for a pod.
+type PodProtectionPolicy string
```
googs1025, line 67: Change to one type and use ExtraEnabled Disabled to decide

pkg/framework/plugins/defaultevictor/defaultevictor.go
```diff
@@ -77,139 +73,161 @@ func New(args runtime.Object, handle frameworktypes.Handle) (frameworktypes.Plug
...
-			if err != nil {
-				return fmt.Errorf("unable to list pods for minReplicas filter in the policy parameter")
-			}
+	// determine whether to use PodProtectionPolicies config
+	useNewConfig := len(args.PodProtectionPolicies.Disabled) > 0 || len(args.PodProtectionPolicies.ExtraEnabled) > 0
 
-			if uint(len(objs)) < defaultEvictorArgs.MinReplicas {
-				return fmt.Errorf("owner has %d replicas which is less than minReplicas of %d", len(objs), defaultEvictorArgs.MinReplicas)
-			}
+	if !useNewConfig {
```
googs1025, line 171: We are here to maintain compatibility with the old config logic.

pkg/framework/plugins/defaultevictor/defaultevictor.go
```diff
@@ -77,139 +73,161 @@ func New(args runtime.Object, handle frameworktypes.Handle) (frameworktypes.Plug
...
-				return fmt.Errorf("unable to check if pod is covered by PodDisruptionBudget: %w", err)
+	return effective
+}
+
+// legacyGetProtectedPolicies returns protected policies using old-style boolean flags.
+func legacyGetProtectedPolicies(args *DefaultEvictorArgs) []PodProtectionPolicy {
+	var policies []PodProtectionPolicy
+
+	defaultPolicies := args.defaultPodProtectionPolicies
+	for _, policy := range defaultPolicies {
```
ingvagabund, line 202: I am unable to see benefit of iterating through the list of default policies vs. checking the deprecated args directly. I.e.:
```
if !args.EvictLocalStoragePods {
	policies = append(policies, policy)
}
if !args.EvictDaemonSetPods {
	policies = append(policies, policy)
}
...
```
Would you please elaborate more on why the for loop is better?

pkg/framework/plugins/defaultevictor/defaultevictor.go
```diff
@@ -77,139 +73,161 @@ func New(args runtime.Object, handle frameworktypes.Handle) (frameworktypes.Plug
...
 
-			if len(pod.OwnerReferences) > 1 {
-				klog.V(5).InfoS("pod has multiple owner references which is not supported for minReplicas check", "size", len(pod.OwnerReferences), "pod", klog.KObj(pod))
-				return nil
-			}
+// getEffectiveProtectedPolicies determines which policies are currently active.
+// It supports both new-style (PodProtectionPolicies) and legacy-style flags.
+func getEffectiveProtectedPolicies(args *DefaultEvictorArgs) []PodProtectionPolicy {
+	if args.defaultPodProtectionPolicies == nil {
+		args.defaultPodProtectionPolicies = BuiltInPodProtectionPolicies
```
ingvagabund, line 165: `defaultPodProtectionPolicies` is used only for internal purposes so the same list of default policies does not have to be duplicated. If you plan to define `BuiltInPodProtectionPolicies` as well the new variable can be used instead of the internal `defaultPodProtectionPolicies` field.

pkg/framework/plugins/defaultevictor/defaultevictor.go
```diff
@@ -77,139 +73,161 @@ func New(args runtime.Object, handle frameworktypes.Handle) (frameworktypes.Plug
...
 	}
-	if defaultEvictorArgs.LabelSelector != nil && !selector.Empty() {
-		ev.constraints = append(ev.constraints, func(pod *v1.Pod) error {
-			if !selector.Matches(labels.Set(pod.Labels)) {
-				return fmt.Errorf("pod labels do not match the labelSelector filter in the policy parameter")
-			}
-			return nil
-		})
+
+	for _, policy := range AllPolicies {
```
ingvagabund, line 127: The same question here. Can you more elaborate on why iterating through `AllPolicies` is better than directly invoking the individual eviction constraint functions? I.e.
```go
addEvictionConstraintsForLocalStoragePods(ev, !policyMap[PodsWithLocalStorage])
addEvictionConstraintsForDaemonSetPods(ev, !policyMap[DaemonSetPods])
...

```

pkg/framework/plugins/defaultevictor/constraints.go
```diff
@@ -0,0 +1,188 @@
...
+func addEvictionConstraintsForFailedBarePods(ev *DefaultEvictor, evictFailedBarePods bool) {
```
ingvagabund, line 32: All the functions could return the constraint(s) and have it/them added outside. I.e.:
```go
func evictionConstraintsForFailedBarePods(evictFailedBarePods bool) constraint {
	if evictFailedBarePods {
		klog.V(1).InfoS("Warning: EvictFailedBarePods is set to True. This could cause eviction of pods without ownerReferences.")
		return func(pod *v1.Pod) error {
			if len(ownerRefList) == 0 && pod.Status.Phase != v1.PodFailed {
				return fmt.Errorf("pod does not have any ownerRefs and is not in failed phase")
			}
			return nil
		}
	}
	return func(pod *v1.Pod) error {
		if len(podutil.OwnerRef(pod)) == 0 {
			return fmt.Errorf("pod does not have any ownerRefs")
		}
		return nil
	}
}
...
ev.constraints = append(ev.constraints, evictionConstraintsForFailedBarePods(...))
ev.constraints = append(ev.constraints, evictionConstraintsForSystemCriticalPods(...)...)
ev.constraints = append(ev.constraints, evictionConstraintsForLocalStoragePods(...))
```

ingvagabund (commented):
To make the review and refactoring easier to follow would you please move the changes in constaints.go into a separate PR? Without introducing the pod protection policy. Just to refactor the code is it's easier to introduce the pod protection policies later.

pkg/framework/plugins/defaultevictor/defaultevictor.go
googs1025: removed

README.md
```diff
@@ -148,22 +148,66 @@ In general, each plugin can consume metrics from a different provider so multipl
...
-| `priorityThreshold`       |`priorityThreshold`|| (see [priority filtering](#priority-filtering))                                                                             |
-| `nodeFit`                 |`bool`|`false`| (see [node fit filtering](#node-fit-filtering))                                                                             |
-| `minReplicas`             |`uint`|`0`| ignore eviction of pods where owner (e.g. `ReplicaSet`) replicas is below this threshold                                    |
-| `minPodAge`               |`metav1.Duration`|`0`| ignore eviction of pods with a creation time within this threshold                                                          |
-| `ignorePodsWithoutPDB`    |`bool`|`false`| set whether pods without PodDisruptionBudget should be evicted or ignored                                                   |
-
-### Example policy
+| Name                      | Type                     | Default Value | Description                                                                                                                                                                                                           |
```
googs1025, line 151: see https://github.com/googs1025/descheduler/tree/refator/evict_arg?tab=readme-ov-file#evictor-plugin-configuration-default-evictor 

This can see more clearer.

ingvagabund:
Apologies for the delays. I will not be able to finish the review. I am leaving for a vacation until June 8. I will resume once I am back.

googs1025:
@ingvagabund /PTAL when you have time 😄

pkg/framework/plugins/defaultevictor/defaults.go
ricardomaraschini: https://github.com/kubernetes-sigs/descheduler/pull/1717

README.md
```diff
@@ -148,22 +148,66 @@ In general, each plugin can consume metrics from a different provider so multipl
...
+|---------------------------|--------------------------|---------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `nodeSelector`            | `string`                 | `nil`        | Limits the nodes that are processed.                                                                                                                                                                                  |
+| `evictLocalStoragePods`   | `bool`                   | `false`      | **[Deprecated: Use `protectionPolicies` with `"podsWithLocalStorage"` instead]**<br>Allows eviction of pods using local storage.                                                                                          |
```
ricardomaraschini: I would expect the protections to be capitalized, e.g. `PodsWithLocalStorage` instead of `podsWithLocalStorage`.

README.md
```diff
@@ -148,22 +148,66 @@ In general, each plugin can consume metrics from a different provider so multipl
...
+| `nodeFit`                 | `bool`                   | `false`      | (See [node fit filtering](#node-fit-filtering))                                                                                                                                                                       |
+| `minReplicas`             | `uint`                   | `0`          | Ignores eviction of pods where the owner (e.g., `ReplicaSet`) replicas are below this threshold.                                                                                                                      |
+| `minPodAge`               | `metav1.Duration`        | `0`          | Ignores eviction of pods with a creation time within this threshold.                                                                                                                                                  |
+| `ignorePodsWithoutPDB`    | `bool`                   | `false`      | **[Deprecated: Use `protectionPolicies` with `"podsWithoutPDB"` instead]**<br>Sets whether pods without PodDisruptionBudget should be evicted or ignored.                                                                 |
```
ricardomaraschini: Can we move this closer to the other protections on this table ?

README.md
```diff
@@ -148,22 +148,66 @@ In general, each plugin can consume metrics from a different provider so multipl
...
+#### Supported Values for `protectionPolicies.Disabled`
+
+| Value                    | Meaning                                                                 |
+|--------------------------|-------------------------------------------------------------------------|
+| `"podsWithLocalStorage"` | Prevents eviction of Pods using local storage. Equivalent to old field `evictLocalStoragePods: true`. |
+| `"daemonSetPods"`        | Prevents eviction of DaemonSet-managed Pods. Equivalent to old field `evictDaemonSetPods: true`.     |
+| `"systemCriticalPods"`   | Prevents eviction of system-critical Pods. Equivalent to old field `evictSystemCriticalPods: true`.  |
+| `"failedBarePods"`       | Prevents eviction of failed bare Pods (without controllers). Equivalent to old field `evictFailedBarePods: true`. |
+
+> Setting these values will **disable the default protection rules**, meaning these Pods may be evicted.
```
ricardomaraschini: This should be placed before the table of possible values I guess. Could we also be more descriptive on what Protections are ?

README.md
```diff
@@ -148,22 +148,66 @@ In general, each plugin can consume metrics from a different provider so multipl
...
+| `ignorePodsWithoutPDB`    | `bool`                   | `false`      | **[Deprecated: Use `protectionPolicies` with `"podsWithoutPDB"` instead]**<br>Sets whether pods without PodDisruptionBudget should be evicted or ignored.                                                                 |
+| `protectionPolicies`      | `PodProtections`         | `{}`          | Holds the list of enabled and disabled protection policies.<br>Users can selectively disable certain default protection rules or enable extra ones. See below for supported values.                                   |### Example policy
+
+#### Supported Values for `protectionPolicies.Disabled`
+
+| Value                    | Meaning                                                                 |
+|--------------------------|-------------------------------------------------------------------------|
+| `"podsWithLocalStorage"` | Prevents eviction of Pods using local storage. Equivalent to old field `evictLocalStoragePods: true`. |
```
ricardomaraschini: I think we need to invert the logic of the "Meaning" column here. If one enters these values here he or she is in fact "allowing" and not "preventing" these pods to be evicted. 

Something like:

`Allow eviction of Pods using Local Storage`.

I would also remove the reference to the old field here.

README.md
```diff
@@ -148,22 +148,66 @@ In general, each plugin can consume metrics from a different provider so multipl
...
+---
+
+#### Supported Values for `protectionPolicies.ExtraEnabled`
+
+| Value              | Meaning                                                                 |
+|--------------------|-------------------------------------------------------------------------|
+| `"podsWithPVC"`    | Prevents eviction of Pods using Persistent Volume Claims (PVCs). Equivalent to old field `ignorePvcPods: true`. |
+| `"podsWithoutPDB"` | Prevents eviction of Pods without a PodDisruptionBudget (PDB). Equivalent to old field `ignorePodsWithoutPDB: true`. |
+
+> Setting these values will **enable additional protection rules**, meaning these Pods will not be evicted.
```
ricardomaraschini: Ditto, move this before the table of possible values.

README.md
```diff
@@ -148,22 +148,66 @@ In general, each plugin can consume metrics from a different provider so multipl
...
+
+---
+
+#### Example: Migrating from Old Fields to New Fields
```
ricardomaraschini: I don't think this portion is necessary. We already have a configuration example below.

README.md
```diff
@@ -148,22 +148,66 @@ In general, each plugin can consume metrics from a different provider so multipl
...
+| `minPodAge`               | `metav1.Duration`        | `0`          | Ignores eviction of pods with a creation time within this threshold.                                                                                                                                                  |
+| `ignorePodsWithoutPDB`    | `bool`                   | `false`      | **[Deprecated: Use `protectionPolicies` with `"podsWithoutPDB"` instead]**<br>Sets whether pods without PodDisruptionBudget should be evicted or ignored.                                                                 |
+| `protectionPolicies`      | `PodProtections`         | `{}`          | Holds the list of enabled and disabled protection policies.<br>Users can selectively disable certain default protection rules or enable extra ones. See below for supported values.                                   |### Example policy
+
+#### Supported Values for `protectionPolicies.Disabled`
```
ricardomaraschini: I think that `DefaultDisabled` (instead of `Disabled`) conveys a more accurate meaning and make it easier to differentiate from the `ExtraEnabled`.

README.md
```diff
@@ -193,11 +237,29 @@ profiles:
...
+        # evictFailedBarePods: true
+        # Deprecated: Use `disabledDefaultPodProtections` with "withLocalStorage" instead.
+        # evictLocalStoragePods: true
         nodeFit: true
         minReplicas: 2
+        # PodProtectionPolicies holds the list of enabled and disabled protection policies.
+        # Users can selectively disable certain default protection rules or enable extra ones.
+        protectionPolicies:
+          # ExtraEnabled specifies additional protection policies that should be enabled.
+          # Supports: podsWithPVC, podsWithoutPDB
```
ricardomaraschini: We already have a table with supported values. Having the same information in this line here makes things harder to maintain.

README.md
```diff
@@ -193,11 +237,29 @@ profiles:
...
+          extraEnabled:
+          #  - podsWithPVC
+          #  - podsWithoutPDB 
+          # Disabled specifies which default protection policies should be disabled.
+          # Supports: podsWithLocalStorage, daemonSetPods, systemCriticalPods, failedBarePods
```
ricardomaraschini: We already have a table with supported values. Having the same information in this line here makes things harder to maintain.

README.md
```diff
@@ -148,22 +148,66 @@ In general, each plugin can consume metrics from a different provider so multipl
...
+| `minReplicas`             | `uint`                   | `0`          | Ignores eviction of pods where the owner (e.g., `ReplicaSet`) replicas are below this threshold.                                                                                                                      |
+| `minPodAge`               | `metav1.Duration`        | `0`          | Ignores eviction of pods with a creation time within this threshold.                                                                                                                                                  |
+| `ignorePodsWithoutPDB`    | `bool`                   | `false`      | **[Deprecated: Use `protectionPolicies` with `"podsWithoutPDB"` instead]**<br>Sets whether pods without PodDisruptionBudget should be evicted or ignored.                                                                 |
+| `protectionPolicies`      | `PodProtections`         | `{}`          | Holds the list of enabled and disabled protection policies.<br>Users can selectively disable certain default protection rules or enable extra ones. See below for supported values.                                   |### Example policy
```
ricardomaraschini: Why isn't this simply called `podProtections` instead of `protectionPolicies` ? I mean, make the config concise with the struct naming makes everything easier in the long run.

pkg/descheduler/policyconfig.go
```diff
@@ -108,6 +108,10 @@ func setDefaultEvictor(profile api.DeschedulerProfile, client clientset.Interfac
 			IgnorePvcPods:           false,
 			EvictFailedBarePods:     false,
 			IgnorePodsWithoutPDB:    false,
+			PodProtectionPolicies: defaultevictor.PodProtections{
```
ricardomaraschini: This seems confusing, we should stick with one naming. I propose `PodProtection`. It could be then singular or plural. What do you think ?

pkg/descheduler/policyconfig_test.go
```diff
@@ -153,7 +155,7 @@ profiles:
 					t.Errorf("unexpected error: %s. Was expecting %s", err.Error(), tc.err.Error())
 				}
 			}
-			diff := cmp.Diff(tc.result, result)
+			diff := cmp.Diff(tc.result, result, cmpopts.IgnoreFields(defaultevictor.DefaultEvictorArgs{}, "defaultPodProtectionPolicies"))
```
ricardomaraschini: OK, this does not seem correct. Why are we introducing a private configuration here ? This should not be like that.

pkg/framework/plugins/defaultevictor/types.go
```diff
@@ -25,16 +25,62 @@ import (
...
+	// defaultPodProtectionPolicies is a private field that contains the list of PodProtectionPolicy values
+	// that are enabled by default. These policies represent the built-in, default protections applied to
+	// certain types of Pods, such as DaemonSet pods or SystemCritical pods. Users can selectively disable
+	// any of these using the "disabled" list in PodProtectionPolicies.
+	// The following four policies are included by default:
+	//   - PodsWithLocalStorage: Protects pods with local storage.
+	//   - DaemonSetPods: Protects DaemonSet managed pods.
+	//   - SystemCriticalPods: Protects system-critical pods.
+	//   - FailedBarePods: Protects failed bare pods (not part of any controller).
+	defaultPodProtectionPolicies []PodProtectionPolicy
```
ricardomaraschini: I don't think this is right. We don't need to have these defined in the config do we ? Can't we have those defined globally as discussed [here](https://github.com/kubernetes-sigs/descheduler/issues/1664#issuecomment-2818107014) ? I mean something like:

```go
// PodProtection is a name of a policy that protects pods against eviction.
type PodProtection string

// This is a list of all supported protections, some are enabled by default
// some may be enabled by the user through the configuration.
const (
        PodsWithLocalStorage PodProtection = "PodsWithLocalStorage"
        SystemCriticalPods   PodProtection = "SystemCriticalPods"
        FailedBarePods       PodProtection = "FailedBarePods"
        DaemonSetPods        PodProtection = "DaemonSetPods"
        PodsWithoutPDB       PodProtection = "PodsWithoutPDB"
        PodsWithPVC          PodProtection = "PodsWithPVC"
)

// DefaultPodProtections holds the list of the protections that are enabled by
// default. User can use disabledDefaultPodProtections evictor arguments to
// disable these.
var DefaultPodProtections = []PodProtection{
        PodsWithLocalStorage,
        SystemCriticalPods,
        FailedBarePods,
        DaemonSetPods,
}

// ExtraPodProtections holds a list of the protections the user can enable
// through the configuration. See extraPodProtections below. These are
// not enabled by default.
var ExtraPodProtections = []PodProtection{
        PodsWithPVC,
        PodsWithoutPDB,
}
```

I mean, I don't see why we need this as part of the configuration as these are fixed and known in advance. Is there any special reason for this ?

ricardomaraschini (changes requested):
I am going to stop reviewing this half away through as some of my comments here may ended up reshaping the PR a little bit. Please take a look at my comments and let's discuss them.

ricardomaraschini:
@googs1025 Do you think you will be able to work on this ? I would like to get it in as soon as possible. Let me know if you can't work on it (I may take this one over).

googs1025:
> @googs1025 Do you think you will be able to work on this ? I would like to get it in as soon as possible. Let me know if you can't work on it (I may take this one over).

Sorry for being late. Thank you for your review, I plan to make some changes based on your suggestions this week.

googs1025:
I will rebase and  change codes acording  review first. I will ping you again after finishing above. Since you have more experience, if you have a better way of writing, you may be able to make improvements directly in this PR, which is faster. 😄

README.md
googs1025: how about change to: 

> Setting a value in `disabled` **disables the corresponding default protection rule**. This means the specified type of Pods will **no longer be protected** from eviction and may be evicted if they meet other criteria.

README.md
googs1025: ok

README.md
googs1025: got this.

pkg/descheduler/policyconfig_test.go
googs1025: removed.

pkg/framework/plugins/defaultevictor/types.go
googs1025: ok. added this

README.md
googs1025: change to: 

`| `podProtections`          | `PodProtections`        | `{}`          | Holds the list of enabled and disabled protection pod policies.<br>Users can selectively disable certain default protection rules or enable extra ones. See below for supported values.                             |
`

README.md
googs1025: removed

README.md
googs1025: removed

README.md
googs1025: removed

README.md
googs1025: done

README.md
googs1025: done

README.md
googs1025: change to capitalized

README.md
ricardomaraschini: Sounds good.

README.md
ricardomaraschini: This sounds way better! Thanks.

pkg/descheduler/policyconfig_test.go
```diff
@@ -496,21 +497,257 @@ profiles:
...
+        nodeFit: true
+    plugins:
+      filter:
+        enabled:
+          - "DefaultEvictor"
+      preEvictionFilter:
+        enabled:
+          - "DefaultEvictor"
+`),
+			result: &api.DeschedulerPolicy{
```
ricardomaraschini: Is a DeschedulerPolicy returned even when an error like `cannot use Deprecated fields alongside PodProtections.ExtraEnabled or PodProtections.DefaultDisabled` happens ? One would expect the return to be `nil` and not an object.

pkg/descheduler/policyconfig_test.go
```diff
@@ -496,21 +497,257 @@ profiles:
...
+							},
+							PreEvictionFilter: api.PluginSet{
+								Enabled: []string{defaultevictor.PluginName},
+							},
+						},
+					},
+				},
+			},
+			err: fmt.Errorf("in profile ProfileName: cannot use Deprecated fields alongside PodProtections.ExtraEnabled or PodProtections.DefaultDisabled"),
+		},
```
ricardomaraschini, line 806: I missed a few test cases here (they all should result in a failure):

- Disables a default protection that does not exist.
- Enables an extra protection that does not exist.
- Disables an extra protection.
- Enables a default protection.

pkg/framework/plugins/defaultevictor/defaultevictor.go
```diff
@@ -106,10 +104,107 @@ func (d *DefaultEvictor) addAllConstraints(logger klog.Logger, handle frameworkt
...
+
+	if !useNewConfig {
+		// fall back to the Deprecated config
+		return legacyGetPodProtections(args)
+	}
+
+	effective := make([]PodProtection, 0)
+	effective = append(effective, DefaultPodProtections...)
+
+	for _, disabledPolicy := range args.PodProtections.DefaultDisabled {
```
ricardomaraschini: These loops can be replaced by something like this I guess:

```go
effective = slices.DeleteFunc(effective, func(protection PodProtection) bool {
    return slices.Index(args.PodProtections.DefaultDisabled, protection) >= 0
})
```

pkg/framework/plugins/defaultevictor/defaultevictor.go
```diff
@@ -106,10 +104,107 @@ func (d *DefaultEvictor) addAllConstraints(logger klog.Logger, handle frameworkt
...
+				// Remove the policy
+				effective = append(effective[:i], effective[i+1:]...)
+				break
+			}
+		}
+	}
+
+	// Add extra enabled policies if not included
+	for _, policy := range args.PodProtections.ExtraEnabled {
+		if !contains(effective, policy) {
```
ricardomaraschini: I don't think this is necessary. There is no way effective will contain anything other than DefaultPodProtections, right ? Assuming the configuration has been sanitized this could be replaced with the way simpler:

```go
effective = append(effective, args.PodProtections.ExtraEnabled...)
```

pkg/framework/plugins/defaultevictor/defaultevictor.go
```diff
@@ -106,10 +104,107 @@ func (d *DefaultEvictor) addAllConstraints(logger klog.Logger, handle frameworkt
...
+	d.constraints = append(d.constraints, evictionConstraintsForLocalStoragePods(!policyMap[PodsWithLocalStorage])...)
+	d.constraints = append(d.constraints, evictionConstraintsForDaemonSetPods(!policyMap[DaemonSetPods])...)
+
+	d.constraints = append(d.constraints, evictionConstraintsForPvcPods(policyMap[PodsWithPVC])...)
+	d.constraints = append(d.constraints, evictionConstraintsForIgnorePodsWithoutPDB(policyMap[PodsWithoutPDB], handle)...)
+	return nil
+}
+
+// contains checks if a slice of PodProtectionPolicy contains a given policy.
+func contains(list []PodProtection, item PodProtection) bool {
```
ricardomaraschini: This isn't necessary but now there is a `slices.Contains` function in the standard library that can be used.

pkg/framework/plugins/defaultevictor/defaultevictor.go
```diff
@@ -106,10 +104,107 @@ func (d *DefaultEvictor) addAllConstraints(logger klog.Logger, handle frameworkt
...
-	d.constraints = append(d.constraints, evictionConstraintsForIgnorePodsWithoutPDB(args.IgnorePodsWithoutPDB, handle)...)
 	return nil
 }
 
+// applyEffectivePodProtections applies all relevant protection pod policies to the plugin.
+// For each:
+// - If it's in the list, we disallow eviction (shouldEvict = false).
+// - If it's NOT in the list, we allow eviction (shouldEvict = true).
+func applyEffectivePodProtections(d *DefaultEvictor, podProtections []PodProtection, handle frameworktypes.Handle) error {
+	policyMap := make(map[PodProtection]bool)
```
ricardomaraschini: can we use `protection` instead of `policy` in this function ?

pkg/framework/plugins/defaultevictor/defaultevictor.go
```diff
@@ -106,10 +104,107 @@ func (d *DefaultEvictor) addAllConstraints(logger klog.Logger, handle frameworkt
...
+	if constraints, err := evictionConstraintsForSystemCriticalPods(d.logger, !policyMap[SystemCriticalPods], d.args.PriorityThreshold, handle); err != nil {
+		return err
+	} else {
+		d.constraints = append(d.constraints, constraints...)
+	}
+	d.constraints = append(d.constraints, evictionConstraintsForLocalStoragePods(!policyMap[PodsWithLocalStorage])...)
+	d.constraints = append(d.constraints, evictionConstraintsForDaemonSetPods(!policyMap[DaemonSetPods])...)
+
+	d.constraints = append(d.constraints, evictionConstraintsForPvcPods(policyMap[PodsWithPVC])...)
+	d.constraints = append(d.constraints, evictionConstraintsForIgnorePodsWithoutPDB(policyMap[PodsWithoutPDB], handle)...)
```
ricardomaraschini: Can we try to simplify this function with a `for` and a `switch` statement ?  I think this is confusing to understand.

ricardomaraschini (changes requested):

pkg/framework/plugins/defaultevictor/defaultevictor.go
googs1025: done

pkg/framework/plugins/defaultevictor/defaultevictor.go
googs1025: yes. sgtm

pkg/framework/plugins/defaultevictor/defaultevictor.go
googs1025: done

pkg/framework/plugins/defaultevictor/defaultevictor.go
googs1025: done

pkg/descheduler/policyconfig_test.go
googs1025: added

pkg/descheduler/policyconfig_test.go
googs1025: done

googs1025:
> This is looking better! There is some inconsitencies in the following functions:
> 
> ```go
> evictionConstraintsForSystemCriticalPods()
> evictionConstraintsForFailedBarePods()
> evictionConstraintsForLocalStoragePods()
> evictionConstraintsForDaemonSetPods()
> evictionConstraintsForPvcPods()
> evictionConstraintsForIgnorePodsWithoutPDB()
> ```
> 
> All these functions receive a `bool` argument but the argument semantics change depending on the function. Take for example the function `evictionConstraintsForSystemCriticalPods()`. We already negate the bool when calling it just for it to negate it once more. **NOTE**: it also still logs a message referring to the old configuration and this needs to be changed.
> 
> What I would like to suggest is that all these functions receive a `bool` indicating **whether the protection is enabled or not**. If it is true then it is enable, if false then it is disabled. And each function then deals with the logic on its own way. In other words, let's standardize the semantics of the `bool` argument across all these functions.
> 
> What do you think ?
> 
> If you don't mind please keep your next changes in separated commits as it makes it easier to review. Later we can squash them all together.

Yes, the reason I used negation bool is due to historical logic... I prefer not to modify the existing logic but instead adapt it at a higher level. A better way would be for me to submit another small PR to address this issue.  🤔

pkg/framework/plugins/defaultevictor/types.go
```diff
@@ -25,19 +25,82 @@ import (
...
+// DefaultPodProtections holds the list of protection policies that are enabled by default.
+// User can use the 'disabledDefaultPodProtections' evictor arguments (via PodProtections.DefaultDisabled)
+// to disable any of these default protections.
+//
+// The following four policies are included by default:
+//   - PodsWithLocalStorage: Protects pods with local storage.
+//   - DaemonSetPods: Protects DaemonSet managed pods.
+//   - SystemCriticalPods: Protects system-critical pods.
+//   - FailedBarePods: Protects failed bare pods (not part of any controller).
+var DefaultPodProtections = []PodProtection{
```
ingvagabund: `DefaultPodProtections` does not seem to be used outside of the package. Making the variable not-exported will easy the maintenance guarantees.

pkg/framework/plugins/defaultevictor/types.go
```diff
@@ -25,19 +25,82 @@ import (
...
+	DaemonSetPods,
+}
+
+// ExtraPodProtections holds a list of protection policies that the user can optionally enable
+// through the configuration (via PodProtections.ExtraEnabled). These policies are not enabled by default.
+//
+// Currently supported extra policies:
+//   - PodsWithPVC: Protects pods using PersistentVolumeClaims.
+//   - PodsWithoutPDB: Protects pods lacking a PodDisruptionBudget.
+var ExtraPodProtections = []PodProtection{
```
ingvagabund: `ExtraPodProtections` does not seem to be used outside of the package. Making the variable not-exported will easy the maintenance guarantees.

pkg/framework/plugins/defaultevictor/defaultevictor_test.go
```diff
@@ -330,11 +331,13 @@ func TestDefaultEvictorFilter(t *testing.T) {
 					pod.Status.Phase = v1.PodFailed
 				}),
 			},
-		}, {
+		},
```
ingvagabund, line 334: Nit: Keeping the line with `}, {` make the diff smaller. Also, there are many convention used across unit tests. Both `}, {` and `},>NL>{` are acceptable as long as they are not mixed within the same unit test.

pkg/framework/plugins/defaultevictor/validation.go
```diff
@@ -38,5 +39,32 @@ func ValidateDefaultEvictorArgs(obj runtime.Object) error {
...
+			}
+		}
+
+		for _, policy := range args.PodProtections.DefaultDisabled {
+			if !slices.Contains(DefaultPodProtections, policy) {
+				return fmt.Errorf("invalid pod protection policy in DefaultDisabled: %q. Valid options are: %v",
+					string(policy), DefaultPodProtections)
+			}
+		}
+	}
```
ingvagabund, line 75: Checking for duplicates might be another validation step. I.e. `DefaultDisabled = ["FailedBarePods", "FailedBarePods"]` should be invalid to avoid creating a too long list. The same for `ExtraEnabled`.

pkg/framework/plugins/defaultevictor/defaultevictor.go
```diff
@@ -107,10 +105,123 @@ func (d *DefaultEvictor) addAllConstraints(logger klog.Logger, handle frameworkt
...
+	applyDaemonSetPodsProtection(d, protectionMap)
+	applyPvcPodsProtection(d, protectionMap)
+	applyPodsWithoutPDBProtection(d, protectionMap, handle)
+
+	return nil
+}
+
+func applyFailedBarePodsProtection(d *DefaultEvictor, protectionMap map[PodProtection]bool) {
+	isProtectionEnabled := protectionMap[FailedBarePods]
+	d.constraints = append(d.constraints, evictionConstraintsForFailedBarePods(d.logger, !isProtectionEnabled)...)
```
ingvagabund, line 133: Given you are already introducing new `applyFailedBarePodsProtection` what about to inline `evictionConstraintsForFailedBarePods` here? To remove the extra invocation to avoid passing `!isProtectionEnabled`? The same for other `applyXXX` functions.

pkg/framework/plugins/defaultevictor/defaultevictor_test.go
googs1025: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_descheduler/1665/pull-descheduler-verify-master/1950927774013198336

This was not my original intention to do, but I encountered several CI failed. Similar to this. 🤔

googs1025: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_descheduler/1665/pull-descheduler-verify-master/1950932234429009920

pkg/framework/plugins/defaultevictor/validation.go
googs1025: added

pkg/framework/plugins/defaultevictor/types.go
googs1025: done

pkg/framework/plugins/defaultevictor/types.go
googs1025: done

pkg/framework/plugins/defaultevictor/defaultevictor.go
googs1025: in this pr will fix this: https://github.com/kubernetes-sigs/descheduler/pull/1725

pkg/framework/plugins/defaultevictor/defaultevictor_test.go
ingvagabund: Oh, the lint :-|. Gotcha. Please ignore my comment then :)

ingvagabund:
Will leave the final lgtm for @ricardomaraschini

PR #1666, opened 2025-04-15
Bump latest go version

sammedsingalkar09:
Bump the latest go version


sammedsingalkar09:
> @sammedsingalkar09 Descheduler's Go version is in sync with upstream Kubernetes. Please note that Kubernetes 1.32 is using Go 1.23 https://github.com/kubernetes/kubernetes/blob/release-1.32/go.mod#L9

To fix vulnerabilities in
golang.org/x/crypto and golang.org/x/net,
is it sufficient to just update these specific dependencies ?

PR #1672, opened 2025-04-23
add miss evictDaemonSetPods param in README.md

googs1025:
- add miss evictDaemonSetPods param in README.md


googs1025:
@a7i can you take a quick look this? I think this is just a miss

PR #1673, opened 2025-04-24
Use Boolean logic for descheduler.alpha.kubernetes.io/evict annotation

tiraboschi:
In all the example and tests descheduler.alpha.kubernetes.io/evict annotation was used with = true value to force a pod to be considered evictable regardless of all the other checks. Surprisingly descheduler.alpha.kubernetes.io/evict=false was also producing the same effect.
Properly implement Boolean logic so that
descheduler.alpha.kubernetes.io/evict=false can be used to configure a pod to be ignored by the descheduler regardless of all the other checks.


pkg/framework/plugins/defaultevictor/defaultevictor.go
```diff
@@ -58,10 +59,20 @@ func IsPodEvictableBasedOnPriority(pod *v1.Pod, priority int32) bool {
 	return pod.Spec.Priority == nil || *pod.Spec.Priority < priority
 }
 
-// HaveEvictAnnotation checks if the pod have evict annotation
-func HaveEvictAnnotation(pod *v1.Pod) bool {
-	_, found := pod.ObjectMeta.Annotations[evictPodAnnotationKey]
-	return found
+// HaveEvictAnnotation checks if the pod have evict annotation and its value
+func HaveEvictAnnotation(pod *v1.Pod) (bool, bool) {
+	svalue, found := pod.ObjectMeta.Annotations[evictPodAnnotationKey]
```
a7i, line 64: thoughts on using `strconv.ParseBool`?

tiraboschi: I think it depends on how much we want to push in terms of backward compatibility: the current code is checking only the presence of the annotation.
`strconv.ParseBool` will return a `syntaxError` for values (like an empty string) that fails to be cleanly converted to a Bool. Using `strconv.ParseBool`  and catching the error to consider it as a True to mimic backward compatibility is also an option.

tiraboschi: Yes, this is also a great suggestion.

tiraboschi: Done

pkg/framework/plugins/defaultevictor/defaultevictor.go
```diff
@@ -55,10 +57,27 @@ func IsPodEvictableBasedOnPriority(pod *v1.Pod, priority int32) bool {
 	return pod.Spec.Priority == nil || *pod.Spec.Priority < priority
 }
 
-// HaveEvictAnnotation checks if the pod have evict annotation
-func HaveEvictAnnotation(pod *v1.Pod) bool {
-	_, found := pod.ObjectMeta.Annotations[evictPodAnnotationKey]
-	return found
+// HaveEvictAnnotation checks if the pod have evict annotation and its value
+func HaveEvictAnnotation(pod *v1.Pod) (bool, bool, string) {
```
googs1025, line 60: Can we change it to this? I think it's more readable. 🤔 
```go
func HaveEvictAnnotation(pod *v1.Pod) (bool, bool, string) {
	// Check new (beta) annotation first
	if svalue, found := pod.Annotations[evictPodAnnotationKeyBeta]; found {
		bvalue, err := strconv.ParseBool(svalue)
		if err != nil {
			return false, true, evictPodAnnotationKeyBeta // Return false on parse error but indicate annotation was found
		}
		return bvalue, true, evictPodAnnotationKeyBeta
	}

	// Fallback to old (alpha) annotation
	if _, found := pod.Annotations[evictPodAnnotationKeyAlpha]; found {
		return true, true, evictPodAnnotationKeyAlpha
	}

	// No annotation found
	return false, false, ""
}
```

tiraboschi: done

ingvagabund:
The current issues is about the level of control when excluding pod from eviction. Should users be allowed to decide which pods will not get evicted vs. only admins should be allowed to exclude pods from eviction. I tend to prefer the latter as some plugins might not work properly with user intervention. E.g. a plugin evicting pods that do not comply with company/security policies.

Stateful/pet sets have a special place in Kubernetes. Would it help to implement a protecting mechanism that will exclude any pod owned by a specific owner? And only when such pods are still expected to be evicted a special annotation needs to be introduced to tell "Hey, I am still eligible for eviction"?

E.g.
```
apiVersion: v1
kind: Pod
metadata:
  name: virt-launcher-testvm-abcde
  namespace: default
  labels:
    kubevirt.io: virt-launcher
    kubevirt.io/domain: testvm
  ownerReferences:
  - apiVersion: kubevirt.io/v1
    kind: VirtualMachineInstance
    name: testvm
    uid: 12345678-1234-1234-1234-123456789abc
    controller: true
    blockOwnerDeletion: true
spec:
  containers:
  - name: compute
    image: registry:5000/kubevirt/virt-launcher
    # additional container config here
  # other pod spec details
```

The pod owner is set to `kubevirt.io/v1/VirtualMachineInstance`. With the descheduler introducing a new `excludeKubeVirtPods` (disabled by default) `DefaultEvictor` option. If enabled any pod owned by `VirtualMachineInstance` will be excluded unless `descheduler.alpha.kubernetes.io/evict` (or a different mechanism for ignoring the kube virt check) is present.

tiraboschi:
I see and I agree with your concern about the risk of having unprivileged users being able to arbitrarily opt-out their pods from any plugin.
The descheduler is also providing a label filtering mechanism:
the cluster admin can easily configure the descheduler to ignore by default pods with a specific label (e.g. `kubevirt.io: virt-launcher` in the KubeVirt use case); at that point the owner of the individual VM could rely on the presence of `descheduler.alpha.kubernetes.io/evict` annotation to opt-int a VM or on its absence to opt-out.

Adding another filtering mechanism based on the `ownerReference` sounds overkill to me.
We can admit that the label `kubevirt.io: virt-launcher` can eventually be used to opt-out a generic pod if the cluster admin configured the descheduler to do so, but `ownerReference` is not really providing more protection since a malicious user can easily fool it.

ingvagabund:
Closing in favor of https://github.com/kubernetes-sigs/descheduler/pull/1723.

PR #1482, opened 2024-08-09
Remove descheduler/v1alpha1 type

ingvagabund:
Fixes: https://github.com/kubernetes-sigs/descheduler/issues/1375

PR #1674, opened 2025-04-25
Remove reference to obsolete deschedulerPolicy fields in chart values

meroupatate:
Since the removal of v1alpha1 descheduler policy config type in https://github.com/kubernetes-sigs/descheduler/pull/1482, some values are now obsolete and shouldn't be referenced in the helm chart since they are not applied anymore

### DIFF
--- a/charts/descheduler/values.yaml
+++ b/charts/descheduler/values.yaml
@@ -95,9 +95,6 @@ deschedulerPolicy:
   # maxNoOfPodsToEvictPerNamespace: 10
   # metricsProviders:
   # - source: KubernetesMetrics
-  # ignorePvcPods: true
-  # evictLocalStoragePods: true
-  # evictDaemonSetPods: true
   # tracing:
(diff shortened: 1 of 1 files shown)


charts/descheduler/values.yaml
```diff
@@ -95,9 +95,6 @@ deschedulerPolicy:
   # maxNoOfPodsToEvictPerNamespace: 10
   # metricsProviders:
   # - source: KubernetesMetrics
-  # ignorePvcPods: true
```
googs1025, line 98: thanks for this. 
now, helm chart package uses the descheduler/v1alpha2 configuration. 
These param have been moved to DefaultEvictorArgs 

https://github.com/kubernetes-sigs/descheduler/blob/6e9622bf4180cc2fe869f2a0f84961d9700601d1/pkg/framework/plugins/defaultevictor/types.go#L25-L40

PR #1682, opened 2025-05-06
feat: adding nodelimit to highnodeutil and cordon for underutilized n…

ignaciocaff:
…odes to allow karpenter nuke it

### DIFF
--- a/pkg/framework/plugins/nodeutilization/highnodeutilization.go
+++ b/pkg/framework/plugins/nodeutilization/highnodeutilization.go
@@ -271,13 +271,6 @@ func (h *HighNodeUtilization) Balance(ctx context.Context, nodes []*v1.Node) *fr
 	// sorts the nodes by the usage in ascending order.
 	sortNodesByUsage(lowNodes, true)
 
-	// adding nodeLimit to the function, since it uses
-	// the same one as in the LowNodeUtilization scenario.
-	var nodeLimit *uint
-	if h.args.EvictionLimits != nil {
-		nodeLimit = h.args.EvictionLimits.Node
-	}
-
 	evictPodsFromSourceNodes(
 		ctx,
 		h.args.EvictableNamespaces,
@@ -289,7 +282,7 @@ func (h *HighNodeUtilization) Balance(ctx context.Context, nodes []*v1.Node) *fr
 		h.resourceNames,
 		continueEvictionCond,
 		h.usageClient,
-		nodeLimit,
+		nil,
 	)
 
 	return nil
--- a/pkg/framework/plugins/nodeutilization/highnodeutilization.go
+++ b/pkg/framework/plugins/nodeutilization/highnodeutilization.go
@@ -262,9 +262,9 @@ func (h *HighNodeUtilization) Balance(ctx context.Context, nodes []*v1.Node) *fr
 		if _, err := client.CoreV1().
 			Nodes().
 			Patch(ctx, node.Name, types.MergePatchType, patch, metav1.PatchOptions{}); err != nil {
-			klog.ErrorS(err, "Failed to cordon node", node.Name)
+			klog.ErrorS(err, "Failed to cordon", "node", node.Name)
 		} else {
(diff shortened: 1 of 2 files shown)


pkg/framework/plugins/nodeutilization/highnodeutilization.go
```diff
@@ -249,6 +251,23 @@ func (h *HighNodeUtilization) Balance(ctx context.Context, nodes []*v1.Node) *fr
...
+	// it doesn’t try to place them back on the same nodes.
+	// the risk we’re taking is that if evictPodsFromSourceNodes fails for some reason, we won’t notice.
+	// TODO: uncordon if it fails but to do that, we’d need to change the tool’s error handling.
+	client := h.handle.ClientSet()
+	for _, nodeInfo := range lowNodes {
+		node := nodeInfo.NodeUsage.node
+		patch := []byte(`{"spec":{"unschedulable":true}}`)
+		if _, err := client.CoreV1().
+			Nodes().
+			Patch(ctx, node.Name, types.MergePatchType, patch, metav1.PatchOptions{}); err != nil {
```
ingvagabund, line 264: The descheduler is not expected to change node objects. Also, this is open ended operation. Who/What's going to be responsible for making a node schedulable again. This might quickly make all the nodes unschedulable.

ingvagabund: Also, instead of making a node fully unschedulable an alternative might be to apply soft-taints on the node to make the chance of scheduling pod on it less probable.

googs1025: > Also, instead of making a node fully unschedulable an alternative might be to apply soft-taints on the node to make the chance of scheduling pod on it less probable.

+1 
I also prefer something like this, but I'm not sure if this is what the descheduler should and can do. 🤔

googs1025: but once we add this soft-taints, we have to consider when or how to remove this stain? 🤔

ignaciocaff: Hey folks, this PR was mainly created for internal review. We explored the idea of soft taints, which would also involve modifying the Node object. The use case we’re dealing with is about preventing pods from being scheduled on the Node so Karpenter can clean it up as quickly as possible.
Technically, both operations mutate the node, and the end goal (in our use case) is to terminate it.
On the other hand, if we apply a soft taint and there are no nodes with matching tolerations, the scheduler might still place pods there and that’s exactly what we don’t want.
So you can probably ignore this PR.

ingvagabund: @googs1025 as part of integration of the descheduler with the KubeVirt stack we implemented a downstream soft-tainter: https://github.com/openshift/cluster-kube-descheduler-operator/tree/master/pkg/softtainter. So far the descheduler has not been expected to provide any scheduling guidance for the kube-scheduler. Yet, there's a valid use case for it. Just the ownership is not clear since that would require for the descheduler to have the extra node update permission as @ignaciocaff mentioned. Which I would like to avoid. That's why we have a downstream version for the moment. Yet, extending the descheduler with another controller for a reasonable scheduling guidance might be an option. Something to discuss in the future.

googs1025: > as part of integration of the descheduler with the KubeVirt stack we implemented a downstream soft-tainter: https://github.com/openshift/cluster-kube-descheduler-operator/tree/master/pkg/softtainter

Thanks for the info. We are currently using `NPD`+`descheduler` to perform some inspection tasks, similar to the same problem. Because `NPD` and `descheduler` will not do with scheduling selection (like add taint), this is a good implementation reference for us internally! Thanks.

> Yet, extending the descheduler with another controller for a reasonable scheduling guidance might be an option. Something to discuss in the future.

Of course, if descheduler can provide this feature, I think it would be great. 😄

PR #1683, opened 2025-05-07
Improve percentage calculation in ResourceUsageToResourceThreshold func

googs1025:
- Implemented precise calculation of resource usage percentages, rounding to two decimal places.

### DIFF
--- a/pkg/framework/plugins/nodeutilization/nodeutilization.go
+++ b/pkg/framework/plugins/nodeutilization/nodeutilization.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"maps"
+	"math"
 	"slices"
 	"sort"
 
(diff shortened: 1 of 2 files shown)


pkg/framework/plugins/nodeutilization/nodeutilization.go
```diff
@@ -594,11 +595,18 @@ func ResourceUsageToResourceThreshold(
 			percent = float64(used) / float64(capacity) * 100
 		}
 
-		result[rname] = api.Percentage(percent)
+		roundedPercent := roundFloat(percent, 2)
```
ingvagabund, line 598: There's `thresholdsToKeysAndValues` function used for printing percentages to 2 decimals. Is there another place where printing/computing with rounded numbers is needed?

googs1025: thanks 😄 
I am trying to understand the implementation of nodeutilization. I'm curious Why is float64 not always rounded to two decimal places when used, but only used when printing logs?

ingvagabund: Rounding leads to a loss of precision. The less rounding the better. Though, the losses are probably very small. Yet, why take chances.

googs1025: got this. I have no further questions about this explanation. 😄 

/close

PR #1684, opened 2025-05-07
chore: move namespaces filtering logic to New()

googs1025:
move namespaces filtering logic to `New()` in `RemovePodsViolatingTopologySpreadConstraint` plugin

This logic is determined in `BuildFilterFunc`.

https://github.com/kubernetes-sigs/descheduler/blob/e466307d7cfc3d4ef3f69b27d4d734853b5e6208/pkg/descheduler/pod/pods.go#L94-L120


ingvagabund:
@googs1025 would you please extend the PR description to mention why this is better? To let the broader audience now the intention and why this chance is still correct.

a7i (approved):
/lgtm

PR #1688, opened 2025-05-10
RemovePodsViolatingNodeTaints: list only pods that are not failed/suceeded

ingvagabund:
Listing pods was incorrectly changed to listing all pods during code refactoring in https://github.com/kubernetes-sigs/descheduler/commit/0ff8ecb41ecf90ff2cfaa84a3e763ad0c800cb6d#diff-494cd6260fda05699f68c7da810f7814dea797a006afee5380728a6433206fefL82.

### DIFF
--- a/pkg/framework/plugins/removepodsviolatingnodetaints/node_taint.go
+++ b/pkg/framework/plugins/removepodsviolatingnodetaints/node_taint.go
@@ -106,7 +106,7 @@ func (d *RemovePodsViolatingNodeTaints) Name() string {
 func (d *RemovePodsViolatingNodeTaints) Deschedule(ctx context.Context, nodes []*v1.Node) *frameworktypes.Status {
 	for _, node := range nodes {
 		klog.V(1).InfoS("Processing node", "node", klog.KObj(node))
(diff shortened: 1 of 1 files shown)


a7i (approved):

a7i:
@ingvagabund thoughts on adding a unit test to avoid similar regression in the future?

ingvagabund:
Field selectors are not properly implemented in the fake clientset, resp. informer built on top of a fake clientset. There would be no difference on the level of listing pods. You'd need to simulate the field selector filtering by injecting the right list of pods :(. Instead of relying on the functionality of the field selector to exclude the pods. The field selector filtering is kube-apiserver functionality.

googs1025:
/lgtm 

I will add a case where the pod is in failed success phase and the eviction count is 0 in the unit test of each called plugin. 
After this PRs are merged, the only plugins that use the ListAllPodsOnANode method are: PodLifeTime RemoveFailedPods RemovePodsHavingTooManyRestarts plugins, because these plugins will filter internally or use the pod.Status.Phase status. 😄

googs1025:
> I will add a case where the pod is in failed success phase and the eviction count is 0 in the unit test of each called plugin.

Since I currently have limited bandwidth, I may have to wait for the weekend to work on this.

PR #1691, opened 2025-05-12
fix(example): list only active pod

googs1025:
inspired by this pr: https://github.com/kubernetes-sigs/descheduler/pull/1688/files

- use ListPodsOnANode func in example plugin
- update comment

After the two PRs are merged, the only plugins that use the ListAllPodsOnANode method are: PodLifeTime RemoveFailedPods RemovePodsHavingTooManyRestarts plugins, because these plugins will filter internally or use the pod.Status.Phase status.

### DIFF
--- a/pkg/framework/plugins/example/example.go
+++ b/pkg/framework/plugins/example/example.go
@@ -137,10 +137,23 @@ func (d *Example) Deschedule(ctx context.Context, nodes []*v1.Node) *fwtypes.Sta
 
 	// go node by node getting all pods that we can evict.
 	for _, node := range nodes {
-		// ListAllPodsOnANode is a helper function that retrieves all
-		// pods filtering out the ones we can't evict. We merge the
-		// default filters with the one we created above.
-		pods, err := podutil.ListAllPodsOnANode(
+		// ListAllPodsOnANode is a helper function that retrieves all pods filtering out the ones we can't evict.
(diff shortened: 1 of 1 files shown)


pkg/framework/plugins/example/example.go
```diff
@@ -137,10 +137,23 @@ func (d *Example) Deschedule(ctx context.Context, nodes []*v1.Node) *fwtypes.Sta
...
+		//
+		// In this context, we prefer using ListPodsOnANode because:
+		// 1. It ensures that only active Pods (not in Succeeded or Failed states) are considered for eviction.
+		// 2. This helps avoid unnecessary processing of Pods that no longer consume resources.
+		// 3. By applying an additional filter (d.podFilter and filter), we can further refine which Pods are eligible for eviction,
+		//    ensuring that only Pods meeting specific criteria are selected.
+		//
+		// However, if you need to consider all Pods including those in Succeeded or Failed states for other purposes,
+		// you should use ListAllPodsOnANode instead.
+		pods, err := podutil.ListPodsOnANode(
```
jklaw90, line 156: I wonder if it would be more readable if we made it one function for both and just had an option parameter?  Ie like listoptions for the clientset?

googs1025: sgtm, But I'm not sure whether this package will be referenced downstream 🤔

PR #1705, opened 2025-06-12
feat(prometheus): allow different url schemes

ricardomaraschini:
As per prometheus Golang client implementation: the only url validation done is by means of an `url.Parse()` [call](https://github.com/prometheus/client_golang/blob/main/api/client.go#L90C1-L93C3). We should do the same and not enforce the usage of https scheme.

Our readme even shows an example of Descheduler config using http prometheus url scheme.


googs1025:
/lgtm 

I understand that this seems reasonable. 🤔

PR #1706, opened 2025-06-12
feat: Adding a feature `maxNodesToProcess` to HighNodeUtilization Plugin to limit the nodes processed on each execution

jonahjon:
Adding a new settings on the **HighNodeUtilization** plugin called `maxNodesToProcess` . This plugins purpose is to limit the amount of nodes processed during each descheduler execution. 

This is useful as our company is looking to implement descheduler for improved binpacking, but without this guardrail it can potentially impact several nodes at once which could cause service disruptions.

Similar goal to this other open PR. https://github.com/kubernetes-sigs/descheduler/pull/1616


pkg/framework/plugins/nodeutilization/types.go
```diff
@@ -60,6 +60,11 @@ type HighNodeUtilizationArgs struct {
 	Thresholds    api.ResourceThresholds `json:"thresholds"`
 	NumberOfNodes int                    `json:"numberOfNodes,omitempty"`
 
+	// MaxNodesToProcess limits the number of nodes to process in each
+	// the descheduling execution. This is useful to limit nodes descheduled each run 
+	// when turning this plugin on within a cluster with many underutilized nodes.
+	MaxNodesToProcess int                `json:"maxNodesToProcess,omitempty"`
```
googs1025: can we add some unit test case in `TestHighNodeUtilization`? 😄

googs1025:
I think this change is reasonable to avoid too much change when large-scale nodes run the `HighNodeUtilization` plugin.

README.md
```diff
@@ -424,6 +424,7 @@ actual usage metrics. Implementing metrics-based descheduling is currently TODO
 |---|---|
 |`thresholds`|map(string:int)|
 |`numberOfNodes`|int|
```
ingvagabund, line 488: Worth deprecating `numberOfNodes` and renaming it to `minNodesToProcess`. Alternatively, introducing a new `nodesToProcessLimits` (or a better name) with `min` and `max` fields. Are min and max the only useful limits to introduce? Resp. should only a number be introduced? Or, percentages as well?

pkg/framework/plugins/nodeutilization/highnodeutilization.go
```diff
@@ -208,6 +208,12 @@ func (h *HighNodeUtilization) Balance(ctx context.Context, nodes []*v1.Node) *fr
 
 	lowNodes, schedulableNodes := nodeInfos[0], nodeInfos[1]
 
+    // limit the number of nodes processed each execution if `MaxNodesToProcess` is set
+	if h.args.MaxNodesToProcess > 0 && len(lowNodes) > h.args.MaxNodesToProcess {
+		lowNodes = lowNodes[:h.args.MaxNodesToProcess]
```
ingvagabund: The index needs to rotate so all nodes are eventually processed.

pkg/framework/plugins/nodeutilization/types.go
jonahjon: Sure, i will add those.

pkg/framework/plugins/nodeutilization/highnodeutilization.go
jonahjon: 👀

jonahjon: Looking at possible ways to do this I see a couple ways to solve it, but would love feedback.

1. Add a new field to the struct to remove state from within the function/file
2. Add a new constant within the file to track index
3. Open to any other opinions/suggestions!


```
type HighNodeUtilization struct {
	handle              frameworktypes.Handle
	args                *HighNodeUtilizationArgs
	podFilter           func(pod *v1.Pod) bool
	criteria            []any
	resourceNames       []v1.ResourceName
	highThresholds      api.ResourceThresholds
	usageClient         usageClient
	lastProcessedIndex  int
}

...

if h.args.MaxNodesToProcess > 0 && len(lowNodes) > h.args.MaxNodesToProcess {
	start := h.lastProcessedIndex % len(lowNodes)
	rotated := append(lowNodes[start:], lowNodes[:start]...)

	lowNodes = rotated[:h.args.MaxNodesToProcess]
	h.lastProcessedIndex = (h.lastProcessedIndex + h.args.MaxNodesToProcess) % len(lowNodes)
}
... 
```


```
// rotateStartIdx is a variable to track the rotation index of MaxNodesToProcess
var rotateStartIdx int

...

if h.args.MaxNodesToProcess > 0 && len(lowNodes) > h.args.MaxNodesToProcess {
    start := rotateStartIdx % len(lowNodes)
    end := start + h.args.MaxNodesToProcess

    var selected []NodeInfo
    if end <= len(lowNodes) {
        selected = lowNodes[start:end]
    } else {
        selected = append(lowNodes[start:], lowNodes[:end%len(lowNodes)]...)
    }
    lowNodes = selected

    rotateStartIdx = (rotateStartIdx + h.args.MaxNodesToProcess) % len(lowNodes)
}
```

README.md
jonahjon: For us we're specifically looking to slow this down, so unsure if there is other use cases out there. The default settings are aggressive enough I'm not sure if `min` would be useful for company in any way.

googs1025: Maybe just supporting `max` and `percentage` would be fine? We might care more about avoiding evicting all high utilization nodes. 🤔 
```go
type NodeProcessingLimits struct {
    Max int32  `json:"max,omitempty"`
    MaxPercentage int32 `json:"maxPercentage,omitempty"`
}
```
(The name is just an example, not recommendation.)

pkg/framework/plugins/nodeutilization/highnodeutilization.go
ingvagabund: An observation: it's not guaranteed the list of `nodes` will always be correctly ordered. I.e. even the rotated index may not select the nodes uniformly. If this leads to a sub-efficient behavior we can see if sort the nodes based on their names helps.

ingvagabund: Option 1. looks good. So the second version of the code changes in the comment. Worth moving the code under a function so it can be unit test separately.

Also, `MaxNodesToProcess` needs to be at least as `NumberOfNodes` is if set.

README.md
```diff
@@ -468,7 +469,9 @@ Policy should pass the following validation checks:
 There is another parameter associated with the `HighNodeUtilization` strategy, called `numberOfNodes`.
 This parameter can be configured to activate the strategy only when the number of under utilized nodes
 is above the configured value. This could be helpful in large clusters where a few nodes could go
-under utilized frequently or for a short period of time. By default, `numberOfNodes` is set to zero.
+under utilized frequently or for a short period of time. By default, `numberOfNodes` is set to zero. The parameter `maxNodesToProcess` is used to limit how many nodes should be processed by the descheduler plugin on each execution.
```
ingvagabund, line 534: The current implementation for `MaxNodesToProcess` limits the number of underutilized nodes, not all the nodes. s/is used to limit how many nodes should be processed/is used to limit how many underutilized nodes should be processed/ to make this explicitly clear.

Can you also put the new sentence on a separate line? To limit the number of characters per line.

ingvagabund:
@jonahjon again apologies for the delay. I am currently short on resources to allocate more time for upstream reviews.

README.md
ricardomaraschini: I agree with @ingvagabund here, the `numberOfNodes` configuration relates a lot with the newly introduced `maxNodesToProcess`. To deprecate the first one and to make both part of the same structure (with min and max properties) would be better here. 

I don't think we should be going as far as have "percentages" here except if there is a clear usecase.

pkg/framework/plugins/nodeutilization/highnodeutilization_test.go
```diff
@@ -639,6 +662,7 @@ func TestHighNodeUtilizationWithTaints(t *testing.T) {
 			}
 
 			plugin, err := NewHighNodeUtilization(ctx, &HighNodeUtilizationArgs{
+				MaxNodesToProcess: 1,
```
ricardomaraschini: I would not expect this to be necessary. Have you introduced this by mistake ?

pkg/framework/plugins/nodeutilization/highnodeutilization_test.go
```diff
@@ -481,6 +481,29 @@ func TestHighNodeUtilization(t *testing.T) {
...
+				test.BuildTestPod("p1", 400, 0, "n1", test.SetRSOwnerRef),
+				test.BuildTestPod("p2", 400, 0, "n2", test.SetRSOwnerRef),
+				test.BuildTestPod("p3", 400, 0, "n3", test.SetRSOwnerRef),
+				test.BuildTestPod("p4", 400, 0, "n4", test.SetRSOwnerRef),
+			},
+			expectedPodsEvicted: 1,                                // Only one node's pod should be evicted per run
+			evictedPods:         []string{"p1", "p2", "p3", "p4"}, // Any one of these is valid
+			evictionModes:       nil,
+			// We'll set MaxNodesToProcess in the plugin args below
+		},
```
ricardomaraschini, line 495: This new test is broken. Also, please make "maxNodesToProcess" part of the test definition so each test can set its own value.

ricardomaraschini (changes requested):

ingvagabund:
@jonahjon it seems there are more than one additional functionality changes in this PR:
- node uncording: the descheduler does not have sufficient information about when it's safe to uncordon a node. This is a higher level intel that requires a higher level orchestration.
- applyPodsWithPDBBlockingSingleReplicaOwnerProtection: this new eviction protection exclusion will not work as a pod with a PDB with still be protected by the PDB. The eviction invoked by the descheduler will not succeed.

This PR is expected to focus on adding maxNodesToProcess to HighNodeUtilization. Which is not currently respected.

Please let us know once the PR is ready for review and the PR description aligns with the PR content.

PR #1708, opened 2025-06-25
logger: Align with the previous logger verbosity

ingvagabund:
Changing logger level/message type needs to be discussed separately.


googs1025:
/lgtm 

got this.

Issue #1638, opened 2025-02-27
Update helm chart to allow setting activeDeadlineSeconds on CronJob


willemm:
**Is your feature request related to a problem? Please describe.**
We ran into a problem recently where a Descheduler job insttance had been running for 50 days because of some glitch (It was complaining about access denied, and not exiting).  We had to manually go in and kill the job.

**Describe the solution you'd like**
To prevent any glitch that would cause a job to hang, from completely disabling descheduler operation, I would like to be able to set `cronjob.spec.jobTemplate.spec.activeDeadlineSeconds` in the helm chart, so that it will be terminated by the job controller if it gets into any kind of bad state.

**Describe alternatives you've considered**
I can't think of any alternatives.  Intuitively, "fixing the bug that caused it to loop" seems like a solution at first glance, but there can be any number of unexpected glitches that might cause a job to not terminate.

**What version of descheduler are you using?**

descheduler version: helm chart 0.31.0

PR #1709, opened 2025-06-29
add activeDeadlineSeconds field for cronjob

googs1025:
fix: https://github.com/kubernetes-sigs/descheduler/issues/1638

### DIFF
--- a/charts/descheduler/values.yaml
+++ b/charts/descheduler/values.yaml
@@ -55,7 +55,8 @@ suspend: false
 # startingDeadlineSeconds: 200
 # successfulJobsHistoryLimit: 3
 # failedJobsHistoryLimit: 1
-# ttlSecondsAfterFinished 600
+# ttlSecondsAfterFinished: 600
+# activeDeadlineSeconds: 60 # Make sure this value is SHORTER than the cron interval.
 # timeZone: Etc/UTC
 
 # Required when running as a Deployment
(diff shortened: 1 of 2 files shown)


charts/descheduler/values.yaml
```diff
@@ -56,6 +56,7 @@ suspend: false
 # successfulJobsHistoryLimit: 3
 # failedJobsHistoryLimit: 1
 # ttlSecondsAfterFinished 600
+# activeDeadlineSeconds: 3600
```
ingvagabund: Can you please add a comment saying the value needs to be shorter than the cron interval? Just in case.

googs1025: done

charts/descheduler/values.yaml
```diff
@@ -55,7 +55,8 @@ suspend: false
 # startingDeadlineSeconds: 200
 # successfulJobsHistoryLimit: 3
 # failedJobsHistoryLimit: 1
-# ttlSecondsAfterFinished 600
+# ttlSecondsAfterFinished: 600
+# activeDeadlineSeconds: 60 # Make sure this value is SHORTER than the cron interval to avoid jobs being killed prematurely.
```
ingvagabund: s/to avoid jobs being killed prematurely/to avoid spawning two or more jobs in parallel/
Leaving that part empty will do too.

googs1025: > Leaving that part empty will do too.

sorry I don't get this. What does this mean? Do we need to add the default value directly instead of using the "#" comment?
🤔

ingvagabund: Leaving "Make sure this value is SHORTER than the cron interval" will do. That part of "avoid jobs being killed prematurely." might be misleading. `activeDeadlineSeconds` itself kills a job/pod prematurely. Maybe I don't understand how the cron jobs work properly. Is the current job terminated before a new job is created? Or, can two or more jobs run in parallel?

googs1025: got this.  done

PR #1719, opened 2025-07-10
fix: removepodsviolatingtopologyspreadconstraint to favor evictable pods when balancing domains

a7i:
This PR fixes two issues:

1. Sort pods that are above ideal avg based on the criteria that they fit on other nodes that are below avg. That's because we skip pods here if they are not evictable: https://github.com/kubernetes-sigs/descheduler/blob/e5ab156a992b69019d49535ed474bf60be8f069e/pkg/framework/plugins/removepodsviolatingtopologyspreadconstraint/topologyspreadconstraint.go#L375-L377
This cannot be moved to sortDomains, because we do not know what nodesBelowAvg are until we sort the domains

3. Only add PodNodeAffinity when the inclusion policy is honor. That's because if any of the pods have a different nodeSelector or nodeAffinity, then we end up with multiple namespaceTopologySpreadConstraints, given that we do a deep equal on the object here: https://github.com/kubernetes-sigs/descheduler/blob/e5ab156a992b69019d49535ed474bf60be8f069e/pkg/framework/plugins/removepodsviolatingtopologyspreadconstraint/topologyspreadconstraint.go#L262
An example of this would be pinning 1/3 of replicas to be on Spot, while the rest are on-demand nodes

### DIFF
--- a/pkg/framework/plugins/removepodsviolatingtopologyspreadconstraint/topologyspreadconstraint.go
+++ b/pkg/framework/plugins/removepodsviolatingtopologyspreadconstraint/topologyspreadconstraint.go
@@ -539,16 +539,16 @@ func newTopologySpreadConstraint(constraint v1.TopologySpreadConstraint, pod *v1
 		MaxSkew:            constraint.MaxSkew,
 		TopologyKey:        constraint.TopologyKey,
 		Selector:           selector,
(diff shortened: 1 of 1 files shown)


ingvagabund:
The second commit changes aligns with https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/ upstream documentation.

@a7i this looks great. Thank you for addressing this.

/approve
/lgtm

PR #1722, opened 2025-07-24
Test code refactorings

ingvagabund:
- dropping default values assignments in tests
- replacing `descheduler.alpha.kubernetes.io/evict` literal with `evictPodAnnotationKey` const
- `TestSortPodsBasedOnPriorityLowToHigh`: extending the test to check the whole sorted list instead of a single item


a7i (approved):
/approve

Issue #1659, opened 2025-03-28
Can't use descheduler.alpha.kubernetes.io/evict=false to disable evicton


kreeuwijk:
**Is your feature request related to a problem? Please describe.**
We use VirtualMachine preferences in Kubevirt to default the use of the descheduler for any VMs using the those presets, by setting the `descheduler.alpha.kubernetes.io/evict: 'true'` annotation. But this makes it difficult to turn off the descheduler for specific VMs, as the descheduler actually accepts any value for the annotation. Overriding the annotation from the preset through the VM YAML is only possible if you specify an alternate value.

**Describe the solution you'd like**
I would like to be able to set the annotation
```
descheduler.alpha.kubernetes.io/evict: 'false'
```
and have the descheduler treat that pod as if no annotation was set on it.

**Describe alternatives you've considered**
The only other options that worked were:
* Remove the annotation from the VM preferences altogether and go back to specifying it in the VM YAML. This isn't very user friendly
* Create "No DRS" copies of the VM preferences that don't have the annotation but are the same otherwise. This also isn't very user friendly

**What version of descheduler are you using?**

descheduler version: 0.32.2

**Additional context**
N/A

kreeuwijk:
>> We use VirtualMachine preferences in Kubevirt to default the use of the descheduler

>Would you please elaborate more on this? There are other ways to tell the descheduler to focus on particular pods. E.g. specifying a label selector or a list of included/excluded namespace. Have you explored this option?

Here's an example of a VM preference:
```
apiVersion: instancetype.kubevirt.io/v1beta1
kind: VirtualMachineClusterPreference
metadata:
  name: generic-linux
spec:
  annotations:
    descheduler.alpha.kubernetes.io/evict: "true"
  clock:
    preferredClockOffset:
      utc: {}
      #timezone: America/New_York
  cpu:
    preferredCPUTopology: spread   
    spreadOptions:
      across: SocketsCores
      ratio: 4
  devices:
    preferredDiskBus: virtio
    preferredDiskCache: writethrough
    preferredInterfaceModel: virtio
    preferredRng: {}
  machine:
    preferredMachineType: q35
  preferredTerminationGracePeriodSeconds: 180
```

When you then create a VM, you can link to this VM preference by setting:
```
spec:
  preference:
    name: generic-linux
```
[...]

PR #1723, opened 2025-07-24
Default evictor no eviction policy

ingvagabund:
Define `descheduler.alpha.kubernetes.io/prefer-no-eviction` annotation so pods can  voice their preference not to be evicted. Each plugin decides whether the annotation gets respected or not. When the `DefaultEvictor` plugin sets `noEvictionPolicy` to `Mandatory` all such pods are excluded from eviction. Needs to be used with caution as some plugins may enfore various policies that are expected to be always met.

Alternative for https://github.com/kubernetes-sigs/descheduler/pull/1673.

Fixes: https://github.com/kubernetes-sigs/descheduler/issues/1659

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,6 +162,7 @@ The Default Evictor Plugin is used by default for filtering pods before processi
 | `minReplicas`             |`uint`|`0`| ignore eviction of pods where owner (e.g. `ReplicaSet`) replicas is below this threshold                                    |
 | `minPodAge`               |`metav1.Duration`|`0`| ignore eviction of pods with a creation time within this threshold                                                          |
 | `ignorePodsWithoutPDB`    |`bool`|`false`| set whether pods without PodDisruptionBudget should be evicted or ignored                                                   |
+| `noEvictionPolicy`        |`enum`|``| sets whether a `descheduler.alpha.kubernetes.io/prefer-no-eviction` pod annotation is considered preferred or mandatory. Accepted values: "", "Preferred", "Mandatory". Defaults to "Preferred". |
 
 ### Example policy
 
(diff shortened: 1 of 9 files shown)


ingvagabund:
/hold
for a review round

ingvagabund:
Only admins can enable this new functionality. Also the same admins are aware which plugins need to enforce the eviction. In this case two profiles can be used. One disabling this functionality while configuring plugins enforcing eviction (e.g. enforcing security policies). The other enabling this functionality and allowing some pods to negate the eviction.

README.md
```diff
@@ -162,6 +162,7 @@ The Default Evictor Plugin is used by default for filtering pods before processi
 | `minReplicas`             |`uint`|`0`| ignore eviction of pods where owner (e.g. `ReplicaSet`) replicas is below this threshold                                    |
 | `minPodAge`               |`metav1.Duration`|`0`| ignore eviction of pods with a creation time within this threshold                                                          |
 | `ignorePodsWithoutPDB`    |`bool`|`false`| set whether pods without PodDisruptionBudget should be evicted or ignored                                                   |
+| `noEvictionPolicy`        |`enum`|``| sets whether a `descheduler.alpha.kubernetes.io/prefer-no-eviction` pod annotation is considered preferred or mandatory. Accepted values: "", "Preferred", "Mandatory". Defaults to "Preferred". |
```
ricardomaraschini, line 165: The annotation name here seems odd, it denotes a preference ("prefer-no-eviction") but when `noEvictionPolicy` is mandatory it is not a preference anymore. I have been pounding for a while but I can't come up with a different naming.

ingvagabund: Yeah, I hear you. I can't think of any other name either.

PR #1729, opened 2025-08-06
Improves the validation logic in the plugins

googs1025:
Improves the validation logic in the `ValidatexxxArgs` by changing it from "returning on the first error" to "collecting all validation errors and returning them together," enhancing the user experience when diagnosing config issues.

### DIFF
--- a/pkg/framework/plugins/podlifetime/validation.go
+++ b/pkg/framework/plugins/podlifetime/validation.go
@@ -18,29 +18,31 @@ package podlifetime
 
 import (
 	"fmt"
-
-	"k8s.io/apimachinery/pkg/runtime"
+	"sort"
 
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
 )
 
 // ValidatePodLifeTimeArgs validates PodLifeTime arguments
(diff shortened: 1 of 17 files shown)


pkg/framework/plugins/podlifetime/validation.go
```diff
@@ -72,8 +74,10 @@ func ValidatePodLifeTimeArgs(obj runtime.Object) error {
 	)
 
 	if !podLifeTimeAllowedStates.HasAll(args.States...) {
-		return fmt.Errorf("states must be one of %v", podLifeTimeAllowedStates.UnsortedList())
+		allowed := podLifeTimeAllowedStates.UnsortedList()
```
googs1025, line 77: ensure consistent output

ingvagabund:
/approve
/lgtm

This looks really nice. Thank you.

ingvagabund:
@googs1025 another enhancement might be to refactor the validation code to use `"k8s.io/apimachinery/pkg/util/validation/field"` for providing the [configuration path](https://github.com/kubernetes-sigs/kueue/blob/main/pkg/config/validation.go#L107). E.g. replace
```go
if args.PriorityThreshold != nil && args.PriorityThreshold.Value != nil && len(args.PriorityThreshold.Name) > 0 {
	return fmt.Errorf("priority threshold misconfigured, only one of priorityThreshold fields can be set")
}
```
with
```go
if args.PriorityThreshold != nil && args.PriorityThreshold.Value != nil && len(args.PriorityThreshold.Name) > 0 {
	return field.Invalid(field.NewPath("args").Child("priorityThreshold"), args.PriorityThreshold, fmt.Errorf("priority threshold misconfigured, only one of priorityThreshold fields can be set"))
}
```
If you wanna try it. We have not explored this improvement in the past.

PR #1733, opened 2025-08-12
support include namespaces filtering in node utilization plugin

googs1025:
- support include namespaces filtering in node utilization plugin
- add `ValidateHighNodeUtilizationPluginConfig` unit test

Take over this inactive PR: https://github.com/kubernetes-sigs/descheduler/pull/1643


pkg/framework/plugins/nodeutilization/validation_test.go
```diff
@@ -257,3 +293,153 @@ func TestValidateLowNodeUtilizationPluginConfig(t *testing.T) {
 		})
 	}
 }
+
+func TestValidateHighNodeUtilizationPluginConfig(t *testing.T) {
```
googs1025, line 261: add miss unit test: `ValidateHighNodeUtilizationPluginConfig`

ingvagabund:
I'd like to avoid allowing inclusion of namespaces for either of the node utilization plugins. Including too few namespaces could degrade the effectiveness of both plugins and make the overall balancing less effective. By requiring users to exclude namespaces instead, they have to think more carefully about the scope and are less likely to exclude too many pods. Using a list of excluded namespaces also makes it harder to unintentionally reduce the set of included namespaces.

googs1025:
> I'd like to avoid allowing inclusion of namespaces for either of the node utilization plugins. Including too few namespaces could degrade the effectiveness of both plugins and make the overall balancing less effective. By requiring users to exclude namespaces instead, they have to think more carefully about the scope and are less likely to exclude too many pods. Using a list of excluded namespaces also makes it harder to unintentionally reduce the set of included namespaces.

sgtm. change this pr description @ingvagabund  😄

Issue #1732, opened 2025-08-11
chart: Not possible to add cronjob/job labels or annotations


disconn3ct:
**Is your feature request related to a problem? Please describe.**

The helm chart has `podAnnotations` and `podLabels` but nothing for annotating and labeling the cronjob or the interim jobs.

**Describe the solution you'd like**

Add `cronJobAnnotations`, `cronJobLabels`, `jobAnnotations` and `jobLabels` to the chart cronjob template

**Describe alternatives you've considered**

kustomize patch
**What version of descheduler are you using?**

descheduler version:

**Additional context**

n2h9:
Hey hey hello @disconn3ct  @ingvagabund 👋 😇 

Added describe functionality to helm chart in https://github.com/kubernetes-sigs/descheduler/pull/1736
Could you please take a look 🤗 .

ingvagabund:
Resolved through https://github.com/kubernetes-sigs/descheduler/pull/1736.

PR #1736, opened 2025-08-23
[1732] chore: add cronjob/job labels and annotations to helm chart

n2h9:
This pr implements #1732 suggestion. 

Example result of helm template with default values of new properties ` helm template descheduler . --values values.yaml `:
[output-default-values.yaml.txt](https://github.com/user-attachments/files/21949427/output-default-values.yaml.txt)

Example result of helm template with  test values of new properties ` helm template descheduler . --values test-values.yaml `:
[output-test-values.yaml.txt](https://github.com/user-attachments/files/21949441/output-test-values.yaml.txt)


a7i (approved):
/approve

PR #1738, opened 2025-08-25
optimize: preserving needed fields for memory efficiency

### LINKED ISSUE #1737: [Enhancement] Optimize memory usage
Descheduler use too many memory when runs in `dry-run` mode. In dry-run mode, we perform a full copy of the in-memory data, which results in its memory usage being twice that of the scheduler within the cluster. This is very resource-intensive in large-scale scenarios.
```
	if d.rs.DryRun {
		klog.V(3).Infof("Building a cached client from the cluster for the dry run")
		// Create a new cache so we start from scratch without any leftovers
		fakeClient := fakeclientset.NewSimpleClientset()
		// simulate a pod eviction by deleting a pod
		fakeClient.PrependReactor("create", "pods", d.podEvictionReactionFnc(fakeClient))
		fakeSharedInformerFactory := informers.NewSharedInformerFactory(fakeClient, 0)

		err := d.ir.CopyTo(fakeClient, fakeSharedInformerFactory)
		if err != nil {
			return err
		}

```

Only retain the necessary fields in the cache, removing unnecessary information to optimize memory usage.

Colvin-Y:
fix #1737 
In production scenarios, the fields of pods are usually quite complex. Testing has shown that in our case, the descheduler's memory usage has been reduced by 2/5.

### DIFF
--- a/pkg/descheduler/descheduler.go
+++ b/pkg/descheduler/descheduler.go
@@ -34,7 +34,6 @@ import (
 	policyv1 "k8s.io/api/policy/v1"
 	schedulingv1 "k8s.io/api/scheduling/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -151,6 +150,60 @@ func metricsProviderListToMap(providersList []api.MetricsProvider) map[api.Metri
 		providersMap[provider.Source] = &provider
 	}
 	return providersMap
+}
+
+// preserveNeeded returns the obj preserving fields needed for memory efficiency.
+// Only keep scheduler related fields
+func preserveNeeded(obj interface{}) (interface{}, error) {
+	if pod, ok := obj.(*v1.Pod); ok {
+		preserveContainer := func(c *v1.Container) {
+			c.Command = nil
+			c.Args = nil
+			c.WorkingDir = ""
+			c.Ports = nil
+			c.EnvFrom = nil
+			c.Env = nil
+			c.ResizePolicy = nil
+			c.VolumeMounts = nil
+			c.VolumeDevices = nil
+			c.LivenessProbe = nil
+			c.ReadinessProbe = nil
+			c.StartupProbe = nil
+			c.Lifecycle = nil
+			c.TerminationMessagePath = ""
+			c.TerminationMessagePolicy = ""
+			c.ImagePullPolicy = ""
+			c.SecurityContext = nil
+		}
+
+		// metadata related
+		pod.ManagedFields = nil
+		pod.Finalizers = nil
+
+		// spec related
+		for i := 0; i < len(pod.Spec.InitContainers); i++ {
+			preserveContainer(&pod.Spec.InitContainers[i])
+		}
+		for i := 0; i < len(pod.Spec.Containers); i++ {
+			preserveContainer(&pod.Spec.Containers[i])
+		}
+		pod.Spec.EphemeralContainers = nil
+		pod.Spec.SecurityContext = nil
+		pod.Spec.ImagePullSecrets = nil
+		pod.Spec.Hostname = ""
+		pod.Spec.Subdomain = ""
+		pod.Spec.HostAliases = nil
+		pod.Spec.PriorityClassName = ""
+		pod.Spec.DNSConfig = nil
+		pod.Spec.ReadinessGates = nil
+		pod.Spec.RuntimeClassName = nil
+		pod.Spec.PreemptionPolicy = nil
+	}
+	if node, ok := obj.(*v1.Node); ok {
+		node.ManagedFields = nil
+		node.Status.Images = nil
+	}
+	return obj, nil
 }
 
 func newDescheduler(ctx context.Context, rs *options.DeschedulerServer, deschedulerPolicy *api.DeschedulerPolicy, evictionPolicyGroupVersion string, eventRecorder events.EventRecorder, sharedInformerFactory, namespacedSharedInformerFactory informers.SharedInformerFactory) (*descheduler, error) {
@@ -585,7 +638,7 @@ func RunDeschedulerStrategies(ctx context.Context, rs *options.DeschedulerServer
 	ctx, span = tracing.Tracer().Start(ctx, "RunDeschedulerStrategies")
 	defer span.End()
 
-	sharedInformerFactory := informers.NewSharedInformerFactoryWithOptions(rs.Client, 0, informers.WithTransform(trimManagedFields))
+	sharedInformerFactory := informers.NewSharedInformerFactoryWithOptions(rs.Client, 0, informers.WithTransform(preserveNeeded))
 
 	var nodeSelector string
 	if deschedulerPolicy.NodeSelector != nil {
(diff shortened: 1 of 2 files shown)


pkg/descheduler/descheduler.go
```diff
@@ -153,8 +153,65 @@ func metricsProviderListToMap(providersList []api.MetricsProvider) map[api.Metri
...
+	if node, ok := obj.(*v1.Node); ok {
+		node.ManagedFields = nil
+		node.Status.Images = nil
+	}
+	return obj, nil
+}
+
 func newDescheduler(ctx context.Context, rs *options.DeschedulerServer, deschedulerPolicy *api.DeschedulerPolicy, evictionPolicyGroupVersion string, eventRecorder events.EventRecorder, sharedInformerFactory, namespacedSharedInformerFactory informers.SharedInformerFactory) (*descheduler, error) {
 	podInformer := sharedInformerFactory.Core().V1().Pods().Informer()
+	podInformer.SetTransform(preserveNeeded)
```
googs1025: Using `SetTransform` seems to be very helpful for memory optimization, especially in large clusters. We can delete the unused pod node field. 🤔

googs1025: /lgtm 

cc @ingvagabund

a7i: should this move [here](https://github.com/kubernetes-sigs/descheduler/blob/94d1825c6866623469bf3eaec1b34ada7504d6ae/pkg/descheduler/descheduler.go#L588) instead?

pkg/descheduler/descheduler.go
```diff
@@ -153,8 +153,65 @@ func metricsProviderListToMap(providersList []api.MetricsProvider) map[api.Metri
...
+		pod.Spec.Subdomain = ""
+		pod.Spec.HostAliases = nil
+		pod.Spec.PriorityClassName = ""
+		pod.Spec.DNSConfig = nil
+		pod.Spec.ReadinessGates = nil
+		pod.Spec.RuntimeClassName = nil
+		pod.Spec.PreemptionPolicy = nil
+	}
+	if node, ok := obj.(*v1.Node); ok {
+		node.ManagedFields = nil
```
a7i: this is already being done [here](https://github.com/kubernetes-sigs/descheduler/blob/94d1825c6866623469bf3eaec1b34ada7504d6ae/pkg/descheduler/descheduler.go#L715-L720)

pkg/descheduler/descheduler.go
```diff
@@ -153,8 +153,65 @@ func metricsProviderListToMap(providersList []api.MetricsProvider) map[api.Metri
...
+		pod.Finalizers = nil
+
+		// spec related
+		for i := 0; i < len(pod.Spec.InitContainers); i++ {
+			preserveContainer(&pod.Spec.InitContainers[i])
+		}
+		for i := 0; i < len(pod.Spec.Containers); i++ {
+			preserveContainer(&pod.Spec.Containers[i])
+		}
+		pod.Spec.EphemeralContainers = nil
```
a7i, line 193: ephemeral containers are used in PodLifeTime [here](https://github.com/kubernetes-sigs/descheduler/blob/94d1825c6866623469bf3eaec1b34ada7504d6ae/pkg/framework/plugins/podlifetime/pod_lifetime.go#L100-L105)

pkg/descheduler/descheduler.go
```diff
@@ -153,8 +153,65 @@ func metricsProviderListToMap(providersList []api.MetricsProvider) map[api.Metri
...
+			c.StartupProbe = nil
+			c.Lifecycle = nil
+			c.TerminationMessagePath = ""
+			c.TerminationMessagePolicy = ""
+			c.ImagePullPolicy = ""
+			c.SecurityContext = nil
+		}
+
+		// metadata related
+		pod.ManagedFields = nil
```
a7i: as mentioned for nodes, this is already being done [here](https://github.com/kubernetes-sigs/descheduler/blob/94d1825c6866623469bf3eaec1b34ada7504d6ae/pkg/descheduler/descheduler.go#L715-L720)

pkg/descheduler/descheduler.go
Colvin-Y: @a7i  only status, not spec

pkg/descheduler/descheduler.go
Colvin-Y: > should this move [here](https://github.com/kubernetes-sigs/descheduler/blob/94d1825c6866623469bf3eaec1b34ada7504d6ae/pkg/descheduler/descheduler.go#L588) instead?

ditto

pkg/descheduler/descheduler.go
Colvin-Y: @a7i  thx for your review, plz check them all

a7i:
Nice change @Colvin-Y  - do you have any before/after results on memory footprint you can share?

@ingvagabund How does a global change like this impact a plugin? For example, if I write my custom Descheduler plugin which relies on some of the omitted fields, can I override this behavior?

Colvin-Y:
> 不错的改变@Colvin-Y - 您能否分享关于内存占用的前后结果？
> 

hi @a7i ,
In our prod env, save 47% memory on avg

Colvin-Y:
> If I write my custom Descheduler plugin which relies on some of the omitted fields, can I override this behavior?

hi @a7i 

Before cache start, you can override it. But maybe best practice is to modify func `preserveNeeded` because there are many plugins but using same cache.

ingvagabund:
Each plugin might tell which fields are necessary. I.e. through a new method in the similar fashion as scheduler's `EventsToRegister`. E.g. `FieldsToPreserve` or similar:
```
func (d *PodLifeTime) FieldsToPreserve() ([]fwk.ObjectFields, error) {
	return []fwk.ObjectFields{
		{Resource: fwk.Pod, Fields: {...}},
		{Resource: fwk.Node, Fields: {...}},
	}
}
```
This might be tricky to implement as each plugin will need different fields. So a simple clearing functions as in this PR will not suffice.

ingvagabund:
A naive solution would be to enumerate all fields. E.g.:
```go
c.Command = nil
c.Args = nil
c.WorkingDir = ""
c.Ports = nil
...
```
into
```go
const (
  POD_COMMAND = iota
  POD_ARGS
  POD_WORKING_DIR
  POD_PORTS
  ...
```
The same for nodes. Then, have each plugin to define a list of fields to preserve:
```go
func func (pl *Plugin) PreservedField() [] uint {
  return []uint {POD_COMMAND, POD_ARGS, ...}
}
```
Then iterate through all plugins, make a union of all `PreservedField()` lists and use the list to tell which fields can be cleared:
```go
if _, in := preservedFieldsMap[POD_COMMAND]; !in {
  c.Command = nil
}
if _, in := preservedFieldsMap[POD_ARGS]; !in {
  c.Args = nil
}
...
```

Colvin-Y:
> A naive solution would be to enumerate all fields. E.g.:
> 
> ```go
> c.Command = nil
> c.Args = nil
> c.WorkingDir = ""
> c.Ports = nil
> ...
> ```
> 
> into
> 
> ```go
> const (
>   POD_COMMAND = iota
>   POD_ARGS
>   POD_WORKING_DIR
>   POD_PORTS
>   ...
> ```
> 
> The same for nodes. Then, have each plugin to define a list of fields to preserve:
> 
> ```go
> func func (pl *Plugin) PreservedField() [] uint {
>   return []uint {POD_COMMAND, POD_ARGS, ...}
> }
> ```
> 
> Then iterate through all plugins, make a union of all `PreservedField()` lists and use the list to tell which fields can be cleared:
> 
> ```go
> if _, in := preservedFieldsMap[POD_COMMAND]; !in {
>   c.Command = nil
> }
> if _, in := preservedFieldsMap[POD_ARGS]; !in {
>   c.Args = nil
> }
> ...
> ```

ditto

pkg/descheduler/descheduler.go
```diff
@@ -153,6 +153,61 @@ func metricsProviderListToMap(providersList []api.MetricsProvider) map[api.Metri
...
+		accessor.SetManagedFields(nil)
+		accessor.SetFinalizers(nil)
+	}
+
+	if pod, ok := obj.(*v1.Pod); ok {
+		preserveContainer := func(c *v1.Container) {
+			c.Command = nil
+			c.Args = nil
+			c.WorkingDir = ""
+			c.Ports = nil
```
fanhaouu, line 170: I suggest not removing this c.Ports. The NodePorts plugin in the scheduler needs to use the container’s ports, and the related field values may be required during the node fit phase.

Colvin-Y: ack, I need some time to refact all as 
```
func func (pl *Plugin) PreservedField() [] uint {
  return []uint {POD_COMMAND, POD_ARGS, ...}
}
```

PR #1740, opened 2025-09-09
Expose classification metrics from LowNodeUtilization plugin

tiraboschi:
Enhance the plugin interface introducing an additional `Metrics` interface to be easily implemented by plugins
that intend to register and expose metrics.

Expose two new metrics from the `LowNodeUtilization` plugin:
- `descheduler_low_node_utilization_thresholds`
- `descheduler_low_node_utilization_classification`

These metrics allow users to properly monitor plugin behavior,
which was previously only visible in descheduler logs.

### DIFF
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -17,7 +17,9 @@ limitations under the License.
 package metrics
 
 import (
+	"strings"
 	"sync"
+	"unicode"
 
 	"k8s.io/component-base/metrics"
 	"k8s.io/component-base/metrics/legacyregistry"
@@ -28,6 +30,43 @@ const (
 	// DeschedulerSubsystem - subsystem name used by descheduler
 	DeschedulerSubsystem = "descheduler"
 )
+
+// MetricsHandler provides a smart wrapper for metrics that handles export logic
+type MetricsHandler interface {
+	// WithLabelValues is equivalent to metrics.WithLabelValues but respects ShouldExportMetrics
+	WithLabelValues(lvs ...string) MetricsHandler
+	// Set sets the gauge value (for GaugeVec metrics)
+	Set(val float64)
+	// Inc increments the counter (for CounterVec metrics)
+	Inc()
+	// Add adds a value to the counter (for CounterVec metrics)
+	Add(val float64)
+	// Observe observes a value for histogram (for HistogramVec metrics)
+	Observe(val float64)
+}
+
+// normalizePluginName converts a plugin name to a normalized form for use in metrics subsystems
+// Examples: "LowNodeUtilization" -> "low_node_utilization", "RemovePodsViolatingNodeTaints" -> "remove_pods_violating_node_taints"
+func normalizePluginName(pluginName string) string {
+	var result strings.Builder
+	for i, r := range pluginName {
+		if unicode.IsUpper(r) && i > 0 {
+			result.WriteRune('_')
+		}
+		result.WriteRune(unicode.ToLower(r))
+	}
+	return result.String()
+}
+
+// getPluginSubsystem returns the subsystem name for a plugin, optionally including profile name
+func getPluginSubsystem(profileName, pluginName string) string {
+	normalizedPluginName := normalizePluginName(pluginName)
+	if profileName == "" {
+		return DeschedulerSubsystem + "_" + normalizedPluginName
+	}
+	normalizedProfileName := normalizePluginName(profileName)
+	return DeschedulerSubsystem + "_" + normalizedProfileName + "_" + normalizedPluginName
+}
 
 var (
 	PodsEvicted = metrics.NewCounterVec(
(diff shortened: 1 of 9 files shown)


metrics/metrics.go
```diff
@@ -100,6 +119,8 @@ var (
 		DeschedulerStrategyDuration,
 		LoopDuration,
 		StrategyDuration,
+		LowNodeUtilizationThresholds,
+		LowNodeUtilizationClassification,
```
ingvagabund: This will not scale well for external plugins. Maybe a combination of init() and metrics registration within each plugin? Or, have you tried to invoke `RegisterMetrics` within the plugin instead?

pkg/framework/plugins/nodeutilization/lownodeutilization.go
```diff
@@ -179,6 +181,16 @@ func (l *LowNodeUtilization) Balance(ctx context.Context, nodes []*v1.Node) *fra
 		)
 	}
 
+	if l.handle.ShouldExportMetrics() {
+		for key, rt := range thresholds {
+			for i, t := range rt {
+				for resource, value := range t {
+					metrics.LowNodeUtilizationThresholds.WithLabelValues(key, strconv.Itoa(i), resource.String()).Set(float64(value))
```
ingvagabund: There could be a new handler for setting metrics. E.g. `l.handle.Metric(metricsName).WithLabelValues` or similar that would be a no-op in case metrics are not enabled.

metrics/metrics.go
tiraboschi: Done, I tried adding an optional `Metrics` interface to complement `Handle` without breaking backward compatibility

pkg/framework/plugins/nodeutilization/lownodeutilization.go
tiraboschi: Done, I implemented a new metric handler wrapping all the metric setters embedding there the skip metric setting.

metrics/metrics_test.go
```diff
@@ -0,0 +1,76 @@
+/*
+Copyright 2021 The Kubernetes Authors.
```
ingvagabund: 2025

pkg/framework/plugins/nodeutilization/lownodeutilization.go
```diff
@@ -134,6 +137,16 @@ func (l *LowNodeUtilization) Name() string {
...
 }
 
+// HandleMetric returns a metrics handler for the specified metric name
+func (l *LowNodeUtilization) HandleMetric(metricName string) metrics.MetricsHandler {
+	return metrics.PluginRegistry.HandlePluginMetric(LowNodeUtilizationPluginName, metricName)
+}
+
+// RegisterMetrics registers the plugin's metrics
+func (l *LowNodeUtilization) RegisterMetrics() {
+	RegisterMetrics(metrics.PluginRegistry)
```
ingvagabund: There's probably no way to disallow users to invoke this in the `NewXXX` functions. Also, the descheduler registers the metrics independent of whether metrics server is enabled or not. I wonder what would be the overhead of just registering the metrics directly. Allowing to skip `ShouldExport` logic.

ingvagabund: Also, a single plugin can be initialized in each profile. Thus, a metric entry needs to be labeled with a profile name as well.

pkg/framework/plugins/nodeutilization/lownodeutilization.go
```diff
@@ -179,6 +192,14 @@ func (l *LowNodeUtilization) Balance(ctx context.Context, nodes []*v1.Node) *fra
 		)
 	}
 
+	for key, rt := range thresholds {
+		for i, t := range rt {
+			for resource, value := range t {
+				l.HandleMetric("thresholds").WithLabelValues(key, strconv.Itoa(i), resource.String()).Set(float64(value))
```
ingvagabund, line 198: idea: `l.HandleMetric` could be already injecting the profile name

metrics/metrics_test.go
tiraboschi: done

pkg/framework/plugins/nodeutilization/lownodeutilization.go
tiraboschi: I added a bit of logic to register metrics by profile name and by plugin name.
A metric named `Classification` registered by `LowNodeUtilization` plugin required by `CustomProfile` profile will result in `descheduler_custom_profile_low_node_utilization_classification`.

ingvagabund:
@tiraboschi this can be probably closed

tiraboschi:
> @tiraboschi this can be probably closed

See: https://github.com/kubernetes-sigs/descheduler/pull/1777

PR #1743, opened 2025-09-16
feat(test/e2e): introduce new flags to override RunAsUser and RunAsGroup security context setting

ingvagabund:
So RunAsUser and RunAsGroup can be either omitted and set to different values.

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -19,6 +19,7 @@ package e2e
 import (
 	"bufio"
 	"context"
+	"flag"
 	"fmt"
 	"math"
 	"os"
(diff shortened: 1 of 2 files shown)


test/e2e/e2e_test.go
```diff
@@ -66,6 +67,23 @@ import (
 	"sigs.k8s.io/descheduler/pkg/utils"
 )
 
+var (
+	deschedulerImage = flag.String("descheduler-image", "", "descheduler image to set in the pod spec")
```
ricardomaraschini, line 71: Where is the `deschedulerImage` used ?

ingvagabund: Oh, I forgot.

ingvagabund: Updated PTAL

PR #1751, opened 2025-09-30
helm: Fix liveness probe timeout for descheduler chart

rstribrn:
## Description
This PR fixes an issue where the descheduler pod is frequently restarted due to liveness probe failures.

## Checklist
completeness of your changes:

### DIFF
--- a/charts/descheduler/values.yaml
+++ b/charts/descheduler/values.yaml
@@ -227,8 +227,9 @@ livenessProbe:
     path: /healthz
     port: 10258
     scheme: HTTPS
-  initialDelaySeconds: 3
-  periodSeconds: 10
+  initialDelaySeconds: 5
+  periodSeconds: 20
+  timeoutSeconds: 5
 
 service:
   enabled: false


charts/descheduler/values.yaml
```diff
@@ -227,8 +227,9 @@ livenessProbe:
     path: /healthz
     port: 10258
     scheme: HTTPS
-  initialDelaySeconds: 3
-  periodSeconds: 10
+  initialDelaySeconds: 5
+  periodSeconds: 20
+  timeoutSeconds: 5
```
ricardomaraschini, line 232: Hi, thanks for your contribution. Would you mind to elaborating on how this addresses the problem ?

rstribrn: Current configuration is missing timeoutSeconds, so the default 1 second is used.
- [Reference](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes): timeoutSeconds: Number of seconds after which the probe times out. Defaults to 1 second. Minimum value is 1.

--
I have noticed this 1 second timeout leads to unnecessary descheduler restarts. Especially when K8s node is under pressure. Therefore I changed it to 5s to make it more reliable.
Since I was already there, I also proposed changing the periodSeconds from 10s to 20s, effectively restarting descheduler in 60s rather than 30s periods of its possible unavailability (failureThreshold kept 3).
Its also more reasonable staring checking the descheduler liveness a little bit later (initialDelaySeconds from 3s to 5s).
Overall, changes are focused on fine-tuning the liveness probe values to balance descheduler's restarts against it availability.

To be strict, the most important change is timeoutSeconds (1s->5s), the other two are redundant.

rostislav-stribrny-ness:
According to the test output, I see no relation with my changes. Likely problem with the test itself?

rostislav-stribrny-ness:
While looking at test result, I can see this:
--- FAIL: TestLiveMigrationInBackground (310.09s)

If this is the problem, then I would say it has nothing to do with my changes. These changes are about descheduler's lifecycle when its not available, not its core functionality (managing other pods lifecycle).

ricardomaraschini:
@rstribrn This test seems to be failing across the board. I have seen other PRs with the same failure.

Issue #1758, opened 2025-10-15
Feature Request: Global Namespace Exclusion Configuration


patsevanton:
**Is your feature request related to a problem? Please describe.**
Currently, there is no way to define a global namespace exclusion for all Descheduler strategies. Each strategy must be configured individually with its own `namespaces.exclude` list. This becomes difficult to manage and error-prone in large clusters where multiple strategies are enabled and certain namespaces (e.g., `kube-system`, `monitoring`, or `infra`) should always be ignored.

**Describe the solution you'd like**
I would like to have a global configuration option (for example, `globalNamespaceExclusions`) that allows specifying a list of namespaces that should be excluded from *all* Descheduler strategies. This would make configuration simpler, reduce duplication, and ensure consistent behavior across all strategies.

Example configuration:

```yaml
apiVersion: "descheduler/v1alpha2"
kind: "DeschedulerPolicy"
globalNamespaceExclusions:
  - kube-system
  - monitoring
  - infra
strategies:
  RemoveDuplicates:
    enabled: true
  RemovePodsViolatingInterPodAntiAffinity:
    enabled: true
```

**Describe alternatives you've considered**

* Duplicating the `excludeNamespaces` configuration under each strategy. This works but is hard to maintain and can easily lead to inconsistent configurations.
* Writing an external script or admission controller to patch the Descheduler configuration automatically, but this adds unnecessary operational complexity.

**What version of descheduler are you using?**
descheduler version: v0.32.0

**Additional context**
This feature would be especially useful in environments with many namespaces and teams, ensuring that system or critical namespaces are consistently excluded without having to manually edit every strategy configuration.

ingvagabund:
Hello @patsevanton. Thank you for suggesting the improvements. This is doable. Would the global namespace exclusion be mutually exclusive to per profile namespace exclusion? Or, would the per profile exclusion take precedence over the global one? It would be easier to make it mutually exclusive by the policy validation. On the other hand the restriction would make it less expressive in terms of making the global exclusion for most of the profiles with few exceptions where the exclusion would be different.

ingvagabund:
The first version of the new global might be more strict so we can collect user feedback and later decide whether to make the restriction more relaxed.

I'd also suggest to keep the original naming even for the global exclusion. I.e. `globalNamespaceExclusions` -> `namespace.exclude` so later the `namespace` config struct can be extended with other fields without introducing deprecation and renaming.

patsevanton:
[...]

PR #1786, opened 2025-12-11
feat: add namespace label selector

W1seKappa:
## Description

Added namespace label selector for filtering pods in the DefaultEvictor plugin.
This pull request was created to finalize the initiative from that [pr1501](https://github.com/kubernetes-sigs/descheduler/pull/1501)

DeschedulerPolicy.yaml example
```yaml
kind: "DeschedulerPolicy"
profiles:
  - name: default
    pluginConfig:
      - name: "DefaultEvictor"
        args:
          namespaceLabelSelector:
            matchLabels:
              environment: staging
      - name: "PodLifeTime"
        args:
          maxPodLifeTimeSeconds: 86400
    plugins:
      deschedule:
        enabled:
          - "PodLifeTime"
```

Descheduler will evict pods older than 24 hours, but only from namespaces with the label environment: staging.

## Checklist
completeness of your changes:


W1seKappa:
@ingvagabund Hi, I'm happy to continue the discussion based on the initiative from this [pr1501](https://github.com/kubernetes-sigs/descheduler/pull/1501).

pkg/framework/plugins/defaultevictor/defaultevictor.go
```diff
@@ -473,6 +496,41 @@ func getPodIndexerByOwnerRefs(indexName string, handle frameworktypes.Handle) (c
 	}); err != nil {
 		return nil, err
 	}
+	return indexer, nil
+}
+
+func getNamespacesListByLabelSelector(indexName string, labelSelector *metav1.LabelSelector, handle frameworktypes.Handle) (cache.Indexer, error) {
```
ingvagabund: I understand the idea here. If a namespace is matched by the label selector it's returned. If it is not an empty list is returned. I.e.:
- indexer.ByIndex(indexName, NSmatched) -> {NSmatched} // list with a single ns
- indexer.ByIndex(indexName, NSNotmatched) -> {} // empty list

On the other hand `getNamespacesListByLabelSelector` gives the belief all namespaces matching a label selector are returned. I.e. a label selector is the key, not a namespace. I.e.
- indexer.ByIndex(indexName, LS1) -> list of namespaces matching LS1
- ...
- indexer.ByIndex(indexName, LSN) -> list of namespaces matching LSN

Given each profile can have its own DefaultEvictor configuration with a different label selector the current global "namespaceWithLabelSelector" indexer will not work correctly as once the first indexer gets registered, any second attempt returns the first indexer. So the second label selector will get ignored and all DefaultEvictors will share the first label selector.

ingvagabund: @W1seKappa thank you for taking over. As the next step I suggest to create another unit test that have two profiles configured. Each with a different label selector and to make sure both evictors evict distinct set of namespaces to correctly validate the functionality. There's no need to create actual profiles. Just mimicking it through creating two default evictors over the same namespace informer.

W1seKappa: @ingvagabund Hello, I manually tested the functionality you mentioned.  Indeed, it is currently not working correctly. I see two ways to solve this problem:

1. Create a separate index for each set of mathLabels.
```yaml
namespaceLabelSelector:
    matchLabels:
            env: prod
            animal: cat
```
It will create its own index `namespaceWithLabelSelector-animal=cat,env=prod`.

```yaml
namespaceLabelSelector:
    matchLabels:
         env: prod 
```
It will create its own index `namespaceWithLabelSelector-env=prod`.
This solution is currently implemented in the code. I understand the need for some optimizations, but I wanted to develop the core idea first.
2. Create an index for each pair in matchLabels.
These same configuration files will generate 3 indexes instead of 2:
`namespaceWithLabelSelector-env=prod`
`namespaceWithLabelSelector-env=dev`
`namespaceWithLabelSelector-animal=cat`

ingvagabund: One drawback of creating indexer for each defaultplugin/nslabelselector is the number of indexers will grow with each profile. In addition, https://github.com/kubernetes-sigs/descheduler/issues/1499 is asking for the same. So each plugin will have its own indexer in general. With https://github.com/kubernetes-sigs/descheduler/issues/1758 getting later extended as another increase.

In the worst case all ns label selectors will be different. Allowing each plugin to choose the indexer name may lead to name collisions. Luckily, the label selector plays no role in the naming as each plugin has a unique position in the configuration. To make the ns indexer registration collision free the framework itself can expose a new method for the registration.

E.g.
```go
type Handle interface {
	...
	RegisterNSIndexer(indexer func(obj interface{}) ([]string, error)) error
	GetNSIndexer() (cache.Indexer, error)
	...
}
```
or similar construct. Each plugin is injected with a `Handle` instance. So the indexer name can get generated based on the profile/plugin configuration indices. E.g. "profile_idx_pluginname_idx_nslabelselector".

In the best case all the ns label selectors can be identical. So the framework could do some smart indexing of the label selectors. With a match expression as another way. Leaving admins to take the responsibility of making the label selectors/match expressions properly chosen. E.g. avoiding "env=prod,animal=cat" and "animal=cat,env=prod" case.

ingvagabund: An indexer is a nice way to quickly detect whether a namespace matches a label selector. With the current approach the indexer creates an optimal distribution of keys in a hash table. Either a namespace is matched (the ns has a hash key) or is not matched (the ns does not have a hash key). Trying to combine multiple label selectors into the same indexer would make the searching and updating less efficient.

ingvagabund: Note: right now the configuration offers only a global node selector. Yet, it's also a valid case to have two different profiles targeting two distinct node pools. So the concept of growing the number of indexers with the number of profiles spans beyond the namespace label selector.

W1seKappa: @ingvagabund, Hello!
If I understood correctly, at this stage it was necessary to add tests to this solution. I've done that.
Next, the indexing framework needs to be refactored. Right?
I would be happy to participate.

W1seKappa:
@ingvagabund Hi, can you review pr please?

ingvagabund:
In general, each plugin can have many indexers—for example, for nodes, namespaces, pods, and other resources. Each plugin, depending on its internal optimizations, may choose a different indexing function.

Indexers are currently used very rarely in the framework and default plugins. Therefore, extending the framework interface with additional helpers for registering and retrieving indexers might introduce an unnecessary and overly restrictive layer without first understanding how indexers will be used.

For the moment, I suggest avoiding any restrictions on how many indexers can be registered or which ones can be registered. Instead, we should extend the framework handle to provide a unique ID for each profile, so that indexers within the same profile share a unique prefix. This avoids collisions when the same profile is instantiated more than once.

Later, once we learn more about indexer usage, we can revisit whether it makes sense to impose additional restrictions.

ingvagabund:
@W1seKappa once https://github.com/kubernetes-sigs/descheduler/pull/1808 is merged you should be good to proceed here.

W1seKappa:
@ingvagabund HI, I changed the logic for creating index names, please review it.

W1seKappa:
@ingvagabund Hello, sorry for the frequent mentions, I really want to continue working on PR)

ingvagabund:
@W1seKappa this project has been understaffed for many years. Unfortunately, I can't accelerate my company priorities. Yet, this PR is on my todo list. I'd like to first refactor the code again so each plugin's New is invoked only once. So you can register the indexer only once and simplify the logic.

pkg/framework/plugins/defaultevictor/defaultevictor.go
```diff
@@ -473,6 +499,36 @@ func getPodIndexerByOwnerRefs(indexName string, handle frameworktypes.Handle) (c
...
+	nsInformer := handle.SharedInformerFactory().Core().V1().Namespaces().Informer()
+	indexer := nsInformer.GetIndexer()
+	// do not reinitialize the indexer, if it's been defined already
+	for name := range indexer.GetIndexers() {
+		if name == indexName {
+			return indexer, nil
+		}
+	}
+
+	if err := nsInformer.AddIndexers(cache.Indexers{
```
ingvagabund: No matter how the main descheduler loop refactorings ends `New` is the right place for registering any indexer. So later on you can access it through `handle.SharedInformerFactory().Core().V1().Namespaces().Informer().ByIndex(...)` directly. Even now you could invoke getNamespacesListByLabelSelector in New. That will register the indexer. Right now `New` gets invoked during each descheduling cycle. So the condition whether the indexer exists is still needed. Once I am done with improving the code the condition can be dropped.

ingvagabund:
One step closer with this: https://github.com/kubernetes-sigs/descheduler/pull/1812.

@W1seKappa This PR looks good. Just one nit. If you could move the indexer registration into the New function?

W1seKappa:
> One step closer with this: #1812.
> 
> @W1seKappa This PR looks good. Just one nit. If you could move the indexer registration into the New function?

@ingvagabund Hi, done, i hope so. Can you please check the PR?

ingvagabund:
@W1seKappa thank you. The PR is good to go. Can you please squash all the commit before merging?

W1seKappa:
> @W1seKappa thank you. The PR is good to go. Can you please squash all the commit before merging?

@ingvagabund Done, let's go)

Issue #1558, opened 2024-11-19
optional Restart instead of Eviction


B1F030:
**Is your feature request related to a problem? Please describe.**

Is there a graceful resolution for Eviction?
For now, descheduler evicts the pods which will cause a service interruption.
And I wonder that can I customize the config to make it **Restart** the pods instead of **Eviction**?

**Describe the solution you'd like**

Provide an optional config to restart pods instead of evict pods, so that the service will not be interrupted.

**Describe alternatives you've considered**

**What version of descheduler are you using?**

descheduler version: v0.31.0

**Additional context**

a7i:
>For now, descheduler evicts the pods which will cause a service interruption.

Would you please elaborate on why that is, given that it uses the eviction api. Do you define a `PodDisruptionBudget`?

B1F030:
> Would you please elaborate on why that is, given that it uses the eviction api. Do you define a `PodDisruptionBudget`?

Sure, I'll provide more details about our scenario:

We have two kubernetes gpu node pools, A as monthly(one node) and B as elastic(zero node, but with autoscaler).
now a deployment with one replicas using gpu is running on A(exclusive to all resources of one node), when we rolling update it, it will trigger the autoscaler, and be scheduled to B, then A will be in low usage.

Since monthly node is cheaper, we want the pod to be rescheduled and go back to A, so that the elastic node can be recovered to zero.

In conclusion, this workload takes up almost all of resources on one node, and there's only one replicas so we can't use PDB(using multiple replicas will increase cost).
We hope that, when rolling update, it will be scheduled to the elastic node. After rolling update is done, trigger the reschedule and evict the workload to monthly node(depends on `preferredDuringSchedulingIgnoredDuringExecution`).

Also we don't want the service interrupted, so I'm looking for a graceful method to reschedule(create the pod before evict it, just like rolling update or `kubectl rollout restart deployment`).

googs1025:
I understand this scenario is more like a migration, not a simple "eviction" scenario

PR #1841, opened 2026-02-24
feat: rollout restart single-replica Deployments instead of evicting

### LINKED ISSUE #786: descheduler when ReplicaSet=1

Deployment often act as backend providing service. If one deployment only has single pod(replicas: 1), descheduler can still evict it. At that time, the service will be stopped while recreating.

Add a filter to count the deployments' or jobs' replicas. If replicas=1, this pod can not be evict.

dmitriy-myz:
## Description
For Deployments with replicas=1 and RollingUpdate strategy, trigger a rollout restart (patch pod template annotation) instead of using the Pod Eviction API. This avoids downtime by letting the Deployment controller create a new pod before terminating the old one.

Falls through to normal eviction on errors, non-Deployment pods, multi-replica Deployments, or Recreate strategy.
Fixes #786 #1558
## Checklist
completeness of your changes:

### DIFF
--- a/pkg/descheduler/evictions/evictions.go
+++ b/pkg/descheduler/evictions/evictions.go
@@ -25,10 +25,12 @@ import (
 
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
+	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
 	policy "k8s.io/api/policy/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
@@ -229,6 +231,11 @@ type PodEvictor struct {
 	erCache                          *evictionRequestsCache
 	featureGates                     featuregate.FeatureGate
 
+	// restartedDeployments tracks "namespace/name" of Deployments already rollout-restarted this cycle
+	restartedDeployments map[string]bool
+	// lastRolloutRestart indicates the most recent evictPod call used rollout restart
+	lastRolloutRestart bool
+
 	// registeredHandlers contains the registrations of all handlers. It's used to check if all handlers have finished syncing before the scheduling cycles start.
 	registeredHandlers []cache.ResourceEventHandlerRegistration
 }
@@ -258,6 +265,7 @@ func NewPodEvictor(
 		metricsEnabled:                   options.metricsEnabled,
 		nodePodCount:                     make(nodePodEvictedCount),
 		namespacePodCount:                make(namespacePodEvictCount),
+		restartedDeployments:             make(map[string]bool),
 		featureGates:                     featureGates,
 	}
 
@@ -401,6 +409,7 @@ func (pe *PodEvictor) ResetCounters() {
 	pe.nodePodCount = make(nodePodEvictedCount)
 	pe.namespacePodCount = make(namespacePodEvictCount)
 	pe.totalPodCount = 0
+	pe.restartedDeployments = make(map[string]bool)
 }
 
 func (pe *PodEvictor) evictionRequestsTotal() uint {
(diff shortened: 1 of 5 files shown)


vazyzy (approved):

pkg/descheduler/evictions/evictions.go
```diff
@@ -552,24 +561,124 @@ func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod, opts EvictOptio
...
+		klog.V(1).InfoS("Evicted pod", "pod", klog.KObj(pod), "reason", opts.Reason, "strategy", opts.StrategyName, "node", pod.Spec.NodeName, "profile", opts.ProfileName, "method", method)
 		reason := opts.Reason
 		if len(reason) == 0 {
 			reason = opts.StrategyName
 			if len(reason) == 0 {
 				reason = "NotSet"
 			}
 		}
-		pe.eventRecorder.Eventf(pod, nil, v1.EventTypeNormal, reason, "Descheduled", "pod eviction from %v node by sigs.k8s.io/descheduler", pod.Spec.NodeName)
+		if pe.lastRolloutRestart {
```
ingvagabund, line 580: Is it always guaranteed `pe.lastRolloutRestart` was set by `pod.Namespace`/`pod.Name`?

pkg/descheduler/evictions/evictions.go
```diff
@@ -552,24 +561,124 @@ func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod, opts EvictOptio
...
+			pe.eventRecorder.Eventf(pod, nil, v1.EventTypeNormal, reason, "Descheduled", "pod rollout-restarted (single-replica) from %v node by sigs.k8s.io/descheduler", pod.Spec.NodeName)
```
ingvagabund, line 581: Annotating the corresponding deployment does not always guarantee the pod will get rolled out. Recording the event might be confusing.

pkg/descheduler/evictions/evictions.go
```diff
@@ -552,24 +561,124 @@ func (pe *PodEvictor) EvictPod(ctx context.Context, pod *v1.Pod, opts EvictOptio
...
+		return true, true
+	}
+	if !pe.dryRun {
+		if err := pe.rolloutRestartDeployment(ctx, deploy); err != nil {
+			klog.V(1).InfoS("Failed to rollout restart Deployment, falling through to normal eviction", "deployment", deployKey, "pod", klog.KObj(pod), "err", err)
+			return false, false
+		}
+	}
+	klog.V(1).InfoS("Triggered rollout restart for single-replica Deployment instead of eviction", "deployment", deployKey, "pod", klog.KObj(pod), "dryRun", pe.dryRun)
+	pe.restartedDeployments[deployKey] = true
```
ingvagabund, line 670: The older deployment keys are not getting garbage collected.

EDIT: `ResetCounters()` resets it.

ingvagabund: This might not work well when the descheduling cycle is short.

ingvagabund:
In general the descheduler does not assume anything about a deployment's lifecycle. A pod is either evicted or it is not. What if annotating a deployment is not enough? What if there are other requirements to be met before a pod gets rolled out? The recommended solution here is to build a validating admission webhook that can handle all these concerns.

I understand the current use case is a deployment with a single replica with goal to trigger the rolling update. Yet, not even this operation is atomic.

dmitriy-myz:
@ingvagabund some context for why this matters to us, then a response to the design question.

We run on-demand environments where ~60 pods are scheduled at once. They frequently land on a single node and overload it. 
We rely on the descheduler to rebalance, and our QA tests run against these envs immediately after provisioning - so eviction-driven downtime on single-replica deployments broke the test runs. This PR is the patch we've been running internally to make rebalance safe for that workflow.

On the validating webhook recommendation - I want to make sure I'm not missing something, because as I understand it the webhook path doesn't fully replace this PR:

- A validating webhook can block an eviction, but it can't trigger a rollout restart. Patching the Deployment still has to happen somewhere. So the webhook approach is really "webhook + a separate controller that watches blocked evictions and patches Deployments" - strictly more components than this PR, not fewer.
- Operationally, a webhook adds a stateful HA service to every cluster running descheduler, with cluster-wide blast radius: if the webhook pod is down with failurePolicy:
Fail, all evictions break (including kubectl drain and node upgrades). With failurePolicy: Ignore it silently does nothing, which defeats the safety property. Plus cert routine.

So the trade is "~600 lines additive in descheduler with fall-through-to-eviction on any failure" vs. "a webhook + controller deployment, with new failure modes for cluster eviction." I'd argue the former is the lighter footprint for users.

That said - your concerns about lifecycle assumptions and non-atomicity are fair, and I'm willing to address them concretely:

1. Opt-in via a feature gate, default off. Users who don't enable it get exactly today's behavior.
2. Pre-flight guards before patching: skip the rollout-restart path when spec.paused, maxSurge: 0, or status.ObservedGeneration != generation, and fall through to normal
eviction.
3. Fix the inline comments: move lastRolloutRestart off the receiver, switch the dedup map to TTL-keyed by observedGeneration so it self-expires on short cycles, and only emit the rollout event after verifying progress.

Would the opt-in + pre-flight + the inline fixes be acceptable to you? If you'd still rather see this live entirely outside descheduler, I'll close in favor of that path - but I want to confirm that's the direction before reshaping the PR.

ismaelpuerto:
more context: this PR would help us operate critical workloads directly connected to [PLCs (Programmable Logic Controllers)](https://en.wikipedia.org/wiki/Programmable_logic_controller) in industrial automation environments, where an eviction-induced disruption could stop the production line itself.

PR #1844, opened 2026-03-01
Extend PodLifeTime with condition, exit code, owner kind, and transition time filters

a7i:
### Extend PodLifeTime with new filtering fields

Adds new fields to `PodLifeTimeArgs` so it can handle fine-grained, transition-based eviction:

- **`conditions`**: Filter pods by `status.conditions` entries (`type`, `status`, `reason`) with an optional `minTimeSinceLastTransitionSeconds` per condition filter
- **`exitCodes`**: Filter by container terminated exit codes
- **`ownerKinds`**: Include or exclude pods by owner reference kind (`include`/`exclude`)
- **`states` enhanced**: Now also checks container *terminated* reasons (previously only checked waiting reasons)

All non-empty filter categories are ANDed together (a pod must satisfy every specified filter). Within each category, entries are ORed. Pods are processed from oldest to newest based on their creation time. Condition filters, including `minTimeSinceLastTransitionSeconds`, determine which pods are eligible for eviction but do not affect this ordering.

### Extract shared container state matching helpers into podutil

Moves the container waiting/terminated state checking logic from `PodLifeTime` and `RemovePodsHavingTooManyRestarts` into `podutil` as separate exported helpers: `HasMatchingContainerWaitingState` and `HasMatchingContainerTerminatedState`. Each plugin composes only the helpers it needs — `TooManyRestarts` uses only the waiting helper, while `PodLifeTime` uses both.

## Checklist

- [x] Code Readability
- [x] Naming Conventions
- [x] Code Duplication
- [x] Function/Method Size
- [x] Comments & Documentation
- [x] Error Handling
- [x] Testing
- [x] Performance
- [x] Dependencies
- [x] Logging & Monitoring
- [x] Backward Compatibility
- [x] Resource Management
- [x] PR Description
- [x] Documentation & Changelog

### DIFF
--- a/pkg/framework/plugins/removepodshavingtoomanyrestarts/toomanyrestarts.go
+++ b/pkg/framework/plugins/removepodshavingtoomanyrestarts/toomanyrestarts.go
@@ -84,19 +84,13 @@ func New(ctx context.Context, args runtime.Object, handle frameworktypes.Handle)
 			if states.Has(string(pod.Status.Phase)) {
 				return true
 			}
-
-			containerStatuses := pod.Status.ContainerStatuses
-
-			if tooManyRestartsArgs.IncludingInitContainers {
-				containerStatuses = append(containerStatuses, pod.Status.InitContainerStatuses...)
+			if podutil.HasMatchingContainerWaitingState(pod.Status.ContainerStatuses, states) {
+				return true
 			}
-
-			for _, containerStatus := range containerStatuses {
-				if containerStatus.State.Waiting != nil && states.Has(containerStatus.State.Waiting.Reason) {
-					return true
-				}
+			if tooManyRestartsArgs.IncludingInitContainers &&
+				podutil.HasMatchingContainerWaitingState(pod.Status.InitContainerStatuses, states) {
+				return true
 			}
-
 			return false
 		})
 	}
--- a/pkg/descheduler/pod/pods.go
+++ b/pkg/descheduler/pod/pods.go
@@ -292,6 +292,28 @@ func SortPodsBasedOnAge(pods []*v1.Pod) {
 	})
 }
 
+// HasMatchingContainerWaitingState returns true if any container status has a
+// waiting reason present in the given set.
+func HasMatchingContainerWaitingState(statuses []v1.ContainerStatus, states sets.Set[string]) bool {
+	for _, cs := range statuses {
+		if cs.State.Waiting != nil && states.Has(cs.State.Waiting.Reason) {
+			return true
+		}
+	}
+	return false
+}
+
+// HasMatchingContainerTerminatedState returns true if any container status has
+// a terminated reason present in the given set.
+func HasMatchingContainerTerminatedState(statuses []v1.ContainerStatus, states sets.Set[string]) bool {
+	for _, cs := range statuses {
+		if cs.State.Terminated != nil && states.Has(cs.State.Terminated.Reason) {
+			return true
+		}
+	}
+	return false
+}
+
 func GroupByNodeName(pods []*v1.Pod) map[string][]*v1.Pod {
 	m := make(map[string][]*v1.Pod)
 	for i := 0; i < len(pods); i++ {


pkg/framework/plugins/podlifetime/pod_lifetime.go
```diff
@@ -69,92 +81,111 @@ func New(ctx context.Context, args runtime.Object, handle frameworktypes.Handle)
...
 			}
 		}
-
-		nodeMap[node.Name] = node
 		podsToEvict = append(podsToEvict, pods...)
 	}
 
-	// Should sort Pods so that the oldest can be evicted first
-	// in the event that PDB or settings such maxNoOfPodsToEvictPer* prevent too much eviction
 	podutil.SortPodsBasedOnAge(podsToEvict)
```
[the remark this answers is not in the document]
a7i: This is somewhat unrelated to changes in this PR and should not change existing behavior of PodLifeTime

README.md
```diff
@@ -785,30 +785,52 @@ profiles:
...
+      args:
+        states:
+        - "Succeeded"
+        conditions:
+        - reason: "PodCompleted"
+          status: "True"
+          minTimeSinceLastTransitionSeconds: 14400
+        ownerKinds:
+          exclude:
+          - "Job"
```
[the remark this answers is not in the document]
a7i: The YAML is valid — this indentation style (list items at the same level as the key) is used consistently throughout the entire README and all other example files in the project. Verified with `yaml.safe_load` / Ruby `YAML.load_file`. Changing this would be inconsistent with the existing convention.

examples/pod-life-time-transition.yml
```diff
@@ -0,0 +1,20 @@
...
+      args:
+        states:
+        - "Succeeded"
+        conditions:
+        - reason: "PodCompleted"
+          status: "True"
+          minTimeSinceLastTransitionSeconds: 14400 # 4 hours
+        namespaces:
+          include:
+          - "default"
```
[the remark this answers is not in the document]
a7i: Same as above — the YAML is valid and parses correctly. This indentation style is consistent with all other examples in the project.

README.md
```diff
@@ -785,30 +785,52 @@ profiles:
...
-  - [Pod Reason](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-conditions) reasons of: `NodeAffinity`, `NodeLost`, `Shutdown`, `UnexpectedAdmissionError`
-  - [Container State Waiting](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#container-state-waiting) condition of: `PodInitializing`, `ContainerCreating`, `ImagePullBackOff`, `CrashLoopBackOff`, `CreateContainerConfigError`, `ErrImagePull`, `ImagePullBackOff`, `CreateContainerError`, `InvalidImageName`
+All non-empty filter categories are ANDed (a pod must satisfy every specified filter). Within each category, items are ORed (matching any one entry satisfies that filter). Pods are processed from oldest to newest based on their creation time. Condition filters, including `minTimeSinceLastTransitionSeconds`, determine which pods are eligible for eviction but do not affect this ordering.
 
-If a value for `states` or `podStatusPhases` is not specified,
-Pods in any state (even `Running`) are considered for eviction.
+See the [plugin README](pkg/framework/plugins/podlifetime/README.md) for detailed documentation and advanced use cases.
```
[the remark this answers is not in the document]
a7i: Updated the PR description to match the implementation: pods are sorted by creation time, condition filters only determine eligibility.

pkg/framework/plugins/podlifetime/pod_lifetime.go
```diff
@@ -69,92 +80,114 @@ func New(ctx context.Context, args runtime.Object, handle frameworktypes.Handle)
...
-	})
+	for _, f := range extraFilters {
+		podFilter = podutil.WrapFilterFuncs(podFilter, f)
+	}
+
+	if podLifeTimeArgs.MaxPodLifeTimeSeconds != nil {
+		podFilter = podutil.WrapFilterFuncs(podFilter, func(pod *v1.Pod) bool {
+			podAge := metav1.Now().Sub(pod.GetCreationTimestamp().Local())
+			if podAge < 0 {
+				return false
```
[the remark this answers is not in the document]
a7i: Fixed — now uses signed `time.Duration` and returns `false` when negative (clock skew guard).

a7i:
/cc @jklaw90 @ingvagabund 

🙏🏼

pkg/framework/plugins/podlifetime/pod_lifetime.go
```diff
@@ -170,3 +203,74 @@ loop:
 
 	return nil
 }
+
+// matchesAnyPodConditionFilter returns true if the pod has at least one
```
jklaw90, line 198: are some of these filters already done in another plugin? wonder if we should move these to a shared package

a7i: Good point. As far as I know PodLifeTime and RemoveFailedPods shared a lot of common logic (including this), but now that RemoveFailedPods calls PodLifeTIme under the hood, that duplication is no longer there.

a7i: found one instance for podshavintoomanyrestarts, I extract a common util method, but kept it as a separate commit

ingvagabund:
/hold
to avoid accidental approval

README.md
```diff
@@ -785,30 +785,52 @@ profiles:
...
-|--------------------------------|---------------------------------------------------|--------------------------|
-| `maxPodLifeTimeSeconds`        | int                                               |                          |
-| `states`                       | list(string)                                      | Only supported in v0.25+ |
-| `includingInitContainers`      | bool                                              | Only supported in v0.31+ |
-| `includingEphemeralContainers` | bool                                              | Only supported in v0.31+ |
-| `namespaces`                   | (see [namespace filtering](#namespace-filtering)) |                          |
-| `labelSelector`                | (see [label filtering](#label-filtering))         |                          |
+| Name | Type | Notes |
+|------|------|-------|
+| `maxPodLifeTimeSeconds` | uint | Pods older than this many seconds are evicted |
```
ingvagabund: nit: Worth sorting the parameters alphabetically

ingvagabund:
Instead of delegating `RemoveFailedPods` -> `PodLifeTime` let's deprecate `RemoveFailedPods` (mark it as critical fixes only). So users are not surprised by seeing different logs and aware they need to update their configuration if they wanna use the new `PodLifeTime`'s parameters.

Resp. do you still see any benefit maintaining `RemoveFailedPods`?

pkg/framework/plugins/removepodshavingtoomanyrestarts/toomanyrestarts.go
```diff
@@ -81,23 +81,7 @@ func New(ctx context.Context, args runtime.Object, handle frameworktypes.Handle)
...
-			}
-
-			containerStatuses := pod.Status.ContainerStatuses
-
-			if tooManyRestartsArgs.IncludingInitContainers {
-				containerStatuses = append(containerStatuses, pod.Status.InitContainerStatuses...)
-			}
-
-			for _, containerStatus := range containerStatuses {
-				if containerStatus.State.Waiting != nil && states.Has(containerStatus.State.Waiting.Reason) {
```
ingvagabund, line 95: This currently only includes waiting state, not terminating. The new `HasMatchingContainerState` includes terminating as well. It's true terminating pods are no longer relevant for the plugin so there's no harm checking their state. Yet, these extra pods will increase the list of pods to evict if a user includes the terminated reasons (e.g. by accident) and might reach the eviction limits sooner.

Also, this will add this extra reasoning for "why do we check terminated states for TooManyRestarts plugin too?" question.

Might work better to create a separate helpers for checking individual parts rather than the whole. E.g. `HasMatchingContainerWaitingState` and `HasMatchingContainerTerminatedState` and allow each plugin to combine only what is needed.

Also, `MatchesPodState` adds `states.Has(pod.Status.Reason)` check which is not used in `TooManyRestarts`.

a7i:
> Instead of delegating `RemoveFailedPods` -> `PodLifeTime` let's deprecate `RemoveFailedPods` (mark it as critical fixes only). So users are not surprised by seeing different logs and aware they need to update their configuration if they wanna use the new `PodLifeTime`'s parameters.
> 
> Resp. do you still see any benefit maintaining `RemoveFailedPods`?

Hey @ingvagabund  - That can certainly be a follow-up. but keep in mind that RemoveFailedPods _only_ proceesses failed pods, but PodLifeTime uses `reasons` which does an OR on status phase and container conditions. Hence, why it's passing in its own PodFilter.

a7i:
I can go ahead and remove the commit that refactors RemoveFailedPods to use PodLifeTime to keep this PR moving. The intention was to keep the codebase DRY.

ingvagabund:
I'd rather focus this PR on one intention. New parameters/functionality. If there's a code duplication (or will get created) it's better to first refactor the code so it's easy to follow the changes.

README.md
a7i: not the case with any other strategy in README, but will do

ingvagabund: up to you, it could be done for all in a separate PR. Just nice to have in here.

pkg/framework/plugins/podlifetime/pod_lifetime.go
```diff
@@ -69,50 +112,78 @@ func New(ctx context.Context, args runtime.Object, handle frameworktypes.Handle)
...
 	}
 
-	podFilter = podutil.WrapFilterFuncs(podFilter, func(pod *v1.Pod) bool {
-		podAgeSeconds := int(metav1.Now().Sub(pod.GetCreationTimestamp().Local()).Seconds())
-		return podAgeSeconds > int(*podLifeTimeArgs.MaxPodLifeTimeSeconds)
-	})
+	if podLifeTimeArgs.MaxPodLifeTimeSeconds != nil {
+		podFilter = podutil.WrapFilterFuncs(podFilter, func(pod *v1.Pod) bool {
+			podAge := metav1.Now().Sub(pod.GetCreationTimestamp().Local())
+			if podAge < 0 {
```
a7i, line 75: this is odd, but copilot requested a change for code that I did not touch

pkg/framework/plugins/podlifetime/pod_lifetime.go
```diff
@@ -26,30 +26,74 @@ import (
...
+	return newPlugin(ctx, podLifeTimeArgs, handle)
+}
+
+// Name retrieves the plugin name
+func (d *PodLifeTime) Name() string {
+	return PluginName
+}
+
+// Deschedule extension point implementation for the plugin
+func (d *PodLifeTime) Deschedule(ctx context.Context, nodes []*v1.Node) *frameworktypes.Status {
```
ingvagabund: Can you keep the original order of where `Deschedule` was defined? To keep the diff smaller.  Also for others. E.g. `Name()`

README.md
a7i: Done — parameters are now sorted alphabetically.

pkg/framework/plugins/removepodshavingtoomanyrestarts/toomanyrestarts.go
a7i: Good points. Split into `HasMatchingContainerWaitingState` and `HasMatchingContainerTerminatedState` in podutil. `TooManyRestarts` now only uses the waiting helper, and `PodLifeTime` composes both. Removed `MatchesPodState` entirely so each plugin controls exactly which checks it needs.

pkg/framework/plugins/podlifetime/pod_lifetime.go
```diff
@@ -69,50 +68,78 @@ func New(ctx context.Context, args runtime.Object, handle frameworktypes.Handle)
...
-			}
-
-			// Ephemeral Container Status Reason
-			if podLifeTimeArgs.IncludingEphemeralContainers {
-				for _, containerStatus := range pod.Status.EphemeralContainerStatuses {
-					if containerStatus.State.Waiting != nil && states.Has(containerStatus.State.Waiting.Reason) {
-						return true
+				return false
+			})
+		}
```
ingvagabund: `} else if` just in case the validation breaks

pkg/framework/plugins/podlifetime/pod_lifetime.go
```diff
@@ -170,3 +191,60 @@ loop:
...
+	if filter.Type != "" && string(cond.Type) != filter.Type {
+		return false
+	}
+	if filter.Status != "" && string(cond.Status) != filter.Status {
+		return false
+	}
+	if filter.Reason != "" && cond.Reason != filter.Reason {
+		return false
+	}
+	return true
```
ingvagabund: The case where all fields are unset is validated. So in case someone invokes this without the validation a condition with all fields unset matches every pod? Maybe revert the logic?

pkg/framework/plugins/podlifetime/pod_lifetime_test.go
```diff
@@ -804,3 +819,534 @@ func TestPodLifeTime_PodPhaseStates(t *testing.T) {
...
+			pods: []*v1.Pod{
+				buildPod("p1", "node1", func(p *v1.Pod) {
+					p.Status.Phase = v1.PodFailed
+					p.ObjectMeta.OwnerReferences = []metav1.OwnerReference{{Kind: "Job", Name: "job1", APIVersion: "batch/v1"}}
+				}),
+				buildPod("p2", "node1", func(p *v1.Pod) {
+					p.Status.Phase = v1.PodFailed
+				}),
+			},
+			expectedEvictedPodCount: 1,
```
ingvagabund, line 1089: This should not evict a pod if a job is excluded.

README.md
```diff
@@ -785,30 +785,52 @@ profiles:
...
+apiVersion: "descheduler/v1alpha2"
+kind: "DeschedulerPolicy"
+profiles:
+  - name: ProfileName
+    pluginConfig:
+    - name: "PodLifeTime"
+      args:
+        states:
+        - "Succeeded"
+        conditions:
```
ingvagabund, line 820: `orConditions` maybe? To make it explicit? or `disjointConditions`?

a7i: this would deviate away from other plugins that use reasons. I think we can update the docs to say it's ORed

pkg/framework/plugins/podlifetime/pod_lifetime_test.go
a7i: isn't p2 a non-Job pod? hence why it's expecting p2 pod to be evicted

pkg/framework/plugins/podlifetime/pod_lifetime.go
a7i: Technically only one need to be set so this condition check is still needed

a7i: validation:
```
args.MaxPodLifeTimeSeconds != nil ||
		len(args.States) > 0 ||
		len(args.Conditions) > 0 ||
		len(args.ExitCodes) > 0
```

pkg/framework/plugins/podlifetime/pod_lifetime_test.go
a7i: I added a separate test to focus on a Job not getting evicted, but this test-case is valid and the assertion is correct

README.md
a7i: README has been updated accordingly (and matches `reasons` field logic)

PR #1856, opened 2026-04-08
evictions: fix missing observability for background evictions

tiraboschi:
## Description
Background evictions were completely invisible in metrics: the `ignore=true` path caused `EvictPod` to return before incrementing any counter, leaving operators with no signal that a background eviction had been triggered or completed.

Add a "background" result label emitted at eviction request time and a "success" label emitted from the informer `DeleteFunc` when the pod is actually gone. The two labels together give a complete picture: "background" is always recorded (surviving descheduler pod restarts), while "success" confirms the eviction completed within the same lifecycle.

## Checklist
completeness of your changes:

- [X] **Code Readability**: Is the code easy to understand, well-structured, and consistent with project conventions?
- [X] **Naming Conventions**: Are variable, function, and structs descriptive and consistent?
- [X] **Code Duplication**: Is there any repeated code that should be refactored?
- [X] **Function/Method Size**: Are functions/methods short and focused on a single task?
- [X] **Comments & Documentation**: Are comments clear, useful, and not excessive? Were comments updated where necessary?
- [X] **Error Handling**: Are errors handled appropriately ?
- [X] **Testing**: Are there sufficient unit/integration tests?
- [X] **Performance**: Are there any obvious performance issues or unnecessary computations?
- [X] **Dependencies**: Are new dependencies justified ?
- [X] **Logging & Monitoring**: Is logging used appropriately (not too verbose, not too silent)?
- [X] **Backward Compatibility**: Does this change break any existing functionality or APIs?
- [X] **Resource Management**: Are resources (files, connections, memory) managed and released properly?
- [X] **PR Description**: Is the PR description clear, providing enough context and explaining the motivation for the change?
- [X] **Documentation & Changelog**: Are README and docs updated if necessary?

### DIFF
--- a/pkg/descheduler/evictions/evictions.go
+++ b/pkg/descheduler/evictions/evictions.go
@@ -57,6 +57,7 @@ var (
 
 type evictionRequestItem struct {
 	podName, podNamespace, podNodeName string
+	strategyName, profileName          string
 	evictionAssumed                    bool
 	assumedTimestamp                   metav1.Time
 }
@@ -149,17 +150,20 @@ func (erc *evictionRequestsCache) addPod(pod *v1.Pod) {
 	erc.requestsTotal++
 }
 
-func (erc *evictionRequestsCache) assumePod(pod *v1.Pod) {
+func (erc *evictionRequestsCache) assumePod(pod *v1.Pod, profileName, strategyName string) {
 	erc.mu.Lock()
 	defer erc.mu.Unlock()
 	uid := getPodKey(pod)
 	if _, exists := erc.requests[uid]; exists {
+		// Pod already assumed by a previous strategy/profile; first one wins.
 		return
 	}
 	erc.requests[uid] = evictionRequestItem{
 		podNamespace:     pod.Namespace,
 		podName:          pod.Name,
 		podNodeName:      pod.Spec.NodeName,
+		strategyName:     strategyName,
+		profileName:      profileName,
 		evictionAssumed:  true,
 		assumedTimestamp: metav1.NewTime(time.Now()),
 	}
@@ -189,6 +193,14 @@ func (erc *evictionRequestsCache) deletePod(pod *v1.Pod) {
 	if _, exists := erc.requests[uid]; exists {
 		erc.deleteItem(uid)
 	}
+}
+
+func (erc *evictionRequestsCache) getPod(pod *v1.Pod) (evictionRequestItem, bool) {
+	erc.mu.RLock()
+	defer erc.mu.RUnlock()
+	uid := getPodKey(pod)
+	item, exists := erc.requests[uid]
+	return item, exists
 }
 
 func (erc *evictionRequestsCache) hasPod(pod *v1.Pod) bool {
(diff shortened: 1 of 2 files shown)


pkg/descheduler/evictions/evictions.go
```diff
@@ -608,7 +625,11 @@ func (pe *PodEvictor) evictPod(ctx context.Context, pod *v1.Pod, opts EvictOptio
 					return true, nil
 				}
 				klog.V(3).InfoS("Eviction in background assumed", "pod", klog.KObj(pod))
-				pe.erCache.assumePod(pod)
+				pe.erCache.assumePod(pod, opts.StrategyName, opts.ProfileName)
+				if pe.metricsEnabled {
+					metrics.PodsEvicted.With(map[string]string{"result": "background", "strategy": opts.StrategyName, "namespace": pod.Namespace, "node": pod.Spec.NodeName, "profile": opts.ProfileName}).Inc()
```
ingvagabund, line 629: This part is tricky as the pod might not end up getting evicted at the end. So unless a pod is evicted the counters should not incremented.

pkg/descheduler/evictions/evictions.go
```diff
@@ -347,10 +361,13 @@ func NewPodEvictor(
 						klog.ErrorS(nil, "Cannot convert to *v1.Pod", "obj", t)
 						return
 					}
-					if erCache.hasPod(pod) {
+					if item, exists := erCache.deleteAndGet(pod); exists {
```
ingvagabund: What about to add `getPod` method instead? 
```go
if item, exists := erCache.getPod(pod); exists {
	...
}
```

The pod will get deleted anyway. If it no longer exists after `getPod` succeeds it's a no-op. Otherwise, the pod will get deleted from the cache.

pkg/descheduler/evictions/evictions.go
```diff
@@ -160,6 +161,8 @@ func (erc *evictionRequestsCache) assumePod(pod *v1.Pod) {
 		podNamespace:     pod.Namespace,
 		podName:          pod.Name,
 		podNodeName:      pod.Spec.NodeName,
+		strategyName:     strategyName,
```
ingvagabund, line 165: Note: a pod might get assumed by multiple strategies/profiles so the first wins here

pkg/descheduler/evictions/evictions.go
```diff
@@ -149,7 +150,7 @@ func (erc *evictionRequestsCache) addPod(pod *v1.Pod) {
 	erc.requestsTotal++
 }
 
-func (erc *evictionRequestsCache) assumePod(pod *v1.Pod) {
+func (erc *evictionRequestsCache) assumePod(pod *v1.Pod, strategyName, profileName string) {
```
ingvagabund: nit: it's more intuitive to first set profile, then strategy.

pkg/descheduler/evictions/evictions.go
tiraboschi: The idea is to use `"result": "background"` label to signal that a background eviction got initiated.
The same metric with label `"result": "success"` will only be incremented when the pod that was going to be evicted in background is removed from the cache (assuming that the descheduler instance is still alive at that time).
Tracking the initiated background evictions (while "success" tracks confirmed ones) can be useful in terms of observability on its own: for instance with the two values the user can compute the background vs simple eviction rate or the background eviction completion rate.

pkg/descheduler/evictions/evictions.go
tiraboschi: done

pkg/descheduler/evictions/evictions.go
tiraboschi: Right, I added a comment above

pkg/descheduler/evictions/evictions.go
tiraboschi: Good suggestion, done.

pkg/descheduler/evictions/evictions.go
ingvagabund: I missed the "background" value :) All good then.

PR #1857, opened 2026-04-10
feat: add HTTP trigger API for on-demand descheduling cycles

### LINKED ISSUE #1855: Proposal: Add http handler for manual descheduling process launch
During testing and in some situations, we'd like to be able to initiate the descheduling process manually, for example, via an API request. Is this currently possible?

If this functionality doesn't exist, we're open to developing it. Would it be useful to you?

ergoz:
### Problem

The descheduler today supports two operational modes if runs in deployment mode:

1. **One-shot** (`--descheduling-interval=0`): runs a single eviction pass at startup and exits.
2. **Continuous** (`--descheduling-interval=5m`): runs passes on a fixed time interval.

Neither mode gives operators any way to **manually trigger** an eviction cycle without restarting the process or waiting for the next scheduled tick. This creates friction in several real-world scenarios:

- **Post-incident remediation**: a node drain or topology imbalance just occurred and you need eviction to run *right now*, not in 4 minutes when the timer fires.
- **CI/CD pipelines**: a deployment job wants to explicitly rebalance pods after a rollout and needs to know when descheduling is complete before proceeding.
- **Debugging & dry-run validation**: an SRE wants to trigger a dry-run pass interactively, inspect logs, and iterate — without restarting the descheduler pod.
- **Long interval + emergency**: the cluster runs descheduler with a 30-minute interval to reduce churn, but an on-call engineer needs to act immediately.

There was no escape hatch. The only workaround was to delete the descheduler pod and let it restart, which is disruptive and causes a gap in leader election.

---

### Solution

This PR introduces an optional HTTP trigger API that allows any authorized caller to **synchronously** initiate a descheduling cycle at any time via a single `curl` command.

The API is disabled by default and must be explicitly opted into with `--enable-trigger-api`, keeping backward compatibility complete.

Closes #1855

---

#### `cmd/descheduler/app/options/options.go`

- Added `EnableTriggerAPI bool` field to `DeschedulerServer` to control the feature.
- Added `TriggerCh chan chan error` — the communication channel between the HTTP handler and the main scheduling loop. Using `chan chan error` allows the handler to receive the cycle result and propagate it synchronously to the HTTP caller.
- Registered the `--enable-trigger-api` CLI flag.

#### `cmd/descheduler/app/server.go`

- When `EnableTriggerAPI` is set, a buffered `chan chan error` is created and `POST /api/v1/descheduler/run` is registered on the existing `PathRecorderMux` (alongside `/metrics` and `/healthz`), so it is served by the same HTTPS server — no new port, no new listener, no new TLS config.
- The handler is **synchronous**: it blocks until the descheduling cycle finishes and returns the result as JSON. The caller gets a definitive success/failure signal.
- Concurrency protection: the trigger channel has a buffer of 1. If a cycle is already queued or running, a subsequent request receives **429 Too Many Requests** immediately instead of blocking indefinitely or spawning a second concurrent cycle.
- Request cancellation is respected: if the HTTP client disconnects, the handler unblocks via `r.Context().Done()` and returns **504 Gateway Timeout**.

#### `pkg/descheduler/descheduler.go`

- Replaced `wait.NonSlidingUntil` with an explicit `select`-based event loop. The loop listens on three channels simultaneously:
  - `ticker.C` — fires at `--descheduling-interval` (existing behavior, unmodified)
  - `rs.TriggerCh` — receives a manual trigger; executes the cycle and writes the result back to the caller's response channel
  - `ctx.Done()` — graceful shutdown
- When `TriggerCh` is `nil` (feature disabled), it is never selected in `select`, so the scheduler behaves exactly as before.
- When `--descheduling-interval=0` and `--enable-trigger-api` is set, the process now stays alive after the initial cycle and responds to trigger calls indefinitely — enabling a **daemon-without-tick** mode useful for purely on-demand operation.

---

### API Reference

| Method | Path | Description |
|--------|------|-------------|
| `POST` | `/api/v1/descheduler/run` | Trigger a descheduling cycle synchronously |

**Success (200)**
```json
{"message": "descheduling cycle completed successfully", "status": "ok"}
```

**Cycle already running (429)**
```json
{"message": "descheduling cycle already in progress or pending", "status": "error"}
```

**Internal error (500)**
```json
{"message": "failed to run descheduler loop: ...", "status": "error"}
```

**Client disconnected (504)**
```json
{"message": "request cancelled or timed out", "status": "error"}
```

---

#### Start descheduler with the trigger API enabled

```bash
descheduler \
  --policy-config-file=/etc/descheduler/policy.yaml \
  --descheduling-interval=30m \
  --enable-trigger-api
```

#### Trigger a cycle immediately from inside the cluster

```bash
kubectl exec -n kube-system deploy/descheduler -- \
  curl -sk -X POST https://localhost:10258/api/v1/descheduler/run
```

#### Trigger via port-forward from a local machine

```bash
kubectl port-forward -n kube-system deploy/descheduler 10258:10258 &
curl -sk -X POST https://localhost:10258/api/v1/descheduler/run
```

#### Use in a CI/CD step after a deployment

```bash
curl -sk --max-time 300 -X POST \
  https://descheduler.kube-system.svc.cluster.local:10258/api/v1/descheduler/run \
  | jq -e '.status == "ok"'
```

#### Daemon-without-tick mode (purely on-demand, no automatic interval)

```bash
descheduler \
  --policy-config-file=/etc/descheduler/policy.yaml \
  --descheduling-interval=0 \
  --enable-trigger-api
# The initial cycle runs at startup.
# The process stays alive and responds only to explicit POST /api/v1/descheduler/run calls.
```

---

### Backward Compatibility

- `--enable-trigger-api` defaults to `false`. Existing deployments are unaffected.
- The main scheduling loop (`select`-based replacement of `wait.NonSlidingUntil`) is functionally equivalent for all existing flag combinations when the trigger API is disabled.
- No new ports, certificates, or RBAC resources are introduced.

---

## Checklist
completeness of your changes:

- [x] **Code Readability**: Is the code easy to understand, well-structured, and consistent with project conventions?
- [x] **Naming Conventions**: Are variable, function, and structs descriptive and consistent?
- [x] **Code Duplication**: Is there any repeated code that should be refactored?
- [x] **Function/Method Size**: Are functions/methods short and focused on a single task?
- [x] **Comments & Documentation**: Are comments clear, useful, and not excessive? Were comments updated where necessary?
- [x] **Error Handling**: Are errors handled appropriately ?
- [x] **Testing**: Are there sufficient unit/integration tests?
- [x] **Performance**: Are there any obvious performance issues or unnecessary computations?
- [x] **Dependencies**: Are new dependencies justified ?
- [x] **Logging & Monitoring**: Is logging used appropriately (not too verbose, not too silent)?
- [x] **Backward Compatibility**: Does this change break any existing functionality or APIs?
- [x] **Resource Management**: Are resources (files, connections, memory) managed and released properly?
- [x] **PR Description**: Is the PR description clear, providing enough context and explaining the motivation for the change?
- [x] **Documentation & Changelog**: Are README and docs updated if necessary?

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,8 @@ kubectl create -f kubernetes/base/configmap.yaml
 kubectl create -f kubernetes/deployment/deployment.yaml
 ```
 
+> When running as a Deployment you can optionally enable the [Trigger API](#trigger-api) to manually trigger descheduling cycles on demand via an HTTP endpoint.
+
 ### Install Using Helm
 
 Starting with release v0.18.0 there is an official helm chart that can be used to install the
@@ -1147,6 +1149,108 @@ To get best results from HA mode some additional configurations might require:
 
 The metrics are served through https://localhost:10258/metrics by default.
 The address and port can be changed by setting `--binding-address` and `--secure-port` flags.
+
+## Trigger API
+
+The descheduler exposes an optional HTTP endpoint that allows any authorized caller to **manually trigger a descheduling cycle on demand** without waiting for the next scheduled interval or restarting the process.
+
+This is useful when:
+- A node drain or topology imbalance just occurred and eviction must run immediately, not in N minutes when the timer fires.
+- A CI/CD pipeline needs to rebalance pods after a rollout and must know the cycle has completed before proceeding.
+- An on-call engineer needs an emergency eviction pass without touching the descheduler deployment.
+- A cluster runs with a long `--descheduling-interval` to minimize churn, but requires an escape hatch for urgent situations.
+
+### Enabling the Trigger API
+
+Pass the `--enable-trigger-api` flag when starting the descheduler:
+
+```bash
+descheduler \
+  --policy-config-file=/etc/descheduler/policy.yaml \
+  --descheduling-interval=30m \
+  --enable-trigger-api
+```
+
+When using Helm, add the flag via `deschedulerCommandArguments`:
+
+```yaml
+deschedulerCommandArguments:
+  - "--enable-trigger-api"
+```
+
+The endpoint is registered on the existing HTTPS server (default port `10258`) — no additional ports, listeners, or TLS configuration is required.
+
+### Endpoint Reference
+
+| Method | Path | Description |
+|--------|------|-------------|
+| `POST` | `/api/v1/descheduler/run` | Synchronously trigger a full descheduling cycle |
+
+The request **blocks until the cycle completes** and returns a JSON response with the result. Only one trigger can be queued at a time; concurrent requests are rejected with `429`.
+
+#### Response codes
+
+| Code | Meaning |
+|------|---------|
+| `200 OK` | Cycle completed successfully |
+| `405 Method Not Allowed` | Non-POST request |
+| `429 Too Many Requests` | A trigger is already queued or a cycle is in progress |
+| `500 Internal Server Error` | Cycle failed; error detail in `message` field |
+| `504 Gateway Timeout` | HTTP client disconnected before the cycle finished |
+
+**Example success response:**
+```json
+{"message": "descheduling cycle completed successfully", "status": "ok"}
+```
+
+**Example rejection response (cycle already running):**
+```json
+{"message": "descheduling cycle already in progress or pending", "status": "error"}
+```
+
+### Usage Examples
+
+**Trigger from inside the cluster (exec into the descheduler pod):**
+
+```bash
+kubectl exec -n kube-system deploy/descheduler -- \
+  curl -sk -X POST https://localhost:10258/api/v1/descheduler/run
+```
+
+**Trigger via port-forward from a local machine:**
+
+```bash
+kubectl port-forward -n kube-system deploy/descheduler 10258:10258 &
+curl -sk -X POST https://localhost:10258/api/v1/descheduler/run
+```
+
+**Trigger with a timeout (useful in scripts):**
+
+```bash
+curl -sk --max-time 300 -X POST \
+  https://localhost:10258/api/v1/descheduler/run | jq .
+```
+
+**Use in a CI/CD step and assert success:**
+
+```bash
+curl -sk --max-time 300 -f -X POST \
+  https://localhost:10258/api/v1/descheduler/run \
+  | jq -e '.status == "ok"'
+```
+
+### On-Demand-Only Mode (no automatic interval)
+
+Setting `--descheduling-interval=0` together with `--enable-trigger-api` runs one cycle at startup and then keeps the process alive, responding exclusively to manual trigger requests. This is useful for environments where eviction should only happen on explicit operator request:
+
+```bash
+descheduler \
+  --policy-config-file=/etc/descheduler/policy.yaml \
+  --descheduling-interval=0 \
+  --enable-trigger-api
+```
+
+> **Note:** The trigger API is disabled by default. Existing deployments without `--enable-trigger-api` are completely unaffected.
 
 ## Compatibility Matrix
 The below compatibility matrix shows the k8s client package(client-go, apimachinery, etc) versions that descheduler
--- a/README.md
+++ b/README.md
@@ -1155,10 +1155,11 @@ The address and port can be changed by setting `--binding-address` and `--secure
 The descheduler exposes an optional HTTP endpoint that allows any authorized caller to **manually trigger a descheduling cycle on demand** without waiting for the next scheduled interval or restarting the process.
 
 This is useful when:
-- A node drain or topology imbalance just occurred and eviction must run immediately, not in N minutes when the timer fires.
-- A CI/CD pipeline needs to rebalance pods after a rollout and must know the cycle has completed before proceeding.
-- An on-call engineer needs an emergency eviction pass without touching the descheduler deployment.
-- A cluster runs with a long `--descheduling-interval` to minimize churn, but requires an escape hatch for urgent situations.
+
+* A node drain or topology imbalance just occurred and eviction must run immediately, not in N minutes when the timer fires.
+* A CI/CD pipeline needs to request a rebalance after a rollout and know the cycle was accepted.
+* An on-call engineer needs an emergency eviction pass without touching the descheduler deployment.
+* A cluster runs with a long `--descheduling-interval` to minimize churn, but requires an escape hatch for urgent situations.
 
 ### Enabling the Trigger API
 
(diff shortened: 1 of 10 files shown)


ergoz:
@googs1025 @JaneLiuL Hello! I hope you're all doing well. I wanted to gently follow up on the pull request I submitted a little while ago — I'd really appreciate it if someone could take a look when they have a free moment.

No rush at all, just wanted to make sure it didn't get lost in the shuffle! Happy to answer any questions or make changes based on your feedback.

Thanks so much in advance! 🙏

ergoz:
@googs1025 @JaneLiuL Hello! Kindly remind :) we really need this feature :)

googs1025:
will check this pr tomorrow

ergoz:
@googs1025 great news! Thank you very much!

pkg/descheduler/descheduler.go
```diff
@@ -532,23 +532,44 @@ func RunDeschedulerStrategies(ctx context.Context, rs *options.DeschedulerServer
...
+	if rs.DeschedulingInterval.Seconds() > 0 {
+		ticker := time.NewTicker(rs.DeschedulingInterval)
+		defer ticker.Stop()
+		tickerC = ticker.C
+	}
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
```
googs1025, line 581:
```go
case <-ctx.Done():                                                                                                                                                                    
      // Drain any pending trigger so the caller doesn't hang.                                                                                                                          
      select {                                                                                                                                                                          
      case resultCh := <-rs.TriggerCh:                                                                                                                                                
          resultCh <- fmt.Errorf("descheduler shutting down")                                                                                                                           
      default:                                                                                                                                                                          
      }
      return nil  
```

edge case: if a request is sitting in TriggerCh (buffer 1) when shutdown fires, this returns and the handler's <-resultCh blocks until the request context is torn down — which usually shows up as a 504 to the caller during a graceful shutdown. Maybe drain it on the way out?    🤔

googs1025:
Thanks for the detailed write-up 
                                                                                                                                                                                                                                                                                                                                           
 A few things I'd want to see before this lands, though:                                                                                                                                                                                                                                                                                                            
1. The endpoint mutates cluster state but doesn't go through any auth filter — it's mounted on the same mux as /metrics and /healthz, which are read-only by convention. Anyone with network reachability to :10258 can trigger an eviction pass. I think we need either an auth filter (SAR/bearer token, like other K8s components do) or at minimum a loopback-only bind so this is kubectl exec-only.                                                                                                                                                  
2. No tests in the diff — the main loop rewrite (wait.NonSlidingUntil → custom select) really needs coverage for the four flag combinations (interval 0/+, trigger on/off).

README.md
```diff
@@ -1148,6 +1150,108 @@ To get best results from HA mode some additional configurations might require:
 The metrics are served through https://localhost:10258/metrics by default.
 The address and port can be changed by setting `--binding-address` and `--secure-port` flags.
 
+## Trigger API
+
+The descheduler exposes an optional HTTP endpoint that allows any authorized caller to **manually trigger a descheduling cycle on demand** without waiting for the next scheduled interval or restarting the process.
```
ricardomaraschini, line 1155: What sort of authorization is done ? i.e. how does this code distinguishes between authorized and unauthorized callers ?

ricardomaraschini (changes requested):

pkg/descheduler/descheduler.go
ergoz: Oh, missed this case :) i will fix it!

ergoz:
> Thanks for the detailed write-up
> 
> A few things I'd want to see before this lands, though:
> 
> 1. The endpoint mutates cluster state but doesn't go through any auth filter — it's mounted on the same mux as /metrics and /healthz, which are read-only by convention. Anyone with network reachability to :10258 can trigger an eviction pass. I think we need either an auth filter (SAR/bearer token, like other K8s components do) or at minimum a loopback-only bind so this is kubectl exec-only.
> 2. No tests in the diff — the main loop rewrite (wait.NonSlidingUntil → custom select) really needs coverage for the four flag combinations (interval 0/+, trigger on/off).

1. Yes, you are right! I will add auth like in k8s. But should i add another port listener / mux / something else to split it with /healthz and /metrics? or auth is enough?
2. Ok, will be added :)

README.md
ergoz: The meaning here is ambiguous; I meant that only those with access to the cluster and port can run it. But as @googs1025 asked, I'll add authorization. :)

ingvagabund:
In the past we discussed the option of running the descheduler async based on e.g. node condition changes. E.g. everytime a node goes unready for some time or nodes scale down/up. Observing cluster objects via informers and custom indexers.

> real-world scenarios

Can you please more elaborate on how this would work from the user perspective? A single endpoint for every user with a single role? Or, is there a possibility of creating multiple roles? E.g assigning users permissions for individual actions?

ergoz:
Code reworked and has some new logic and features:

A descheduling cycle has been started:
- It no longer blocks until the cycle completes.
- If any descheduling cycle is already running, the endpoint returns `409 Conflict`.
- Startup, scheduled, and manually triggered cycles now all go through the same single-flight cycle manager, so they cannot overlap.

I also added Kubernetes-style authentication and authorization:

- The caller must send a service account bearer token.
- The token is authenticated with `TokenReview`.
- The authenticated user is authorized with `SubjectAccessReview`.
- The authorization check is for the non-resource URL `/api/v1/descheduler/run` with verb `post`.

### What this gives us

This addresses the main security concern from the review: the endpoint mutates cluster state, so network reachability alone is not enough.

Now access is controlled by Kubernetes RBAC. Operators can decide exactly which users or service accounts are allowed to trigger descheduler manually, without exposing the endpoint to every caller that can reach port `10258`.

The endpoint also no longer has the shutdown edge case from the previous `chan chan error` design. Since the HTTP request only waits until the cycle is accepted, it does not hang waiting for cycle completion during graceful shutdown.

### User perspective and roles

From the user perspective there is a single action: manually request a descheduling cycle. 

That action is represented as:

```yaml
nonResourceURLs:
  - /api/v1/descheduler/run
verbs:
  - post
```

Example role for callers:

```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: descheduler-trigger
rules:
- nonResourceURLs:
  - /api/v1/descheduler/run
  verbs:
  - post
```

Different users or service accounts can be granted access with normal `RoleBinding` / `ClusterRoleBinding` objects. Since there is currently only one manual action, there is only one permission to grant. If we add more manual actions later, they can be represented as separate non-resource URLs and granted independently.

### Usage example

A caller with an authorized service account token can trigger a cycle like this:

```bash
TOKEN="$(cat /var/run/secrets/kubernetes.io/serviceaccount/token)"

curl -sk -X POST \
  -H "Authorization: Bearer ${TOKEN}" \
  https://descheduler.kube-system.svc:10258/api/v1/descheduler/run
```

Accepted response:

```json
{"message":"descheduling cycle started","status":"accepted"}
```

If another startup, scheduled, or manual cycle is already running:

```json
{"message":"descheduling cycle already running","status":"error"}
```

with HTTP status `409 Conflict`.

### Tests added

I added coverage for:

* the single-flight cycle manager
* rejecting concurrent cycles
* HTTP `202`, `409`, `401`, `403`, and `405` behavior
* TokenReview / SubjectAccessReview authorization flow
* all four `--descheduling-interval` / `--enable-trigger-api` combinations:
  * interval `0`, trigger API disabled
  * interval `0`, trigger API enabled
  * interval `>0`, trigger API disabled
  * interval `>0`, trigger API enabled

All tests in project are success (only unit tests, without e2e)

```
go test $(go list ./... | rg -v '/test/e2e$')
```

Result: `Go test: 889 passed in 42 packages`

ergoz:
Hello! I made some changes. Could you review them, please?

Issue #1478, opened 2024-08-04
Migrate all e2es to run descheduler as an image


ingvagabund:
With https://github.com/kubernetes-sigs/descheduler/pull/1474 we can run the descheduler through a pod. End-to-end testing is not intended to run the descheduler code directly, as this does not accurately reflect its operation in production.
Each e2e test can be refactored separately. https://github.com/kubernetes-sigs/descheduler/pull/1472 is an example of the first step running the descheduler as a whole. While https://github.com/kubernetes-sigs/descheduler/pull/1474 gives idea on how to run it as a deployment.

The following e2es need to be migrated:
- [ ] TestRemoveDuplicates
- [ ] TestFailedPods
- [ ] TestLeaderElection
- [ ] TestTopologySpreadConstraint
- [ ] TestClientConnectionConfiguration

ingvagabund:
@jklaw90 @JaneLiuL @fanhaouu @pravarag @eminaktas @zhifei92 @googs1025 in case you would like to contribute

fanhaouu:
> @jklaw90 @JaneLiuL @fanhaouu @pravarag @eminaktas @zhifei92 @googs1025 in case you would like to contribute

Not sure if there is an ETA, I can make some contributions. @ingvagabund

ingvagabund:
There's no ETA. Also, there's no guarantee the issue will stay unresolved for long. I will eventually migrate all the e2es to simplify https://github.com/kubernetes-sigs/descheduler/pull/1466.

fanhaouu:
> There's no ETA. Also, there's no guarantee the issue will stay unresolved for long. I will eventually migrate all the e2es to simplify #1466.

ok

hsunwenfang:
Hi @ingvagabund 
I would like to contribute, thx!

ingvagabund:
@hsunwenfang thank you. Any of the listed items is currently for taking.

fanhaouu:
I have completed some E2E tests, but encountered some issues when running the E2E tests locally. I am still debugging and working on resolving them.
@ingvagabund @ingvagabund

hsunwenfang:
Hi @ingvagabund 
I'm working on TestLeaderElection and is refering your commit
https://github.com/kubernetes-sigs/descheduler/pull/1472/commits/ab467a5dd29925eb3fd794d2d93965f4a58281fc
[...]

PR #1861, opened 2026-04-25
test/e2e: migrate TestDeschedulingInterval to run descheduler as an image

Sanil2108:
## Summary

Per #1478, the e2e suite is being moved off in-process descheduler runs and onto deployments of the descheduler image so the tests exercise the real shipped artifact. \`TestDeschedulingInterval\` was one of the two remaining cases (along with \`TestClientConnectionConfiguration\`).

## What this changes

\`TestDeschedulingInterval\` verifies the run-once-and-exit lifecycle when no \`--descheduling-interval\` is set. The previous version called \`descheduler.RunDeschedulerStrategies\` directly, which doesn't reflect the production code path. The new version:

- Builds a one-shot Pod (\`restartPolicy=Never\`, no \`--descheduling-interval\` flag, no liveness probe — irrelevant for a one-shot run) via a new \`newDeschedulerOneShotPod\` helper.
- Mounts an empty \`DeschedulerPolicy\` ConfigMap (we're testing the lifecycle, not any specific descheduling decision).
- Asserts the Pod reaches \`Succeeded\` phase within 3 minutes — which is what \`pkg/descheduler/descheduler.go:546\` (\`if rs.DeschedulingInterval.Seconds() == 0 { cancel() }\`) guarantees.

The helper derives the Pod from the existing \`deschedulerDeployment\` template so the container image, security context, volume mounts, and SA stay in lock-step with the long-running tests.

## Removed

The old in-process \`TestDeschedulingInterval\` in \`test/e2e/e2e_test.go\` and the now-unused \`options\` and \`pkg/descheduler\` imports it pulled in.

## Out of scope

\`TestClientConnectionConfiguration\` is intentionally not migrated here — the maintainer noted it's \"tricky\" because the descheduler doesn't currently expose the custom client configuration. Happy to take that as a follow-up if you'd like to scope it.

## Test plan

- [x] \`go build -tags e2e ./test/e2e/...\` clean
- [x] \`go vet -tags e2e ./test/e2e/...\` clean
- [ ] Full e2e run on a real cluster (will rely on CI)

Refs #1478

### DIFF
--- a/test/e2e/e2e_deschedulinginterval_test.go
+++ b/test/e2e/e2e_deschedulinginterval_test.go
@@ -1,0 +1,141 @@
+/*
+Copyright 2024 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package e2e
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"testing"
+	"time"
+
+	batchv1 "k8s.io/api/batch/v1"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/wait"
+	clientset "k8s.io/client-go/kubernetes"
+	componentbaseconfig "k8s.io/component-base/config"
+
+	apiv1alpha2 "sigs.k8s.io/descheduler/pkg/api/v1alpha2"
+	"sigs.k8s.io/descheduler/pkg/descheduler/client"
+)
+
+// TestDeschedulingInterval verifies that with `--descheduling-interval` set to
+// 0 (the default), the descheduler binary runs a single pass and then exits
+// cleanly.
+func TestDeschedulingInterval(t *testing.T) {
+	ctx := context.Background()
+
+	clientSet, err := client.CreateClient(componentbaseconfig.ClientConnectionConfiguration{Kubeconfig: os.Getenv("KUBECONFIG")}, "")
+	if err != nil {
+		t.Fatalf("Error during kubernetes client creation with %v", err)
+	}
+
+	deschedulerPolicyConfigMapObj, err := deschedulerPolicyConfigMap(&apiv1alpha2.DeschedulerPolicy{})
+	if err != nil {
+		t.Fatalf("Error building descheduler policy configmap: %v", err)
+	}
+
+	t.Logf("Creating %q policy CM ...", deschedulerPolicyConfigMapObj.Name)
+	if _, err := clientSet.CoreV1().ConfigMaps(deschedulerPolicyConfigMapObj.Namespace).Create(ctx, deschedulerPolicyConfigMapObj, metav1.CreateOptions{}); err != nil {
+		t.Fatalf("Error creating %q CM: %v", deschedulerPolicyConfigMapObj.Name, err)
+	}
+	defer func() {
+		t.Logf("Deleting %q CM...", deschedulerPolicyConfigMapObj.Name)
+		if err := clientSet.CoreV1().ConfigMaps(deschedulerPolicyConfigMapObj.Namespace).Delete(ctx, deschedulerPolicyConfigMapObj.Name, metav1.DeleteOptions{}); err != nil {
+			t.Fatalf("Unable to delete %q CM: %v", deschedulerPolicyConfigMapObj.Name, err)
+		}
+	}()
+
+	job := deschedulerRunOnceJob("descheduling-interval")
+	jobClient := clientSet.BatchV1().Jobs(job.Namespace)
+
+	t.Logf("Creating a run-once descheduler job %q", job.Name)
+	if _, err := jobClient.Create(ctx, job, metav1.CreateOptions{}); err != nil {
+		t.Fatalf("Error creating descheduler job %q: %v", job.Name, err)
+	}
+	defer func() {
+		printJobPodLogs(ctx, t, clientSet, job)
+		t.Logf("Deleting descheduler job %q...", job.Name)
+		deletePropagationPolicy := metav1.DeletePropagationForeground
+		if err := jobClient.Delete(ctx, job.Name, metav1.DeleteOptions{PropagationPolicy: &deletePropagationPolicy}); err != nil {
+			t.Fatalf("Unable to delete descheduler job %q: %v", job.Name, err)
+		}
+	}()
+
+	t.Logf("Waiting for descheduler job %q to complete", job.Name)
+	if err := wait.PollUntilContextTimeout(ctx, 5*time.Second, 3*time.Minute, true, func(ctx context.Context) (bool, error) {
+		current, err := jobClient.Get(ctx, job.Name, metav1.GetOptions{})
+		if err != nil {
+			return false, err
+		}
+		for _, condition := range current.Status.Conditions {
+			if condition.Status != v1.ConditionTrue {
+				continue
+			}
+			switch condition.Type {
+			case batchv1.JobComplete:
+				return true, nil
+			case batchv1.JobFailed:
+				return false, fmt.Errorf("job %q failed: %v: %v", job.Name, condition.Reason, condition.Message)
+			}
+		}
+		t.Logf("Job %q has not completed yet, waiting...", job.Name)
+		return false, nil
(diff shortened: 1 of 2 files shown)


test/e2e/e2e_deschedulinginterval_test.go
```diff
@@ -0,0 +1,144 @@
...
+// by deschedulerDeployment. Differences from the Deployment helper:
+//
+//   - RestartPolicy is Never (Pods default to Always otherwise).
+//   - The `--descheduling-interval=100m` argument is dropped so the descheduler
+//     follows the run-once-and-exit code path.
+//   - No liveness probe — a one-shot run isn't long enough to make probing
+//     meaningful, and a slow first probe could race the clean exit.
+func newDeschedulerOneShotPod(testName, policyConfigMapName string) *v1.Pod {
+	deploymentTemplate := deschedulerDeployment(testName)
+	podTemplate := deploymentTemplate.Spec.Template
```
ingvagabund: Used only once, can be inlined

test/e2e/e2e_deschedulinginterval_test.go
```diff
@@ -0,0 +1,144 @@
...
+
+	podSpec := podTemplate.Spec
```
ingvagabund: This variable can be inlined as well. Even though this makes the path a bit longer keeping it makes the code easier to read.

test/e2e/e2e_deschedulinginterval_test.go
```diff
@@ -0,0 +1,144 @@
...
+	podSpec.RestartPolicy = v1.RestartPolicyNever
+
+	container := podSpec.Containers[0]
```
ingvagabund: The same here

test/e2e/e2e_deschedulinginterval_test.go
```diff
@@ -0,0 +1,144 @@
...
+	ctx := context.Background()
+
+	clientSet, err := client.CreateClient(componentbaseconfig.ClientConnectionConfiguration{Kubeconfig: os.Getenv("KUBECONFIG")}, "")
+	if err != nil {
+		t.Errorf("Error during kubernetes client creation with %v", err)
+	}
+
+	// Empty policy is fine — we are testing the run-once-and-exit lifecycle, not
+	// any specific descheduling decision.
+	deschedulerPolicyConfigMapObj, err := deschedulerPolicyConfigMap(&apiv1alpha2.DeschedulerPolicy{})
```
ingvagabund, line 49: `deschedulerPolicyConfigMapObj`'s  `.Name`, resp. `.Namespace` are not changed. `newDeschedulerOneShotPod` can be simplified to keep the original CM name.

test/e2e/e2e_deschedulinginterval_test.go
```diff
@@ -0,0 +1,144 @@
...
+// newDeschedulerOneShotPod builds a Pod that runs the descheduler image once
+// and exits, mirroring the container/security/volume configuration produced
+// by deschedulerDeployment. Differences from the Deployment helper:
+//
+//   - RestartPolicy is Never (Pods default to Always otherwise).
+//   - The `--descheduling-interval=100m` argument is dropped so the descheduler
+//     follows the run-once-and-exit code path.
+//   - No liveness probe — a one-shot run isn't long enough to make probing
+//     meaningful, and a slow first probe could race the clean exit.
+func newDeschedulerOneShotPod(testName, policyConfigMapName string) *v1.Pod {
```
ingvagabund: Alternatively, you could define a new `deschedulerPodSpec` from `deschedulerDeployment` and have `deschedulerDeployment` invoke it internally. Then consume it here as well to avoid creating the whole deployment and dropping everything but the pod template.

test/e2e/e2e_deschedulinginterval_test.go
```diff
@@ -0,0 +1,144 @@
...
+	deploymentTemplate := deschedulerDeployment(testName)
+	podTemplate := deploymentTemplate.Spec.Template
+
+	podSpec := podTemplate.Spec
+	podSpec.RestartPolicy = v1.RestartPolicyNever
+
+	container := podSpec.Containers[0]
+	container.Args = []string{
+		"--policy-config-file", "/policy-dir/policy.yaml",
```
ingvagabund: Instead of resetting the args you could have `deschedulerPodSpec` take the descheduling interval value (set to 0 in this case). To avoid exposing the actual arguments and keeping only a single place where the binary args are set.

There's already a test under e2e_evictioninbackground_test.go that resets the args. So you are not breaking any convention here. Yet, if there's a way to avoid exposing the arguments/implementation details it's still preferable.

test/e2e/e2e_deschedulinginterval_test.go
```diff
@@ -0,0 +1,144 @@
...
+	container.LivenessProbe = nil
+	podSpec.Containers[0] = container
+
+	for i, vol := range podSpec.Volumes {
+		if vol.ConfigMap != nil && vol.ConfigMap.LocalObjectReference.Name == "descheduler-policy-configmap" {
+			podSpec.Volumes[i].ConfigMap.LocalObjectReference.Name = policyConfigMapName
+		}
+	}
+
+	return &v1.Pod{
```
ingvagabund: Have you considered running it as a Job? To avoid the cases where a pod gets evicted in-flight before it finishes? E.g. due to resource constraints. Creating a job can help with re-creating the pod in these cases. To avoid some of the potential flakes.

Sanil2108:
Thanks for the thorough review, @ingvagabund! Pushed a follow-up that addresses the feedback:

- Removed the `policyConfigMapName` parameter from `newDeschedulerOneShotPod` — `deschedulerPolicyConfigMap` always returns a CM named `"descheduler-policy-configmap"`, which is exactly what `deschedulerDeployment` already mounts, so the volume-name replacement loop was a no-op.
- Inlined the intermediate `deploymentTemplate` and `podTemplate` variables, each used only once to reach the next field in the chain.

On the two alternative suggestions:
- **`deschedulerPodSpec` helper**: happy to go that route if you'd prefer — would mean touching `e2e_test.go` to extract the pod spec from `deschedulerDeployment`. Let me know and I'll follow up with that refactor.
- **Job instead of Pod**: noted, and it's a good point about pod eviction. I kept it as a Pod to stay consistent with the existing pattern in `e2e_evictioninbackground_test.go`, but can convert to a Job if you'd prefer the extra resilience.

ingvagabund:
Thank you for keeping the PR in a good shape :)

> deschedulerPodSpec helper: happy to go that route if you'd prefer — would mean touching e2e_test.go to extract the pod spec from deschedulerDeployment. Let me know and I'll follow up with that refactor.

By all means :)

> Job instead of Pod: noted, and it's a good point about pod eviction. I kept it as a Pod to stay consistent with the existing pattern in e2e_evictioninbackground_test.go, but can convert to a Job if you'd prefer the extra resilience.

Please proceed.

Sanil2108:
Apologies for the long delay here. Both suggestions are now in: extracted a deschedulerPodSpec(deschedulingInterval) helper that deschedulerDeployment consumes internally, and converted the one-shot run to a Job (dropped the liveness probe for the run-once case since it races the clean exit).

Allso squashed the branch to fix the CLA check on the earlier commit. Thanks for the patience!

neoLsH:
Hi @Sanil2108, friendly ping — it looks like the last round left a failing `pull-descheduler-test-e2e-k8s-master-1-35` run. Are you still working on this? If you are short on bandwidth I would be happy to help investigate the failure and push it over the line — just let me know. No rush!

Sanil2108:
> Hi @Sanil2108, friendly ping — it looks like the last round left a failing pull-descheduler-test-e2e-k8s-master-1-35 run. Are you still working on this? If you are short on bandwidth I would be happy to help investigate the failure and push it over the line — just let me know. No rush!

@neoLsH I am really sorry for the long silence here, working on it right away and will be more active on this, and will get it through the lines.

Sanil2108:
Rebased onto master and resolved the conflict, CI's running again. Everything from the earlier review is in (the deschedulerPodSpec helper + the run-once Job). @ingvagabund could you take another pass when you get a chance?

Sanil2108:
Seems like an issue with the tests. @ingvagabund should I take a look on why they are breaking?

ingvagabund:
From https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_descheduler/1861/pull-descheduler-test-e2e-k8s-master-1-36/2099772571728744448:
```
+ go test sigs.k8s.io/descheduler/test/e2e/ -v -timeout 0 --args --descheduler-image docker.io/library/descheduler:v20260915-descheduler-helm-chart-0.36.0-37-g90320005 --kubevirt-version-tag v1.9.0 --pod-run-as-user-id 1000 --pod-run-as-group-id 1000
# k8s.io/kube-openapi/pkg/internal/third_party/go-json-experiment/json
vendor/k8s.io/kube-openapi/pkg/internal/third_party/go-json-experiment/json/alias.go:618:21: undefined: json.SkipFunc
vendor/k8s.io/kube-openapi/pkg/internal/third_party/go-json-experiment/json/alias.go:957:14: undefined: json.DiscardUnknownMembers
FAIL	sigs.k8s.io/descheduler/test/e2e [build failed]
```

PR #1864, opened 2026-05-01
fix(helm): remove default cpu resource limits to allow user-defined sizing

### LINKED ISSUE #1788: Helm Chart: make default resources empty so users can omit cpu request/limit
## Description
The [Helm chart](https://github.com/kubernetes-sigs/descheduler/tree/master/charts/descheduler) currently sets default container resources (including CPU request/limit). This makes it impossible to configure only memory (or to "unset" CPU limits) via values overrides, because Helm values merging won't remove keys from the chart's default map.

### Current behaviour [[link](https://github.com/kubernetes-sigs/descheduler/blob/b6e81fdd4b508f3abaf75e53d28c90b69455bf29/charts/descheduler/values.yaml#L17)]
Chart defaults include CPU, e.g.:

```
resources:
  requests:
    cpu: 500m
    memory: 256Mi
  limits:
    cpu: 500m
    memory: 256Mi
```

If a user tries to omit CPU by removing it from their override file, or setting cpu: "", Helm's merge results in CPU being retained from the chart defaults (or producing an invalid quantity).

### Expected behavior
Chart should not force CPU requests/limits by default. Users should decide their own sizing.

### Why this matters
Many clusters rely on defaults (or Vertical Pod Autoscaler, LimitRanges, etc.) and do not want to pin CPU limits for this workload.
With current defaults, you can't express "set memory but don't set cpu" using values alone.

### Proposed change
Change chart values.yaml default from a populated map to an empty one:

```
resources: {}
```

Sanil2108:
## Problem

The chart currently ships with hard-coded CPU and memory defaults for the `resources` field:

```yaml
resources:
  requests:
    cpu: 500m
    memory: 256Mi
  limits:
    cpu: 500m
    memory: 256Mi
```

Because Helm uses a **recursive map merge** for values overrides, individual keys inside a map cannot be removed by a user override. For example, passing `--set resources.limits.cpu=""` or providing:

```yaml
resources:
  limits:
    memory: 512Mi
```

in a custom `values.yaml` will still result in `cpu: 500m` being present (inherited from the chart defaults). There is no way for users to deploy the descheduler *without* CPU limits without forking the chart or using post-render hooks.

This is especially problematic in environments where CPU limits are explicitly discouraged (e.g. clusters following the [No CPU Limits](https://home.robusta.dev/blog/stop-using-cpu-limits) pattern), as well as in resource-constrained setups where the default `500m` CPU request may be too high.

Fixes #1788

## Solution

Change the default value of `resources` from the hard-coded map to an empty map (`{}`):

```yaml
resources: {}
```

This is consistent with how most upstream Helm charts (e.g. cert-manager, ingress-nginx, kube-prometheus-stack) handle resources — ship with no defaults and let operators supply their own sizing.

Users who want the previous behaviour can still pin it explicitly in their `values.yaml`:

```yaml
resources:
  requests:
    cpu: 500m
    memory: 256Mi
  limits:
    cpu: 500m
    memory: 256Mi
```

## Testing

- `helm lint charts/descheduler` passes with no warnings or errors.
- Verified that rendering `resources: {}` in the cronjob/deployment templates produces valid Kubernetes manifests (Kubernetes accepts an empty `resources:` block as "no constraints").

### DIFF
--- a/charts/descheduler/README.md
+++ b/charts/descheduler/README.md
@@ -66,7 +66,7 @@ The following table lists the configurable parameters of the _descheduler_ chart
 | `cmdOptions`                        | The options to pass to the _descheduler_ command                                                                      | _see values.yaml_                         |
 | `priorityClassName`                 | The name of the priority class to add to pods                                                                         | `system-cluster-critical`                 |
 | `rbac.create`                       | If `true`, create & use RBAC resources                                                                                | `true`                                    |
-| `resources`                         | Descheduler container CPU and memory requests/limits                                                                  | _see values.yaml_                         |
+| `resources`                         | Descheduler container CPU and memory requests/limits                                                                  | `{}`                                      |
 | `serviceAccount.create`             | If `true`, create a service account for the cron job                                                                  | `true`                                    |
 | `serviceAccount.name`               | The name of the service account to use, if not set and create is true a name is generated using the fullname template | `nil`                                     |
 | `serviceAccount.annotations`        | Specifies custom annotations for the serviceAccount                                                                   | `{}`                                      |
--- a/charts/descheduler/values.yaml
+++ b/charts/descheduler/values.yaml
@@ -14,13 +14,7 @@ image:
 imagePullSecrets:
 #   - name: container-registry-secret
 
-resources:
-  requests:
-    cpu: 500m
-    memory: 256Mi
-  limits:
-    cpu: 500m
-    memory: 256Mi
+resources: {}
 
 ports:
   - containerPort: 10258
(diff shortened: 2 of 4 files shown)


ricardomaraschini:
@Sanil2108 Would you mind giving some context on why e2e tests were also changed as part of this ?

Edit: I just saw you also opened https://github.com/kubernetes-sigs/descheduler/pull/1861 with what seems to be the same e2e test changes. Can we keep this one restricted to the default resources ?

ricardomaraschini:
@Sanil2108 Using latest helm (v4) I think it is possible to omit these with: 

```
--set resources.requests.cpu=null \
--set resources.requests.memory=null \
--set resources.limits.cpu=null \
--set resources.limits.memory=null
```

PR #1865, opened 2026-05-02
fix(removefailedpods): allow opt-in eviction of failed system-critical pods

### LINKED ISSUE #1775: Failed Pods with system critical priority are note removed by RemoveFailedPods plugin

descheduler version: v0.34.0

yes

            - "--policy-config-file=/policy-dir/policy.yaml"
            - "--v=3"

```YAML
kind: "DeschedulerPolicy"
maxNoOfPodsToEvictPerNamespace: 5
maxNoOfPodsToEvictTotal: 30
profiles:
- name: default
  pluginConfig:
  - args:
      podProtections:
        defaultDisabled:
        - PodsWithLocalStorage
        - FailedBarePods
        extraEnabled: null
    name: DefaultEvictor
  - args:
      includingInitContainers: true
      podRestartThreshold: 20
    name: RemovePodsHavingTooManyRestarts
  - name: RemovePodsViolatingNodeTaints
  - name: RemovePodsViolatingInterPodAntiAffinity
  - args:
      constraints:
      - DoNotSchedule
    name: RemovePodsViolatingTopologySpreadConstraint
  - args:
      nodeAffinityType:
    name: RemovePodsViolatingNodeAffinity
  - args:
      excludeOwnerKinds:
      - Job
      minPodLifetimeSeconds: 120
      reasons:
      - Evicted
      - Shutdown
      - NodeLost
      - NodeAffinity
      - UnexpectedAdmissionError
      - Preempting
    name: RemoveFailedPods
  plugins:
    balance:
      enabled: []
    deschedule:
      enabled:
      - RemovePodsHavingTooManyRestarts
      - RemovePodsViolatingNodeAffinity
      - RemovePodsViolatingInterPodAntiAffinity
      - RemoveFailedPods

```

$ kubectl version
Client Version: v1.33.5
Kustomize Version: v5.6.0
Server Version: v1.33.5-eks-3025e55
</pre></details>

**What did you do?**

`Failed` Pods are removed, also those with with system critical priority.

While not evicting Pods with system critical priority makes sense for eviction plugins, this filter logic might not make sense for the `RemoveFailedPods` pluign. The `RemoveFailedPods` plugin is about `Failed` Pods, which means Pods which are not running anymore and removing `Failed` Pods should not break anything.

**What did you see instead?**

Not all `Failed` Pods are removed. Failed Pods with system critical priority are not removed.

```
I1117 09:24:03.412833       1 defaultevictor.go:356] "Pod fails the following checks" plugin="DefaultEvictor" ExtensionPoint="Filter" pod="kube-system/cert-manager-fb975885b-clf87" checks="[pod has system critical priority and is protected against eviction, pod has higher priority than specified priority class threshold]"
```

I suspect this is caused by https://github.com/kubernetes-sigs/descheduler/blob/v0.34.0/pkg/framework/plugins/removefailedpods/failedpods.go#L63

```Go
	// We can combine Filter and PreEvictionFilter since for this strategy it does not matter where we run PreEvictionFilter
		WithFilter(podutil.WrapFilterFuncs(handle.Evictor().Filter, handle.Evictor().PreEvictionFilter)). // <- maybe we should remove this line
		WithNamespaces(includedNamespaces).
		WithoutNamespaces(excludedNamespaces).
		WithLabelSelector(failedPodsArgs.LabelSelector).
		BuildFilterFunc()
```

---

Note, disabling the [SystemCriticalPods](https://github.com/kubernetes-sigs/descheduler?tab=readme-ov-file#supported-values-for-podprotectionsdefaultdisabled) default pod protection is not an acceptable work-around, as this would also allow for eviction of Pods with system critical priority. The goal is to protect system critical Pods from eviction but to allow removal of `Failed` Pods.

jawwad-ali:
## Description

Fixes #1775.

The `RemoveFailedPods` plugin chains `handle.Evictor().Filter` into its candidate-listing filter, which protects all system-critical-priority pods regardless of `pod.Status.Phase`. As reported in #1775, this prevents cleanup of `Failed`-phase system-critical pods (e.g. a stuck `cert-manager` pod). Disabling `SystemCriticalPods` protection at the `DefaultEvictor` level is not an acceptable workaround because it would also expose **living** system-critical pods to eviction.

This PR adds an opt-in `IncludingSystemCriticalPods bool` field to `RemoveFailedPodsArgs`. When the operator explicitly sets it, the plugin replaces `handle.Evictor().Filter` with an unconditional permit at candidate-listing time. The eviction is still gated by:

- the existing `PodFailed` phase check
- `validateCanEvict` (`Reasons`, `ExitCodes`, `MinPodLifetimeSeconds`, `ExcludeOwnerKinds`, namespaces, labels)
- `handle.Evictor().PreEvictionFilter`

**Default behaviour is unchanged** — both living and failed system-critical pods stay protected unless the operator opts in.

This addresses @googs1025's "by design" comment on the issue: the framework-level protection stays the framework default; the new flag is a per-plugin, per-operator escape valve scoped specifically to the cleanup-stuck-failed-pods use case, without the broader blast radius of disabling `SystemCriticalPods` in `DefaultEvictor`.

Posted the design proposal on the issue thread first (https://github.com/kubernetes-sigs/descheduler/issues/1775#issuecomment-4363120984) — happy to make the toggle finer-grained (e.g. a per-protection struct) if reviewers prefer.

### Trade-off

The flag also bypasses the framework's other default protections (DaemonSet, StaticPod, LocalStorage, FailedBarePods) **for failed pods** when set. The argument: failed pods are by definition not running, so DaemonSet/StaticPod semantics don't apply the same way, and the operator has explicitly opted in. If reviewers prefer to keep those protections enforced even when this flag is set, the implementation can be tightened to bypass only the system-critical check.

### Test plan

`go test ./pkg/framework/plugins/removefailedpods/... -v` — all existing 25 cases pass plus 2 new cases:

- `system-critical priority failed pod is protected by default` (asserts default behaviour is unchanged)
- `includingSystemCriticalPods=true, system-critical priority failed pod is evicted` (asserts the opt-in works)

`go vet ./pkg/framework/plugins/removefailedpods/...` clean. `go build ./...` clean.

## Checklist
- [x] Code Readability: Is the code easy to understand, well-structured, and consistent with project conventions?
- [x] Naming Conventions: Are variable, function, and structs descriptive and consistent?
- [x] Code Duplication: Is there any repeated code that should be refactored?
- [x] Function/Method Size: Are functions/methods short and focused on a single task?
- [x] Comments & Documentation: Are comments clear, useful, and not excessive?
- [x] Error Handling: Are errors handled appropriately?
- [x] Testing: Are there sufficient unit/integration tests?
- [x] Performance: Are there any obvious performance issues or unnecessary computations?
- [x] Dependencies: Are new dependencies justified?
- [x] Logging & Monitoring: Is logging used appropriately?
- [x] Backward Compatibility: Does this change break any existing functionality or APIs?
- [x] Resource Management: Are resources managed and released properly?
- [x] PR Description: Is the PR description clear, providing enough context?
- [ ] Documentation & Changelog: Are README and docs updated if necessary? (No CHANGELOG.md in repo; docs unchanged.)

### DIFF
--- a/pkg/framework/plugins/removefailedpods/types.go
+++ b/pkg/framework/plugins/removefailedpods/types.go
@@ -25,12 +25,12 @@ import (
 type RemoveFailedPodsArgs struct {
 	metav1.TypeMeta `json:",inline"`
 
-	Namespaces                  *api.Namespaces       `json:"namespaces,omitempty"`
-	LabelSelector               *metav1.LabelSelector `json:"labelSelector,omitempty"`
-	ExcludeOwnerKinds           []string              `json:"excludeOwnerKinds,omitempty"`
-	MinPodLifetimeSeconds       *uint                 `json:"minPodLifetimeSeconds,omitempty"`
-	Reasons                     []string              `json:"reasons,omitempty"`
-	ExitCodes                   []int32               `json:"exitCodes,omitempty"`
-	IncludingInitContainers     bool                  `json:"includingInitContainers,omitempty"`
-	IncludingSystemCriticalPods bool                  `json:"includingSystemCriticalPods,omitempty"`
+	Namespaces                    *api.Namespaces       `json:"namespaces,omitempty"`
+	LabelSelector                 *metav1.LabelSelector `json:"labelSelector,omitempty"`
+	ExcludeOwnerKinds             []string              `json:"excludeOwnerKinds,omitempty"`
+	MinPodLifetimeSeconds         *uint                 `json:"minPodLifetimeSeconds,omitempty"`
+	Reasons                       []string              `json:"reasons,omitempty"`
+	ExitCodes                     []int32               `json:"exitCodes,omitempty"`
+	IncludingInitContainers       bool                  `json:"includingInitContainers,omitempty"`
+	EvictFailedSystemCriticalPods bool                  `json:"evictFailedSystemCriticalPods,omitempty"`
 }
--- a/pkg/framework/plugins/removefailedpods/types.go
+++ b/pkg/framework/plugins/removefailedpods/types.go
@@ -25,11 +25,12 @@ import (
 type RemoveFailedPodsArgs struct {
 	metav1.TypeMeta `json:",inline"`
 
-	Namespaces              *api.Namespaces       `json:"namespaces,omitempty"`
-	LabelSelector           *metav1.LabelSelector `json:"labelSelector,omitempty"`
-	ExcludeOwnerKinds       []string              `json:"excludeOwnerKinds,omitempty"`
-	MinPodLifetimeSeconds   *uint                 `json:"minPodLifetimeSeconds,omitempty"`
-	Reasons                 []string              `json:"reasons,omitempty"`
-	ExitCodes               []int32               `json:"exitCodes,omitempty"`
-	IncludingInitContainers bool                  `json:"includingInitContainers,omitempty"`
+	Namespaces                  *api.Namespaces       `json:"namespaces,omitempty"`
+	LabelSelector               *metav1.LabelSelector `json:"labelSelector,omitempty"`
+	ExcludeOwnerKinds           []string              `json:"excludeOwnerKinds,omitempty"`
+	MinPodLifetimeSeconds       *uint                 `json:"minPodLifetimeSeconds,omitempty"`
+	Reasons                     []string              `json:"reasons,omitempty"`
+	ExitCodes                   []int32               `json:"exitCodes,omitempty"`
+	IncludingInitContainers     bool                  `json:"includingInitContainers,omitempty"`
+	IncludingSystemCriticalPods bool                  `json:"includingSystemCriticalPods,omitempty"`
 }
--- a/pkg/framework/plugins/removefailedpods/failedpods.go
+++ b/pkg/framework/plugins/removefailedpods/failedpods.go
@@ -30,6 +30,7 @@ import (
 	"sigs.k8s.io/descheduler/pkg/descheduler/evictions"
 	podutil "sigs.k8s.io/descheduler/pkg/descheduler/pod"
 	frameworktypes "sigs.k8s.io/descheduler/pkg/framework/types"
+	"sigs.k8s.io/descheduler/pkg/utils"
 )
 
 const PluginName = "RemoveFailedPods"
@@ -60,13 +61,21 @@ func New(ctx context.Context, args runtime.Object, handle frameworktypes.Handle)
 
 	// We can combine Filter and PreEvictionFilter since for this strategy it does not matter where we run PreEvictionFilter
 	defaultFilter := handle.Evictor().Filter
-	if failedPodsArgs.IncludingSystemCriticalPods {
-		// The operator has explicitly opted in to evicting failed pods that are
-		// otherwise protected by the framework default-evictor (system-critical
-		// priority, daemon set, etc.). Bypass the default-evictor filter at
-		// candidate-listing time; the PodFailed phase check (below),
-		// validateCanEvict, and PreEvictionFilter still gate eviction.
-		defaultFilter = func(pod *v1.Pod) bool { return true }
+	if failedPodsArgs.EvictFailedSystemCriticalPods {
+		// The operator has explicitly opted in to evicting failed pods that
+		// would otherwise be protected by the framework default-evictor's
+		// SystemCriticalPods check. Wrap the default-evictor filter so that
+		// pods which are both PodFailed and system-critical are admitted at
+		// candidate-listing time; all other DefaultEvictor protections
+		// (DaemonSet, StaticPod, LocalStorage, ...) still apply, as do
+		// PreEvictionFilter, the PodFailed phase check (below), and
+		// validateCanEvict.
+		defaultFilter = func(pod *v1.Pod) bool {
+			if pod.Status.Phase == v1.PodFailed && utils.IsCriticalPriorityPod(pod) {
+				return true
+			}
+			return handle.Evictor().Filter(pod)
+		}
 	}
 
 	podFilter, err := podutil.NewOptions().
--- a/pkg/framework/plugins/removefailedpods/failedpods.go
+++ b/pkg/framework/plugins/removefailedpods/failedpods.go
@@ -59,8 +59,18 @@ func New(ctx context.Context, args runtime.Object, handle frameworktypes.Handle)
 	}
 
 	// We can combine Filter and PreEvictionFilter since for this strategy it does not matter where we run PreEvictionFilter
+	defaultFilter := handle.Evictor().Filter
+	if failedPodsArgs.IncludingSystemCriticalPods {
+		// The operator has explicitly opted in to evicting failed pods that are
+		// otherwise protected by the framework default-evictor (system-critical
+		// priority, daemon set, etc.). Bypass the default-evictor filter at
+		// candidate-listing time; the PodFailed phase check (below),
+		// validateCanEvict, and PreEvictionFilter still gate eviction.
+		defaultFilter = func(pod *v1.Pod) bool { return true }
+	}
+
 	podFilter, err := podutil.NewOptions().
-		WithFilter(podutil.WrapFilterFuncs(handle.Evictor().Filter, handle.Evictor().PreEvictionFilter)).
+		WithFilter(podutil.WrapFilterFuncs(defaultFilter, handle.Evictor().PreEvictionFilter)).
 		WithNamespaces(includedNamespaces).
 		WithoutNamespaces(excludedNamespaces).
 		WithLabelSelector(failedPodsArgs.LabelSelector).
--- a/pkg/framework/plugins/removefailedpods/failedpods_test.go
+++ b/pkg/framework/plugins/removefailedpods/failedpods_test.go
@@ -354,9 +354,9 @@ func TestRemoveFailedPods(t *testing.T) {
 			},
 		},
 		{
-			description: "includingSystemCriticalPods=true, system-critical priority failed pod is evicted",
+			description: "evictFailedSystemCriticalPods=true, system-critical priority failed pod is evicted",
 			args: RemoveFailedPodsArgs{
-				IncludingSystemCriticalPods: true,
+				EvictFailedSystemCriticalPods: true,
 			},
 			nodes:                   []*v1.Node{test.BuildTestNode("node1", 2000, 3000, 10, nil)},
 			expectedEvictedPodCount: 1,
@@ -387,14 +387,14 @@ func TestRemoveFailedPods(t *testing.T) {
 			}
 
 			plugin, err := New(ctx, &RemoveFailedPodsArgs{
-				Reasons:                     tc.args.Reasons,
-				ExitCodes:                   tc.args.ExitCodes,
-				MinPodLifetimeSeconds:       tc.args.MinPodLifetimeSeconds,
-				IncludingInitContainers:     tc.args.IncludingInitContainers,
-				IncludingSystemCriticalPods: tc.args.IncludingSystemCriticalPods,
-				ExcludeOwnerKinds:           tc.args.ExcludeOwnerKinds,
-				LabelSelector:               tc.args.LabelSelector,
-				Namespaces:                  tc.args.Namespaces,
+				Reasons:                       tc.args.Reasons,
+				ExitCodes:                     tc.args.ExitCodes,
+				MinPodLifetimeSeconds:         tc.args.MinPodLifetimeSeconds,
+				IncludingInitContainers:       tc.args.IncludingInitContainers,
+				EvictFailedSystemCriticalPods: tc.args.EvictFailedSystemCriticalPods,
+				ExcludeOwnerKinds:             tc.args.ExcludeOwnerKinds,
+				LabelSelector:                 tc.args.LabelSelector,
+				Namespaces:                    tc.args.Namespaces,
 			},
 				handle,
 			)
--- a/pkg/framework/plugins/removefailedpods/failedpods_test.go
+++ b/pkg/framework/plugins/removefailedpods/failedpods_test.go
@@ -342,6 +342,30 @@ func TestRemoveFailedPods(t *testing.T) {
 				buildTestPod("p2", "node1", newPodStatus("Shutdown", v1.PodFailed, nil, nil), nil),
 			},
 		},
+		{
+			description:             "system-critical priority failed pod is protected by default",
+			args:                    createRemoveFailedPodsArgs(false, nil, nil, nil, nil),
+			nodes:                   []*v1.Node{test.BuildTestNode("node1", 2000, 3000, 10, nil)},
+			expectedEvictedPodCount: 0,
+			pods: []*v1.Pod{
+				buildCriticalPriorityPod("p1", "node1", newPodStatus("", "", nil, &v1.ContainerState{
+					Terminated: &v1.ContainerStateTerminated{Reason: "NodeAffinity"},
+				})),
+			},
+		},
+		{
+			description: "includingSystemCriticalPods=true, system-critical priority failed pod is evicted",
+			args: RemoveFailedPodsArgs{
+				IncludingSystemCriticalPods: true,
+			},
+			nodes:                   []*v1.Node{test.BuildTestNode("node1", 2000, 3000, 10, nil)},
+			expectedEvictedPodCount: 1,
+			pods: []*v1.Pod{
+				buildCriticalPriorityPod("p1", "node1", newPodStatus("", "", nil, &v1.ContainerState{
+					Terminated: &v1.ContainerStateTerminated{Reason: "NodeAffinity"},
+				})),
+			},
+		},
 	}
 	for _, tc := range tests {
 		t.Run(tc.description, func(t *testing.T) {
@@ -363,13 +387,14 @@ func TestRemoveFailedPods(t *testing.T) {
 			}
 
 			plugin, err := New(ctx, &RemoveFailedPodsArgs{
-				Reasons:                 tc.args.Reasons,
-				ExitCodes:               tc.args.ExitCodes,
-				MinPodLifetimeSeconds:   tc.args.MinPodLifetimeSeconds,
-				IncludingInitContainers: tc.args.IncludingInitContainers,
-				ExcludeOwnerKinds:       tc.args.ExcludeOwnerKinds,
-				LabelSelector:           tc.args.LabelSelector,
-				Namespaces:              tc.args.Namespaces,
+				Reasons:                     tc.args.Reasons,
+				ExitCodes:                   tc.args.ExitCodes,
+				MinPodLifetimeSeconds:       tc.args.MinPodLifetimeSeconds,
+				IncludingInitContainers:     tc.args.IncludingInitContainers,
+				IncludingSystemCriticalPods: tc.args.IncludingSystemCriticalPods,
+				ExcludeOwnerKinds:           tc.args.ExcludeOwnerKinds,
+				LabelSelector:               tc.args.LabelSelector,
+				Namespaces:                  tc.args.Namespaces,
 			},
 				handle,
 			)
@@ -412,5 +437,14 @@ func buildTestPod(podName, nodeName string, podStatus v1.PodStatus, deletionTime
 	pod.ObjectMeta.OwnerReferences = test.GetReplicaSetOwnerRefList()
 	pod.ObjectMeta.SetCreationTimestamp(metav1.Now())
 	pod.DeletionTimestamp = deletionTimestamp
+	return pod
+}
+
+// buildCriticalPriorityPod builds a failed pod whose priority matches the
+// system-critical threshold checked by utils.IsCriticalPriorityPod.
+func buildCriticalPriorityPod(podName, nodeName string, podStatus v1.PodStatus) *v1.Pod {
+	pod := buildTestPod(podName, nodeName, podStatus, nil)
+	priority := int32(2000000000)
+	pod.Spec.Priority = &priority
 	return pod
 }


pkg/framework/plugins/removefailedpods/types.go
```diff
@@ -25,11 +25,12 @@ import (
...
-	IncludingInitContainers bool                  `json:"includingInitContainers,omitempty"`
+	Namespaces                  *api.Namespaces       `json:"namespaces,omitempty"`
+	LabelSelector               *metav1.LabelSelector `json:"labelSelector,omitempty"`
+	ExcludeOwnerKinds           []string              `json:"excludeOwnerKinds,omitempty"`
+	MinPodLifetimeSeconds       *uint                 `json:"minPodLifetimeSeconds,omitempty"`
+	Reasons                     []string              `json:"reasons,omitempty"`
+	ExitCodes                   []int32               `json:"exitCodes,omitempty"`
+	IncludingInitContainers     bool                  `json:"includingInitContainers,omitempty"`
+	IncludingSystemCriticalPods bool                  `json:"includingSystemCriticalPods,omitempty"`
 }
```
googs1025, line 36: Small consistency thing: we already have EvictSystemCriticalPods on DefaultEvictor, and the docs/issues all use that wording. Could we name this one along the same lines — e.g.
EvictFailedSystemCriticalPods — so users don't have to keep two mental models for what's essentially the same concept at different scopes?

pkg/framework/plugins/removefailedpods/types.go
```diff
@@ -25,11 +25,12 @@ import (
 type RemoveFailedPodsArgs struct {
```
googs1025, line 25: Have you double-checked that the new field is wired through any versioned API types as well? If RemoveFailedPodsArgs has a counterpart under pkg/api/v1alphaX/, the YAML decode path  will silently drop includingSystemCriticalPods and the flag will look like it does nothing in real configs even though tests pass. Probably fine, but worth a grep -r RemoveFailedPodsArgs pkg/api/ to confirm.

googs1025:
Reading the code I think this flag as written is named IncludingSystemCriticalPods but actually replaces the whole  
  DefaultEvictor.Filter (DaemonSet/StaticPod/LocalStorage all fall away too) — but more importantly, even if we tighten the implementation, putting this knob at the plugin level creates a precedence problem: a plugin-level flag would be able to override the framework-level EvictSystemCriticalPods decision, which inverts the usual "framework is the ceiling, plugins can only be more restrictive" convention. 
If we ever later add a framework-level EvictFailedSystemCriticalPods to address the same need at the right layer, the two flags would conflict in confusing ways (plugin says yes, framework says no — who wins?).

pkg/framework/plugins/removefailedpods/types.go
jawwad-ali: Done in f05b631a — renamed to `EvictFailedSystemCriticalPods` (and `evictFailedSystemCriticalPods` JSON tag) so it parallels `EvictSystemCriticalPods` on DefaultEvictor.

pkg/framework/plugins/removefailedpods/types.go
jawwad-ali: Checked — `grep -rn RemoveFailedPodsArgs pkg/api/` returns nothing. Descheduler doesn't use the apimachinery per-plugin versioned-types pattern; plugin args are registered via `pluginregistry.Register` and decoded through `pluginArgConversionScheme` with the internal `RemoveFailedPodsArgs` as the target type, so the new field flows through the YAML decode path without a v1alpha2 shadow. The existing `zz_generated.deepcopy.go` uses `*out = *in` which already covers the new bool.

jawwad-ali:
Thanks for the careful read @googs1025 — both points landed.

**Over-broad bypass**: fixed in f05b631a. The filter override is now scoped to pods that are *both* `PodFailed` *and* `utils.IsCriticalPriorityPod`; everything else (DaemonSet, StaticPod, LocalStorage, FailedBarePods) still goes through `handle.Evictor().Filter` unchanged. The PR body's earlier "this also bypasses other protections" trade-off no longer applies.

**Precedence inversion**: this one is worth a real decision before I push more code, so I want to ask rather than guess.

The remaining concern as I read it: even with the implementation tightened, a *plugin-level* `EvictFailedSystemCriticalPods` knob lets a plugin override a framework-level `EvictSystemCriticalPods=false` decision, which inverts the usual "framework is the ceiling, plugins can only narrow" convention. If you later add a framework-level `EvictFailedSystemCriticalPods` to `DefaultEvictorArgs`, the two flags collide.

Two paths from here, happy to take whichever you prefer:

1. **Keep this PR plugin-scoped**, document the precedence interaction (plugin opt-in implies overriding the framework's `SystemCriticalPods` protection *for failed pods only*), and treat a future framework-level toggle as an additive change rather than a conflict (when present, the framework-level flag is the source of truth and the plugin-level one becomes a no-op or is deprecated).

2. **Move the toggle to the framework** — close this PR (or rebuild it), add `EvictFailedSystemCriticalPods` to `DefaultEvictorArgs`, and have `applySystemCriticalPodsProtection` skip its check when `pod.Status.Phase == v1.PodFailed && args.EvictFailedSystemCriticalPods`. `RemoveFailedPods` then needs no plugin-level flag at all; users opt in once at the evictor level. This is the architecturally cleaner direction and aligns with the existing `EvictSystemCriticalPods` knob's location.

I'm leaning toward (2) because it sidesteps the precedence-inversion question entirely and the call-site reads more naturally (`evictFailedSystemCriticalPods` lives next to `evictSystemCriticalPods` in the same args block). Path (1) keeps this PR small but inherits the issue you raised.

Which way do you want me to take it? Happy to redo this PR end-to-end at the framework level if (2) is the answer.

ingvagabund:
Dropped comment in https://github.com/kubernetes-sigs/descheduler/issues/1775#issuecomment-4382341259

jawwad-ali:
Thanks @ingvagabund — that profile-scoped workaround is a great pointer; I hadn't fully appreciated that disabling `SystemCriticalPods` in a dedicated `RemoveFailedPods`-only profile sidesteps the issue today without any code change.

Your last line is the part I want to make sure I'm reading right:

> Every pod identified by RemoveFailedPods plugin as evictable is a failed pod by the definition. So the SystemCriticalPods gating can be lifted.

Taken at face value, that argument applies just as cleanly to the plugin's own filter as it does to a user-supplied profile: if every candidate the plugin will ever evict is, by definition, in `PodFailed`, the framework-level `SystemCriticalPods` gate is logically redundant *for this plugin*. Which opens a third option I hadn't considered:

**Option C — unconditional lift, no flag.** Drop `EvictFailedSystemCriticalPods` from `RemoveFailedPodsArgs` entirely; in `New()`, the filter for `RemoveFailedPods` always bypasses the `SystemCriticalPods` check (only — DaemonSet, StaticPod, LocalStorage, FailedBarePods stay enforced). Smallest possible diff, no precedence-inversion question, and the behaviour matches your "by definition" argument.

That sidesteps both your point and @googs1025's earlier precedence-inversion concern, at the cost of a silent behaviour change for any operator currently relying on `SystemCriticalPods` to also block *failed* system-critical pods from `RemoveFailedPods` (release-note worthy, but I'd argue arguably a bug fix rather than a breaking change, given the plugin's contract).

So three paths I can see, want me to take whichever you and @googs1025 converge on:

1. **Close this PR**, document the dedicated-profile workaround in the issue / README, leave plugin code untouched.
2. **Option C above** — unconditional lift, no flag. Smallest code, behaviour change called out in release notes.
3. **Keep this PR's shape** (opt-in `EvictFailedSystemCriticalPods` flag, tightened to system-critical-only as in `f05b631a`) — preserves backwards compatibility but accepts the precedence-inversion @googs1025 flagged.

I lean toward (2) given your "by definition" framing — it's the cleanest read of the plugin's contract — but happy to go (1) or (3) if either of you sees a reason I'm missing.

ingvagabund:
The current policy API is designed to allow customization of what is considered a failed pod and whether even failed critical pods should be evicted or not. It's perfectly fine to keep the defaults and not evicting failed critical pods as in some cases a different eviction mechanism might be needed to deployed due to different e.g. lifecycle requirements. E.g. some application installation may have different company policies for evicting critical pods.

ingvagabund:
Maybe I misunderstood the description of the issue in https://github.com/kubernetes-sigs/descheduler/issues/1775. Yet, the goal is to allow eviction of failed critical pods by RemoveFailedPods plugin. Which is already possible by configuring two profiles each with different DefaultEvictor arguments.

jawwad-ali:
Thanks for the clarification @ingvagabund — that's a useful correction. I over-read your earlier *"the SystemCriticalPods gating can be lifted"* as a code-change recommendation; in context (and reading your two follow-up messages together) it's clearly justification for why the dedicated-profile workaround is safe, not a steer to change the plugin's filter chain.

Reflecting your position back to make sure I have it right:

- The current policy API is intentional — defaulting to *protect* failed critical pods is the right default because operators have varying compliance/lifecycle policies.
- The use-case in #1775 is already addressable today by configuring a `RemoveFailedPods`-only profile with `SystemCriticalPods` in `PodProtections.DefaultDisabled`.
- The plugin's filter chain therefore doesn't need a per-plugin escape valve.

If that read is correct, the right outcome is for me to **close this PR** and (optionally) leave a short comment on #1775 pointing the reporter at your YAML example so the issue can also be closed. Happy to do both.

Want me to wait for @googs1025 to weigh in before closing, or should I close it now? Either is fine — just want to follow whichever cadence the team prefers.

ingvagabund:
As long as you can validate the suggested existing solution works there's nothing else to do. Maybe adding an example for the plugin under https://github.com/kubernetes-sigs/descheduler#removefailedpods? At this point it is up to @alex-berger to tell us whether the suggested solution works.

alex-berger:
The suggested solution looks good to me and once it is released I, I can test it on our clusters.

ingvagabund:
The currently suggested solution is already part of the releases: https://github.com/kubernetes-sigs/descheduler/issues/1775#issuecomment-4382341259. Can you please take a look and see if it works for your use case?

jawwad-ali:
Got it @ingvagabund — clear direction. I'll leave this PR open so it stays the natural place to revisit if @alex-berger hits anything unexpected with the workaround, and close it once he confirms it works on his clusters (also happy to close earlier if you'd rather not have it sit).

On the README example — happy to send a small, separate **docs-only** PR adding the dedicated-`RemoveFailedPods`-profile YAML you posted to the [RemoveFailedPods section of the README](https://github.com/kubernetes-sigs/descheduler#removefailedpods), so future operators hitting #1775's symptom find the pattern documented next to the plugin itself. Want me to:

1. Open that docs PR now (so it can be reviewed in parallel with @alex-berger's validation), or
2. Wait until he confirms the workaround works, then open the docs PR?

Either is fine — flag whichever fits your review cadence.

jawwad-ali:
@alex-berger — gentle follow-up on validating the workaround 🙂

The two-profile approach @ingvagabund outlined (a dedicated `RemoveFailedPods` profile with `SystemCriticalPods` placed in `DefaultEvictor`'s `PodProtections.DefaultDisabled`) is already shipped in current releases — full YAML here: https://github.com/kubernetes-sigs/descheduler/issues/1775#issuecomment-4382341259

Whenever you get a chance, could you try it on your clusters and confirm it cleans up the stuck `Failed`-phase system-critical pods (the cert-manager case from #1775)? Once you confirm it works as expected, I'll close this PR in favour of the config-only solution. If you hit anything unexpected, this PR stays open as the natural place to pick the code path back up.

No rush — thanks!

alex-berger:
@jawwad-ali My tests failed, the failed Pod is not removed. 

I tested this configuration 

```
kind: "DeschedulerPolicy"
maxNoOfPodsToEvictPerNamespace: 5
maxNoOfPodsToEvictTotal: 30
profiles:
- name: default
  pluginConfig:
  - args:
      podProtections:
        defaultDisabled:
        - PodsWithLocalStorage
        - FailedBarePods
        extraEnabled: null
    name: DefaultEvictor
  - args:
      includingInitContainers: true
      podRestartThreshold: 20
    name: RemovePodsHavingTooManyRestarts
  - name: RemovePodsViolatingNodeTaints
  - name: RemovePodsViolatingInterPodAntiAffinity
  - args:
      constraints:
      - DoNotSchedule
    name: RemovePodsViolatingTopologySpreadConstraint
  - args:
      nodeAffinityType:
    name: RemovePodsViolatingNodeAffinity
  plugins:
    balance:
      enabled: []
    deschedule:
      enabled:
      - RemovePodsHavingTooManyRestarts
      - RemovePodsViolatingNodeAffinity
      - RemovePodsViolatingInterPodAntiAffinity
- name: RemoveFailedPods
  pluginConfig:
  - args:
      podProtections:
        defaultDisabled:
        - PodsWithLocalStorage
        - FailedBarePods
        - SystemCriticalPods
    name: DefaultEvictor
  - args:
      excludeOwnerKinds:
      - Job
      minPodLifetimeSeconds: 120
      reasons:
      - Evicted
      - Shutdown
      - NodeLost
      - NodeAffinity
      - UnexpectedAdmissionError
      - Preempting
      - PodFailed
    name: RemoveFailedPods
  plugins:
    deschedule:
      enabled:
      - RemoveFailedPods
```
against this failed Pod:
```
apiVersion: v1
kind: Pod
metadata:
  name: test-failed-pod
  namespace: kube-system
  uid: f68c3387-6827-4729-a512-fb1ad5adb05d
  resourceVersion: "80927848"
  generation: 1
  creationTimestamp: "2026-06-03T12:07:28Z"
  labels:
    run: test-failed-pod
  selfLink: /api/v1/namespaces/kube-system/pods/test-failed-pod
status:
  observedGeneration: 1
  phase: Failed
  conditions:
  - type: PodReadyToStartContainers
    observedGeneration: 1
    status: "False"
    lastProbeTime: null
    lastTransitionTime: "2026-06-03T13:20:44Z"
  - type: Initialized
    observedGeneration: 1
    status: "True"
    lastProbeTime: null
    lastTransitionTime: "2026-06-03T13:20:40Z"
  - type: Ready
    observedGeneration: 1
    status: "False"
    lastProbeTime: null
    lastTransitionTime: "2026-06-03T13:20:40Z"
    reason: PodFailed
  - type: ContainersReady
    observedGeneration: 1
    status: "False"
    lastProbeTime: null
    lastTransitionTime: "2026-06-03T13:20:40Z"
    reason: PodFailed
  - type: PodScheduled
    observedGeneration: 1
    status: "True"
    lastProbeTime: null
    lastTransitionTime: "2026-06-03T13:20:40Z"
  hostIP: 10.176.29.207
  hostIPs:
  - ip: 10.176.29.207
  podIP: 10.176.23.28
  podIPs:
  - ip: 10.176.23.28
  startTime: "2026-06-03T13:20:40Z"
  containerStatuses:
  - name: test-failed-pod
    state:
      terminated:
        exitCode: 1
        reason: Error
        startedAt: "2026-06-03T13:20:42Z"
        finishedAt: "2026-06-03T13:20:42Z"
        containerID: >-
          containerd://b62c4e602f2f6e2589617634291908b340766bfa2d15ba798110082d531a6ef4
    lastState: {}
    ready: false
    restartCount: 0
    ...
    containerID: >-
      containerd://b62c4e602f2f6e2589617634291908b340766bfa2d15ba798110082d531a6ef4
    started: false
    resources: {}
    volumeMounts:
    - name: kube-api-access-8hbkh
      mountPath: /var/run/secrets/kubernetes.io/serviceaccount
      readOnly: true
      recursiveReadOnly: Disabled
    user:
      linux:
        uid: 0
        gid: 0
        supplementalGroups:
        - 0
        - 10
  qosClass: BestEffort
spec:
  ...
  containers:
  - name: test-failed-pod
    ...
    args:
    - /bin/sh
    - "-c"
    - exit 1
    resources: {}
    volumeMounts:
    - name: kube-api-access-kfbkf
      readOnly: true
      mountPath: /var/run/secrets/kubernetes.io/serviceaccount
    terminationMessagePath: /dev/termination-log
    terminationMessagePolicy: File
    imagePullPolicy: Always
  restartPolicy: Never
  schedulerName: default-scheduler
  priorityClassName: system-cluster-critical
  priority: 2000000000
  enableServiceLinks: true
  preemptionPolicy: PreemptLowerPriority
  ...
```

alex-berger:
@jawwad-ali Actually, my test scenario was wrong, with the below test scenario (I had to add `Error` to the reasons list) **it works** and the Pod got removed.

I tested this configuration 

```
kind: "DeschedulerPolicy"
maxNoOfPodsToEvictPerNamespace: 5
maxNoOfPodsToEvictTotal: 30
profiles:
- name: default
  pluginConfig:
  - args:
      podProtections:
        defaultDisabled:
        - PodsWithLocalStorage
        - FailedBarePods
        extraEnabled: null
    name: DefaultEvictor
  - args:
      includingInitContainers: true
      podRestartThreshold: 20
    name: RemovePodsHavingTooManyRestarts
  - name: RemovePodsViolatingNodeTaints
  - name: RemovePodsViolatingInterPodAntiAffinity
  - args:
      constraints:
      - DoNotSchedule
    name: RemovePodsViolatingTopologySpreadConstraint
  - args:
      nodeAffinityType:
    name: RemovePodsViolatingNodeAffinity
  plugins:
    balance:
      enabled: []
    deschedule:
      enabled:
      - RemovePodsHavingTooManyRestarts
      - RemovePodsViolatingNodeAffinity
      - RemovePodsViolatingInterPodAntiAffinity
- name: RemoveFailedPods
  pluginConfig:
  - args:
      podProtections:
        defaultDisabled:
        - PodsWithLocalStorage
        - FailedBarePods
        - SystemCriticalPods
    name: DefaultEvictor
  - args:
      excludeOwnerKinds:
      - Job
      minPodLifetimeSeconds: 120
      reasons:
      - Evicted
      - Shutdown
      - NodeLost
      - NodeAffinity
      - UnexpectedAdmissionError
      - Preempting
      - Error
    name: RemoveFailedPods
  plugins:
    deschedule:
      enabled:
      - RemoveFailedPods
```
against this failed Pod:
```
apiVersion: v1
kind: Pod
metadata:
  name: test-failed-pod
  namespace: kube-system
  uid: f68c3387-6827-4729-a512-fb1ad5adb05d
  resourceVersion: "80927848"
  generation: 1
  creationTimestamp: "2026-06-03T12:07:28Z"
  labels:
    run: test-failed-pod
  selfLink: /api/v1/namespaces/kube-system/pods/test-failed-pod
status:
  observedGeneration: 1
  phase: Failed
  conditions:
  - type: PodReadyToStartContainers
    observedGeneration: 1
    status: "False"
    lastProbeTime: null
    lastTransitionTime: "2026-06-03T13:20:44Z"
  - type: Initialized
    observedGeneration: 1
    status: "True"
    lastProbeTime: null
    lastTransitionTime: "2026-06-03T13:20:40Z"
  - type: Ready
    observedGeneration: 1
    status: "False"
    lastProbeTime: null
    lastTransitionTime: "2026-06-03T13:20:40Z"
    reason: PodFailed
  - type: ContainersReady
    observedGeneration: 1
    status: "False"
    lastProbeTime: null
    lastTransitionTime: "2026-06-03T13:20:40Z"
    reason: PodFailed
  - type: PodScheduled
    observedGeneration: 1
    status: "True"
    lastProbeTime: null
    lastTransitionTime: "2026-06-03T13:20:40Z"
  hostIP: 10.176.29.207
  hostIPs:
  - ip: 10.176.29.207
  podIP: 10.176.23.28
  podIPs:
  - ip: 10.176.23.28
  startTime: "2026-06-03T13:20:40Z"
  containerStatuses:
  - name: test-failed-pod
    state:
      terminated:
        exitCode: 1
        reason: Error
        startedAt: "2026-06-03T13:20:42Z"
        finishedAt: "2026-06-03T13:20:42Z"
        containerID: >-
          containerd://b62c4e602f2f6e2589617634291908b340766bfa2d15ba798110082d531a6ef4
    lastState: {}
    ready: false
    restartCount: 0
    ...
    containerID: >-
      containerd://b62c4e602f2f6e2589617634291908b340766bfa2d15ba798110082d531a6ef4
    started: false
    resources: {}
    volumeMounts:
    - name: kube-api-access-8hbkh
      mountPath: /var/run/secrets/kubernetes.io/serviceaccount
      readOnly: true
      recursiveReadOnly: Disabled
    user:
      linux:
        uid: 0
        gid: 0
        supplementalGroups:
        - 0
        - 10
  qosClass: BestEffort
spec:
  ...
  containers:
  - name: test-failed-pod
    ...
    args:
    - /bin/sh
    - "-c"
    - exit 1
    resources: {}
    volumeMounts:
    - name: kube-api-access-kfbkf
      readOnly: true
      mountPath: /var/run/secrets/kubernetes.io/serviceaccount
    terminationMessagePath: /dev/termination-log
    terminationMessagePolicy: File
    imagePullPolicy: Always
  restartPolicy: Never
  schedulerName: default-scheduler
  priorityClassName: system-cluster-critical
  priority: 2000000000
  enableServiceLinks: true
  preemptionPolicy: PreemptLowerPriority
  ...
```

jawwad-ali:
Thanks for the precise repro and for circling back to validate, @alex-berger 🙏

For anyone who finds this later: the workaround was correct; the first run missed only because of the `reasons` list. `RemoveFailedPods` matches `reasons` against the pod's actual failure reason — for your pod that's the container's terminated `reason: Error` (exit 1). `PodFailed` is the pod *phase* (`status.phase: Failed`), not a value the `reasons` filter matches, so listing it caught nothing; swapping in `Error` made the candidate match and the pod was evicted.

That confirms @ingvagabund's two-profile, config-only approach (a dedicated `RemoveFailedPods` profile with `SystemCriticalPods` in `DefaultEvictor`'s `PodProtections.DefaultDisabled`) solves #1775 end-to-end with no code changes.

So I'm closing this PR in favour of the config-only solution, as discussed. I'll follow up with a small docs-only PR adding this profile example — plus the `reasons`-vs-phase gotcha above — to the `RemoveFailedPods` README section, so future operators find the pattern (and avoid the trap). Thanks @ingvagabund and @googs1025 for steering this to the right layer 🙌

PR #1883, opened 2026-06-19
Register the PodDisruptionBudget informer so PodsWithoutPDB protection works

### LINKED ISSUE #1882: DefaultEvictor: PodsWithoutPDB protection silently blocks all evictions

  descheduler version: v0.36.0 (Helm chart 0.36.0)
  
  Yes, reproduced on v0.36.0.

  Running as a CronJob via Helm chart (`kind: CronJob`, default configuration).

  ```yaml
  kind: "DeschedulerPolicy"
  profiles:
    - name: default
      pluginConfig:
        - name: DefaultEvictor
          args:
            nodeFit: true
            podProtections:
              extraEnabled:
                - "PodsWithoutPDB"
              defaultDisabled:
                - "PodsWithLocalStorage"
          args:
            topologyBalanceNodeFit: true
            constraints:
              - DoNotSchedule
              - ScheduleAnyway
      plugins:
        evictor:
          enabled:
            - DefaultEvictor
        balance:
          enabled:
            - RemovePodsViolatingTopologySpreadConstraint
  ```

  <details><summary><code>kubectl version</code> Output </summary>

```bash
  $ kubectl version
  Client Version: v1.33.7
  Kustomize Version: v5.6.0
  Server Version: v1.33.11-eks-40737a8
```
</details>

  **What did you do?**

  Configured `DefaultEvictor` with `podProtections.extraEnabled: ["PodsWithoutPDB"]` and enabled
  `RemovePodsViolatingTopologySpreadConstraint`. Pods were clearly violating topology spread constraints — all pods were
  concentrated in a single availability zone while other AZs had available capacity.

  Pods violating `TopologySpreadConstraints` to be evicted, with `evictedPods >= 1` reported in the logs.

  **What did you see instead?**

  `evictedPods=0` — no pods were evicted despite clear topology violations. No errors were reported in the logs.

  Removing `PodsWithoutPDB` from `extraEnabled` restored normal eviction behavior (`evictedPods=1` confirmed).

yugstar:
Fixes #1882

## Problem

The `DefaultEvictor` `PodsWithoutPDB` protection reads the PodDisruptionBudget lister (`handle.SharedInformerFactory().Policy().V1().PodDisruptionBudgets().Lister()`) during the descheduling cycle. That cycle runs **after** `sharedInformerFactory.Start()` / `WaitForCacheSync()`, and the factory only starts informers that were registered before `Start()`.

The PDB informer was never registered in production code, so the lister returns an empty list and `IsPodCoveredByPDB` reports `false` for **every** pod. With `PodsWithoutPDB` enabled, that means every pod looks "without a PDB" and is protected — silently blocking **all** evictions (the reported symptom: `evictedPods=0`, no errors).

## Fix

Register the PDB informer up front in `newDescheduler`, exactly mirroring the existing workaround already in place for the `PersistentVolumeClaims` informer used by the same plugin (a few lines above).

## Test

Added `TestPodsWithoutPDBProtectionRespectsPDBs`, which runs through the **real** `newDescheduler` / `RunDeschedulerStrategies` startup path with `PodsWithoutPDB` enabled and duplicate pods covered by a PDB, asserting they are still evicted.

This must go through the real startup path — the `defaultevictor` unit harness pre-registers the PDB lister, which masks the bug. Verified:

- **without the fix:** `FAIL … PodsWithoutPDB protection blocked all evictions`
- **with the fix:** `ok`

`go build` / `go vet` / the new test all pass.

### DIFF
--- a/pkg/descheduler/descheduler.go
+++ b/pkg/descheduler/descheduler.go
@@ -126,6 +126,11 @@ func newDescheduler(ctx context.Context, rs *options.DeschedulerServer, deschedu
 	// Temporarily register the PVC because it is used by the DefaultEvictor plugin during
 	// the descheduling cycle, where informer registration is ignored.
 	_ = sharedInformerFactory.Core().V1().PersistentVolumeClaims().Informer()
+	// Temporarily register the PDB for the same reason: the DefaultEvictor PodsWithoutPDB
+	// protection reads the PDB lister during the descheduling cycle, which runs after the
+	// factory has started, so the informer must be registered up front or the lister is
+	// empty and every pod is treated as not covered by a PDB.
+	_ = sharedInformerFactory.Policy().V1().PodDisruptionBudgets().Informer()
 
 	getPodsAssignedToNode, err := podutil.BuildGetPodsAssignedToNodeFunc(podInformer)
 	if err != nil {
--- a/pkg/descheduler/descheduler_test.go
+++ b/pkg/descheduler/descheduler_test.go
@@ -389,6 +389,85 @@ func TestDuplicate(t *testing.T) {
 	}
 }
 
+// TestPodsWithoutPDBProtectionRespectsPDBs is a regression test for #1882.
+//
+// The DefaultEvictor PodsWithoutPDB protection reads the PDB lister during the
+// descheduling cycle, which runs after the shared informer factory has started.
+// If the PDB informer is not registered up front, the lister is empty, every pod
+// looks "without PDB", the protection blocks all evictions, and nothing is evicted.
+//
+// Here the duplicate pods ARE covered by a PDB, so with PodsWithoutPDB enabled they
+// must still be evictable by RemoveDuplicates. This fails without the informer
+// registration (the pods are wrongly treated as uncovered and protected). It must run
+// through the real newDescheduler/RunDeschedulerStrategies path: the defaultevictor
+// unit harness pre-registers the PDB lister and would mask the bug.
+func TestPodsWithoutPDBProtectionRespectsPDBs(t *testing.T) {
+	initPluginRegistry()
+
+	ctx := context.Background()
+	node1 := test.BuildTestNode("n1", 2000, 3000, 10, nil)
+	node2 := test.BuildTestNode("n2", 2000, 3000, 10, nil)
+
+	ownerRef := test.GetReplicaSetOwnerRefList()
+	buildPod := func(name string) *v1.Pod {
+		p := test.BuildTestPod(name, 100, 0, node1.Name, nil)
+		p.Namespace = "default"
+		p.Labels = map[string]string{"app": "foo"}
+		p.ObjectMeta.OwnerReferences = ownerRef
+		return p
+	}
+	p1, p2, p3 := buildPod("p1"), buildPod("p2"), buildPod("p3")
+	pdb := test.BuildTestPDB("foo-pdb", "foo")
+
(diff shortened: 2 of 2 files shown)


pkg/descheduler/descheduler.go
```diff
@@ -126,6 +126,11 @@ func newDescheduler(ctx context.Context, rs *options.DeschedulerServer, deschedu
 	// Temporarily register the PVC because it is used by the DefaultEvictor plugin during
 	// the descheduling cycle, where informer registration is ignored.
 	_ = sharedInformerFactory.Core().V1().PersistentVolumeClaims().Informer()
+	// Temporarily register the PDB for the same reason: the DefaultEvictor PodsWithoutPDB
+	// protection reads the PDB lister during the descheduling cycle, which runs after the
+	// factory has started, so the informer must be registered up front or the lister is
+	// empty and every pod is treated as not covered by a PDB.
+	_ = sharedInformerFactory.Policy().V1().PodDisruptionBudgets().Informer()
```
googs1025, line 133: Could we consider whether the PDB informer should be registered only when `PodsWithoutPDB` protection is enabled?

  This fix is correct and mirrors the existing PVC workaround, but it also means every descheduler run will start a PDB informer, even for policies that do not use `PodsWithoutPDB`.
  RBAC already grants `get/list/watch` on PDBs, so this is not a correctness issue, but on large clusters it adds an unconditional watch.

  If keeping the minimal fix is preferred, I think that is reasonable. Just wanted to call out the tradeoff between the current simple approach and conditionally registering the
  informer based on the active profiles.

pkg/descheduler/descheduler_test.go
```diff
@@ -389,6 +389,85 @@ func TestDuplicate(t *testing.T) {
...
+				},
+			},
+		},
+	}
+
+	if err := RunDeschedulerStrategies(ctx, rs, policy, "v1"); err != nil {
+		t.Fatalf("Unable to run descheduler strategies: %v", err)
+	}
+
+	if len(evictedPods) == 0 {
```
googs1025: Could we make this assertion a bit stricter?

  In this setup there are three duplicate pods on `node1` and two feasible nodes, so `RemoveDuplicates` should evict exactly one pod. Checking only `len(evictedPods) == 0` verifies the
  regression, but would still pass if this path started over-evicting later.

  ```go
  if len(evictedPods) != 1 {
      t.Fatalf("expected exactly one duplicate pod covered by a PDB to be evicted, got %d", len(evictedPods))
  }
```

pkg/descheduler/descheduler.go
yugstar: Thanks for flagging the tradeoff. I'd lean toward keeping the unconditional registration here, for consistency with the PVC informer registered the same way right above it — making this profile-conditional would be the only such case in the function. As you note the watch is already covered by RBAC, and a single PDB informer is lightweight next to the pod/node informers the descheduler already runs. Happy to move both informers to conditional registration as a separate follow-up if you'd prefer that direction, so the two stay consistent.

pkg/descheduler/descheduler_test.go
yugstar: Good call — done in 5f6c6d7. Tightened it to assert exactly one eviction (`!= 1`, with the count in the failure message), since the setup has three duplicates on `node1` and two feasible nodes. Verified it still passes, so it now also guards against over-eviction, not just the original block-everything regression.

PR #1890, opened 2026-07-02
feat: add opt-in schedulerName and runtimeClassName to the Helm chart

somaz94:
## Description

Adds two optional pod-spec fields to the descheduler chart, mirroring the existing `priorityClassName` / `topologySpreadConstraints` knobs across both the Deployment and CronJob workloads:

- `schedulerName` — schedule the descheduler pods with a non-default scheduler
- `runtimeClassName` — run the descheduler pods under a specific RuntimeClass (e.g. gVisor / Kata)

Both default to an empty string, so the rendered output is unchanged for existing users (default render is byte-identical). This follows the same opt-in pattern as the recently merged PodDisruptionBudget support (#1885).

## Validation

- `helm unittest charts/descheduler --strict` — 22 passed (6 new cases across `deployment_test.yaml` and `cronjob_test.yaml`)
- `ct lint --config=.github/ci/ct.yaml` — passed
- `helm template` default render is byte-identical to `master` for both `kind=Deployment` and `kind=CronJob`
- With the fields set, both workloads render `schedulerName:` and `runtimeClassName:` at the pod spec level

## Checklist

- [x] Testing: helm-unittest cases for both fields (set + default-absent) on Deployment and CronJob
- [x] Backward Compatibility: default empty → render unchanged for existing users
- [x] Documentation & Changelog: README parameters table updated

PR #1892, opened 2026-07-07
feat: add opt-in hostUsers to the Helm chart

somaz94:
Adds an opt-in `hostUsers` field to the descheduler Helm chart so the Pod spec `hostUsers` (user-namespace sharing, Kubernetes 1.28+) can be configured. Mirrors the existing `automountServiceAccountToken` bool pattern (`{{- if kindIs "bool" .Values.hostUsers }}`), so an unset value leaves the render byte-identical and both `true` and `false` are honored. Applied to both the Deployment and CronJob workloads.

Validation (local):
- `helm unittest charts/descheduler --strict`: 22 passed (6 new cases: set-true / set-false / omit-default x Deployment + CronJob)
- `ct lint --config .github/ci/ct.yaml --charts charts/descheduler --validate-maintainers=false`: all charts linted successfully
- Default render byte-identical to master for both `kind=Deployment` and `kind=CronJob`
- `--set hostUsers=false` renders `hostUsers: false`; `--set hostUsers=true` renders `hostUsers: true` on both workloads

### DIFF
--- a/charts/descheduler/README.md
+++ b/charts/descheduler/README.md
@@ -71,6 +71,7 @@ The following table lists the configurable parameters of the _descheduler_ chart
 | `priorityClassName`                 | The name of the priority class to add to pods                                                                         | `system-cluster-critical`                 |
 | `schedulerName`                     | The name of the scheduler used to schedule the descheduler cronjob/deployment pods                                    | `""`                                      |
 | `runtimeClassName`                  | The RuntimeClass applied to the descheduler cronjob/deployment pods                                                   | `""`                                      |
+| `hostUsers`                         | If set (`true`/`false`), sets `hostUsers` on the pod spec to control user-namespace sharing                           | `nil`                                     |
 | `rbac.create`                       | If `true`, create & use RBAC resources                                                                                | `true`                                    |
 | `resources`                         | Descheduler container CPU and memory requests/limits                                                                  | _see values.yaml_                         |
 | `serviceAccount.create`             | If `true`, create a service account for the cron job                                                                  | `true`                                    |
--- a/charts/descheduler/templates/cronjob.yaml
+++ b/charts/descheduler/templates/cronjob.yaml
@@ -93,6 +93,9 @@ spec:
           {{- if .Values.runtimeClassName }}
           runtimeClassName: {{ .Values.runtimeClassName }}
           {{- end }}
+          {{- if kindIs "bool" .Values.hostUsers }}
+          hostUsers: {{ .Values.hostUsers }}
+          {{- end }}
           serviceAccountName: {{ template "descheduler.serviceAccountName" . }}
           {{- if kindIs "bool" .Values.automountServiceAccountToken }}
           automountServiceAccountToken: {{ .Values.automountServiceAccountToken }}
--- a/charts/descheduler/templates/deployment.yaml
+++ b/charts/descheduler/templates/deployment.yaml
@@ -47,6 +47,9 @@ spec:
       {{- if .Values.runtimeClassName }}
       runtimeClassName: {{ .Values.runtimeClassName }}
       {{- end }}
+      {{- if kindIs "bool" .Values.hostUsers }}
+      hostUsers: {{ .Values.hostUsers }}
+      {{- end }}
       serviceAccountName: {{ template "descheduler.serviceAccountName" . }}
       {{- if kindIs "bool" .Values.automountServiceAccountToken }}
       automountServiceAccountToken: {{ .Values.automountServiceAccountToken }}
--- a/charts/descheduler/tests/cronjob_test.yaml
+++ b/charts/descheduler/tests/cronjob_test.yaml
@@ -41,3 +41,27 @@ tests:
           path: spec.jobTemplate.spec.template.spec.schedulerName
       - notExists:
           path: spec.jobTemplate.spec.template.spec.runtimeClassName
+
+  - it: sets hostUsers to false when configured
+    set:
+      hostUsers: false
+    template: templates/cronjob.yaml
+    asserts:
+      - equal:
+          path: spec.jobTemplate.spec.template.spec.hostUsers
+          value: false
+
+  - it: sets hostUsers to true when configured
+    set:
+      hostUsers: true
+    template: templates/cronjob.yaml
+    asserts:
+      - equal:
+          path: spec.jobTemplate.spec.template.spec.hostUsers
+          value: true
+
+  - it: omits hostUsers by default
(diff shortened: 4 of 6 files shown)


a7i:
Why do we need this? Descheduler doesn't run as root anyways

somaz94:
Thanks for taking a look! The motivation here is user-namespace isolation rather than the container's own user: with `hostUsers: false` the pod's UIDs are mapped to unprivileged UIDs on the host, so even a non-root process gets an extra layer of protection against container-escape and host-UID collisions — it's orthogonal to whether descheduler runs as root inside the container.

That said it's a pure opt-in — the default is unchanged and the rendered output is byte-identical when the field is unset, so it only affects users who explicitly want userns isolation, and it mirrors the other pod-spec fields the chart already exposes. Happy to close it if you'd rather not carry it, though — just let me know.

somaz94:
Good question — the value isn't about *who the container runs as*, it's about the host-level isolation boundary.

`hostUsers: false` opts the pod into a **user namespace**, which is orthogonal to `runAsNonRoot`. Even though descheduler already runs as a non-root UID, without a user namespace that UID is the *same* UID on the host, and any Linux capabilities the process holds are real capabilities in the host's initial user namespace. With `hostUsers: false`, the pod's UIDs/GIDs and capabilities are remapped into an unprivileged range that has no meaning on the host — so it's defense-in-depth: if a container-escape or kernel CVE is ever hit, the blast radius on the node is bounded by the mapping rather than by "we happened to run as non-root."

Concretely it buys:
- host UID/GID remapping (container identities ≠ host identities),
- capabilities scoped to the namespace instead of the host,
- alignment for operators who mandate user namespaces fleet-wide via Pod Security / a hardened baseline.

The change is fully opt-in — `hostUsers` defaults to `nil`, so the field is omitted from the pod spec entirely and there's zero behavior change for anyone who doesn't set it (covered by the true/false/omitted unit tests). It's the same pattern the chart already uses for other pod-spec security toggles like `automountServiceAccountToken` and `priorityClassName`: expose the knob so security-conscious users don't have to patch the rendered output via kustomize/postrender.

googs1025:
flake should be fix in https://github.com/kubernetes-sigs/descheduler/pull/1891

somaz94:
Thanks @googs1025 — now that #1891 has merged, retriggering the e2e jobs.

/retest

ingvagabund:
I am not aware of any need for `hostUsers: true`. We might as well set it to `false` by default. The descheduler is not expected to access a host system. Only consume from servers (apiserver, metrics, ...). Happy to be proved wrong.

somaz94:
Thanks for taking a look, @ingvagabund — I think there may be a misread of what this PR does, so let me clarify.

**The chart does not set `hostUsers: true`.** The value is unset by default and the field is only rendered when it is explicitly a bool:

```gotemplate
{{- if kindIs "bool" .Values.hostUsers }}
hostUsers: {{ .Values.hostUsers }}
{{- end }}
```

With default values the rendered Deployment/CronJob is byte-identical to today, and there is a unit test asserting exactly that (`omits hostUsers by default` → `notExists: spec.template.spec.hostUsers`). Nothing changes for existing users — that is what "opt-in" in the title refers to.

I also agree with your reasoning, and the `values.yaml` comment already points people in that direction — the example shown is `false`, not `true`:

```yaml
# Set the Pod spec `hostUsers` field to control user-namespace sharing.
# Default is not set (the cluster default applies). Set to `false` to opt into
# user-namespace isolation, or `true` to share the host user namespace.
# hostUsers: false
```

The use case is the reverse of how it may read: operators who want the descheduler pod to run in its own user namespace need a way to set `hostUsers: false`, and today the chart gives them no way to do that short of a post-render patch.

On actually defaulting it to `false`: I would rather not do that in this PR. `hostUsers: false` requires the `UserNamespacesSupport` feature gate plus a CRI/kernel combination supporting idmapped mounts; on clusters without it, pods requesting it fail to start. Making that the chart default would change behavior for every consumer and break the ones that cannot support it. I am happy to do it as a follow-up if you want it as the project's stance — I just think it should be a deliberate decision rather than a side effect of adding the knob.

Let me know which you prefer and I will adjust.

somaz94:
@ingvagabund Thanks for taking a look — I think there may be a misunderstanding about what this PR sets, so let me clarify.

**This PR never sets `hostUsers: true`.** The value defaults to `nil`, and the template only renders the field when the user explicitly sets a boolean:

```gotemplate
{{- if kindIs "bool" .Values.hostUsers }}
hostUsers: {{ .Values.hostUsers }}
{{- end }}
```

So with the default values the rendered manifest is byte-identical to today — the key is simply absent, and the kubelet default applies. The purpose of the knob is the opposite of what you described: it lets an operator set `hostUsers: **false**` to opt into a user namespace, which remaps container UID/GID away from the host and confines capabilities to that namespace. That is defense-in-depth against container escape, and it is orthogonal to `runAsNonRoot` (which is about which UID the process runs as, not which user namespace it lives in).

I would push back gently on defaulting it to `false`: user namespaces require kernel and runtime support and are not available on every distribution, so flipping the default would be a behaviour change for existing users. Leaving it unset is the conservative option.

For precedent, the same design has already landed in three other kubernetes-sigs charts, all from an unrelated contributor:

- kubernetes-sigs/metrics-server#1751 — merged 2026-07-22
- kubernetes-sigs/node-feature-discovery#2411 — merged 2026-01-19
- kubernetes-sigs/headlamp#4371 — merged 2026-02-04

All three use the identical shape: `hostUsers: nil` in `values.yaml`, rendered only when non-nil, documented with a `nil` default.

One thing metrics-server does that this PR does not, and which I am happy to adopt if you would like it — it wraps the field in a capability guard so the manifest stays valid on older clusters:

```gotemplate
{{- if (semverCompare ">= 1.33-0" .Capabilities.KubeVersion.Version) }}
```

Say the word and I will add that. And if you still think the knob is not worth carrying, I am fine closing it — just wanted to make sure the decision is based on what the PR actually does.

somaz94:
Rebased onto master now that #1890 has landed — the two PRs touched the same six chart files, so this one went conflicting the moment #1890 merged. Resolution keeps both feature sets side by side; the net diff against master is additions only, with no changes to the `schedulerName` / `runtimeClassName` work from #1890.

Verified locally: `helm lint` clean, `helm unittest` 28/28 across 5 suites, and rendering confirms all three fields are omitted from the PodSpec by default and emitted only when explicitly set (both the CronJob and Deployment paths).

The push dropped the `lgtm` label — @ingvagabund could you take another look when you have a moment?

PR #1898, opened 2026-07-21
fix(nodeutilization): evict pods that tolerate destination taints via a broad toleration

Mukuwul:
**What this does**

`PodToleratesTaints` decides, for a set of candidate destination nodes, whether a pod tolerates at least one of them. It does this by comparing the pod's tolerations to each node's taints — but only when:

```go
if len(pod.Spec.Tolerations) >= len(taintsForNode) {
```

That precondition isn't correct. A single toleration can tolerate any number of taints — an `Exists` toleration with an empty key tolerates *every* taint. So a pod with fewer tolerations than a node has taints is skipped by the guard and reported as not tolerating the node, even when it actually tolerates all of its taints.

The only caller is the Low/HighNodeUtilization eviction loop (`evictPods` in `pkg/framework/plugins/nodeutilization/nodeutilization.go`), which uses this to decide whether a pod can be moved to a destination node. The effect is that pods with a broad toleration are silently skipped whenever the destination nodes carry more taints than the pod has tolerations — so the plugins quietly stop rebalancing them.

**Repro**

- Destination (under-utilized) nodes tainted with two `NoSchedule` taints, e.g. `nvidia.com/gpu=:NoSchedule` and `dedicated=team-a:NoSchedule`.
- A pod on an over-utilized node with `tolerations: [{operator: Exists}]` (tolerates everything).

The guard evaluates `1 >= 2` → false → the real toleration check is skipped → the pod is treated as fitting nowhere and is never evicted, even though it fully tolerates the destination nodes.

**The fix**

Drop the count precondition and always evaluate the tolerations against the taints (which is what `TolerationsTolerateTaintsWithFilter` already does correctly, and matches how the scheduler reasons about tolerations). The verbose per-taint diagnostic logging is kept.

Added `TestPodToleratesTaints`, including the regression case above. It fails on `master` (`= false, want true`) and passes with this change.

None filed — found while reading the nodeutilization eviction path. Happy to open a tracking issue if maintainers prefer.

### DIFF
--- a/pkg/utils/pod.go
+++ b/pkg/utils/pod.go
@@ -148,28 +148,24 @@ func PodRequestsAndLimits(pod *v1.Pod) (v1.ResourceList, v1.ResourceList) {
 // PodToleratesTaints returns true if a pod tolerates one node's taints
 func PodToleratesTaints(ctx context.Context, pod *v1.Pod, taintsOfNodes map[string][]v1.Taint) bool {
 	for nodeName, taintsForNode := range taintsOfNodes {
-		if len(pod.Spec.Tolerations) >= len(taintsForNode) {
+		// A single toleration can tolerate any number of taints (for example an
+		// Exists toleration with an empty key tolerates every taint), so we must
+		// always evaluate the tolerations against the taints rather than gating
+		// on the toleration/taint counts.
+		if TolerationsTolerateTaintsWithFilter(ctx, pod.Spec.Tolerations, taintsForNode, nil) {
+			return true
+		}
 
-			if TolerationsTolerateTaintsWithFilter(ctx, pod.Spec.Tolerations, taintsForNode, nil) {
-				return true
-			}
-
-			if klog.V(5).Enabled() {
-				for i := range taintsForNode {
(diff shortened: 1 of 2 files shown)


yogeshbendre:
Thank you @Mukuwul for this PR.

Independently found and reproduced this same bug before searching and finding this PR.
I can confirm both the root cause and the fix seem right.

Verified against a fresh clone at the v0.36.0 release tag. The len(tolerations) >= len(taints) guard in PodToleratesTaints does block the real toleration check runs. 
I tried with the same kind of repro as described above having a single {operator: Exists} toleration against two destination-node taints. 

In addition, I had independently applied a fix similar to this PR's fix locally and confirmed:

 - The new TestPodToleratesTaints regression test passes.
 - The full pkg/utils suite passes — nothing else relies on the removed count guard.
 - pkg/framework/plugins/nodeutilization and its classifier/normalizer subpackages pass unchanged.


 This matches how TolerationsTolerateTaintsWithFilter is used everywhere else in the codebase (no count precondition there either), so the fix looks correct and safe to merge. 

Hope a maintainer can get eyes on this soon. :)

Issue #1896, opened 2026-07-19
Multiple logging issues


mattburgess:
**What version of descheduler are you using?**

descheduler version: v0.35.1

**Does this issue reproduce with the latest release?** Untested, but I presume it's still happening as the release notes for v0.36.0 don't show any logging related fixes.

**Which descheduler CLI options are you using?**

```text
--policy-config-file=/policy-dir/policy.yaml
--descheduling-interval=5m
--logging-format=json
--v=3
```

**Please provide a copy of your descheduler policy config file**
N/A

**What k8s version are you using (`kubectl version`)?**

<details><summary><code>kubectl version</code> Output</summary><br><pre>
$ kubectl version
Client Version: v1.34.3
Kustomize Version: v5.7.1
Server Version: v1.35.6-eks-8f14419
</pre></details>

**What did you do?**
Inspected descheduler's logs in Datadog

**What did you expect to see?**
1. Logs are shown as being pulled in from descheduler's stdout and stderr streams
2. Logs have the correct categorization (debug/info/warn/emergency)

**What did you see instead?**
1. Logs all come in from the stderr stream.
2. Many logs display as `emergency`

For issue 1:

* I tried setting `--log-json-split-stream`, which in turn requires turning on the `LoggingAlphaOptions` feature. However, enabling that, the pod goes into CrashLoopBackOff with this error:

   ```text
   E0719 09:24:16.305765       1 server.go:69] "failed to apply" err="unrecognized feature gate: LoggingAlphaOptions"
   E0719 09:24:16.305884       1 run.go:72] "command failed" err="unrecognized feature gate: LoggingAlphaOptions"`
   ```
* As an aside, the help text for `--log-json-split-stream` says `The default is to write a single stream to stdout` but that's incorrect. The current default writes everything to the `stderr` stream.

For issue 2:

* This is caused by the `status` field in some of the logs, e.g.:

   ```json
   {
     "service": "descheduler",
     "message": "Ignoring node",
     "attributes": {
       "caller": "node/node.go:107",
       "condition": "Ready",
       "node": {
         "name": "<redacted>"
       },
       "status": "False",
       "ts": 1784445642611.5325,
       "v": 1
   }
   ```

[...]

PR #1899, opened 2026-07-21
fix(nodeutilization): log prometheus warnings via a structured key

Mukuwul:
**What this does**

`NodeUsageFromPrometheusMetrics` logs Prometheus query warnings with:

```go
logger.Info("prometheus metrics warnings: %v", warnings)
```

`logger` here is a `klog.Logger` (logr), whose signature is `Info(msg string, keysAndValues ...any)` — it doesn't interpret printf verbs. So the `%v` is printed literally as part of the message, and `warnings` is left as a key with no value, which produces a malformed log line (and broken JSON when `--logging-format=json` is used) every time a query returns warnings.

This logs a plain message and attaches the warnings as a structured `warnings` key instead.

None directly. It's in the same spirit as the structured-logging concerns raised in #1896, but that issue is about JSON stream splitting and node-condition field naming — this is a separate call site, so I've kept it as its own small fix rather than folding it in.

### DIFF
--- a/pkg/framework/plugins/nodeutilization/usageclients_test.go
+++ b/pkg/framework/plugins/nodeutilization/usageclients_test.go
@@ -22,8 +22,10 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"strings"
 	"testing"
 
+	"github.com/go-logr/logr/funcr"
 	"github.com/prometheus/common/model"
 
 	v1 "k8s.io/api/core/v1"
@@ -32,6 +34,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/informers"
 	fakeclientset "k8s.io/client-go/kubernetes/fake"
+	"k8s.io/klog/v2"
 	"k8s.io/metrics/pkg/apis/metrics/v1beta1"
 	fakemetricsclient "k8s.io/metrics/pkg/client/clientset/versioned/fake"
 
@@ -146,11 +149,13 @@ func TestActualUsageClient(t *testing.T) {
 type fakePromClient struct {
 	result   interface{}
 	dataType model.ValueType
+	warnings []string
 }
 
 type fakePayload struct {
-	Status string      `json:"status"`
-	Data   queryResult `json:"data"`
+	Status   string      `json:"status"`
+	Data     queryResult `json:"data"`
+	Warnings []string    `json:"warnings,omitempty"`
 }
 
 type queryResult struct {
@@ -169,6 +174,7 @@ func (client *fakePromClient) Do(ctx context.Context, request *http.Request) (*h
 			Type:   client.dataType,
 			Result: client.result,
 		},
+		Warnings: client.warnings,
 	})
 
 	return &http.Response{StatusCode: 200}, jsonData, err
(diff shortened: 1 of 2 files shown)


PKNaveen:
This does fix the logging issue which was initially added under #1655, hoping we could also write a test to mock the warnings as well under `TestPrometheusUsageClient`, However I'm not sure if this falls within our scope.

Mukuwul:
Good call, and thanks for linking #1655 — I hadn't spotted that it was the same call site.

I've pushed a test. `fakePromClient` can now carry warnings in its payload, and `TestNodeUsageFromPrometheusMetricsLogsWarnings` captures the logger output through a `funcr` sink (already vendored, so no new dependency) and asserts the warnings come through as a structured key.

It's a real regression test rather than just coverage — with the old line it fails and prints exactly what the bug looks like:

```
"msg"="prometheus metrics warnings: %v" "<non-string-key: [\"query timed ou>"="<no-value>"
```

so the verb stays literal and the warnings end up as a mangled key with no value.

I put it in its own test rather than inside `TestPrometheusUsageClient`, since that one is table-driven around `sync()` and node utilisation, and this needed to hook the logger and call `NodeUsageFromPrometheusMetrics` directly. Happy to fold it into the table if you'd rather keep it all in one place.

pkg/framework/plugins/nodeutilization/usageclients_test.go
```diff
@@ -297,3 +303,47 @@ func TestPrometheusUsageClient(t *testing.T) {
...
+		t.Fatalf("unexpected error: %v", err)
+	}
+
+	var line string
+	for _, l := range logged {
+		if strings.Contains(l, "warnings") {
+			line = l
+			break
+		}
+	}
```
PKNaveen, line 332:
```suggestion
l:= strings.Join(logged, "\n")
```

You can replace this with strings.join

pkg/framework/plugins/nodeutilization/usageclients_test.go
```diff
@@ -297,3 +303,47 @@ func TestPrometheusUsageClient(t *testing.T) {
...
+		t.Fatalf("expected the prometheus warnings to be logged, got %q", logged)
+	}
+	// The warnings have to be attached as a structured key/value pair. A printf
+	// style call leaves the verb in the message and drops the value.
+	if strings.Contains(line, "%v") {
+		t.Errorf("log message contains an uninterpreted format verb: %q", line)
+	}
+	if !strings.Contains(line, `"warnings"=`) {
+		t.Errorf("expected a structured \"warnings\" key, got %q", line)
+	}
```
PKNaveen, line 343: You can join both these conditions as we are expected to return same log structure and error out if any other is provided

pkg/framework/plugins/nodeutilization/usageclients_test.go
```diff
@@ -146,11 +149,13 @@ func TestActualUsageClient(t *testing.T) {
...
 	dataType model.ValueType
+	warnings []string
 }
 
 type fakePayload struct {
-	Status string      `json:"status"`
-	Data   queryResult `json:"data"`
+	Status   string      `json:"status"`
+	Data     queryResult `json:"data"`
+	Warnings []string    `json:"warnings,omitempty"`
```
ingvagabund, line 158: Is this part of an actual payload type?

pkg/framework/plugins/nodeutilization/usageclients_test.go
```diff
@@ -297,3 +303,47 @@ func TestPrometheusUsageClient(t *testing.T) {
 		})
 	}
 }
+
+func TestNodeUsageFromPrometheusMetricsLogsWarnings(t *testing.T) {
```
ingvagabund, line 307: I don't think we have a precedence for validating any logging functionality. Though, it's an interesting area to explore. Are there any other projects that validate their logging code?

PR #1900, opened 2026-07-21
fix(nodeutilization): avoid appending onto a shared-cache pod's Containers slice

Mukuwul:
**What this does**

`withResourceRequestForAny` builds a combined container list with:

```go
all := append(pod.Spec.Containers, pod.Spec.InitContainers...)
```

The pods here come from the shared informer cache. When `cap(pod.Spec.Containers) > len(pod.Spec.Containers)`, `append` writes the init containers into the spare capacity of the cached slice's backing array — i.e. it mutates a shared cache object from inside what is meant to be a read-only filter. In the common case the slice has `len == cap` and `append` allocates, so it's latent, but it's the kind of aliasing bug that's hard to reproduce and easy to trip later.

This switches to `slices.Concat`, which always returns a fresh slice and never aliases its inputs. `slices` is already imported in this file.

None filed — noticed while reading the nodeutilization pod filters.

### DIFF
--- a/pkg/framework/plugins/nodeutilization/nodeutilization.go
+++ b/pkg/framework/plugins/nodeutilization/nodeutilization.go
@@ -758,7 +758,7 @@ func assessAvailableResourceInNodes(
 // has a resource request specified for any of the given resources names.
 func withResourceRequestForAny(names ...v1.ResourceName) podutil.FilterFunc {
 	return func(pod *v1.Pod) bool {
-		all := append(pod.Spec.Containers, pod.Spec.InitContainers...)
+		all := slices.Concat(pod.Spec.Containers, pod.Spec.InitContainers)
 		for _, name := range names {
 			for _, container := range all {
 				if _, ok := container.Resources.Requests[name]; ok {
--- a/pkg/framework/plugins/nodeutilization/nodeutilization_test.go
+++ b/pkg/framework/plugins/nodeutilization/nodeutilization_test.go
@@ -28,6 +28,7 @@ import (
 	"sigs.k8s.io/descheduler/pkg/api"
 	"sigs.k8s.io/descheduler/pkg/framework/plugins/nodeutilization/classifier"
 	"sigs.k8s.io/descheduler/pkg/framework/plugins/nodeutilization/normalizer"
+	"sigs.k8s.io/descheduler/test"
 )
 
 func BuildTestNodeInfo(name string, apply func(*NodeInfo)) *NodeInfo {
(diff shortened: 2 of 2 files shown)


pkg/framework/plugins/nodeutilization/nodeutilization.go
```diff
@@ -758,7 +758,7 @@ func assessAvailableResourceInNodes(
 // has a resource request specified for any of the given resources names.
 func withResourceRequestForAny(names ...v1.ResourceName) podutil.FilterFunc {
 	return func(pod *v1.Pod) bool {
-		all := append(pod.Spec.Containers, pod.Spec.InitContainers...)
+		all := slices.Concat(pod.Spec.Containers, pod.Spec.InitContainers)
```
ricardomaraschini, line 761: Can you please add a unit test for this ? I could not reproduce this so I most likely misunderstood you.

ricardomaraschini (changes requested):

pkg/framework/plugins/nodeutilization/nodeutilization.go
a7i: https://go.dev/play/p/0I-oFc4w8Pf

```
len=2 cap=2 [1 2]
len=2 cap=2 [3 4]
len=4 cap=4 [1 2 3 4]
```

Mukuwul: Fair point — objects that come straight from the API decoder have `len == cap`, which is the case your playground shows @a7i: the append allocates and nothing is shared. The case I was worried about is spare capacity (`cap > len`), where append writes through into the existing backing array instead of allocating:

```go
backing := make([]int, 3)
backing[0], backing[1], backing[2] = 1, 2, 99

containers := backing[:2] // len=2 cap=3
all := append(containers, 7)

fmt.Println(all)     // [1 2 7]
fmt.Println(backing) // [1 2 7] — 99 was overwritten in place
```

To be upfront about it: I can't point at a code path in the descheduler today that hands this filter a spare-capacity slice — decoded and DeepCopy'd objects are built exact-capacity — so this is defensive hardening for a filter that should be read-only, not a bug I can reproduce in production.

I've added the unit test you asked for: it builds a `Containers` slice with len 2 / cap 3 and asserts the element beyond len stays untouched. It fails with the old append (the init container gets written into the backing array) and passes with `slices.Concat`.

If you feel the hardening isn't worth carrying I'm happy to close, but the test seemed worth having either way.

ricardomaraschini: Giving the small change I am OK with this (even though I don't see this as something we would hit).  Can you please fill in the CLA Authorization ?

Mukuwul: done

Mukuwul:
Looked into the e2e failures — they're not specific to this PR. The same three jobs are red on every recent run across #1890, #1891 and #1904 as well; I've written up the details (RBAC bootstrap failure on cluster bring-up, plus leftover-state cascade once one test fails) in #1905.

For what it's worth, this PR's changed line can't be exercised by the e2e suite at all: `withResourceRequestForAny` is only used by HighNodeUtilization, which has no e2e test.

Will `/retest` once #1905 clears up.

PR #1902, opened 2026-07-21
Always emit CronJob suspend in Helm chart when value is a bool

dpacgdm:
## Summary
- CronJob Helm template used `{{- if .Values.suspend }}`, which skips `suspend: false` (Go template zero-value), so the field never appeared in the rendered manifest
- After an external `kubectl patch` set `suspend: true`, Helm/GitOps had nothing to reconcile against and could not restore `suspend: false`
- Emit `suspend` whenever the value is a bool (`kindIs bool`), matching the existing pattern for `automountServiceAccountToken`, and cover default `false` / `true` in helm-unittest

## Test plan
- [ ] `helm unittest charts/descheduler` (new CronJob suspend cases)
- [ ] `helm template` with default values shows `suspend: false`
- [ ] `helm template --set suspend=true` shows `suspend: true`
- [ ] Chart lint / Helm CI workflow green

### DIFF
--- a/charts/descheduler/templates/cronjob.yaml
+++ b/charts/descheduler/templates/cronjob.yaml
@@ -15,7 +15,7 @@ metadata:
     {{- end }}
 spec:
   schedule: {{ .Values.schedule | quote }}
-  {{- if .Values.suspend }}
+  {{- if kindIs "bool" .Values.suspend }}
   suspend: {{ .Values.suspend }}
   {{- end }}
   concurrencyPolicy: "Forbid"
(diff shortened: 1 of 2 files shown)


a7i:
@dpacgdm would you rebase the PR please?

dpacgdm:
Done :)

a7i:
Looks like we still have a conflict in `charts/descheduler/tests/cronjob_test.yaml`

once resolved, I will add lgtm again

PR #1926, opened 2026-09-21
Limit pods removed per deployment

yuvraj9:
Limit pods removed per deployment


yuvraj9:
Still WIP, will raise again after final testing

